PR #38, opened 2025-10-07
Feature/scaffold plugin structure issue 24

### LINKED ISSUE #24: Scaffold the initial plugin structure
### Description

Set up the foundational plugin structure for the AI Experiments plugin within the WordPress/ai repository.  This includes creating the base plugin files, organizing directories, configuring build tooling, and establishing a structure that makes it easy for individual features to be developed, reviewed, and potentially migrated into other WordPress projects (for example, the Gutenberg plugin or WordPress core).

### Goals

- [ ] Create the plugin bootstrap file (`ai.php`) with standard WordPress plugin headers.
- [ ] Establish clear directories for `/includes`, `/admin`, `/features`, `/assets`, `/languages`, and `/tests`.
- [ ] Each feature (for example, Title Rewriting, Excerpt Generation, Alt Text Generation) should live within its own folder or class namespace to ensure it is self-contained and independently reviewable.
- [ ] Configure Composer for dependency management, including the WP AI Client SDK.
- [ ] Implement a lightweight feature registration or loader system so new experiments can be added without editing core plugin files.
- [ ] Add initial documentation describing the directory layout, how to add a new feature, and expectations for encapsulation.
- [ ] Include (or update existing) standard files such as README.md, DEVELOPERS.md, and `.editorconfig` following canonical plugin practices.

### Design Considerations

- Encapsulation is key.  Each feature should function with minimal coupling to the rest of the plugin to simplify future elevation into Gutenberg or WordPress core.
- Maintain alignment with WordPress coding standards and plugin architecture patterns.
- Scaffold with testing hooks and localization in place to reduce rework later.

### Acceptance Criteria

- Plugin activates successfully with no errors or notices.
- Directory structure and feature organization follow the documented pattern.
- Code for each experimental feature can be isolated and reviewed independently.
- Initial documentation is available describing how to extend or contribute new features.

Ref34t:
## Scaffold Initial Plugin Structure

  Implements #24 – Foundational plugin structure for the WordPress AI Experiments plugin.

  ## Overview

  This work delivers the core architecture for a modular, feature-based AI plugin. Features can be
  registered, toggled, and tested in isolation, and the repository now follows the conventions we’ll
  reuse as we add more experiments.

  ## What’s Included

  ### Core Architecture

  - Plugin bootstrap (ai.php) – WordPress headers, version guards (PHP 7.4+, WP 6.7+), and
    initialization wiring.
  - Feature system – Registry refactored around a reusable Feature_Collection, maintaining enable/
    disable logic and conditional feature support.
  - PSR-4 autoloading – Composer-managed namespaces for WordPress\AI core code and feature modules.
  - Interface-driven design – Feature and Conditional_Feature interfaces ensure every experiment
    exposes the same surface.

  ### Plugin Structure

  ai/
  ├── ai.php
  ├── includes/
  │   ├── Plugin.php
  │   ├── Feature_Collection.php
  │   ├── Feature_Registry.php
  │   ├── Interfaces/
  │   └── Abstracts/
  ├── features/
  │   └── Example_Feature/
  ├── tests/
  │   ├── bootstrap.php
  │   ├── Integration/
  │   │   ├── Features/
  │   │   └── Includes/
  │   └── Unit/
  │       └── Includes/
  ├── docs/
  │   ├── CONTRIBUTING.md
  │   └── TESTING.md
  ├── phpcs.xml.dist
  └── phpunit.xml.dist

  ### Development Infrastructure

  - WordPress Coding Standards (PHPCS) plus PHPCompatibility checks.
  - Testing – PHPUnit configuration, bootstrap wiring, and refreshed integration/unit suites in PSR-
    4 directories.
  - Static analysis – PHPStan bootstrap and composer stan script with lightweight WordPress stubs.
  - Composer scripts – composer lint, composer format, composer stan, composer test.

  ### Documentation

  - docs/CONTRIBUTING.md – Expanded developer guide covering setup, feature creation, testing, and
    contribution workflow.
  - docs/TESTING.md – Updated strategy with new test locations and commands.
  - README – Points to the new docs and outlines the development experience.

  ## Key Features

  ### Encapsulation

  Every experiment lives under its own directory in features/, making reviews, toggles, and
  migrations painless.

  ### Extensibility

  Third-party developers can add custom features by hooking into the registry:
````
  add_action( 'ai_register_features', function ( $registry ) {
      $registry->register_feature( new My_Custom_Feature() );
  } );

````

  ### Feature Control

  Features remain fully overridable:
````
  add_filter( 'ai_feature_enabled', function ( $enabled, $feature_id ) {
      return 'example-feature' === $feature_id ? false : $enabled;
  }, 10, 2 );

 ````

### CODE REVIEW

**Comment by jeffpaul** [comment]
@Ref34t thanks for picking up #24 and working on this PR!  I see that the PR is still in draft, so is there more work you have pending or should this be considered ready for review?

**Comment by Ref34t** [comment]
@jeffpaul you're welcome. I'll assign it for review today after I finish my iteration on it.

**Comment by Ref34t** [comment]
@jeffpaul it's ready now

File: ai.php [hunk 1]
```diff
@@ -0,0 +1,179 @@
...
+ * Deactivation hook callback.
+ *
+ * @since 0.1.0
+ */
+function ai_deactivate() {
+	// Flush rewrite rules on deactivation.
+	flush_rewrite_rules();
+}
+
+register_deactivation_hook( __FILE__, __NAMESPACE__ . '\ai_deactivate' );
```
felixarntz: Why are these needed? What does this plugin do that requires flushing rewrite rules?

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,464 @@
+# Developer Guide
```
felixarntz: This is _a lot_ of documentation. I'm by no means opposed to it, but this PR is already huge without it, so I think we should break things down a bit. At least extensive documentation like this should be in a separate PR so that it can be reviewed separately without holding up the code in this PR.

Same goes for the `TESTING.md` file and some of the other documentation. Could you please remove them from here and instead open a separate PR to add them?

includes/Abstracts/Abstract_Feature.php
```diff
@@ -0,0 +1,176 @@
...
+	public function is_enabled(): bool {
+		/**
+		 * Filters whether a specific feature is enabled.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param bool   $enabled    Whether the feature is enabled.
+		 * @param string $feature_id The feature identifier.
+		 */
+		return apply_filters( 'ai_feature_enabled', $this->enabled, $this->id );
```
felixarntz: Consider replacing this with a `"ai_feature_{$this->id}_enabled"` filter, or alternatively add such a filter to run before this one.

It allows simple one liners like `add_filter( 'ai_feature_example_enabled', '__return_false' );`.

includes/Abstracts/Abstract_Feature.php
```diff
@@ -0,0 +1,176 @@
...
+	 *
+	 * Useful for debugging and testing.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return array Array of registered hooks.
+	 */
+	public function get_hooks(): array {
+		return $this->hooks;
+	}
```
felixarntz: Why are these needed? It seems a bit excessive and theoretical in nature, and there's no prior art I'd be aware of in WordPress projects.

includes/Feature_Collection.php
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Manages a collection of features.
+ *
+ * This class is responsible for storing and retrieving feature instances
+ * in a WordPress-independent manner.
+ *
+ * @since 0.1.0
+ */
+class Feature_Collection {
```
felixarntz: IMO there's no need to have a separate `Feature_Collection` class. All of what it does could easily be incorporated into `Feature_Registry` without losing any flexibility.

includes/Feature_Registry.php
```diff
@@ -0,0 +1,189 @@
...
+
+	/**
+	 * Private constructor to enforce singleton pattern.
+	 *
+	 * @since 0.1.0
+	 */
+	private function __construct() {
+		$this->feature_collection = new Feature_Collection();
+		$this->register_default_features();
+	}
```
felixarntz: The singleton pattern is an anti-pattern that has several problems such as testability.

If we want to have a single central instance to operate on, I'd recommend storing the instance somewhere else, e.g. a `wp_ai_feature_registry()` function with a `static` variable that returns that instance (and creates it the first time the function is called). This gives us something similar to the singleton pattern, without the associated drawbacks.

includes/Feature_Registry.php
```diff
@@ -0,0 +1,189 @@
...
+
+	/**
+	 * Initializes all enabled features.
+	 *
+	 * Loops through all registered features and calls their register() method
+	 * if they are enabled and meet requirements (for conditional features).
+	 *
+	 * @since 0.1.0
+	 */
+	public function initialize_features(): void {
```
felixarntz: The methods for initializing and registering the features (this method here and everything below) would be better off in a separate class, which could act as the central orchestrator. Maybe call it `Feature_Loader` or `Feature_Initializer` or something like that.

It could internally use the `Feature_Registry`. But firing all the hooks on the registry itself and registering features on itself is not a good pattern, as it couples the concrete usage with the flexible implementation.

includes/Interfaces/Conditional_Feature.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Interface for features with activation requirements.
+ *
+ * Features implementing this interface can define requirements that must be met
+ * before the feature can be enabled. Examples include PHP extensions, WordPress
+ * version, or other plugin dependencies.
+ *
+ * @since 0.1.0
+ */
+interface Conditional_Feature extends Feature {
```
felixarntz: This seems a bit premature to have at this point? Unless it was discussed as a requirement before that I missed. Yes, there may at some point be features that have specific requirements that go beyond the plugin's own requirements, but I'm not sure it makes sense to introduce and think about it just yet? It's another thing that increases complexity of this PR, which we could probably avoid for now.

includes/Interfaces/Feature.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/**
+ * Feature interface.
+ *
+ * @package WordPress\AI\Interfaces
+ */
+
+namespace WordPress\AI\Interfaces;
```
felixarntz: Nit: Consider calling the namespace `Contracts` instead of `Interfaces`, it's a bit more commonplace and in line with our other AI PHP projects.

includes/Plugin.php
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	/**
+	 * Private constructor to enforce singleton pattern.
+	 *
+	 * @since 0.1.0
+	 */
+	private function __construct() {
+		$this->feature_registry = Feature_Registry::instance();
+	}
```
felixarntz: See my other comment about avoiding the singleton pattern.

In any case, even if we wanted to keep the singleton pattern, we should only use it on the single entry point class that we really only want to initialize once. Classes that are instantiated within that single instance don't need to _also_ use the singleton pattern.

As a higher-level feedback though, I don't really see why we need this class at all. All that it does is set up the `Feature_Registry`. I'd suggest replacing this class with the `Feature_Loader` (or similar) that I mentioned above. This gives it a concrete purpose, not just loading something else.

includes/Plugin.php
```diff
@@ -0,0 +1,99 @@
...
+	 *
+	 * @since 0.1.0
+	 */
+	private function register_hooks(): void {
+		/**
+		 * Fires after the AI plugin has been initialized.
+		 *
+		 * @since 0.1.0
+		 */
+		do_action( 'ai_plugin_initialized' );
```
felixarntz: I don't think there's a need for this action. There are already concrete actions for registering features, which is the only thing that has a concrete need at this point.

languages/readme.txt
```diff
@@ -0,0 +1,3 @@
+This directory contains translation files for the WordPress AI Experiments plugin.
+
+WordPress will automatically load translations from WordPress.org when the plugin is hosted there.
```
felixarntz: No need for this directory. This will be on WordPress.org, and translation files will be provided from there, not bundled within the plugin.

phpstan.neon.dist
```diff
@@ -0,0 +1,14 @@
...
+		- includes
+		- features
+	parallel:
+		maximumNumberOfProcesses: 1
+	bootstrapFiles:
+		- tests/phpstan-bootstrap.php
+	ignoreErrors:
+		-
+			message: '#^Call to undefined function [A-Za-z0-9_\\\\]+\(.*\)$#'
+			path: .*
```
felixarntz: Why is this exclusion needed?

readme.txt
```diff
@@ -0,0 +1,89 @@
+=== AI Experiments ===
```
felixarntz: See above, adding the `readme.txt` should be its own PR, also as it requires a completely different expertise to review.

felixarntz (changes requested):
@Ref34t Thank you for kick-starting this! I think a lot of it goes in the right direction, but a few pieces seem overly complicated or catering for theoretical problems that we don't know are relevant.

I encourage you to simplify the PR according to my feedback and break out all the documentation into separate PRs to streamline the review process and decouple documentation and marketing language from source code.

readme.txt
jeffpaul: Note that I added the `readme.txt` file as some changes were made to `README.md` that are more properly formatted for the WPORG txt version.  Happy to strip that out and roll into a separate PR if helpful.

ai.php
Ref34t: I think it originates from the possibility of needing it for a certain feature, but I do agree with you, this can be removed as it doesn't add any additional functionality now. 

handled here https://github.com/WordPress/ai/pull/38/commits/99fe2cbc8af4dea8b45973cd75222d889263ceed

docs/CONTRIBUTING.md
Ref34t: Alright sounds very reasonable, handled here https://github.com/WordPress/ai/pull/38/commits/7926793f087609efbccff806b7ee2ed9e6538aff

includes/Abstracts/Abstract_Feature.php
Ref34t: I like that alot, resolved here [29959e4](https://github.com/WordPress/ai/pull/38/commits/29959e4b1cfbf7e977bd9d4a09bbf58e94f283a5)

includes/Abstracts/Abstract_Feature.php
Ref34t: resolved here [3c920d1](https://github.com/WordPress/ai/pull/38/commits/3c920d166a3f08e4913ca33881c79634820ffb9b)

README.md
```diff
@@ -43,9 +55,10 @@ The AI Experiments plugin is meant to be studied, forked, and extended.  If you
 
 If you’re a plugin developer, you’ll be able to:
 
-* Register new AI abilities
-* Override default behavior with custom filters
-* Reuse the same building blocks in your own plugins
+*   Read the [Contributing Guide](docs/CONTRIBUTING.md) for detailed development information.
```
swissspidy: This file doesn't exist. It should be in the root folder anyway though.

Ref34t: It was added to another PR for separation of concerns https://github.com/WordPress/ai/pull/42

includes/Feature_Collection.php
Ref34t: good organization, handled here https://github.com/WordPress/ai/pull/38/commits/ca6e6918657505f8b57eafa7213d09c43a29c9aa

includes/Feature_Registry.php
Ref34t: I like that!
handled here https://github.com/WordPress/ai/pull/38/commits/6b122a9ce9e49e84fbee9fa2e6d89fa2ba41c37d

phpstan.neon.dist
Ref34t: was included during some unit testing not needed now resolved https://github.com/WordPress/ai/pull/38/commits/434a9694233c62a418e444d10428db2d2fae78b3

languages/readme.txt
Ref34t: https://github.com/WordPress/ai/pull/38/commits/4669873a145c148cb4d43f31af8317e6b807cf62

includes/Interfaces/Feature.php
Ref34t: handled https://github.com/WordPress/ai/pull/38/commits/35af80ab255791ba775c374f344e4fef6670b8dc

includes/Interfaces/Conditional_Feature.php
Ref34t: I agree with you handled here https://github.com/WordPress/ai/pull/38/commits/35af80ab255791ba775c374f344e4fef6670b8dc

includes/Plugin.php
Ref34t: handled https://github.com/WordPress/ai/pull/38/commits/ba83023efddd00907053efebd7244875e6199cc8

includes/Plugin.php
Ref34t: handlded https://github.com/WordPress/ai/pull/38/commits/ba83023efddd00907053efebd7244875e6199cc8

README.md
```diff
@@ -1,6 +1,18 @@
-# AI Experiments Plugin for WordPress
+=== WordPress AI Experiments ===
+Contributors: wordpressorg
+Tags: ai, artificial intelligence, experiments, features
+Requires at least: 6.7
+Tested up to: 6.8
+Stable tag: 0.1.0
+Requires PHP: 7.4
+License: GPLv2 or later
+License URI: https://www.gnu.org/licenses/gpl-2.0.html
```
swissspidy: IMO this is something for `readme.txt` for the plugin directory, not the GitHub readme

tests/phpstan-bootstrap.php
```diff
@@ -0,0 +1,54 @@
...
+ * Provides light-weight stubs for common WordPress functions so that
+ * static analysis can run without loading the full WP stack.
+ *
+ * @package WordPress\AI\Tests
+ */
+
+$autoloader = __DIR__ . '/../vendor/autoload.php';
+if ( file_exists( $autoloader ) ) {
+	require_once $autoloader;
+}
```
swissspidy: PHPStan should already check this. If not, specify as a separate file in the PHPStan config instead.

tests/phpstan-bootstrap.php
```diff
@@ -0,0 +1,54 @@
+<?php
+/**
+ * Bootstrap file for PHPStan analysis.
+ *
+ * Provides light-weight stubs for common WordPress functions so that
+ * static analysis can run without loading the full WP stack.
```
swissspidy: Let's just use the wordpress-stubs package, it's easier than having to maintain this ourselves :-)

Ref34t: thanks for that handled it https://github.com/WordPress/ai/pull/38/commits/b10d728f102b6f2305f05bd091c42d6369f9d1cc

tests/phpstan-bootstrap.php
Ref34t: handled https://github.com/WordPress/ai/pull/38/commits/57c7e7ae44938d0cc7566ef0eea13cf82a6bf95a

README.md
Ref34t: I've created it and moved it to the other pr

.gitignore
```diff
@@ -0,0 +1,33 @@
+# Composer
+/vendor/
+composer.lock
```
dkotter: I think we want to allow this, as we normally do commit lock files

.gitignore
```diff
@@ -0,0 +1,33 @@
...
+
+# Node
+node_modules/
+package-lock.json
```
dkotter: Same comment here, not sure this should be ignored

.gitignore
```diff
@@ -0,0 +1,33 @@
...
+
+# WordPress
+wp-config.php
+wp-content/uploads/
+wp-content/cache/
```
dkotter: Do we need to ignore these? Is there a scenario where these files would ever end up nested within this repo?

.phpunit.result.cache
dkotter, line 1: Seems like this file should be ignored?

README.md
dkotter: Looking at that PR, seems we do move this file up to the root so the URL here can be modified now:
```suggestion
*   Read the [Contributing Guide](CONTRIBUTING.md) for detailed development information.
```

ai.php
```diff
@@ -0,0 +1,178 @@
+<?php
+/**
+ * Plugin Name: AI
```
dkotter: Mostly leaving this to open conversation. Are we wanting this plugin to just be named `AI`? We talk about it elsewhere as `AI Experiments` or `WordPress AI Experiments` so wondering if we want to update the plugin name here to be more descriptive?

ai.php
```diff
@@ -0,0 +1,178 @@
...
+ * Plugin URI: https://github.com/WordPress/ai
+ * Description: Experimental AI features for WordPress
+ * Version: 0.1.0
+ * Requires at least: 6.7
```
dkotter: This might be a larger conversation but since this plugin will rely on the Abilities API, do we want the minimum version here set to 6.9? Or are we wanting to bundle the Abilities API code into this plugin?

ai.php
```diff
@@ -0,0 +1,178 @@
...
+ * Requires PHP: 7.4
+ * Author: WordPress Contributors
```
dkotter: Looking at a few other plugins, seems this is normally set to `WordPress.org Contributors`, though the WP AI Client uses `WordPress AI Team`

ai.php
```diff
@@ -0,0 +1,178 @@
...
+ */
+
+namespace WordPress\AI;
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// Define plugin constants.
```
dkotter: While there's nothing necessarily wrong with the rest of the code here, I'd suggest we follow the same approach that either the [MCP Adapter plugin](https://github.com/WordPress/mcp-adapter/blob/trunk/mcp-adapter.php) does or the [Abilities API plugin](https://github.com/WordPress/abilities-api/blob/trunk/abilities-api.php) does, meaning most of the functionality isn't in this main file and is instead in a bootstrap file or a base plugin class

ai.php
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/**
+ * Displays an admin notice for version requirement failures.
+ *
+ * @since 0.1.0
+ *
+ * @param string $message The error message to display.
+ */
+function ai_version_notice( $message ) {
```
dkotter: Since we're already in a namespaced file, don't think we need the `ai_` namespace on each function here

includes/Features/Example_Feature/Example_Feature.php
dkotter, line 1: So this gets into opinion territory but not sure I agree with using `features` as a directory. Seems like every other canonical plugin just uses an `includes` directory and everything is nested within there. That's what I'd suggest here so we're keeping parity with how things are done elsewhere

includes/Features/Example_Feature/Example_Feature.php
```diff
@@ -0,0 +1,113 @@
...
+namespace WordPress\AI\Features\Example_Feature;
+
+use WordPress\AI\Abstracts\Abstract_Feature;
+
+/**
+ * Reference feature demonstrating hooks and REST endpoints.
+ *
+ * @since 0.1.0
+ */
+class Example_Feature extends Abstract_Feature {
```
dkotter, line 17: Another opinion but seems like we'd be better using the name `Ability` over `Feature`, since we are building on the Abilities API. As an example, I would look to follow more or less what [Felix did in his chatbot demo](https://github.com/felixarntz/wp-ai-sdk-chatbot-demo/tree/main/includes/Abilities), as far as how the code structure would be set up.

dkotter (changes requested):

includes/Features/Example_Feature/Example_Feature.php
felixarntz: +1

I'd say move the `features` folder to become `includes/Features` (with proper namespace).

includes/Feature_Loader.php
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		$this->initialized = true;
+
+		/**
+		 * Fires after all features have been initialized.
+		 *
+		 * @since 0.1.0
+		 */
+		do_action( 'ai_features_initialized' );
```
felixarntz: Probably the order of these should be reversed: If logic happens on the action, it's also part of the initialization IMO.
```suggestion
		/**
		 * Fires after all features have been initialized.
		 *
		 * @since 0.1.0
		 */
		do_action( 'ai_features_initialized' );

		$this->initialized = true;
```

includes/Feature_Loader.php
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/**
+ * Gets the global Feature_Registry instance.
+ *
+ * @since 0.1.0
+ *
+ * @return Feature_Registry The Feature_Registry instance.
+ */
+function wp_ai_feature_registry(): Feature_Registry {
```
felixarntz: Why is this "singleton alternative" in this file? Classes and functions shouldn't be mixed within the same file. I'd say you can move this function to the plugin main file, or (based on another comment from @dkotter) maybe a separate `load.php` file or similar.

includes/Feature_Registry.php
```diff
@@ -0,0 +1,160 @@
...
+use WordPress\AI\Contracts\Feature;
+
+/**
+ * Manages a collection of features.
+ *
+ * This class is responsible for storing and retrieving feature instances.
+ *
+ * @since 0.1.0
+ */
+class Feature_Collection {
```
felixarntz: Similar to the above, there shouldn't be multiple classes in a single file.

Following up on the feedback from my previous review, I'm still curious why we need `Feature_Collection`. All of this can be centralized in `Feature_Registry` itself, no need for the extra class which is mostly just additional code without a particular benefit IMO.

felixarntz (commented):
@Ref34t This is looking almost good to go, though a few pieces to adjust remain (including @dkotter's feedback).

includes/Features/Example_Feature/Example_Feature.php
felixarntz: I disagree on this one, because the AI Experiments plugin would have different features, but those will not all be abilities. For example, if we implement a button in Gutenberg to generate a post titles with AI, that's a feature. It's not an ability that we would provide as a tool for an LLM. A tool for an LLM would be e.g. "set_post_title", and then it could obviously generate a title on its own and call the tool to save it.

The latter is for agentic experiences, the former is for user-invoked "just use AI for this specific purpose" features.

dkotter: > For example, if we implement a button in Gutenberg to generate a post titles with AI, that's a feature. It's not an ability that we would provide as a tool for an LLM

So I guess the way I was thinking of this is the UI part of this would be a feature separate from an ability (in this example, the button you'd click to trigger things). But I was assuming the underlying functionality that is triggered when the button is clicked (making a request to an AI provider to generate a post title) would be something we register as an ability, which allows others to trigger this functionality however they see fit.

If I'm reading this correctly, you wouldn't want to expose functionality like that as an ability?

felixarntz: I think abilities are primarily meant to expose features of WordPress to autonomous AI agents, equivalent to "tools" (when using the jargon more common in the AI ecosystem). In other words, most abilities won't by themselves use AI, but it would _be used_ by AI. Of course there are exceptions, e.g. there could be an ability to generate an image using a specific prompt, which would effectively allow a regular text-based model to invoke an image generation model. But usually abilities wouldn't need to use AI.

On the other end of the spectrum are user-facing features like a button to generate a post title using AI. To me, based on the above "generate and set post title with AI" doesn't sound like something we should have an ability for:

* Sure, we could have a "generate and set post title with AI" ability. But that would mean calling another LLM model just to generate the post title and then set it, which would also be a hyper-specific ability.
* Since abilities are meant to be used by AI models or agents, there could simply be a "set post title" ability, and that would be enough, since the LLM could generate a post title anyway on its own and then call that "set post title" ability.

.gitignore
Ref34t: that's right 
resolved here https://github.com/WordPress/ai/pull/38/commits/f7c34182e801251f77da64a89965e195a7e30dac

.gitignore
Ref34t: done here https://github.com/WordPress/ai/pull/38/commits/f7c34182e801251f77da64a89965e195a7e30dac

.gitignore
Ref34t: done here https://github.com/WordPress/ai/pull/38/commits/f7c34182e801251f77da64a89965e195a7e30dac

.phpunit.result.cache
Ref34t: done https://github.com/WordPress/ai/pull/38/commits/f7c34182e801251f77da64a89965e195a7e30dac

ai.php
Ref34t: I think this can be discussed in another issue what name we are going to give this plugin

ai.php
Ref34t: makes sense I think will update it

Ref34t: The Abilities API will be merged into the core by 6.9, so I think the minimum version needs to be 6.9 as you said

ai.php
Ref34t: updated to `WordPress.org Contributors`

ai.php
Ref34t: resolved here https://github.com/WordPress/ai/pull/38/commits/26b3736736abca5231e692a911881ab8c5eda112

ai.php
Ref34t: resolved here https://github.com/WordPress/ai/pull/38/commits/26b3736736abca5231e692a911881ab8c5eda112

includes/Features/Example_Feature/Example_Feature.php
Ref34t: resolved

ai.php
Ref34t: resolved

ai.php
Ref34t: resolved

includes/Features/Example_Feature/Example_Feature.php
Ref34t: I agree with @felixarntz  the term `ability` is already used in the `Abilities API` so not a good idea to use it here, I'm fine with `Feature`, open to other synonyms if you have another suitable one

includes/Feature_Loader.php
Ref34t: resolved https://github.com/WordPress/ai/pull/38/commits/3830c16fde39b5ade0a9652df13aafbc99be72d8

includes/Feature_Loader.php
Ref34t: https://github.com/WordPress/ai/pull/38/commits/bc0b265b97ffc59339ec7fd3f7fd6adb74fd7fa8 resolved

includes/Feature_Registry.php
Ref34t: yes I agree with you
handled here https://github.com/WordPress/ai/pull/38/commits/25418e932cb2d183409d280185ccd8795aec3f85
I think I was arranging tests but made it more complicated with no gain 👍

includes/Features/Example_Feature/Example_Feature.php
```diff
@@ -0,0 +1,113 @@
...
+	 * Sets up metadata for the example feature.
+	 *
+	 * @since 0.1.0
+	 */
+	public function __construct() {
+		$this->id          = 'example-feature';
+		$this->label       = __( 'Example Feature', 'ai' );
+		$this->description = __( 'Demonstrates the AI feature system with example hooks and functionality.', 'ai' );
+		$this->enabled     = true;
+	}
```
felixarntz, line 31: I think we should streamline this to be enforced by the `Abstract_Feature` class. Right now, there's no contract in the implementation that makes it obvious that these 4 properties need to be set.

I suggest the following:
* Implement a `__construct()` on the abstract base class.
* Implement a new method `abstract protected load_feature_metadata(): array` which must return an associative array with keys `id`, `label`, `description`.
* Call that method in the base class constructor, then set the `id`, `label`, and `description` properties based on the values.
* Set `enabled` to `true` automatically, this does not make sense to be controlled by the individual feature class implementations.
* Then each feature class, including `Example_Feature`, should implement the `load_feature_metadata()` method instead of a custom constructor.

includes/Abstracts/Abstract_Feature.php
```diff
@@ -0,0 +1,125 @@
...
+	 */
+	protected $description;
+
+	/**
+	 * Whether the feature is enabled.
+	 *
+	 * @since 0.1.0
+	 * @var bool
+	 */
+	protected $enabled = true;
```
felixarntz: I think we should make this private, at least for now: The individual feature class implementation shouldn't be able to control whether it's active - it'll be something that developers should control via the available filter, or the plugin's users via settings UI.

By making it private, we ensure nobody can touch it.

includes/Abstracts/Abstract_Feature.php
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	/**
+	 * Checks if feature is enabled.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return bool True if enabled, false otherwise.
+	 */
+	public function is_enabled(): bool {
```
felixarntz: Similarly to the above, let's make this method `final` to ensure nobody can tamper with whether a feature is active or not in unexpected ways. It should be controlled via the filters implemented here.

tests/Integration/Includes/Feature_LoaderTest.php
```diff
@@ -0,0 +1,268 @@
...
+	 * Constructor.
+	 *
+	 * @since 0.1.0
+	 */
+	public function __construct() {
+		$this->id          = 'mock-feature';
+		$this->label       = 'Mock Feature';
+		$this->description = 'A mock feature for testing';
+		$this->enabled     = true;
+	}
```
felixarntz: This would need to be updated as well, based on my other comment.

felixarntz (commented):
@Ref34t Getting close!

My last remaining concern is to make the feature abstraction a bit easier and also more error-proof to use.

tests/Integration/Includes/Feature_LoaderTest.php
Ref34t: resolved

includes/Abstracts/Abstract_Feature.php
Ref34t: resolved

includes/Abstracts/Abstract_Feature.php
Ref34t: resolved

includes/Features/Example_Feature/Example_Feature.php
Ref34t: resolved

Ref34t:
@felixarntz I've pushed a new commit covering the points you mentioned yesterday with some improvements over them. Here is what this last iteration does: 

Introducing a comprehensive validation system for the plugin's feature architecture, ensuring all features have valid metadata before they can be loaded. The changes centralize metadata management in the `Abstract_Feature` base class and add proper exception handling throughout the feature loading pipeline.

## Summary of Changes

### New Exception Classes
- `Invalid_Feature_Exception` - for invalid feature instances
- `Invalid_Feature_Metadata_Exception` - for missing/invalid metadata

### Abstract_Feature
- Added `final __construct()` that validates and assigns  metadata (`id`, `label`, `description`)
- Added `abstract load_feature_metadata()` method that child classes must implement
- Changed `$enabled` from `protected` to `private`
- Made `is_enabled()` final

### Feature_Loader
- Wrapped feature loading in try-catch to handle validation exceptions
- Returns `WP_Error` when features fail validation

### Example_Feature
- Implemented `load_feature_metadata()` returning feature  metadata array
- Removed direct property assignments (now handled by parent  constructor)
 
### bootstrap.php
 - Added exception handling that displays admin notices when  Features fail to load

dkotter (approved):
Thanks for all the work here @Ref34t! Overall this looks good to me, noting I'd expect there to be some changes as we actually start building things out but I feel this is a good base to start with

jeffpaul:
@dkotter I'll defer to you on merge order here between this and #46 to minimize conflicts

includes/Abstracts/Abstract_Feature.php
```diff
@@ -0,0 +1,172 @@
...
+
+		/**
+		 * Filters the enabled status across all features.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param bool   $enabled    Whether the feature is enabled.
+		 * @param string $feature_id The feature identifier.
+		 */
+		return apply_filters( 'ai_feature_enabled', $enabled, $this->id );
```
mindctrl: Is there a need for two filters here? I didn't see any discussion around it, so was trying to understand the need for both.

Ref34t: the second one is a master filter to disable all the features which your comment showed me that it shouldn't be here so I refractored it to be in the loader https://github.com/WordPress/ai/commit/bb89550310c6b1166ab31746363123f8c6e2c2f6

dkotter:
> @dkotter I'll defer to you on merge order here between this and #46 to minimize conflicts

I would merge this one in first as it contains all the scaffold work. And then I can rebase #46 to fix conflicts

includes/Abstracts/Abstract_Feature.php
```diff
@@ -0,0 +1,164 @@
...
+		 *
+		 * The dynamic portion of the hook name, `$this->id`, refers to the feature ID.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param bool $enabled Whether the feature is enabled.
+		 */
+		$enabled = apply_filters( "ai_feature_{$this->id}_enabled", $enabled );
+
+		return $enabled;
```
felixarntz: Always best to sanitize all filter results, since filters don't guarantee anything:
```suggestion
		return (bool) apply_filters( "ai_feature_{$this->id}_enabled", $enabled );
```

felixarntz (approved):
@Ref34t Thank you for the iterations, LGTM!

Just one minor thing that we can fix inline.

felixarntz:
@jeffpaul @dkotter @Ref34t Happy for this to be merged! I agree it should probably be merged before #46.

mindctrl:
I'm seeing a PHP notice due to translations being loaded too early.

> PHP Notice:  Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>ai</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the <code>init</code> action or later. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.7.0.)

This happens when `WordPress\AI\load()`, which runs on `plugins_loaded`, registers the Example Feature and ends up running `__()` in `Example_Feature::load_feature_metadata()`.

Maybe the feature initialization could be deferred to the `init` hook?

includes/Feature_Loader.php
```diff
@@ -0,0 +1,229 @@
...
+	 *
+	 * This is where built-in features are registered. Third-party features
+	 * should use the 'ai_register_features' action hook.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param array<mixed> $features Optional. Array of feature instances to register. Defaults to built-in features.
+	 * @throws Invalid_Feature_Exception If a feature does not implement the Feature interface.
+	 */
+	public function register_default_features( array $features = array() ): void {
```
felixarntz: Why is there a `$features` parameter? Seems unnecessary, given the method is explicitly called `register_default_features`.

includes/Feature_Loader.php
```diff
@@ -0,0 +1,229 @@
...
+		// Use provided features or load defaults.
+		if ( empty( $features ) ) {
+			$features = $this->get_default_features();
+		}
+
+		// Register all features with type validation.
+		foreach ( $features as $feature ) {
+			// Skip invalid feature instances.
+			if ( ! is_a( $feature, Feature::class ) ) {
```
felixarntz: Better:
```suggestion
			if ( ! $feature instanceof Feature ) {
```

includes/bootstrap.php
```diff
@@ -0,0 +1,179 @@
...
+		_doing_it_wrong(
+			__NAMESPACE__ . '\load',
+			sprintf(
+				/* translators: %s: Error message. */
+				esc_html__( 'AI Plugin initialization failed: %s', 'ai' ),
+				esc_html( $e->getMessage() )
+			),
+			'0.1.0'
+		);
+	}
```
felixarntz, line 185: +1 in regards to @mindctrl's comment, this should probably only happen on `init`:
* Only before `init` the user gets initialized by default, and that influences translation strings. It's a best practice to not run any translation functions before the user is normally initialized.
* It's common in WordPress for things to be registered on `init`, so this would tie in nicely with that practice.
* It would enforce that no feature can break these patterns either. `init` is sufficiently early in the WordPress bootstrap process for probably any kind of interaction a specific feature may need to support.

includes/bootstrap.php
```diff
@@ -0,0 +1,179 @@
...
+}
```
mindctrl: re: my earlier comment about loading translations too early... @felixarntz had a good idea to just move it all to `init`.

Another idea I had originally was to do something like this:

Extract the `try/catch` block to a new function, like `initialize_features()`, and attach it to `init` like so within `load()`:

`add_action( 'init', __NAMESPACE__ . '\initialize_features' );`

This would allow the plugin to bootstrap its main dependencies and non-translation related functionality on `plugins_loaded`, possibly to be used by other integrations before `init`, and then allow the features to be loaded at the appropriate time.

I think either approach works fine.

includes/bootstrap.php
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Checks if the PHP version meets the minimum requirement.
+ *
+ * @since 0.1.0
+ *
+ * @return bool True if PHP version is sufficient, false otherwise.
+ */
+function check_php_version(): bool {
+	if ( version_compare( phpversion(), AI_MIN_PHP_VERSION, '<' ) ) {
```
mindctrl, line 63: A small thing but when testing this PR against WP 6.9-beta1, using this constant set to `6.9` causes the plugin to show an admin notice and stops the bootstrap process.

> AI plugin requires WordPress version 6.9 or higher. You are running WordPress version 6.9-beta1-61040-src.

I had to edit the value to make the plugin work during beta testing. Of course it's only a temporary inconvenience, but it made me wonder if we should use a different way, like checking if the `wp_register_ability()` function exists. Thoughts?

includes/Features/Example_Feature/Example_Feature.php
justlevine: If this conversation is still open, then I recommend we avoid both `Ability` (domain specific) and `Feature` (limiting, will probably cause sorting/grouping problems when trying to delineate what is a "feature"), and go with something broader/more generic like `Experiment` or even `Module`.

includes/Features/Example_Feature/Example_Feature.php
```diff
@@ -0,0 +1,116 @@
...
+	 * Modifies the document title parts when debugging.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param array $title Title parts.
+	 * @return array
+	 */
+	public function modify_title( array $title ): array {
+		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+			$title['site'] = $title['site'] . ' [AI]';
```
mindctrl, line 67: Just noticed this triggers a PHP warning when viewing the front page of a site. (I'm using 6.9-beta1 with default 2025 theme).

> Warning: Undefined array key "site" in /wp-content/plugins/ai/includes/Features/Example_Feature/Example_Feature.php on line 67

This is because WordPress doesn't set `site` when `is_front_page()` returns true. Source: https://github.com/WordPress/wordpress-develop/blob/aae3ac32369a78cbf40472aa8cd39a74481cf687/src/wp-includes/general-template.php#L1253-L1257

includes/Feature_Loader.php
Ref34t: good catch removed it https://github.com/WordPress/ai/pull/38/commits/90579ca4f234bafd076e2d2a6d34c9d7e8337453

includes/bootstrap.php
Ref34t: @mindctrl Thank you for your insights. I went with this option. resolved here https://github.com/WordPress/ai/pull/38/commits/65139be45cdccb31d8092653a0f00ed480913784

includes/bootstrap.php
Ref34t: @felixarntz I went with John's approach, is this satisfying for you? https://github.com/WordPress/ai/pull/38/commits/65139be45cdccb31d8092653a0f00ed480913784

includes/bootstrap.php
Ref34t: @mindctrl I went with using a lower version for now until the release of 6.9, we can pump it up then

includes/Features/Example_Feature/Example_Feature.php
Ref34t: Good catch! I added a validation check using `isset()` before changing the `$title['site']` https://github.com/WordPress/ai/pull/38/commits/721d8a184ab64d3b4447b6a37fc9c3af7cba8455

dkotter (approved):

PR #42, opened 2025-10-12
Add documentation files to docs branch

### LINKED ISSUE #2: Add CONTRIBUTING.md
The initial version can probably be based on https://github.com/WordPress/php-ai-client/blob/trunk/CONTRIBUTING.md as a foundation, I suspect, to align.

For labels, I recommend that this has 'documentation' when triaged.

Ref34t:


CONTRIBUTING.md
swissspidy, line 1: `CONTRIBUTING.md` is usually in the project root, where GitHub will have special handling for it. See https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#about-contributing-guidelines

Ref34t: done it here https://github.com/WordPress/ai/pull/42/commits/6a06df20bdcf44c608f79195f4c2fab48df93197

CONTRIBUTING.md
```diff
@@ -0,0 +1,464 @@
...
+
+> **Note:** The `wordpress/wp-ai-client` package will be added to `composer.json` once it's officially released. For now, the plugin scaffolding is ready for integration.
+
+3. **Link to WordPress:**
+
+Symlink the plugin directory into your WordPress installation:
+
+```bash
+ln -s /path/to/ai /path/to/wordpress/wp-content/plugins/ai
+```
```
dkotter: This seems like a very specific development step that won't apply to everyone and is not one I see documented on other plugins. Isn't it just as easy to clone this repo into your `wp-content` directory? I'd maybe just remove this as a step

CONTRIBUTING.md
dkotter, line 1: Overall this document is very detailed, to the point where I think it could be broken down into separate files. There's a lot more here than what you normally find in a standard `CONTRIBUTING.md` file, to the point where it's a little overwhelming if I am a new contributor looking for instructions

Ref34t: @dkotter I've streamlined `CONTRIBUTING.md` to match other repos standard and separate the `DEVELOPER_GUIDE.md` so it has the way forward

CONTRIBUTING.md
Ref34t: @dkotter resolved

CONTRIBUTING.md
mindctrl: I had a similar thought about `DEVELOPER_GUIDE.md`, because it has info for both contributing to the plugin and for extending it in your own plugin. The flow starts out talking about contributing back, then there's a Plugin API section for extending, then there's git branching instructions (which are for contributing back and not doing your own extension).

Having said that, I wouldn't be opposed to rolling with what's been done and improving it as we go.

theaminulai, line 1: You should take a look at this issue:  
[GitHub - WordPress/ai Issues #2](https://github.com/WordPress/ai/issues/2).

theaminulai (changes requested):

readme.txt
jeffpaul, line 1: For now, I think we should spin this file out as its own PR as I think we're good to get some sort of base `readme.txt` file merged in.

jeffpaul (changes requested):

jeffpaul:
For the non-readme.txt changes here I'd defer to @karmatosed @felixarntz desire to mimic what's been done with Performance Lab primarily and if no example there then what Gutenberg has in place.

jeffpaul (approved):
I generally concur with the feedback here that these docs should be a bit more concise, either spin content out to other docs where its not directly related to the specific docs' focus or remove that content as unhelpful, but for now am going to proceed with merging this and #53 in and will follow-up with @karmatosed after about getting some help on iterating on the docs.

Issue #41, opened 2025-10-10
Add workflow to build plugin and add WordPress Playground preview link


jeffpaul:
### Description

Create a GitHub Action that runs on every PR that will install dependencies, build the plugin from the PR branch, package it, and post a comment on the PR with a link to a WordPress Playground instance that loads the built plugin for an easy click-to-test scenario for contributors.

### Goals

* Trigger on `pull_request` events for `open`, `synchronize`, and `reopen`
* Install and cache dependencies (Composer and Node)
* Run lint and build steps
* Package the plugin into a ZIP artifact suitable for WordPress installation
* Spin up a Playground preview that loads the built plugin
* Post a PR comment with:
* a Playground link that auto-installs or preloads the plugin
  * build metadata (commit SHA, build time)
  * links to build artifacts and logs

### Scope

* One reusable workflow file in `.github/workflows`
* Minimal secrets, prefer public Playground tooling that does not require credentials
* Support re-runs and PR updates without leaving stale comments

### Out of scope

* Release packaging for WPORG
* E2E test authoring beyond a sanity check that WordPress boots and the plugin activates

### Requirements

* Workflow runs on ubuntu-latest
* Uses `actions/cache` for Composer and Node to speed up builds
* Produces a single plugin ZIP uploaded as a workflow artifact
* Updates the PR comment with the Playground link on each new commit to the PR
* Handles forks safely.  If a forked PR cannot use write permissions, the workflow should skip comment posting and still upload artifacts

### Implementation notes

* Typical steps
  * checkout repo
  * set up PHP and Composer
  * set up Node and npm or yarn
  * composer install with no dev if that matches your packaging rules
  * npm ci and npm run build
  * create plugin zip from the correct root directory
  * upload artifact
* Playground link generation options
  * Use official WordPress Playground tooling that can generate a share link preloading a plugin ZIP from the workflow artifact or from a temporary URL
  * If a direct share link is not possible, publish the ZIP to a temporary URL via GitHub Pages for the repo or via the workflow’s artifact URL, then construct a Playground query that installs from that URL
* PR comment format should be stable and updatable.  Use a comment identifier so the job edits the existing comment rather than creating duplicates

### Security considerations

* Do not expose secrets in logs or PR comments
* Restrict permissions in the workflow to the minimum needed
* Avoid running arbitrary commands from PR content when generating the preview

### Acceptance criteria

* On a PR, the workflow builds the plugin and uploads a ZIP artifact
[...]

PR #49, opened 2025-10-23
Add in various test GitHub Action workflows

dkotter:
### Description of the Change

This PR adds in the following GitHub Action workflows:

- Run PHPCS
- Run PHPStan
- Run PHPUnit
- Run JS linting (disabled for now)

Also adds in / updates all necessary config files for this workflows and adds in / updates all necessary dependencies.

In addition, there were a handful of failures the above workflows flagged that have now been fixed up, including some PHPCS issues, PHPStan issues and some failing PHPUnit tests.

And last, tries to fix a couple issues with the `Pull Request Comments` workflow that was merged in #41. This is unfortunately hard to test until these files are merged in as GitHub won't run these workflows from a fork it seems. But I'm hoping these are the last couple changes required there.

> [!Note]
> The above workflows were all copied from either the Abilities API repo or the MCP Adapter repo. I modified them slightly to work here but for the most part, left things alone. So if there are any changes we would like to make to the workflows or to the configs they use, that can be discussed.

### How to test the Change

Ensure all workflows are running and passing on this PR

### Test using WordPress Playground

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%20Request%20Comments&artifact=ai&pr=49%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2049%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).

### DIFF
--- a/.github/workflows/build-plugin-zip.yml
+++ b/.github/workflows/build-plugin-zip.yml
@@ -27,7 +27,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
-          ref: ${{ github.ref }}
           show-progress: ${{ runner.debug == '1' && 'true' || 'false' }}
           persist-credentials: false
 
--- a/.github/workflows/pull-request-comments.yml
+++ b/.github/workflows/pull-request-comments.yml
@@ -87,7 +87,7 @@ jobs:
                   path: '/tmp/pr/pr.zip',
                   data: {
                     resource: 'url',
-                    url: `/plugin-proxy.php?org=WordPress&repo=ai&workflow=Build%20Plugin%20Zip&artifact=ai&pr=${prNumber}`,
+                    url: `/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%20Request%20Comments&artifact=ai&pr=${prNumber}`,
                     caption: `Downloading AI Experiments PR ${prNumber}`,
                   },
                 },
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,75 @@
+name: Test
+
+on:
+    workflow_dispatch:
+    push:
+        branches:
+            - trunk
+    pull_request:
+        types:
+            - opened
+            - synchronize
+            - ready_for_review
+
+# Cancels all previous workflow runs for pull requests that have not completed.
+concurrency:
+    # The concurrency group contains the workflow name and the branch name for pull requests
+    # or the commit hash for any other events.
+    group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
+    cancel-in-progress: true
+
+jobs:
+    # Runs the PHP coding standards checks.
+    #
+    # Violations are reported inline with annotations.
+    #
+    # Performs the following steps:
+    # - Checks out the repository.
+    # - Sets up PHP.
+    # - Configures caching for PHPCS scans.
+    # - Installs Composer dependencies.
+    # - Runs PHPCS on the full codebase.
+    # - Generate a report for displaying issues as pull request annotations.
+    phpcs:
+        name: Run PHPCS coding standards checks
+        runs-on: ubuntu-24.04
+        permissions:
+            contents: read
+        timeout-minutes: 20
+
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+              with:
+                  show-progress: ${{ runner.debug == '1' && 'true' || 'false' }}
+                  persist-credentials: false
+
+            - name: Set up PHP
+              uses: shivammathur/setup-php@ec406be512d7077f68eed36e63f4d91bc006edc4 # v2.35.4
+              with:
+                  php-version: '8.3'
+                  coverage: none
+                  tools: cs2pr
+
+            # This date is used to ensure that the PHPCS cache is cleared at least once every week.
+            # http://man7.org/linux/man-pages/man1/date.1.html
+            - name: "Get last Monday's date"
+              id: get-date
+              run: echo "date=$(/bin/date -u --date='last Mon' "+%F")" >> "$GITHUB_OUTPUT"
+
+            - name: Cache PHPCS scan cache
+              uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+              with:
+                  path: tests/_output/phpcs-cache.json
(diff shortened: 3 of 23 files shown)


.github/workflows/build-plugin-zip.yml
```diff
@@ -27,7 +27,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
-          ref: ${{ github.ref }}
```
dkotter, line 30: I think with this in place the zip we build is always from `trunk` instead of the branch. Can't test that without merging this in due to this PR being opened from a fork but if you download the artifact from this PR, you'll see none of the changes made here reflected in that (and this breaks our Playground integration as we had a requirement of WP 6.9, which isn't out yet)

.github/workflows/pull-request-comments.yml
```diff
@@ -87,7 +87,7 @@ jobs:
                   path: '/tmp/pr/pr.zip',
                   data: {
                     resource: 'url',
-                    url: `/plugin-proxy.php?org=WordPress&repo=ai&workflow=Build%20Plugin%20Zip&artifact=ai&pr=${prNumber}`,
+                    url: `/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%20Request%20Comments&artifact=ai&pr=${prNumber}`,
```
dkotter, line 90: Because we call our `Build Plugin Zip` workflow from within this `Pull Request Comments` workflow, looks like the workflow name that shows in the API is `Pull Request Comments`, so updating that to ensure our Playground integration will work (though will also require the change referenced [here](https://github.com/WordPress/ai/pull/49#discussion_r2457543183))

.github/workflows/test.yml
```diff
@@ -0,0 +1,285 @@
...
+    # - Installs NPM dependencies.
+    # - Runs Prettier formatting check.
+    # - Runs JavaScript linting.
+    jslint:
+        name: Run JavaScript code quality checks
+        runs-on: ubuntu-24.04
+        permissions:
+            contents: read
+        timeout-minutes: 20
+        if: false  # Temporarily disabled
```
dkotter, line 263: Disabled for now as we have no JS files and so this workflow will fail

ai.php
```diff
@@ -11,7 +11,7 @@
  * Plugin Name:       AI
  * Plugin URI:        https://github.com/WordPress/ai
  * Description:       Experimental AI features for WordPress
- * Requires at least: 6.9
+ * Requires at least: 6.8
```
dkotter, line 14: In #24 we updated in one spot the minimum WP version to be 6.8 but didn't update that here, so taking care of that now

includes/Features/Example_Feature/README.md
```diff
@@ -33,17 +33,10 @@ Use the feature-specific filter:
 add_filter( 'ai_feature_example-feature_enabled', '__return_false' );
 ```
 
-Or use the generic filter:
+Or use the generic filter to disable all features:
 
 ```php
-add_filter(
-	'ai_feature_enabled',
```
dkotter, line 40: Looks like at some point the name of this filter was changed and no longer allows you to disable individual features. Didn't look through commit/comment history but wondering if that was intentional?

tests/Integration/Includes/Features/Example_Feature/Example_FeatureTest.php
```diff
@@ -107,7 +107,7 @@ public function test_modify_title_with_debug_true() {
 
 		$modified_title = apply_filters( 'document_title_parts', $title_parts );
 
-		$this->assertEquals( 'My Site [AI]', $modified_title['site'] );
+		$this->assertEquals( 'My Site [AI] [AI]', $modified_title['site'] );
```
dkotter: This fixes the test but not sure why the `[AI]` part gets added twice. Seems like a likely bug in our Example Feature. Need to look closer at that

includes/Features/Example_Feature/README.md
dkotter: To be more clear, you can still disable an individual feature using the filter `ai_feature_{$this->id}_enabled`. But `ai_feature_enabled` no longer exists, seems to have been replaced with `ai_features_enabled` (`feature` vs `features`) and that filter doesn't pass the individual feature to it, so it's a universal disablement.

Ref34t: @dkotter indeed, the `ai_feature_enabled` filter was refactored to the `features` version and now lives in the loader, due to finding that having two different filters was not good enough.

.github/workflows/test.yml
jeffpaul: Might be ok to enable unless we feel confident we'll enable once we get JS in here?

ai.php
jeffpaul, line 14: I'll open a separate issue for us to handle the 6.9 compat testing + bump once core hits RC1

jeffpaul (approved):

ai.php
jeffpaul: See #54

.github/workflows/test.yml
dkotter: My only concern with enabling it is it causes this workflow to always fail even though nothing is actually wrong. I have this noted to come back to so no concern with forgetting to enable this but happy to turn this on now if we want

tests/Integration/Includes/Features/Example_Feature/Example_FeatureTest.php
dkotter: Fixed in https://github.com/WordPress/ai/pull/49/commits/a5c7acc111e6d9bd7938d21e44dca3e7f7dd05cb. Features were being initialized twice when running tests, resulting in the feature being applied twice (and thus seeing the double `[AI]` label). Seems to just be an issue in the tests but if anyone else wants to take a look, may be worth a second set of eyes to ensure there isn't a bug in our feature loader that would allow features to be loaded twice in normal circumstances

PR #50, opened 2025-10-24
Pull in the various AI dependencies we'll be using

dkotter:
### Description of the Change

The AI Experiments plugin will be building on top of the other AI building blocks, namely the Abilities API, the MCP Adapter and the WordPress/PHP AI Client.

This PR brings all of those in as composer dependencies, though worth a discussion here on if that's the way we want to handle things. We could instead require those to be installed as separate plugins (probably using the `Requires Plugins` header) or even just rely on what ends up in WordPress 6.9 (at least for the Abilities API).

My opinion here is that by bundling, we make the user experience nicer (don't require multiple additional plugins to be installed) and we control what versions we're relying on (allowing us to use newer things that may not end up in WordPress 6.9). But open to other thoughts.

At the moment, this is what's pulled in:

- v0.3.0 of the Abilities API
- The `trunk` version of MCP Adapter, as the only current release (v0.1.0) requires PHP 8.1+
- The `trunk` version of WP AI Client, noting I debated pulling in [this branch](https://github.com/WordPress/wp-ai-client/pull/6) so we could take advantage of the settings that introduces, but hoping that can be merged in soon
    - Also noting since this plugin isn't recognized by packagist, had to set the `repositories` config to pull this in

In addition, the MCP Adapter suggests using the Jetpack Autoloader so also pulling that in and switching to loading that over the default composer autoloader.

### How to test the Change

1. Checkout this branch locally
2. Run `composer install`
5. Activate the plugin. Ensure no errors occur. Note that we currently have a bug where our minimum WordPress version is set to 6.9, which isn't out yet. This means WordPress won't allow you to activate the plugin. Can either force this using WP-CLI or temporarily modify `ai.php` to change 6.9 to 6.8 (this is fixed in #49)

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Build%2520Plugin%2520Zip&artifact=ai&pr=50%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2050%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,8 +4,9 @@
   "type": "wordpress-plugin",
   "license": "GPL-2.0-or-later",
   "require": {
-    "php": ">=7.4",
+    "automattic/jetpack-autoloader": "^5.0",
     "ext-json": "*",
+    "php": ">=7.4",
     "wordpress/abilities-api": "^0.3.0",
     "wordpress/mcp-adapter": "dev-trunk"
   },
@@ -30,6 +31,7 @@
   },
   "config": {
     "allow-plugins": {
+      "automattic/jetpack-autoloader": true,
       "dealerdirect/phpcodesniffer-composer-installer": true
     },
     "platform": {
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,8 @@
   "require": {
     "php": ">=7.4",
     "ext-json": "*",
-    "wordpress/abilities-api": "^0.3.0"
+    "wordpress/abilities-api": "^0.3.0",
+    "wordpress/mcp-adapter": "dev-trunk"
   },
   "require-dev": {
     "wp-coding-standards/wpcs": "^3.2.0",
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "automattic/jetpack-autoloader": "^5.0",
     "ext-json": "*",
     "php": ">=7.4",
-    "wordpress/abilities-api": "^0.3.0",
+    "wordpress/abilities-api": "^0.4.0-rc",
     "wordpress/mcp-adapter": "dev-trunk",
     "wordpress/wp-ai-client": "dev-trunk"
   },
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,8 @@
   "license": "GPL-2.0-or-later",
   "require": {
     "php": ">=7.4",
-    "ext-json": "*"
+    "ext-json": "*",
+    "wordpress/abilities-api": "^0.3.0"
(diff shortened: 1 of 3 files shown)


jeffpaul (approved):

gziolo:
https://packagist.org/packages/wordpress/abilities-api#v0.4.0-rc is available now. It contains all the latest and greatest from WordPress 6.9 beta. It will be promoted to the stable release as soon as we confirm it integrates seamlessly with plugins.

dkotter:
> https://packagist.org/packages/wordpress/abilities-api#v0.4.0-rc is available now. It contains all the latest and greatest from WordPress 6.9 beta. It will be promoted to the stable release as soon as we confirm it integrates seamlessly with plugins.

Thanks for the heads up on that. I've updated this PR to pull the 0.4.0-rc version, noting all of these versions will continually be updating as needed going forward but might as well use the latest and greatest from the start

composer.json
```diff
@@ -4,8 +4,12 @@
   "type": "wordpress-plugin",
   "license": "GPL-2.0-or-later",
   "require": {
+    "automattic/jetpack-autoloader": "^5.0",
```
felixarntz, line 7: Just chiming in here with a thought (not reason to change anything at all, just something to think about): An alternative to trying to only load one version of the dependency would be to prefix the bundled dependencies.

**Benefit:** More robust, as it prevents conflicts due to minor differences between different versions if one plugin used an API slightly differently.
**Drawback:** Code will be "unnecessarily" (at least _almost_ unnecessarily) loaded multiple times, and the initial setup is a little more tricky.

It's certainly a tradeoff. But I'd be inclined to eventually document both as reasonable alternative paths for pulling in and loading those dependencies.

PR #51, opened 2025-10-26
Admin Settings Screen Implementation

### LINKED ISSUE #25: Create a basic admin settings screen
### Description

Introduce a basic admin interface that provides a single toggle to enable or disable AI experiments.  This screen will serve as the entry point for configuring and discovering AI-powered capabilities.

### Goals

- [ ] Add a new “AI Experiments” submenu item under “Settings” the admin.
- [ ] Display a global toggle labeled “Enable Experiments” with persistent option storage.
- [ ] Provide a modular system for registering settings sections so that each Experiment can surface its own configuration independently.
- [ ] Use the same encapsulation model defined during the scaffolding stage, ensuring Experiment-specific settings are contained within that Experiment’s own component or namespace.
- [ ] Follow modern WordPress design patterns and block/site editor UI conventions.

### Acceptance Criteria

- Admin screen appears under Settings and loads without errors.
- The toggle persists correctly using the WordPress options API.
- Experiments can register their own settings panels without directly modifying the core settings page.
- The UI reflects block-based styling consistent with modern WordPress admin patterns.

Ref34t:
## Summary

Builds a complete admin settings screen with React UI for managing AI experimental features.

## Key Components

- **Admin Settings Page** - React-based UI using WordPress components
- **Settings Toggle** - Global on/off switch for experimental features
- **Feature Toggles** - Per-feature enable/disable functionality
- **Settings Registry** - Allows features to register their own settings sections
- **Settings Service** - Coordinates all settings components
- **REST API** - Backend data layer for settings persistence

## Architecture

- Direct instantiation in bootstrap
- Filter-based dependency injection for Feature_Toggles service
- Features receive Settings_Registry as parameter
- Render callbacks accept Settings_Toggle directly

## Feature Development Pattern

Features extend Abstract_Feature and follow these rules:

1. Implement `load_feature_metadata()` returning id, label, description
2. Always register settings sections via `ai_register_settings_sections` hook
3. Check `is_enabled()` before registering functional hooks
4. Pass services as parameters
5. Use `Provides_Settings_Section` trait for settings panels

See Example_Feature for complete reference implementation.

## Files Added

- Admin settings infrastructure (9 PHP classes)
- React/TypeScript UI (src/ directory)
- Feature settings trait (Provides_Settings_Section)
- Comprehensive test suite

Resolves #25


includes/bootstrap.php
```diff
@@ -159,6 +164,9 @@ function load(): void {
 
 	// Hook feature initialization to init.
 	add_action( 'init', __NAMESPACE__ . '\initialize_features' );
+	add_action( 'admin_init', __NAMESPACE__ . '\initialize_admin' );
```
mindctrl: `admin_init` doesn't work here because when `Settings_Page::register()` runs, `admin_menu` has already fired. I tested with it attached to `init` and it works fine.

```suggestion
	add_action( 'init', __NAMESPACE__ . '\initialize_admin' );
```

Ref34t: @mindctrl good catch handled it

jeffpaul:
@Ref34t is this PR ready for review or is it still in-progress?

Ref34t:
@jeffpaul it's ready

.github/workflows/pull-request-comments.yml
dkotter, line 1: Looking at the changes here, it seems to avoid permission issues we're adding a comment instead of updating the PR description. I have a separate PR (https://github.com/WordPress/ai/pull/65) that fixes permission issues. I'd suggest we remove these changes from this PR and can continue discussion on that other PR to keep things more focused here

.gitignore
```diff
@@ -13,8 +13,8 @@ node_modules/
 .DS_Store
 
 # Build
-/build/
 /dist/
+/build/
```
dkotter, line 16: Seems this change just swaps the position of these two, any reason for that?

composer.json
```diff
@@ -77,7 +77,7 @@
   "scripts": {
     "format": "phpcbf --standard=phpcs.xml.dist",
     "lint": "phpcs --standard=phpcs.xml.dist",
-    "phpstan": "phpstan analyse --memory-limit=1G",
+    "stan": "phpstan analyse --memory-limit=1G",
```
dkotter: Is there a reason for this naming change?

includes/Admin/Admin_Settings_Page.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Registers the submenu item under Settings.
+	 *
+	 * @since 0.1.0
+	 */
+	public function register_menu(): void {
+		$this->hook_suffix = add_options_page(
+			__( 'AI Experiments', 'ai' ),
+			__( 'AI Experiments', 'ai' ),
```
dkotter: Might be worth setting a constant for this instead of repeating it

src/index.tsx
dkotter, line 1: Since we almost certainly will have other JS/TS files to support each individual feature, wondering if all this code should be nested under `src/admin` or `src/settings` or similar just to make it clear as we start introducing other code?

dkotter (changes requested):

dkotter (commented):
In testing this out, I think we're likely to have some iterations on the actual design of the settings page though overall it looks good to me. My one note that I'd suggest we fix in this PR is the first time I enable/disable something, a notice is shown. If I don't close that message out and then enable/disable again, the page content jumps around. I think the issue here is the notice is removed then added back so quickly and causes things to shift up and down.

.github/workflows/pull-request-comments.yml
dkotter: Note I've brought the code changes here that ensure we have a PR description over to #65, though I didn't bring over the changes that add a comment instead of updating the PR description

includes/Features/Example_Feature/Example_Feature.php
```diff
@@ -36,11 +42,63 @@ protected function load_feature_metadata(): array {
 	 * @since 0.1.0
 	 */
 	public function register(): void {
+		// Always register settings sections so the feature appears in admin.
+		add_action(
+			'ai_register_settings_sections',
+			array( $this, 'register_settings_sections' )
+		);
```
dkotter, line 49: Thinking about this a bit more, wondering if it would be better to have this handled within the `Abstract_Feature` class instead of requiring each individual feature class to have this exact same code?

So basically the abstract class would run this and then each extending class would have to implement their own `register_settings_sections` method?

includes/Features/Example_Feature/Example_Feature.php
```diff
@@ -36,11 +42,63 @@ protected function load_feature_metadata(): array {
...
+
+		// Only register functional hooks if the feature is enabled.
+		if ( ! $this->is_enabled() ) {
+			return;
+		}
```
dkotter: Similar to the above, right now each feature class would need to have these exact same lines of code. Instead of that duplication, wondering if we introduce a new method in the parent class that is only called if `is_enabled` returns true. So if you need to register functionality that is always there, even if the feature is disabled, you could do that in the `register` method (or we could rename that to something else) and if you want to run functionality only if the feature is enabled, you could do that in a new method, something like `register_when_enabled`

.gitignore
jeffpaul: I think there was a commit that removed the build directory and then in a later commit it was added it back, just in a slightly different location

includes/Admin/Settings_Page_Assets.php
```diff
@@ -0,0 +1,131 @@
...
+ * Manages scripts and styles for the admin settings page.
+ *
+ * @since 0.1.0
+ */
+class Settings_Page_Assets {
+	private const SCRIPT_FILE = 'build/index.js';
+
+	private const STYLE_FILE = 'build/style-index.css';
+
+	private const STYLE_FILE_PATH = __DIR__ . '/../../build/style-index.css';
```
mindctrl: In places where we're traversing the path upwards like this, I wonder if we should use `WP_AI_DIR` instead?

```suggestion
	private const STYLE_FILE_PATH = WP_AI_DIR . '/build/style-index.css';
```

includes/Abstracts/Abstract_Feature.php
```diff
@@ -132,12 +200,19 @@ public function get_description(): string {
...
 	 *
 	 * @return bool True if enabled, false otherwise.
 	 */
 	final public function is_enabled(): bool {
-		$enabled = $this->enabled;
+		if ( null !== $this->feature_toggles ) {
+			$enabled = $this->feature_toggles->is_feature_enabled( $this->id );
+		} else {
+			$enabled = $this->enabled;
+		}
```
mindctrl, line 216: I'm a bit confused on the enablement checks. We have:

- `$this->feature_toggles->is_feature_enabled( $this->id );`
- `$this->enabled`
- `apply_filters( "ai_feature_{$this->id}_enabled", $enabled )`

It's hard to know what the source of truth is. Do have needs for this complexity already? I wonder if the UI could store the enabled/disabled options somewhere and just use those with the existing filter and avoid all the new classes? Admittedly I don't have a full picture of the next steps and/or plans for this.

includes/Admin/Settings_Page_Assets.php
```diff
@@ -0,0 +1,131 @@
...
+ */
+
+namespace WordPress\AI\Admin;
+
+/**
+ * Manages scripts and styles for the admin settings page.
+ *
+ * @since 0.1.0
+ */
+class Settings_Page_Assets {
```
dkotter: Thinking about this more, I almost think we're better off having a generic assets loader as we'll need to load CSS and JS files for all other features, not just for the settings screen. I'd say we should repurpose this to support all asset loading

dkotter: I've started work on the title generation feature and here's an asset loader I've set up for that (still a WIP): https://github.com/dkotter/ai/blob/feature/title-generation-ui/includes/Asset_Loader.php. This could be used here as well

src/admin/settings/app.tsx
justlevine, line 1: Can we kebab-case these typescript, in line with the gutenberg repo?

includes/Features/Example_Feature/Example_Feature.php
justlevine: Conversely, what are we gaining by coupling admin UX into the entrypoint class? AFAIK
1. not all modules need to expose UX/UI/settings beyond being enabled.
2. most modules should execute outside of the admin UI/XI lifecycle. 

I'm thinking dependency injection is doing more harm than good here, and instead, _if_ a specific module needs a special settings and an admin screen, they should just create a co-located `Admin.php` class, and then `add_action( Admin::class, 'register' )` which could self-instantiate, or use DI via your trait, or an `Abstract_Admin` class, or whatever.

includes/Features/Example_Feature/Example_Feature.php
justlevine: What sort of module-specific functionality would we want/expect to hook into WordPress if a plugin was disabled?

Keeping in mind that we control the source code, what if we went with a broader naming that's more flexible, along the lines of (pseudocode)

```php
// Abstract_Feature.php

/**
 * This is what get's called by Abstract_Loader instead of `::register()`.
 * Overload it if you want to do other things
 */
public function init(): void {
  if ( ! $this->is_enabled() }
    return;
  }
  $this->register_hooks();
}

// Where the child class defines their hooks.
abstract protected function register_hooks(): void;
```

This then lets us
```php
// MY_Feature.php

/**
 * {@inheritDoc
 */
protected function register_hooks(): void {
  add_action( ...... );
 }
 
 /**
  * {@inheritDoc}
  * Straightforward and flexible if we need to:
  */
  public function init(): void {
    // do something always, e.g.
    $this->module_specific_dep_check();

    parent::init();

    // do something only if enabled, e.g. a DI'd admin screen class
    $this->admin->init();
  }
```

includes/Feature_Loader.php
```diff
@@ -95,82 +93,46 @@ public function register_default_features(): void {
...
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param \WordPress\AI\Admin\Settings\Feature_Toggles|null $feature_toggles Feature toggles service or null.
+		 */
+		$feature_toggles = apply_filters( 'ai_feature_toggles_service', null );
+
+		// Instantiate default features directly.
+		$features = array(
+			new \WordPress\AI\Features\Example_Feature\Example_Feature( $feature_toggles ),
```
justlevine: 1. Per https://github.com/WordPress/ai/pull/51/files#r2505914125, I'm wondering if DI is beneficial here or unnecessary complexity.

2. If it is justified, could we maybe pass the `class-string<\WordPress\AI\Admin\Settings\Feature_Toggles>` and then only instantiate the class when it's used? IMO would help with future proofing to avoid unnecessary initializing things if a module is disabled.

composer.json
Ref34t: Just to have the same way as the rest. can be reverted if it doesn't make sense for you

includes/Abstracts/Abstract_Feature.php
Ref34t: Thanks for calling that out! I’ve flattened the enablement flow so there’s now a single source of truth:

- `Abstract_Feature::is_enabled()` now starts from the feature’s  `is_enabled_by_default()` value, hands it through the (lazy-loaded) `Feature_Toggles` service, and finally runs the existing `ai_feature_{$this->id}_enabled` filter. The old `$this->enabled` property is gone, so every call site reads the same value.

- The React UI still persists toggles to `wp_ai_feature_toggles`, and `Settings_Payload_Builder` feeds those back into the payload, so admins see the exact stored state.

includes/Features/Example_Feature/Example_Feature.php
Ref34t: I've implemented exactly that split: 

`Abstract_Feature::register()` is now final and always (1) calls register_shared_hooks() for any “always-on” work (e.g. settings sections) before (2) checking `is_enabled()` and then calling `register_enabled_hooks()` for the functional bits. 

Concrete features only override those two protected methods, so there’s no repeated boilerplate or tight coupling to the admin UX unless a module explicitly opts in by  hooking `ai_register_settings_sections`.

 The toggle service itself is still lazily resolved, so modules that never touch admin classes don’t pay for them. 

Happy to iterate further if you’d like to see a different way, but this centralization got rid of the duplication you pointed out while keeping admin pieces optional.

includes/Features/Example_Feature/Example_Feature.php
Ref34t: I’ve folded that pattern into Abstract_Feature now. 

`register()` is final and handles the enablement gate automatically:

it calls `register_shared_hooks()` first (for anything that must always run, like settings sections or dependency checks), then returns immediately if `is_enabled()` is false, otherwise it calls `register_enabled_hooks()` for the functional wiring. 

Concrete features just override those two protected methods, so there’s no need to repeat the boilerplate or worry about coupling admin UI into every feature—the shared phase is optional, the enabled phase only runs when the toggle/filter says so, and the loader still calls a single entrypoint.

Happy to iterate on naming, but this gives us the `register_when_enabled` behavior you outlined without breaking anything

includes/Feature_Loader.php
Ref34t: Resolved by making the toggle service lazy: `Abstract_Feature` now accepts either a concrete `Feature_Toggles` instance or a callable factory, and `Feature_Loader::get_default_features()` wires the `ai_feature_toggles_service` filter to allow returning a `class-string<Feature_Toggles>`.

When a module actually calls `is_enabled()`, the base class resolves the factory on demand (instantiating the class once and caching it), so features pay for the dependency only if they use it. This keeps DI available for testing/extensibility, but avoids eager initialization when modules are disabled.

src/admin/settings/app.tsx
Ref34t: resolved

src/index.tsx
Ref34t: that's right. resolved

.gitignore
Ref34t: resolved

.github/workflows/pull-request-comments.yml
Ref34t: reverted here.

docs/DEVELOPER_GUIDE.md
```diff
@@ -71,7 +71,7 @@ ai/
 │       └── Example_Feature/          # Each feature in own directory
 │           ├── Example_Feature.php
 │           └── README.md
-├── admin/                            # Admin interface (planned)
+├── admin/                            # Admin settings services, controllers, assets
```
justlevine: Is this supposed to be `includes/Admin`?

Would be good to list the subnamespaces and primary files too.

includes/Admin/Admin_Settings_Page.php
Ref34t: resolved

includes/Admin/Settings_Page_Assets.php
Ref34t: resolved

src/admin/settings/components/feature-section.tsx
```diff
@@ -0,0 +1,57 @@
...
+					<div>
+						<h3 className="ai-experiments-settings-app__card-title">
+							{section.title}
+						</h3>
+						{section.description ? (
+							<p className="ai-experiments-settings-app__card-description">
+								{section.description}
+							</p>
+						) : null}
+					</div>
```
mindctrl, line 38: Just noting that the design of these controls doesn't allow for long descriptions that might detail requirements, have an HTML list of items, or anything HTML. 

Screenshot
Image


The fallback method allows for much more control over what is displayed. Looking at the Content Translator demo in the screenshot... without JS, using the `render_settings_section()` fallback way, you would be able to do something like this example code to explain what the feature does:
```
		<p>
			<?php esc_html_e( 'When enabled, this feature provides a block that allows site visitors to translate page content into multiple languages using AI.', 'jp-wp-ai' ); ?>
		</p>
		<p>
			<strong><?php esc_html_e( 'Supported Languages:', 'jp-wp-ai' ); ?></strong>
		</p>
		<ul style="list-style: disc; margin-left: 20px;">
			<?php foreach ( self::SUPPORTED_LANGUAGES as $name ) : ?>
				<li><?php echo esc_html( $name ); ?></li>
			<?php endforeach; ?>
		</ul>
		<p>
			<em><?php esc_html_e( 'Add the Content Translator block to any post or page to enable front-end translation.', 'jp-wp-ai' ); ?></em>
		</p>
```

I can actually see it flash quickly before the React app overwrites the screen. As far as I can tell, it's not possible to do any extended description with the existing React app in its current state. 

I think any experiment more than an example is going to need more space to explain things.

jeffpaul:
@Jameswlepage @JasonTheAdams @felixarntz @swissspidy @dkotter with #61 resolved and merged in, getting y'alls input / reviews / approvals on this PR will help unblock feature development in the plugin; so if you would please chime in here that would be great, thanks!

jeffpaul:
@Ref34t note there are some merge conflicts to review, likely as a result of #61 getting merged in

composer.json
swissspidy: +1 to revert

includes/Features/Example_Feature/Example_Feature.php
justlevine: I definitely like it a lot more - thanks! 

I think I need to review this more closely now that #61 is merged, since my gut is telling me that there's still room to iterate, but so long as the bulk of work stays isolated to the `Admin` namespace I don't see any more real concerns.

swissspidy:
Screenshot 2025-11-13 at 22 18 31

This... is a huge amount of code for something as simple as a settings page 😳 At first glance there's a ton of unused boilerplate code with methods that aren't used except in tests, or reimplement existing core functions such as `rest_sanitize_boolean`. I'd ask an agent to simplify this and then perhaps let Copilot do the initial round of review.

docs/DEVELOPER_GUIDE.md
Ref34t: resolved

Ref34t:
Streamlined and refractored the boilerplate to minimize the size of the PR without losing any functionality, also resolved merge conflicts

docs/DEVELOPER_GUIDE.md
```diff
@@ -51,250 +51,25 @@ wp plugin activate ai
 
 ## Architecture Overview
 
-The plugin follows a modular, feature-based architecture:
```
justlevine, line 54: why replace the tree?

JasonTheAdams:
I'm going to close this PR in favor of #84. This PR does a good job of going after a formal settings registration system, but we're simply not at the point where something like this is necessary. We need something _really_ simple right now that we can improve on later, as we get a clearer sense of what we want the settings to become.

Thank you for your effort on this, @Ref34t! It's sincerely appreciated and helped prod the conversation.

PR #46, opened 2025-10-16
Add new GitHub Action workflows to support WordPress Playground previews

dkotter:
### Description of the Change

As detailed in #41, this PR adds two new GitHub Action workflows:

1. Build the plugin and attach as an artifact
2. Add a link to WordPress Playground to the PR description, allowing for easier testing of individual PRs

The way this works is the following:

1. When a PR is opened, this will trigger the `Pull Request Comments` workflow
2. This workflow will then trigger the `Build Plugin Zip` workflow. This workflow builds the plugin and uploads it as an artifact to the PR
3. If this succeeds, the `Pull Request Comments` workflow will continue.
4. It first checks the contents of the PR description to see if a preview link already exists. If it does, the process stops. If it doesn't, it then continues
5. We then build up the blueprint data we need and add that to the WordPress Playground URL. This is then added to the PR description
6. In theory, when this link is clicked, WordPress Playground has code in place that will download the artifact and install that. This only works for repos in the `WordPress` organization, so this part hasn't been tested yet

Closes #41 

> [!Note]
> This PR contains `.gitignore`, `composer.json`, `composer.lock`, `package.json` and `package-lock.json` files that were copied over from #38, since that PR isn't merged yet. A few minor changes were made to `package.json` to support the build process but everything else matches that PR

### How to test the Change

I was hoping the workflow would run properly on this PR but I think do to this being opened from a fork, it doesn't work.

All of this has been tested on my forked repo, though as noted there in the final step, the Playground link doesn't work properly as my repo isn't within a supported organization. But if desired, can see how things worked there:

- [Successful run](https://github.com/dkotter/ai/actions/runs/18544387650) that added the comment
- [Successful run](https://github.com/dkotter/ai/actions/runs/18544389604) that didn't add a comment because it already existed
- [PR](https://github.com/dkotter/ai/pull/2) that had the comment added

Not sure there's a better way to test without merging this in :(

PR #56, opened 2025-10-28
Change the event the PR Comment workflow runs on

dkotter:
### Description of the Change

In #46 we added a GitHub Action workflow to add a Playground testing link to each PR description. There were a few issues with that and we fixed those in #49. In testing coming out of that merge, noticing a few more problems:

1. The branch used to build the zip was still showing as `trunk`, even though we want it to build from the PR branch
2. The final zip didn't include the `vendor` directory

For the first, in looking at the difference between using `pull_request` and `pull_request_target` as the triggering event, `pull_request_target` (which we were using) will always [reference the PR base branch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) (typically `trunk` in our case) instead of the last commit to the PR branch itself.

I've adjusted that in this PR to use `pull_request` instead. I believe this is the proper event to use, especially based on this comment from GitHub:

> This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request

For the second issue, I've adjusted our build script to ensure the `vendor` directory is included there but only include non-dev dependencies.

### How to test the Change

Ensure the Playground link gets added properly to this PR description and that link works

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=56%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2056%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).

### DIFF
--- a/.github/workflows/pull-request-comments.yml
+++ b/.github/workflows/pull-request-comments.yml
@@ -1,7 +1,7 @@
 name: Pull Request Comments
 
 on:
-  pull_request_target:
+  pull_request:
     types: [ 'opened', 'synchronize', 'reopened', 'edited' ]
 
 # Cancels all previous workflow runs for pull requests that have not completed.
--- a/.github/workflows/pull-request-comments.yml
+++ b/.github/workflows/pull-request-comments.yml
@@ -8,7 +8,7 @@ on:
 concurrency:
   # The concurrency group contains the workflow name and the branch name for pull requests
   # or the commit hash for any other events.
-  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.head_ref || github.sha }}
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
 
 # Disable permissions for all available scopes by default.
 # Any needed permissions should be configured at the job level.
@@ -22,6 +22,7 @@ jobs:
     uses: ./.github/workflows/build-plugin-zip.yml
     permissions:
       contents: read
+      pull-requests: read
 
   # Leaves a comment on a pull request with a link to test the changes in a WordPress Playground instance.
   playground-details:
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request-comments.yml
```diff
@@ -22,6 +23,7 @@ jobs:
     uses: ./.github/workflows/build-plugin-zip.yml
     permissions:
       contents: read
+      pull-requests: read
```
jeffpaul, line 26: This doesn't need `write` in order to edit the PR description or comment?

jeffpaul (approved):
One small question on permissions, otherwise good to go

.github/workflows/pull-request-comments.yml
dkotter: So this controls the permissions for the Build Plugin Zip workflow, which isn't the one editing the PR description. That happens in the next step which does already have the `pull-requests: write` permission.

But in switching from `pull_request_target` to `pull_request`, was getting a permission error when the Build Plugin Zip workflow tried downloading the PR branch to build, so adding this was needed

PR #61, opened 2025-10-30
Add the Title Generation feature base

### LINKED ISSUE #10: Title Generation / Rewriting
### Summary:

### Why it matters:

### Sample Use Cases:

### Why this matters:

3. Prompt construction

### Overview:

### User Actions

4. User selects or edits

6. Failure states

### Trigger Condition

- Title field is empty

### System Behavior

### User Action

### Suggestion UI

### AI Input Requirements:

- Current post/page title

### AI Output Requirements:

- 2–3 suggested titles

### Model Agnosticism:

### Privacy & Config:

### Extensibility Notes:

- Hookable prompt pipeline

### Version 1

Image

### Version 2

Image

### LINKED ISSUE #29: Add UI controls in post editor for rewriting titles
### Description

### Goals

### Acceptance Criteria

dkotter:
## What?

Partially Closes #10

Sets up the base feature class for Title Generation, setting things up for future PRs. Note no UI is added here so the only things to test require direct HTTP requests.

## Why?

We want to add a feature for Title Generation and this is the start of that work. Decided instead of having everything be in a single PR I'd break things down into smaller chunks in order to hopefully speed up reviews and testing. This first chunk contains the initial feature class that registers an ability to generate one or more titles from content.

There will need to be followup PRs to handle (at least) the following:

- Functionality to actually generate titles
- A new Ability to set a title
- Add in needed settings including the ability to disable/enable the feature
- Add in the needed UI to actually run the feature

## How?

- Updates the Abilities API from `0.4.0-rc` to `0.4.0`
- Adds in a new `Title_Generation` class that extends our `Abstract_Feature` class. This feature is registered as one of our defaults and registers one ability right now
- Adds in a new `Abstract_Ability` class that can be used to create new WP Abilities
- Adds in a new `Title_Generation` abilities class that extends the `Abstract_Ability` class. This ability allows you to pass in either some content or a post ID (which we then pull content from) as well as the number of titles you want generated (currently limited from 1 to 10). The actual generation will be handled in another PR

## Testing Instructions

There is no UI yet but can be tested by making direct API requests to the various endpoints:

Make an authenticated `GET` request to the main abilities endpoint and ensure the Title Generation ability shows there
```bash
```

Make an authenticated `GET` request to the Title Generation endpoint and ensure information about the ability shows
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/title-generation' \
```

Make an authenticated `POST` request to the Title Generation run endpoint and ensure the data that is output matches the data you sent in
```bash
--data '{
    "input": {
        "content": "this is some content!",
    }
}'
```

```bash
--data '{
    "input": {
        "post_id": 1
    }
}'
```

```bash
--data '{
    "input": {
        "content": "this is some content!",
        "post_id": 1
    }
}'
```

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=61%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2061%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).


jeffpaul:
Noting that due to waiting on updates in the WP AI client package and the admin settings screen implementation it doesn't yet seem feasible to test this via Playground here.  As such, if @dkotter feels this is `done` then I'm fine with merging this in to unblock other work that branches off this PR.  We can plan to test things more fully in trunk as they come together. Though open to any code review here that folks have time for @felixarntz @swissspidy etc.

jeffpaul (approved):

includes/Abilities/Title_Generation.php
```diff
@@ -0,0 +1,172 @@
...
+
+use WP_Error;
+use WordPress\AI\Abstracts\Abstract_Ability;
+
+/**
+ * Title generation WordPress Ability.
+ *
+ * @since 0.1.0
+ */
+class Title_Generation extends Abstract_Ability {
```
felixarntz, line 20: I think this topic came up in the initial repo scaffolding PR, but I don't understand why we should use abilities for something like this.

My understanding is that abilities are primarily meant for AI models to use WordPress, not that abilities use AI.

As such, I consider title generation a "feature". There will already be an ability to set the post title at some point, and that's the kind of thing I think should be exposed as an ability. That way, an agent could already generate a post title because the generative aspect of it is naturally covered.

Now, I may be mistaken here, and maybe anything you can do in WordPress should be an ability. But thinking from an AI tooling perspective, it feels a bit odd to me to have a tool more or less just to call a generative model when those tools are typically used by generative models anyway.

It would be great to get other leads to chime in here with their perspective on this. @Jameswlepage @swissspidy @JasonTheAdams

includes/Contracts/Feature.php
```diff
@@ -48,6 +48,15 @@ public function get_label(): string;
...
 	public function get_description(): string;
 
+	/**
+	 * Gets the ability slug for the feature.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return string The ability slug for the feature.
+	 */
+	public function get_ability_slug(): string;
```
felixarntz: Related to the above (partially depends on that decision): Do we really know that every single feature will have an ability? And only one? Could there never be a feature that has multiple abilities attached?

felixarntz (changes requested):
@dkotter @jeffpaul From a code perspective this all looks good, but I have a more holistic technical concern that I'd like to get broader consensus on before moving forward.

includes/Contracts/Feature.php
dkotter: Good point though I imagined only a single ability per feature but there could be scenarios where that isn't the case. This could either be renamed (something like `get_base_ability_slug`) or maybe better to just move this to the ability class. I'll hold on changes here pending the large discussion though

includes/Abilities/Title_Generation.php
dkotter: I can easily change this back to where each individual feature registers it's own REST API endpoint that will be used when the feature is triggered (that's [initially](https://github.com/WordPress/ai/pull/61/commits/4a9c4d7448d7b03464c723f3774c998c4ab77f4c) what I started with).

That said, one of the reasons I ended up switching from that approach here is the Abilities API gives us all of that with basically the same code (registering, permission callbacks, execute callbacks, schema, etc). But in going with the ability approach, we accomplish a few additional things:

1. We showcase how the Abilities API can be used, which I believe is one of the goals of this plugin
2. We open up avenues for others to more easily build on top of this plugin. For example, if I wanted to build a plugin on top of this one that allows you to generate titles from the post list screen, I would just need to build my own UI then use the Abilities API to trigger the generate titles ability, either by calling the ability via PHP, calling the REST endpoint directly or eventually calling the client-side ability.
3. If desired, we can expose this ability to an MCP server (that part isn't done in this PR). While I agree it feels odd to have an ability that an AI tool could call that would then itself call an LLM (why not just skip the middleman in most situations) I could see a scenario where someone is wanting to manage their WordPress content within something like Claude, where it can connect to your site, find content, suggest some titles and then update that content with the title you choose

Overall my goal here has been to keep things as extendable as possible, provide examples of how these AI tools can be used and hopefully inspire others to build their own things using these same tools (or build on top of what we do). Worth noting this is a slightly different approach then if I was trying to get something merged to WordPress core but leaning in to the "experiments" label here.

All that said, happy to change this approach if we think that's best. Really just want to get this unblocked so we can start to make some more serious progress here.

felixarntz: This all makes sense. I think the main caveat is regarding:

> We showcase how the Abilities API can be used, which I believe is one of the goals of this plugin

_Is_ the way the Abilities API is used here a good example of how it should be used or not? Going back to my original concern. That's where I'd like other people's feedback so that we can make a decision.

jeffpaul: @felixarntz who else can we tag in for review here to get to a resolution and keep feature development unblocked in the plugin?

felixarntz: @jeffpaul Who I pinged above :)

aaronjorbin: From the [WordPress source code](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/abilities-api.php#L5C1-L8C24)[1]:

>  The Abilities API provides a unified, extensible framework for registering and executing discrete capabilities within WordPress. An "ability" is a self-contained unit of functionality with defined inputs, outputs, permissions, and execution logic.

To me, this makes it clear that abilities are about exposing functionality of WordPress (or plugins) but it leaves open the question of if that functionality can use an external source such as AI or any 3rd party API. Abilities aren't just about exposing functionality to AI, they can also be used (theoretically) in place such as wp-cli, command palate, or in a UI.

I think that generating titles is a discrete activity that can have a defined input, output, permissions, and execution and thus this is an acceptable use of the abilities API. 

[1] I see a few specific word choices that I think need to get cleared up with this definition and will be opening a PR for that, but in general I think this is a good definition to use for now.

swissspidy: It's certainly confusing when a tool or ability itself calls another AI tool. Especially when nothing of that is implemented yet and this is just boilerplate code.

This reminds me of a situation in the OG WP-CLI MCP implementation, where we implemented featured image generation as an MCP tool, and that tool made an AI call to generate an image. Ideally that wouldn't have been an MCP tool though, but done by the AI.

Likewise, the AI can already generate titles itself. It doesn't have to call a tool that calls an AI to generate titles. So in this context, a title generation tool call doesn't make sense. A Title Generation UI feature in the editor could  generate a title using the in-browser AI model or call the WP PHP AI SDK REST API endpoint to do this work and then use the "Update Post" ability to update the title (or just use the editor APIs).

At the same time, however, I can also see how it's useful for a Title Generation ability to do all of that, so that any plugin could easily integrate the same functionality in their UI without reinventing the wheel.

Soo... both could be true? :-)

justlevine: The committed definition that @aaronjorbin referenced is explicit and intentional; at the risk of being a broken record the Abilities API is about much more than AI, and **definitely more** than just MCP.

As far as the delineation between what is an Ability vs a (lowercase) feature vs a(n uppercase) Feature, etc. I think a good part of it is a naming things problem with the word "Features". If we rename to Experiment, we lessen the confusion, and reduce the need for future conversations like this one. 

- #70

> Soo... both could be true? :-)

IMO yes, just because we have an Ability doesn't mean that it needs to be exposed or used by MCP, and abilities can also be made up of other abilities. Long term, we could have a Title Generation Ability (with its own schema) that wraps a Content Generation Ability (that itself internally is just a interpolated prompt to an LLM), that are both toggleable via the same Content Generation ~Feature~ Experiment, with the former using `WP_Ability::$metadata['annotations']['show_in_mcp'] = false`, but `show_in_cp = true` (or whatever, the implementation details are less important than the concept of composability)

includes/Contracts/Feature.php
justlevine: > This could either be renamed (something like get_base_ability_slug) or maybe better to just move this to the ability class.

Third option is not abstracting it at all, and letting the Features decide how/where/when they are using the abilities API. 

Personally, I recommend this _regardless_ of the outcome of the other thread. We control the codebase, IMO there's no reason to abstract and add this level of prescribed coupling. If -after we have a handful of Features implemented - we see this is a pattern worth enforcing, we can do it then.

includes/Feature_Loader.php
```diff
@@ -103,6 +103,7 @@ public function register_default_features(): void {
 	private function get_default_features(): array {
 		$feature_classes = array(
 			'WordPress\AI\Features\Example_Feature\Example_Feature',
+			'WordPress\AI\Features\Title_Generation\Title_Generation',
```
justlevine: nit: referencing the class itself makes the dependency clearer - and lets us know promptly if someone renames the class without updating this array.

```suggestion
			\WordPress\AI\Features\Title_Generation\Title_Generation::class,
```

includes/Abstracts/Abstract_Ability.php
```diff
@@ -0,0 +1,130 @@
...
+namespace WordPress\AI\Abstracts;
+
+use WP_Ability;
+
+/**
+ * Base implementation for a WordPress Ability.
+ *
+ * @since 0.1.0
+ */
+abstract class Abstract_Ability extends WP_Ability {
```
justlevine, line 19: I'm strongly against extending `WP_Ability` like this.

If there's justification for an abstract class, then we should follow the pattern of `Abstract_REST_Controller`, and use this class to build the args that get passed to a `wp_register_ability()` call.

includes/Abstracts/Abstract_Ability.php
```diff
@@ -0,0 +1,130 @@
...
+
+	/**
+	 * The Feature class that the ability belongs to.
+	 *
+	 * @since 0.1.0
+	 * @var \WordPress\AI\Abstracts\Abstract_Feature
+	 */
+	protected $feature;
```
justlevine: Related to https://github.com/WordPress/ai/pull/61/files#r2491071832, I think it's premature to enforce a one-to-one relationship between an Ability and the module(s) that may consume it.

includes/bootstrap.php
```diff
@@ -172,6 +172,19 @@ function initialize_features(): void {
 		$loader   = new Feature_Loader( $registry );
 		$loader->register_default_features();
 		$loader->initialize_features();
+
+		add_action(
+			'wp_abilities_api_categories_init',
+			static function () {
+				wp_register_ability_category(
+					'ai-experiments',
```
justlevine, line 185: What purpose will this category serve?

IMO categories should be semantic, since there can only be one, so if this is a catch-all group for any abilities we register in the plugin, then id avoid against this. We already get the same thing from the (required) ability name prefix.

If we need an identifier, and ability prefix isn't enough, then I recommend `WP_Ability::$metadata['experimental'] = true;`

includes/Abilities/Title_Generation.php
felixarntz: It seems that this is a very open-ended conversation that could go on and on.

I agree that Abilities API is "more than AI", yet here we are building it for AI, and I bet it wouldn't have come up as a new API if we weren't building out AI infra for WordPress. But yes, it can and probably should be used for other use-cases than AI agents.

On the other hand, I also agree with Pascal's point that it's weird to have a tool that purely calls an LLM to generate a title because that doesn't make much sense to provide as a tool to an LLM (since it can already do that on its own). For that context, we'd instead need tools to obtain the necessary context (post content etc.).

Maybe the answer is that it's fine for this to be an ability, but it would be an ability that is primarily relevant for other use-cases than AI agents - e.g. simply a way to implement a new WordPress ~~feature~~ ability using a standard API.

That brings up an important point though, related to the ongoing conversation how to expose the right tools to an agent: Arguably an ability (tool) like this should not be exposed to an agent (because of the above)? Does there need to be a way to mark such abilities? Or should it simply be handled via ability categories? cc @gziolo for awareness

Anyway, due to the nature of this discussion, I'm happy to unblock this and just stick with the ability implementation for now. The one thing we should revise though is to get rid of the `get_ability_slug()` method on the `Feature` interface and base class, to not prematurely couple the two.

includes/Contracts/Feature.php
felixarntz: Agreed, also see https://github.com/WordPress/ai/pull/61#discussion_r2511142951 - the abstraction is premature as long as there's no clear answer to that discussion.

includes/Abstracts/Abstract_Ability.php
felixarntz: We have discussed this many times before: It's fine to extend `WP_Ability`, it's allowed by the API. This is a personal preference.

includes/Abstracts/Abstract_Ability.php
felixarntz: Agreed, let's not couple them.

includes/bootstrap.php
dkotter: My understanding is that every Ability has to have a category associated with it. So I took the approach for now of registering this generic category that all the Abilities we end up with can utilize. Open to better approaches though, if for instance we want more specific categories

includes/Abstracts/Abstract_Ability.php
justlevine: Except we're in the `WordPress/ai` repo so it's not [only](https://make.wordpress.org/project/2025/06/04/criteria-for-creating-or-migrating-repositories-under-the-wordpress-github-organization/) personal preference, and just because something is "allowed" doesn't means it should be done without defined intent, let alone encouraged, or in this case enforced.

(Which I believe is a more accurate summary of the discussion, most recently iirc https://github.com/WordPress/abilities-api/pull/97 ).

**So, for  our specific plugin / this PR**:

1. Why do we need an abstract wrapper for registering abilities at all, especially while there is (currently) only one ability in this codebase?

2. Why is extending `WP_Ability` the ideal path to accomplish No. 1?

The two biggest downsides of extending `WP_Ability` in our particular case (as noted, the general issues with the pattern have been repeatedly discussed):

1. Instead of dogfooding the entire Abilities API, we're diverging from it, which beyond the testing/future-compat issues, also make it significantly harder to "graduate" an experimental ability to Core.
2. Practically with this approach, every feature that registers an ability needs to duplicate the same registration code.

(IMO the `Abstract_REST_Controller` pattern addresses both of those downsides, although any extra layers of prescriptive abstraction inside a Feature this early should still cause some pause, considering the whole purpose of this repo is to "experiment" )

includes/bootstrap.php
justlevine: > My understanding is that every Ability has to have a category associated with it.

Yeah, it's a chicken-egg situation. We wanted to ship a Category API to avoid a generic `uncategorized` scenario by adding it later, but until we establish what those categories should be, we're essentially doing the same generic bucketing 😅

I don't think a premature part of the API should hold up this PR, so if the other folks here can't think of something yet (in all fairness, ideal taxonomies form holistically), then I say we proceed as is and add a comment saying that this is a temporary catch-all and likely to change.

(related: https://wordpress.slack.com/archives/C08TJ8BPULS/p1762568388167959 )

dkotter: Comment added https://github.com/WordPress/ai/pull/61/commits/032a3e7ffc0b9fc6cbdebb8280f31dce153b0b9c

includes/Abstracts/Abstract_Ability.php
dkotter: Removed the coupling of Feature to Ability https://github.com/WordPress/ai/pull/61/commits/550b4fb7e605f153e0a5a58a7d926adc7a575af3

includes/Feature_Loader.php
dkotter: Fixed in https://github.com/WordPress/ai/pull/61/commits/fff6c5d1148598817613ece770df71b0849d22a0

includes/Contracts/Feature.php
dkotter: Removed in https://github.com/WordPress/ai/pull/61/commits/779d3d6c711ef61f134617ae84bacaf651cc86f8

includes/Abstracts/Abstract_Ability.php
felixarntz: If an API allows something, it's fine to do it. It doesn't mean it's encouraged, and it doesn't mean it's discouraged. If the API wants it to be discouraged, it shouldn't even be possible in the first place.

This project, like any other project, has to choose one of those two approaches, and which approach is chosen is personal preference. Your personal preference is to not extend `WP_Ability`, my personal preference is to extend it.

To your questions:
1. This PR implements some scaffolding. Based on the other discussion, it looks like we're leaning more or less any feature in WordPress can be an ability, so it makes sense to provide scaffolding for how to implement abilities in the scope of this plugin.
2. This project is mostly object-oriented, so using classes in general makes sense. If you don't extend `WP_Ability`, we'd probably still want one class per ability to keep things decoupled, especially because the features (or experiments) that make use of the abilities are decoupled. `WP_Ability` provides a possible foundation for this class breakdown, so IMO it makes sense to use it.

Many aspects of coding are subjective. Whether this project uses procedural programming or OOP is subjective, and so is whether it extends `WP_Ability` or does something else. We can reason about it, but there's no "ideal" and no "right or wrong".

felixarntz (approved):
@dkotter LGTM!

includes/Abilities/Title_Generation.php
JasonTheAdams: Hi folks! 👋 

I've somehow missed this PR. Thanks, @jeffpaul, for pinging me!

This is a perfectly valid use of Abilities, from my vantage point. I agree with @swissspidy on this one. It would be weird to expose this to MCP and then have an AI call a function that calls AI. In that case, you'd simply have an Ability for updating the post title (or whatever) which is exposed to MCP, and the model calls that while generating the title itself. But to have an ability that uses AI to generate a title is perfectly fine; I'd even be fine if this was two abilities:
1. An ability for updating posts
2. An ability for generating and returning a title string using AI

The first could potentially even reference the second with a "generate title" boolean input, or something like that. I'm honestly not thinking that hard about it, the point being abilities can be nice and composable. I could even see having a way of piping abilities in the future.

This actually softly brings up something I was [thinking over yesterday](https://github.com/WordPress/mcp-adapter/pull/91#discussion_r2512096585): introducing an `open-world` annotation to abilities. I'm borrowing this concept from MCP, which is simply a way of saying "by the way, this function reaches outside of the application and does something externally". In this case, it would help clarify that this ability uses an external service — the AI model.

In short, I'm good with this!

includes/Abstracts/Abstract_Ability.php
JasonTheAdams: Weighing in here, I kindly suggest we step back from the "is extending `WP_Ability` good or bad" and such debate. To @felixarntz's point, it's part of the API and so everything here is a perfectly sensible use of that, and even a great way of showing how it can be done in a good, object-oriented manner. I like it.

That said, I do think it would be a good idea to also include a procedural example in this project, which we can reference in the documentation as something intentional. That way we can help both procedurally and OOP inclined folks on how they can interact with the Abilities API in their preferred manner.

In short, I think this is a great both-and situation where we can show off how flexibly extensible this API is.

includes/Abilities/Title_Generation.php
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		// TODO: Implement the title generation logic.
+
+		return array(
+			'name'        => $this->get_name(),
+			'label'       => $this->get_label(),
+			'description' => $this->get_description(),
+			'content'     => wp_kses_post( $args['content'] ),
+			'post_id'     => absint( $args['post_id'] ) ?? esc_html__( 'Not provided', 'ai' ),
```
mindctrl: `absint( null )` will return `0` here if no post_id is provided, so `Not provided` won't be used in that case. 

I think it should be something like this (or some variant):

`'post_id' => $args['post_id'] ? absint( $args['post_id'] ) : null,` or 

or

`'post_id'     => absint( $args['post_id'] ) ?: esc_html__( 'Not provided', 'ai' ),`

JasonTheAdams (approved):
Left a few comments, but I think we can merge this! Great work, @dkotter!

includes/Abilities/Title_Generation.php
JasonTheAdams: Good catch! Yeah, let's fix this before merging. 👍

includes/Abilities/Title_Generation.php
```diff
@@ -0,0 +1,172 @@
...
+	/**
+	 * Returns the permission callback of the ability.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param mixed $args The input arguments to the ability.
+	 * @return bool|\WP_Error True if the user has permission, WP_Error otherwise.
+	 */
+	protected function permission_callback( $args ) {
+		if ( ! current_user_can( 'edit_posts' ) ) {
```
mindctrl: I wonder if it makes sense to check if the user can edit the current post, like `current_user_can( 'edit_post', $args['post_id'] )`? Would need to check if `post_id` was supplied. Can be done in the follow up PRs, but wanted to flag for any potential discussion.

JasonTheAdams: Yup! Another good catch. That's exactly how this should be working. 😄

includes/Abilities/Title_Generation.php
dkotter: I can update that but note this entire return statement will be changed in the next PR (where we'll either return the generated titles or a `WP_Error`), this is just here for test purposes

mindctrl: I was just going to comment that the output schema doesn't match the return. If this is just testing purposes, I think it's fine as is for now.

includes/Abilities/Title_Generation.php
dkotter: Yep, was planned for the followup PR but no reason it can't be pulled in here. Have done that in https://github.com/WordPress/ai/pull/61/commits/e0e4873fb33f4fd412b5dbfc9f5a97c1f0fc85c9

includes/Abilities/Title_Generation.php
dkotter: Fixed post ID output here https://github.com/WordPress/ai/pull/61/commits/e0e4873fb33f4fd412b5dbfc9f5a97c1f0fc85c9.

And yeah, the output here will change to match the schema in the next PR. If you want, can see that here (noting there are some changes needed there now after changes made in this PR): https://github.com/WordPress/ai/pull/67/files#diff-8fd34de7d808803d2695dbc4ba5bb8d014520a5063a0a428848f14fb133c6815R141-R149

includes/Abilities/Title_Generation.php
dkotter: Also worth noting I'm open to suggestions on these permissions. Technically this ability will just generate titles, it won't save those titles. So could argue that having edit permissions isn't actually required though feels safer to me

includes/Abstracts/Abstract_Ability.php
justlevine: @JasonTheAdams @felixarntz I think there's been a misunderstanding.
I feel like I'm listing specific and concrete issues with this specific implementation (as I did previously re extending \WP_Ability), and I'm a bit frustrated at what feels like having them repeatedly dismissed as personal preference, or conflated with OOP vs functional programming, instead of them being addressed directly.

Hopefully this (illustrative) diff will make the impact of my suggestion clearer: https://github.com/dkotter/ai/pull/4/files

See how: 
- Abstract_Ability is still OOP, but fully dogfoods the `Abilities_API`.
- Registration logic is DRYed out.
- The DI happens _inside the experiment_ classes, showing how a Feature can be coupled to an Ability, without enforcing all abilities have a 1-to-1 relationship with features.
- If a specific future experiment wants to go either functional or `extends \WP_Ability`, they don't need extra boilerplate
- It also limits the scope/impact of refactoring as we add new experiments or the Abilities API gets new features.

If you don't agree that this a better pattern in our context (or that those aren't quantifiable improvements over the current implementation), I would really appreciate if you could share your specifics thoughts on which and why🙏

felixarntz: The reason I am referring to this as personal preference is precisely because I don't see how the points you're making are an improvement. For example, more concretely:

> Abstract_Ability is still OOP, but fully dogfoods the `Abilities_API`.

What does "fully dogfoods" mean? In your code it dogfoods the `wp_register_ability` function, in the existing code it dogfoods extending the `WP_Ability` class. No win or loss either way.

> Registration logic is DRYed out.

This is also possible using the current pattern of extending `WP_Ability`. The PR doesn't implement it in that way, but let's not conflate whether or not we extend `WP_Ability` with whether it's _possible_ to DRY registration logic.

> The DI happens _inside the experiment_ classes, showing how a Feature can be coupled to an Ability, without enforcing all abilities have a 1-to-1 relationship with features.

I don't see how injecting an experiment into an ability resolves any coupling. As of the latest state of this PR, there's no 1-to-1 coupling between abilities and experiments whatsoever, so arguably this diff adds _more_ coupling.

I don't want to come across like a broken record, but we're comparing two different approaches both of which are reasonable, but one has to be chosen to move forward. I have yet to see any convincing argument that not extending `WP_Ability` is better than extending it.

includes/Abilities/Title_Generation.php
```diff
@@ -0,0 +1,207 @@
...
+		// TODO: Implement the title generation logic.
+
+		return array(
+			'name'        => $this->get_name(),
+			'label'       => $this->get_label(),
+			'description' => $this->get_description(),
+			'content'     => wp_kses_post( $args['content'] ),
+			'post_id'     => $args['post_id'] ? absint( $args['post_id'] ) : esc_html__( 'Not provided', 'ai' ),
+			'n'           => absint( $args['n'] ),
+		);
```
JasonTheAdams, line 138: @dkotter So are we returning this now simply because we're not generating an array of title suggestions, yet? This is temporary?

JasonTheAdams: Ah, you already answered this here: https://github.com/WordPress/ai/pull/61#discussion_r2514976295

includes/Abilities/Title_Generation.php
JasonTheAdams: That's a good point on the capability check, @dkotter. I think we should be using the `read_post` meta cap, instead:
```php
current_user_can( 'read_post', $args[ 'post_id' ] );
```
All we really need to know is that the current user has read access to the post in order to read the content, which will basically always be true unless the post isn't published, the post type isn't public, etc.

includes/Abilities/Title_Generation.php
dkotter: Yep, correct. Am trying to keep each PR as small as possible so this one just lays the foundation, the next PR will actually implement the generation of titles and then a final PR that adds the UI to trigger the generation

includes/Abstracts/Abstract_Ability.php
justlevine: Thanks @felixarntz that was very helpful 🙇

Point by point:

> > Abstract_Ability is still OOP, but fully dogfoods the `Abilities_API`.
> 
> What does "fully dogfoods" mean? In your code it dogfoods the `wp_register_ability` function, in the existing code it dogfoods extending the `WP_Ability` class. No win or loss either way.

By "fully dogfood" I mean we cover more LOC of the API (and have a smaller API contract to maintain at the same time). When we extend `\WP_Ability`, any methods that are overloaded means that's one less method we're dogfooding, and one more piece of tech debt to keep in sync both core and possibly other experiments. That feels like a significant win.

> >  Registration logic is DRYed out.
>
> The PR doesn't implement it in that way, but let's not conflate whether or not we extend WP_Ability with whether it's possible to DRY registration logic.

Possible != ideal. I assumed (hoped) it was left off the current implementation to avoid overloading `\WP_Ability` "unnecessarily" as mentioned. There's a pro v cons discussion to be had about DRY vs adding unnecessary methods , but when it no longer `extends \WP_Ability`, we can be as prescriptive as we want at no cost.

>> The DI happens inside the experiment classes, showing how a Feature can be coupled to an Ability, without enforcing all abilities have a 1-to-1 relationship with features.

> I don't see how injecting an experiment into an ability resolves any coupling.  As of the latest state of this PR,

Sorry I should have been more clear - the entire diff implementation I shared was solely meant to be illustrative. I was showing _how_ an ability could be directly coupled to a feature if someone wanted, not advocating to do so. That was in response to @JasonTheAdams desire to illustrate multiple patterns.

Implied but not included are also an `extends Abstract_Ability` implementation where the register function takes the same args as the current implementation's constructor, and an experiment that chooses to `extends \WP_Ability` instead of `Abstract_Ability`. Or any other future pattern that presents, thats the whole point. That flexibility comes with baggage if our defacto approach is `\WP_Ability`


> I have yet to see any convincing argument that not extending WP_Ability is better than extending it.

I don't want to be a broken record either and repeat other conversations, so at minimum I'll [quote you previously](https://github.com/WordPress/abilities-api/pull/97#issuecomment-3372654254) ;-) : 

> I agree there's a higher risk of developers doing it wrong, if we allow ability_class. But I also agree the flexibility of allowing ability_class gives developers with OOP preferences a simpler integration point - with the caveat that they need to know what they're doing.

As the `Abstract_Ability` class is meant to be used throughout this plugin, and this plugin is both a) meant to serve as an example and b) have low barriers for contributions, i'd think this would minimally be a good enough reason, especially considering (if i understood you) even you agree we don't lose anything by going with an OOP implementation that doesn't extend `WP_Ability`.

Or to put it another way:

> I don't lean strongly either way, 

I _do_, though. Clearly, or I wouldn't be harping on it so much or going into excruciating detail whenever the topic comes up (I'm volunteer hours, nobody's paying me to be long winded 😅😅😅).

So 
- if the above is still true for you
- and nobody else here has strong reasons  _in favor_ of extending `\WP_Ability` inside the initial version of our `Abstract_Ability` class

perhaps we can go ahead with this change anyway, even if you personally don't find my reasons particularly compelling?

felixarntz: Like I said, I don't feel strongly. But there's also something to be said about "strong opinions, loosely held". At some point it's important to make a call and roll with it, even when not everyone agrees. And then it's important to commit to the decision. For example, I still don't fully agree with an ability that's there purely to call an AI model - but I see my argument has some flaws, and while I think the other arguments also have flaws, it just goes to show there's no ideal or perfect solution and it's simply a tradeoff and decision to make. So I'm not going to complain about that further :)

My perspective here is: Because I don't see a real net-positive in either way, it's most pragmatic to go with how @dkotter implemented it. I personally would have implemented it somewhat similarly. Now, if he or someone else wants to invest the additional time to change it, I won't oppose the change. But FWIW I don't see value in it either.

I think from my perspective everything here is said. Curious to get other people's perspective if there are folks with any +/- arguments. But I also think we should unblock this soon.

dkotter: I (probably obviously) like the approach taken here but that said, this isn't a hill I'm going to die on. What I really want is to get things unblocked here so progress can continue and if that means changing the approach, I'm happy to do that.

I will say this thread is pretty long at this point and I'm not sure I fully follow what we consider the "ideal" approach but it seems at a minimum:

- Keep the `Abstract_Ability` class but don't extend `WP_Ability`
- Keep most of the other methods the same in both the `Abstract_Ability` class and any class that extends that (just `Title_Generation` in this case)
- Look at adding a registration method in `Abstract_Ability` to make it easier for Abilities to be registered

Does that seem accurate or are there other parts I'm missing that should be modified?

JasonTheAdams: Lively conversation! Appreciate the passion and continued clarification on both sides! 😄 

Thanks for providing the example and further context, @justlevine! It helps! I took some time to reflect and really tried to think through what you're suggesting.

I think I've somewhat returned back to my original point, but with some clarification. That is, providing two examples as part of this repo for educational purposes:

1. Extending `WP_Ability` as a way of showing how to use `ability_class`
2. Using `wp_register_ability()` in a purely, non-cleverly procedural way

This makes me like the way this works even more. The idea of having an `Abstract_Ability` that doesn't actually extend `WP_Ability` feels, to me, even more abstract than this, and doesn't illustrate using `ability_class`. It's just a procedural way of registering that's wrapped in OOP. I'm also not a fan of bringing the Feature into the Ability level, as I think the Ability is a lower-level concept which should be a dependency of the Feature, not the other way around.

In conclusion, my preference is how this currently works because it illustrates `ability_class`. I would suggest that one of the next features register the Ability in a procedural way, not tucking it into OOP, as a simple example for folks to follow.

includes/Abilities/Title_Generation.php
dkotter: Update in https://github.com/WordPress/ai/pull/61/commits/9b9b12345b20a9cf6d44395c1640317756afacf2

includes/Abstracts/Abstract_Ability.php
justlevine: @dkotter 

>  it seems at a minimum:
> 
> * Keep the `Abstract_Ability` class but don't extend `WP_Ability`
> * Keep most of the other methods the same in both the `Abstract_Ability` class and any class that extends that (just `Title_Generation` in this case)
> * Look at adding a registration method in `Abstract_Ability` to make it easier for Abilities to be registered
> 
> Does that seem accurate or are there other parts I'm missing that should be modified?

Yup, that's all of it 🙇. Basically the [`WP_Rest_Controller`](https://developer.wordpress.org/reference/classes/wp_rest_controller/) pattern, but with ergonomics that make sense for us.

---

> It's just a procedural way of registering that's wrapped in OOP

@JasonTheAdams I'm not entirely sure I understood what you mean here. Can you clarify? 

At risk of being redundant, we wouldn't be testing or demonstrating `ability_class` _inside_ `Abstract_Ability`; instead we're demonstrating `wp_register_ability()` and how to use it in an OOP manner, and testing the _entirety_ of `\WP_Ability` (futureproofing). A specific experiment that wants to `extends \WP_Ability` to use `ability_class` is less code/maintenance. And the _need_ to extend `\WP_Ability` (e.g. to overload behavior, not to have OOP ergonomics when creating new abilities) is much less likely than wanting `WP_Ability` that can opt-into to the entire lifecycle and not worry about future-compat or cross-compat with other experiments.

IOW this doesn't preclude `ability_class`, it's (imo) easier and more robust to illustrate and experiment with `extends \WP_Ability` when it's intentional versus when its enforced plugin-wide as an `Abstract_*` pattern.

(If that was already clear apologies, I'm signing out for the night, and didn't want to hold up the convo with another round trip 😅)

justlevine: > I personally would have implemented it somewhat similarly. 

(  @felixarntz In fairness one of the main reasons I introduced `ability_class` was to keep similar with your prior art and minimal effort in adapting for the WordCamp demo. So really you could say it's _because_ you would have (and did) implemented it similarly that the pattern is supported to begin with 😅)

JasonTheAdams: Appreciate all the input, @justlevine, and referencing this as following the `WP_Rest_Controller` pattern definitely helped in understanding where you're coming from. I confess I'm not quite convinced that should _replace_ this. I can see it as another pattern, and certainly one with history, but I remain in the camp that I like this as an example of an Ability being extended and registered outside of itself as an `ability_class`.

At this point, I think we need to merge this PR and can revisit this in a subsequent PR. As I've said, I'm open to having multiple ways of registering abilities to show folks the versatility of registration. This isn't a dismissal, @justlevine, as we've spent a good amount of time on this and the majority of folks contributing to the conversation remain unconvinced, so we ought to move forward. Please consider making a PR to introduce another Feature which we can contrast this with.

Jameswlepage (approved):

Jameswlepage (approved):

includes/Abstracts/Abstract_Ability.php
justlevine: Thanks @JasonTheAdams , and I agree that iterating quickly is the goal here and happy to kick the discussion to a follow-up PR so we can keep progressing.

Correct me if I'm wrong (cc @jeffpaul ), but my gut says that we have at least until 6.9 ships - if not a bit longer - before any architectural decisions or reversing them have any real impact. (Tangentially, we're also overdue a discussion about WPCS, contribution guidelines/barriers to entry for new modules/experiments, and what a path to core for specific experiments/features would look like)

PR #67, opened 2025-11-03
Add the Title Generation execute functionality

### LINKED ISSUE #10: Title Generation / Rewriting

### LINKED ISSUE #29: Add UI controls in post editor for rewriting titles

dkotter:
## What?

Partially closes #10

Builds off of the Title Generation feature base work done in #61 to wire up the actual execution of title generation. Note the actual UI still isn't in this PR, there will be a close followup that adds that.

## Why?

In PR #61, we introduced the Title Generation Ability and Feature classes but the actual generation of titles wasn't completed there. This PR finishes that piece, though note there still isn't any UI to trigger this functionality, it all needs to be done via direct API requests.

## How?

- Pulls in the latest changes from the `trunk` branch of the WP AI Client repo to take advantage of the new settings screen and helper classes introduced there. This can be changed to pull from an actual release once that is out
- Initializes the WP AI Client
- Adds in a new `API_Request` class that can be used to make requests to an AI Provider. This ultimately utilizes the WP AI Client. Debated whether having this extra abstraction was worth it so can change this if desired. But it felt like the right approach knowing we'll have multiple other AI integrations coming that will all have slightly different requirements (i.e. arguments) they'll need to send
- Adds a `generate_titles` method to the Title Generation Ability and uses that in the execute callback
- Adds some helper functions to normalize content and get additional context we can send to the LLM
- Adds functionality where we can store system instructions in their own PHP file that the Ability can then pull in. This could be hardcoded in the Ability class but felt cleaner to keep these in their own files for easier management

## Testing Instructions

There is no UI yet but can be tested by making direct API requests to the Title Generation Ability run endpoint.

First you'll need to login to WordPress and go to `Settings > AI Credentials` and add in at least one API key.

Then make an authenticated `POST` request to the Title Generation run endpoint and ensure a title is generated based on the content you provide
```bash
--data '{
    "input": {
        "content": "this is some content!",
    }
}'
```

Make an authenticated `POST` request to the Title Generation run endpoint, passing in a valid post ID, and ensure a title is generated
```bash
--data '{
    "input": {
        "post_id": 1,
        "candidates": 1
    }
}'
```

Make an authenticated `POST` request to the Title Generation run endpoint, passing in a valid post ID and number and ensure multiple titles are generated
```bash
--data '{
    "input": {
        "post_id": 1,
        "candidates": 2
    }
}'
```

Make an authenticated `POST` request to the Title Generation run endpoint and don't pass any content or post ID and ensure an error is returned
```bash
--data '{
    "input": {
        "candidates": 2
    }
}'
```

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=67%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2067%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).


jeffpaul:
@dkotter with #61 merged in, looks like you're good to continue here with resolving the merge conflicts and getting this ready for review/testing

includes/Abilities/Title_Generation/system-instruction.php
JasonTheAdams, line 1: @dkotter It would actually be better to make this a PHP file. The reason being the PHP runtime will cache the contents of the file if it's PHP, whereas a .md file will be read every time.

dkotter: Good callout, didn't consider that. Updated in https://github.com/WordPress/ai/pull/67/commits/85503ebfc1972a992e9b29946ad049c3ff543585

includes/Abilities/Title_Generation/system-instruction.php
```diff
@@ -0,0 +1,19 @@
...
+Goal: You will be provided with some context and you should then generate a concise, engaging, and accurate title that reflects that context. This title should be optimized for clarity, engagement, and SEO - while maintaining an appropriate tone for the author\'s intent and audience.
+
+The title suggestion should follow these requirements:
+
+- Be no more than 80 characters
+- Should not contain any markdown, bullets, numbering, or formatting - plain text only
+- Should be distinct in tone and focus
+- Must reflect the actual content and context, not generic clickbait
+
+The context you will be provided is delimited by triple quotes.';
```
JasonTheAdams: Nit: since we're making an example, I think this is a good time to use a [Nowdoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc) to keep the copy clear:
```suggestion
return <<<'INSTRUCTION'
You are an editorial assistant that generates title suggestions for online articles and pages.

Goal: You will be provided with some context and you should then generate a concise, engaging, and accurate title that reflects that context. This title should be optimized for clarity, engagement, and SEO - while maintaining an appropriate tone for the author\'s intent and audience.

The title suggestion should follow these requirements:

- Be no more than 80 characters
- Should not contain any markdown, bullets, numbering, or formatting - plain text only
- Should be distinct in tone and focus
- Must reflect the actual content and context, not generic clickbait

The context you will be provided is delimited by triple quotes.
'INSTRUCTION';
```

includes/Abilities/Title_Generation.php
```diff
@@ -99,7 +93,7 @@ protected function execute_callback( $input ) {
 			array(
 				'content' => null,
 				'post_id' => null,
-				'n'       => 1,
```
JasonTheAdams, line 102: Can we give this `'n'` variable a more meaningful name?

includes/Abilities/Title_Generation/Title_Generation.php
JasonTheAdams: Can we give this `'n'` property a more meaningful name?

includes/API_Request.php
```diff
@@ -0,0 +1,238 @@
...
+	/**
+	 * Check if the AI SDK Client is available.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return bool True if the client is available, false otherwise.
+	 */
+	protected function is_client_available(): bool {
+		return class_exists( AI_Client::class );
+	}
```
JasonTheAdams: Why wouldn't this be available?

includes/API_Request.php
JasonTheAdams, line 1: I'm really confused by this class. Why do we need this? Why not just do `AI_Client::prompt()` in the `Title_Generation` Ability and do everything we need there? Maybe I'm missing something, but the whole abstraction of this class seems unnecessary to me.

I think the biggest thing it's doing is applying the model and provider, if specified. If that's all it's doing, then I'd make a function (or class) somewhere that just creates the `Prompt_Builder`, applies the Provider/Model preferences, and then returns the `Prompt_Builder` instance. That really leans into the point of the builder concept, where you can hand over a partially configured builder intsance.

JasonTheAdams (changes requested):
Excited to see the first usage of the WP AI Client, @dkotter!

I didn't fully review this, but I got confused enough by the architectural direction to pause here for some questions. I feel like this is over abstracted, and if it's not I'm really curious about the needs if not.

includes/API_Request.php
dkotter: So I debated on this so happy to simplify. Few thoughts here:

Was trying to reduce the amount of code we would need to duplicate as we add functionality in the future that needs to make requests. Meaning for generating titles, beyond just passing in a prompt, we set the system instructions, we set the temperature value, we set the candidate count value and we set the default model preferences. I could easily move that out of this class and into the Title Generation Ability but we would then have that exact (or similar) code duplicated each time.

This class also standardizes how we process the response, ensuring we return an error if we get an empty array back and sanitizing the responses before returning. Again, all things that could happen in the Ability but would require us to duplicate that code each time.

That said, I'm fine with the approach of a helper function or two that will build the `Prompt_Builder` and return it and then each Ability/Feature/Experiment/Whatever that wants to make requests can take over from there. Let me make those changes and see if that feels like a better approach

includes/Abilities/Title_Generation.php
dkotter: This is what it's called on the API-spec side (or at least what OpenAI calls it) but renamed to `candidates` to more closely match what our AI Client uses: https://github.com/WordPress/ai/pull/67/commits/556d6dba36fd0b665e47e11bce4efcdaa0b686db

includes/Abilities/Title_Generation/Title_Generation.php
dkotter: Fixed: https://github.com/WordPress/ai/pull/67/commits/556d6dba36fd0b665e47e11bce4efcdaa0b686db

includes/Abilities/Title_Generation/system-instruction.php
dkotter: Updated: https://github.com/WordPress/ai/pull/67/commits/eb226ba437c8d6934150b243a91a3ad8c9f723dd

includes/API_Request.php
JasonTheAdams: I see where you're coming from. My main motivation here is that we want to think of these Abilities as examples, and I want devs to see how easy it is to construct a prompt and get the results in the ability.

If we need to normalize the response for our abilities in some way, then I'd like to do that after using the builder. That will also help us see if there's some sort of sanitization/normalization that we should introduce in the WP AI Client in some way.

Just to make sure you're aware, there's also a `Prompt_Builder_With_WP_Error` that can be used if you want to avoid exceptions.

dkotter: So I've simplified this a bit in https://github.com/WordPress/ai/pull/67/commits/ca74ff17cfad071363cb67974734ed4982ea69d1, getting rid of the `API_Request` class and adding a helper function that builds our prompt builder instead.

If that still feels too abstracted (which I've gone back and forth on and can't decide), we can remove all of that and instead, just do this directly in the `generate_titles` method:

```php
$prompt_builder = AI_Client::prompt_with_wp_error( '"""' . $context . '"""' );
$prompt_builder = $prompt_builder->using_system_instruction( $this->get_system_instruction() );
$prompt_builder = $prompt_builder->using_temperature( 0.7 );
$prompt_builder = $prompt_builder->using_candidate_count( (int) $candidates );
$prompt_builder = $prompt_builder->using_model_preference( ...get_preferred_models() );
```

That's obviously cleaner and a lot less code but does mean we would have very similar (if not exact) code for each experiment we add. But as you've said, would be great if people can look at the code here and easily understand how things work, even if that means we end up with some code duplication

JasonTheAdams: Getting there! A couple notes:
1. If you specify a model you don't need to specify a provider, as a model implies a provider
2. You probably don't need to specify the preferred models array. The Prompt Builder will simply select a model that fits the prompt. Preferring a model is for when you have something like Nano Banana setup and you want to prefer that for image generation.

Also, chaining can be cleaned up a lot (without re-assigning the variable repeatedly):
```php
$prompt_builder = AI_Client::prompt_with_wp_error( '"""' . $context . '"""' );
	->using_system_instruction( $this->get_system_instruction() );
	->using_temperature( 0.7 );
	->using_candidate_count( (int) $candidates );
	->using_model_preference( ...get_preferred_models() );
```

I don't think the helper is really needed, as I'm not really seeing any code duplication that's necessary.

dkotter: Okay, more simplification done in https://github.com/WordPress/ai/pull/67/commits/94dff32f681379cfe6ca33230459c35828190269

> You probably don't need to specify the preferred models array

The reason I added this is because I ran into this issue when testing: https://github.com/WordPress/php-ai-client/issues/117. Basically things worked great one day and the next day things started to fail. In debugging, found out it was because the Prompt Builder started using the `gpt-4o-mini-search-preview` model which doesn't work properly for text generation.

So I can remove this but feels safer to have some defaults to avoid issues where things might break from one day to the next if new models are released or the models API results change.

JasonTheAdams: Ohhh, that's good for us to know. I'm going to ping @felixarntz on this one and get his thoughts. That sounds like a potential bug for that model's capabilities in PHP AI Client.

Definitely let us know if you run into stuff like that. You're the first one bringing this all together, so there are quite likely bugs like this. 😆

includes/Abilities/Title_Generation/Title_Generation.php
```diff
@@ -41,20 +35,21 @@ protected function input_schema(): array {
...
 					'type'              => 'integer',
 					'sanitize_callback' => 'absint',
 					'description'       => esc_html__( 'Content from this post will be used to generate title suggestions. This overrides the content parameter if both are provided.', 'ai' ),
 				),
-				'n'       => array(
+				'candidates' => array(
 					'type'              => 'integer',
 					'minimum'           => 1,
 					'maximum'           => 10,
+					'default'           => 3,
```
JasonTheAdams: Let's make this `3` a class constant so we don't have to sync it in two places.

includes/helpers.php
```diff
@@ -0,0 +1,160 @@
...
+		return $context;
+	}
+
+	/**
+	 * TODO: Might be interesting to add simple Abilities for the following,
+	 * just as a way to demonstrate a different approach to registering Abilities,
+	 * how to call Abilities via PHP and how multiple Abilities can be used together.
+	 *
+	 * Example: Get post content Ability; get post author Ability; get post terms Ability.
+	 */
```
JasonTheAdams, line 83: Agreed! I like the idea of composing abilities!

includes/Abilities/Title_Generation/Title_Generation.php
```diff
@@ -204,4 +224,43 @@ protected function meta(): array {
...
+
+		// Get the prompt builder.
+		$prompt_builder = AI_Client::prompt_with_wp_error( '"""' . $context . '"""' )
+			->using_system_instruction( $this->get_system_instruction() )
+			->using_temperature( 0.7 )
+			->using_candidate_count( (int) $candidates )
+			->using_model_preference( ...get_preferred_models() );
+
+		// Make the request.
+		return $prompt_builder->generate_texts();
```
JasonTheAdams: Comes together nicely! 😃 
```suggestion
		// Generate the text candidates using the AI
		return AI_Client::prompt_with_wp_error( '"""' . $context . '"""' )
			->using_system_instruction( $this->get_system_instruction() )
			->using_temperature( 0.7 )
			->using_candidate_count( (int) $candidates )
			->using_model_preference( ...get_preferred_models() );
			>generate_texts();
```

includes/Abstracts/Abstract_Ability.php
```diff
@@ -97,4 +100,81 @@ abstract protected function permission_callback( $input );
...
+	public function get_system_instruction( ?string $filename = null ): string {
+		return $this->load_system_instruction_from_file( $filename );
+	}
+
+	/**
+	 * Loads system instruction from a PHP file in the feature's directory.
+	 *
+	 * Automatic detection order:
+	 * 1. `system-instruction.php`
+	 * 2. `prompt.php`
```
JasonTheAdams: I'm curious, why have more than one option at this time? Why not just stick with one to stick to a single convention?

includes/Abstracts/Abstract_Ability.php
```diff
@@ -97,4 +100,81 @@ abstract protected function permission_callback( $input );
...
+
+			if ( file_exists( $file_path ) && is_readable( $file_path ) ) {
+				// PHP files should return a string directly.
+				$content = require $file_path; // phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.UsingVariable
+
+				return is_string( $content ) ? wp_strip_all_tags( $content ) : '';
+			}
+		}
+
+		return '';
```
JasonTheAdams, line 167: Do we want silent failure here? Maybe yes, just double-checking. 🙂

includes/helpers.php
JasonTheAdams, line 1: If a file doesn't contain a class, I generally lowercase it to make it clear. Not sure if this is against WPCS?

JasonTheAdams (changes requested):
Finished going through it! Left a few minor suggestions, but it's really close!

includes/API_Request.php
felixarntz: I replied in https://github.com/WordPress/php-ai-client/issues/117#issuecomment-3530661715:

* Concrete problem: Our OpenAI model capability and option assignments are probably inaccurate / out of date.
* Underlying problem: We should not return the alphabetically first eligible model, but rather a flagship model of the provider that makes sense to prefer by default.

felixarntz: I'd say for now let's put a preference list here, and then remove it once we've fixed the upstream bug.

includes/helpers.php
dkotter: I'll open another PR after this merges that adds these in and we can continue conversation on that

includes/Abilities/Title_Generation/Title_Generation.php
dkotter: Good callout, added: https://github.com/WordPress/ai/pull/67/commits/560dc318bc6c36c9f282167b337354c345ce43cc

includes/Abilities/Title_Generation/Title_Generation.php
dkotter: Updated in https://github.com/WordPress/ai/pull/67/commits/919b6609ff62ee19237819cbc9473a2c7690d2e6

includes/Abstracts/Abstract_Ability.php
dkotter: Good catch. Originally had this named as `prompt` and then renamed to match what the Prompt Builder expects and left support for both. But as this is a net-new thing, no need to support both right now. Removed in https://github.com/WordPress/ai/pull/67/commits/8619a7209fb9193db06115d48d63592754bb3fee

includes/Abstracts/Abstract_Ability.php
dkotter: That was the thought, yes, as the Prompt Builder handles that fine. Could change this to return an error that we output, though that would require us to not have the same chaining we have right now on the Prompt Builder as we'd need to check if an error was returned before calling `using_system_instruction`

includes/helpers.php
dkotter: Makes sense, renamed that https://github.com/WordPress/ai/pull/67/commits/145cc150a2422a183305b0c6c463bff2f43ad2c3

includes/Abilities/Title_Generation/Title_Generation.php
```diff
@@ -204,4 +233,41 @@ protected function meta(): array {
...
+			);
+		}
+
+		// Generate the titles using the AI client.
+		return AI_Client::prompt_with_wp_error( '"""' . $context . '"""' )
+			->using_system_instruction( $this->get_system_instruction() )
+			->using_temperature( 0.7 )
+			->using_candidate_count( (int) $candidates )
+			->using_model_preference( ...get_preferred_models() )
+			->generate_texts();
```
JasonTheAdams, line 271: One note I'll make here that I don't think we need to do anything about at this time, is the use of the `Prompt_Builder::is_supporteed_for_text_generation()` method. This is intended to make it clear if there _is not_ a model that supports this prompt's needs, allowing us to do something in that condition.

I want to add a `WP_Ability::is_available()` method in the future for checking ahead of time if an Ability will contextually work. I think this check would fit well there.

JasonTheAdams (approved):
Great work, @dkotter! LGTM! 🎉

Issue #28, opened 2025-10-06
Plugin submission to WordPress.org


jeffpaul:
### Description

Prepare and submit the AI Experiments plugin for publication in the WordPress.org plugin directory to enable broader community testing and feedback.

### Goals

- [x] Verify plugin follows all WPORG guidelines, including naming, security, and licensing.
- [x] Ensure all code is modular, documented, and organized by feature.
- [x] Prepare stable trunk structure for SVN deployment and confirm readme.txt formatting.
- [x] Draft descriptions, banners, icons, and tags for the plugin directory page.
- [x] Run plugin through the Plugin Check Plugin and resolve any issues reported.
- [x] Bump minimum WP version to 6.9 (ensuring this covers any additional concerns from #94)
- [x] Submit to WPORG for review and publish version 0.1.0.
- [x] Before making plugin public on WPORG, add [screenshot(s)](https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/#screenshots)

### Considerations

- Utilizing a GitHub Action for future releases that deploy to WPORG from GitHub releases

### Acceptance Criteria

- Plugin is approved and available in the WPORG directory.
- SVN repository is linked to GitHub trunk and sync verified.
- Plugin listing communicates that all features are experimental and invites testing feedback.

jeffpaul:
@karmatosed do you have time over the next week to work on banner/icon for the plugin?  Alternatively, is there a designer who's expressed interest in contributing here and could help with those images?  If neither, then I can check on pulling design time at Fueled to help on this.

karmatosed:
@jeffpaul yes I will look at doing that today or tomorrow.

karmatosed:
Here are a few options. I went with taking the inspiration of the team logo and adding something instantly gets recognised as experiments. There is one expected color combination of mono, but also two more variations that have different colors. I lean towards those and the cream one.

<img width="5582" height="1591" alt="Image" src="https://github.com/user-attachments/assets/18e449a8-a815-43b8-ae70-9f534dc2d473" />

<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/8761b8f4-ce56-462c-bd09-cef55621fe71" />
<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/61f7b447-4844-43c7-9696-2b5322084a20" />
<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/b3188183-8816-48d0-9e90-23b106de391f" />

[...]

PR #82, opened 2025-11-14
Create `plugin-check.yml`

jeffpaul:
## What?

Relates to #28.

This pull request introduces a new GitHub Actions workflow for automated WordPress plugin checks. The workflow is triggered on pushes and pull requests to the `develop` and `trunk` branches, and it builds the plugin, downloads the build artifact, and runs a plugin compliance check.

New CI workflow for plugin checks:

* Added `.github/workflows/plugin-check.yml` to automate building the plugin and running the WordPress Plugin Check Action on pushes and pull requests to `develop` and `trunk` branches. This ensures plugin quality and compliance before merging changes.

## Why?

This will ensure all future plugin updates stay in line with the WPORG plugin team's expectations.

## Screenshots or screencast 

|Before|After|
|-|-|
|||

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=82%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2082%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).

### DIFF
--- a/.github/workflows/plugin-check.yml
+++ b/.github/workflows/plugin-check.yml
@@ -1,0 +1,39 @@
+name: WordPress Plugin Checks
+
+on:
+  push:
+    branches:
+      - develop
+      - trunk
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  build:
+    uses: WordPress/ai/.github/workflows/build-plugin-zip.yml@trunk
+
+  test:
+    needs: build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Download built zip
+        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+        with:
+          name: ${{ github.event.repository.name }}
+          path: ${{ github.event.repository.name }}
+
+      - name: Display structure of downloaded files
+        run: ls -R
+        working-directory: ${{ github.event.repository.name }}
+
+      - name: Run plugin check
+        uses: wordpress/plugin-check-action@ec9b3fe9beaa76bcc4510b7ba2cb5855a5f80f3f # v1.1.4
+        with:
+          build-dir: ${{ github.event.repository.name }}
+          exclude-directories: 'assets,dist,vendor'
+          ignore-warnings: true


.github/workflows/plugin-check.yml
dkotter, line 1: Worth noting our PHPCS check we already run does pull in the [PHPCS sniffs](https://github.com/WordPress/ai/blob/trunk/phpcs.xml.dist#L83) plugin check uses (this was [copied](https://github.com/WordPress/abilities-api/blob/trunk/phpcs.xml.dist#L77) from how the Abilities API does it). So I'm not sure this check gives us anything new, though nothing wrong with running extra checks

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,39 @@
+name: WordPress Plugin Checks
+
+on:
+  push:
+    branches:
+      - develop
+      - trunk
+  pull_request:
+    branches:
+      - develop
```
dkotter, line 10: Currently `trunk` is our main branch so unless the plan is to switch to `develop` this will need updated

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,39 @@
...
+
```
dkotter, line 11: We'll want to ensure we're setting proper permissions here. The approach we've taken so far is we disable all permissions at the workflow level by doing `permissions: {}` and then we set the proper permissions for each job we run (in this case the `build` and `test` jobs will need those set)

.github/workflows/plugin-check.yml
jeffpaul: Ah ok, is there anything else unique to the plugin check then that's not covered by those sniffs (asked differently, should we just exclude those checks in the action)?

.github/workflows/plugin-check.yml
jeffpaul: My plan was to switch to `develop` after tagging 0.1.0 (or before) as it feels like a safer approach (better?)

.github/workflows/plugin-check.yml
dkotter: Not sure. I know _most_ of what it does is just PHPCS things but there may be other checks it runs 🤷🏻 

I just assumed if both the API Abilities and MCP Adapter repos aren't running this check directly and instead are just running the PHPCS checks it uses, we should do the same and it likely covers everything. But I haven't actually confirmed that

.github/workflows/plugin-check.yml
dkotter: Makes sense. Just noting if we merge this prior to that, this won't actually run on any PRs until after merge

.github/workflows/plugin-check.yml
jeffpaul: Well those other repos aren't WPORG plugins but I'll ask the plugin team if there's anything not already covered by WPCS

jeffpaul:
In chatting with the WPORG Plugin team, it sounds like while there are some things additionally to a WPCS check that are covered by Plugin Check (primarily WPORG repo requirements https://github.com/WordPress/plugin-check/tree/trunk/includes/Checker/Checks/Plugin_Repo) I think we're covered here with a WPCS check and then running Plugin Check before we submit to WPORG (I've already done one check and things looked fine).  As such I'm going to close out this PR.

PR #78, opened 2025-11-14
Add/wporg reqs

jeffpaul:
## What?

Relates to #28.

This pull request updates the plugin's branding and documentation to clarify its purpose as "AI Experiments" for WordPress, improves consistency across files, and updates metadata to match current project standards. The changes mainly affect documentation, plugin headers, and package configuration files.

**Branding and Documentation Updates:**

* Renamed the plugin from "WordPress AI Experiments" to "AI Experiments" throughout `README.md`, `readme.txt`, `ai.php`, `composer.json`, and `package.json` for clarity and consistency. [[1]](diffhunk://#diff-08607de3f8d490d7a94f944d69b5b961055cb74c525e04d51f7b1e9c623ae375L11-R20) [[2]](diffhunk://#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6L1-R13) [[3]](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L2-R3) [[4]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L2-R10)
* Updated plugin descriptions, tags, and changelog entries to refer to "experiments" and "capabilities" instead of "features", and clarified the plugin's experimental nature. [[1]](diffhunk://#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6L1-R13) [[2]](diffhunk://#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6L35-R38) [[3]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L2-R10) [[4]](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L2-R3)
* Added project badges, a banner image, and improved the introductory section in `README.md` for better project visibility and professionalism.

**Metadata and Requirements:**

[...]

PR #83, opened 2025-11-14
Add the Title Generation UI

### LINKED ISSUE #10: Title Generation / Rewriting

### LINKED ISSUE #29: Add UI controls in post editor for rewriting titles

dkotter:
## What?

Partially closes #10

Continues work on the Title Generation Experiment, with initial scaffold work done in #61 and the execution functionality done in #67.

## Why?

We now have a Title Generation Ability that can be triggered to generate title suggestions but that requires you to make direct API requests. This PR adds in the UI that allows users to easily trigger this Ability from the post edit screen.

## How?

- Uses the `Asset_Loader` class within the `Title_Generation` Experiment to load the needed JS file when we're on the post edit screen for a post that has title support.
- Adds in a `TitleToolbar` component and `TitleToolbarWrapper` component (using typescript though can change to plain JS if we want)
- When the title is a block (when using the template preview mode) we modify the Edit component of that block to use the new `TitleToolbar` component
- When the title isn't a block (normal editing mode), we load the `TitleToolbarWrapper` component as a plugin which will then find the title field and inject the `TitleToolbar` component
- In either case, we show a toolbar above the title that has a button in it. When clicked, we show a loading state and when titles are returned, those are loaded in a modal within a textarea. The user can then choose which title (if any) they want to select and optionally make edits before selection. Once they select a title, we set that as the title for the post and trigger an autosave
- Adds in all necessary dependencies

## Testing Instructions

1. Pull changes and run `npm i && npm run build && composer install --no-dev`
2. Go to `Settings > AI Credentials` and ensure you add at least one valid API key
3. Go to `Settings > AI Experiments` and ensure the Title Generation Experiment is enabled
2. Add a new post or edit an existing post
4. Click into the title field and ensure you see a toolbar with a button that says either `Generate` or `Re-generate` depending on if you have an existing title or not
5. Click on that button and ensure a loading state shows
6. Once the request completes, ensure a modal opens with title suggestions. Ensure you can select a title (and optionally edit a title first) and that title is set within the post

## Screenshots

Toolbar with Generate button when no existing title

Toolbar with Re-generate button when there is an existing title

Loading state of the toolbar when the button is clicked

Modal that shows title options, allowing you to edit and select a title

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22https://playground.wordpress.net/plugin-proxy.php?build-ref=trunk%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=83%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2083%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).


src/experiments/title-generation/utils/casing.ts
dkotter, line 1: Note these utilities are only temporary. If we want to keep casing transformation functionality, will likely send that request to AI to do instead of using these. These are just here for testing

src/index.js
dkotter, line 1: Had to add this file otherwise `wp-scripts` wouldn't work

package-lock.json
dkotter, line 1: I know, huge amount of line changes here but is needed to include all the various WordPress components we're using, otherwise linting will complain about importing things that don't exist

.github/workflows/test.yml
```diff
@@ -260,7 +260,6 @@ jobs:
         permissions:
             contents: read
         timeout-minutes: 20
-        if: false  # Temporarily disabled
```
JasonTheAdams, line 263: Temporary for this PR? Or what?

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -0,0 +1,226 @@
...
+			title: selectedTitle,
+		});
+		closeModal();
+		if (!isDirty) {
+			await (dispatch('core') as any).saveEditedEntityRecord(
+				'postType',
+				postType,
+				postId
+			);
+		}
```
JasonTheAdams: I don't think we should be autosaving. In my testing, I meant to change the title, then go back to the posts list without saving. I was really confused as to why the title had changed when I didn't save. Took me a moment to figure out where was no way to change the title this way without it saving.

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Renders the generated title data with editable textareas.
+ *
+ * @param {Object}   props              Component props.
+ * @param {string[]} props.data         The array of titles to render.
+ * @param {Function} props.onDataChange Callback to update the data array.
+ * @param {Function} props.onSelect     Callback when a title is selected.
+ * @return {JSX.Element | null} The rendered data.
+ */
+function RenderData({
```
JasonTheAdams: This reads like a function, not a component. Let's call this something like `TitleOptionsList`.

Same with the rest of this component. We use the term "data" a lot, generically, when we know this a component for listing and modifying title options.

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -0,0 +1,226 @@
...
+							__nextHasNoMarginBottom
+						/>
+						<Button
+							variant="secondary"
+							style={{ marginTop: '15px' }}
+							onClick={() => onSelect(title, i)}
+						>
+							{__('Select', 'ai')}
+						</Button>
+					</FlexItem>
```
JasonTheAdams: Nit: Let's make a `TitleOption` component and use it in this loop. I find it cleans up loops like this and makes things clearly declarative.

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -0,0 +1,226 @@
...
+export default function TitleToolbar(): JSX.Element | null {
+	const postId = select('core/editor').getCurrentPostId();
+	const postType = select('core/editor').getCurrentPostType();
+	const content = select('core/editor').getEditedPostContent();
+	const title = select('core/editor').getEditedPostAttribute('title');
+
+	const { editPost } = useDispatch('core/editor');
+
+	const [isGenerating, setIsGenerating] = useState(false);
+	const [isOpen, setOpen] = useState(false);
```
JasonTheAdams:
```suggestion
	const [isGenerating, setIsGenerating] = useState<boolean>(false);
	const [isOpen, setOpen] = useState<boolean>(false);
```

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+			// Store timeout ID for cleanup
+			const start = () => {
+				if (!timeoutId) {
+					retry();
+				}
+			};
+
+			return start;
+		};
```
JasonTheAdams: Looks like this function can be pulled out of the component entirely since it's not a hook and doesn't reference any component variables.

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
```diff
@@ -0,0 +1,364 @@
...
+			onBlur: () => void,
+			blurDelay: number = 10
+		) => {
+			return {
+				focus: onFocus,
+				blur: () => {
+					setTimeout(onBlur, blurDelay);
+				},
+			};
+		};
```
JasonTheAdams: Looks like this also can be pulled out.

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
```diff
@@ -0,0 +1,364 @@
...
+			const blurEvent = useFocusIn ? 'focusout' : 'blur';
+
+			element.addEventListener(focusEvent, handlers.focus);
+			element.addEventListener(blurEvent, handlers.blur);
+
+			return () => {
+				element.removeEventListener(focusEvent, handlers.focus);
+				element.removeEventListener(blurEvent, handlers.blur);
+			};
+		};
```
JasonTheAdams: This can be pulled out, too.

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
```diff
@@ -0,0 +1,364 @@
...
+			const editorDoc = getEditorDocument();
+			if (editorDoc && !observer) {
+				observer = new MutationObserver(() => {
+					if (
+						!isAttached &&
+						!editorDoc.querySelector('.ai-title-toolbar-wrapper')
+					) {
+						findAndAttachToolbar();
+					}
+				});
```
JasonTheAdams: Let's use the second callback parameter to have the observer disconnect itself when it's no longer needed.

JasonTheAdams (changes requested):
Good work, @dkotter! I left some suggestions. 👍

.github/workflows/test.yml
dkotter: It was disabled when we initially added our default workflows as we didn't have any JS files at the time (and thus it would always fail). Now that this PR is adding JS files, turning this on

src/experiments/title-generation/components/TitleToolbar.tsx
dkotter: I can remove this, though I added based on this comment in the original issue: 

> Upon applying a new title, the system creates a post revision so the change can be undone via the native WordPress revision history

dkotter: Removed in https://github.com/WordPress/ai/pull/83/commits/0390cd4f81a42c767cbadd583691180daff2e9fe

src/experiments/title-generation/components/TitleToolbar.tsx
dkotter: Makes sense, updated in https://github.com/WordPress/ai/pull/83/commits/dea4acbfed3baed20f4bfea27a8809d510e247e0

src/experiments/title-generation/components/TitleToolbar.tsx
dkotter: Updated in https://github.com/WordPress/ai/pull/83/commits/63c9934d5f6b99be6960c34c66352fb30ac0902b

src/experiments/title-generation/components/TitleToolbar.tsx
dkotter: Resolved in https://github.com/WordPress/ai/pull/83/commits/88c8fa2dca1fc00e7983e776fe010931e84a9ba8

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
dkotter: Good catch, fixed in https://github.com/WordPress/ai/pull/83/commits/cf8cdd69d72c836d8cc6781d161a81f6064ada6e

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
dkotter: Fixed in https://github.com/WordPress/ai/pull/83/commits/36fecbac1aa43b1b9bebdd6dce0efc93383c05af

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
dkotter: Fixed in https://github.com/WordPress/ai/pull/83/commits/378116f187322a3262df4da0ed4ef72b53d1abbe

src/experiments/title-generation/components/TitleToolbarWrapper.tsx
dkotter: Good callout, fixed in https://github.com/WordPress/ai/pull/83/commits/ef1b11254543990697bdfc7b3cfa0a97e3a6a2a0

package.json
```diff
@@ -29,17 +29,34 @@
...
   "devDependencies": {
+    "@types/react": "^18.2.47",
+    "@types/react-dom": "^18.2.18",
+    "@types/wordpress__blocks": "^12.5.18",
+    "@wordpress/block-editor": "^15.7.0",
+    "@wordpress/components": "^30.7.0",
+    "@wordpress/compose": "^7.34.0",
+    "@wordpress/data": "^10.34.0",
+    "@wordpress/editor": "^14.34.0",
+    "@wordpress/element": "^6.34.0",
```
felixarntz, line 46: Why are these in `devDependencies`? I know they eventually are served by whatever WordPress Core has for them, but technically I think they're considered dependencies for the production code? Same for a few other `@wordpress/*` dependencies below.

Or the counter question would be: Why do you have only `@wordpress/icons` as a non-dev dependency?

src/experiments/example-experiment/index.tsx
```diff
@@ -6,7 +6,7 @@ import { __ } from '@wordpress/i18n';
 import './index.scss';
 
 const ExampleExperiment = () => {
-	return <div className="test">{ __( 'Example Experiment', 'ai' ) }</div>;
+	return <div className="test">{__('Example Experiment', 'ai')}</div>;
```
felixarntz: Why this "correction"? Before it was using WordPress Coding Standards, and now it isn't.

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -0,0 +1,247 @@
...
+ * Renders a single title option.
+ *
+ * @param {Object}   props          Component props.
+ * @param {string}   props.title    The title value to display.
+ * @param {number}   props.index    The index of this title in the array.
+ * @param {Function} props.onChange Callback to update this title's value.
+ * @param {Function} props.onSelect Callback when this title is selected.
+ * @return {JSX.Element} The rendered title option.
+ */
+function TitleOption({
```
felixarntz: Same here, this is not using WordPress Coding Standards. There is https://www.npmjs.com/package/wp-prettier for exactly this purpose. Since this is a WordPress project, we need to follow WordPress Coding Standards, also in JS/TS.

felixarntz (commented):
@dkotter Overall this is looking good, but a few concerns regarding coding standards and dependencies.

package.json
dkotter: So `@wordpress/icons` isn't something WordPress will serve, we are importing the SVG out of that package so makes sense for this to be a normal dependency. But as you said, the other dependencies are all things WordPress will serve and not things we want bundled. I've included them here just to stop my editor from complaining about importing something that doesn't exist.

That said, I don't think there's any downside to moving this to be normal dependencies as we're using `wp-scripts` to build our code and it will by default remove any WordPress managed dependencies from the final bundle, so if you'd prefer these to be moved, I can do that

src/experiments/example-experiment/index.tsx
dkotter: We use `wp-scripts` here for linting (`npm run lint:js` which will run `wp-scripts lint-js`) and it was flagging this. I've always assumed the linting `wp-scripts` enforces is what WordPress wants (seems it uses https://www.npmjs.com/package/@wordpress/eslint-plugin) but if there's a better/more canonical approach to enforcing JS coding standards, happy to switch up our tooling for that

src/experiments/title-generation/components/TitleToolbar.tsx
dkotter: Commented above but happy to switch to this tool if it makes more sense than the linting `wp-scripts` provides, I've just always defaulted to using `wp-scripts` across all WordPress projects

package.json
felixarntz: Yeah, I think it makes sense to have those kinds of dependencies in non-dev, because they're used by production code. Only tooling dependencies for things like lint, test, etc. (e.g. `@wordpress/scripts` or `@wordpress/env`) I'd put in `devDependencies`.

src/experiments/example-experiment/index.tsx
felixarntz: That is really strange. I don't know if there was a recent update, but I've been using `wp-scripts` myself and it would always enforce the opposite (I.e. WordPress Coding Standards). Is there something else at play here maybe that leads to the regular Prettier taking over the WordPress specific version? 🤔 

This needs some more digging because that's not something that should be changed without broader consensus by the Core developers. A few years ago there was actually a proposal to change it which ended up rejected.

src/experiments/title-generation/components/TitleToolbar.tsx
felixarntz: Definitely +1 to `wp-scripts`, but it's weird that it's flagging WordPress Coding Standards. Let's continue chatting in the above comment thread as needed.

package.json
dkotter: Moved now: https://github.com/WordPress/ai/pull/83/commits/eb8f5233c0c0e0c7d58183bf35e9fa058c0ebc5c

src/experiments/example-experiment/index.tsx
dkotter: Yeah, not sure what's going on. We didn't have an ESLint config file so I was hoping adding that (done in https://github.com/WordPress/ai/pull/83/commits/f755562af83ba596f9675c4bafef8ab8f8b85c1d) would fix it. That did end up flagging additional errors (fixed in https://github.com/WordPress/ai/pull/83/commits/fc4d17891c0f6f016d1809f9603e8d4e0134db09) but still doesn't allow me to add extra spaces around things (I reverted the spacing change in that example file and you can now see the error it's flagging: https://github.com/WordPress/ai/actions/runs/19657884624/job/56298366699?pr=83#step:5:11).

I feel like I'm missing something simple here, a config file or dependency, but I'm not seeing anything obvious

felixarntz: Is there maybe a conflicting version of `prettier` installed that's _not_ the `wp-prettier` fork? If some dependency uses `prettier`, maybe that's what's happening.

[`@wordpress/scripts` uses `wp-prettier`](https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/package.json#L78), so that's all as I thought it would be. But maybe another dependency is pulling in the "real" `prettier`, and that might cause the problem?

dkotter: Okay, think I got this fixed. Had to delete multiple dependencies (and the `package-lock.json` file) but finally got it installing correctly. Doing a quick search, seems the wrong version got installed in https://github.com/WordPress/ai/pull/73 though still not quite sure how. But I've got it working now and I've fixed up all lint errors so things should match proper coding standards now

.prettierrc
```diff
@@ -0,0 +1 @@
+"@wordpress/prettier-config"
```
felixarntz: What does this do? Why is it needed?

felixarntz (approved):
@dkotter LGTM!

There appear to be some transient PHPUnit CI failures though.

src/experiments/title-generation/components/TitleToolbar.tsx
JasonTheAdams: This isn't saving a revision, though, it's saving the whole post. A revision would be fine.

.prettierrc
dkotter: This loads in the WP prettier config to ensure those rules are used when running prettier. But this was part of me experimenting to get things fixed so I've reverted this to how it was initially setup, which removes this file from this PR: https://github.com/WordPress/ai/pull/83/commits/fb62804ac106c6ce3494468d2688d4291774bf1d

JasonTheAdams (approved):
Great work, @dkotter!

dkotter:
> There appear to be some transient PHPUnit CI failures though.

Yeah, the PHPUnit tests are failing on `trunk` since we set our WP minimum to 6.9 in https://github.com/WordPress/ai/pull/78 :(

Getting this error: `Failed to activate plugin. Current WordPress version (6.8.3) does not meet minimum requirements for AI Experiments. The plugin requires WordPress 6.9`

felixarntz:
> > There appear to be some transient PHPUnit CI failures though.
> 
> Yeah, the PHPUnit tests are failing on `trunk` since we set our WP minimum to 6.9 in #78 :(
> 
> Getting this error: `Failed to activate plugin. Current WordPress version (6.8.3) does not meet minimum requirements for AI Experiments. The plugin requires WordPress 6.9`

Why don't we set it to 6.8 for now, so that we don't have these problems? It should still be fine to submit to .org like that. Bumping the minimum requirement to 6.9 probably takes 5 minutes and could be done on release day :)

jeffpaul:
> Why don't we set it to 6.8 for now, so that we don't have these problems? It should still be fine to submit to .org like that. Bumping the minimum requirement to 6.9 probably takes 5 minutes and could be done on release day :)

I went ahead and did that in [77ac717](https://github.com/WordPress/ai/pull/83/commits/77ac717a6cd8a183c531fabc1dedf4bf970e0c9b) and then updated #28 to include:
> Bump minimum WP version to 6.9 (ensuring this covers any additional concerns from https://github.com/WordPress/ai/issues/94)

...as a precursor to submitting to WPORG.

Assuming tests pass here, I'll merge this in.

jeffpaul:
@dkotter perhaps a follow-up PR here to add a docs page like done in https://github.com/WordPress/ai/pull/155/changes#diff-f70665388390657902e73741dea03efcdadc4bd0ceb4432959f8903ae015c042?

PR #84, opened 2025-11-14
Adds basic settings screen

### LINKED ISSUE #25: Create a basic admin settings screen

JasonTheAdams:
## What?
Closes #25 

This introduces a basic settings screen for discovering Experiments, disabling the whole thing, and enabling individual Experiments. It's simple by design, as this is just meant to be a basic starting point as we build things out.

### Experiment Settings

It's possible for an experiment to render its own content within its settings card. The `Abstract_Experiment` class has two methods that can be overloaded: `register_settings()` and `render_settings_fields()`. In the former is when to use the WordPress [register_setting](https://developer.wordpress.org/reference/functions/register_setting/) function, while the latter takes echoed output and places it in the appropriate card, which can include inputs. There's a `Abstract_Experiment::get_field_option_name()` method for registering settings and rendering inputs, as well.

## Why?
Users will need a way of toggling which Experiments they want to use, and configuring Experiment settings.

## How?
This solves this as simply as possible. Later we can do things like differentiate between user-intended and dev-intended Experiments, make the UI much prettier with modern elements, and so forth.

## Testing Instructions
1. Go to Settings > AI Experiments
2. Enable Experiments
3. Toggle individual Experiments
4. In the code, add Experiment settings and try rendering them

### Global settings disabled
image

### Enabled with Experiments toggled
image

### Experiment with no settings but extra content
image

### Experiment with unique setting (saved and persisting)
image

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=84%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%2084%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).


includes/Abstracts/Abstract_Experiment.php
```diff
@@ -148,9 +149,56 @@ final public function is_enabled(): bool {
 		 *
 		 * @since 0.1.0
 		 *
-		 * @param bool $enabled Whether the experiment is enabled.
+		 * @param bool $experiment_enabled Whether the experiment is enabled.
 		 */
-		return (bool) apply_filters( "ai_experiment_{$this->id}_enabled", $enabled );
+		return (bool) apply_filters( "ai_experiment_{$this->id}_enabled", $experiment_enabled );
```
felixarntz: I'm a fan of avoiding to rerun filters and `get_option` calls etc over and over again, if this method is called multiple times. That also helps with weird edge-cases where a filter may return something different at different times in the request lifecycle (which would be `_doing_it_wrong` in the first place).

Maybe keep the `$enabled` class property and use it as an in-memory cache for the result of all this logic?

includes/Abstracts/Abstract_Experiment.php
```diff
@@ -148,9 +149,56 @@ final public function is_enabled(): bool {
...
+	 * Generates a properly namespaced option name for experiment-specific settings.
+	 * Use this when registering and rendering custom settings fields to ensure
+	 * consistent naming across the plugin.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param string $option_name The base option name (e.g., 'api_key', 'temperature').
+	 * @return string The fully namespaced option name.
+	 */
+	protected function get_field_option_name( string $option_name ): string {
```
felixarntz: Best to make this `final` probably to prevent unexpected usage.

includes/Admin/Settings_Page.php
```diff
@@ -0,0 +1,270 @@
...
+	/**
+	 * Initializes the settings page hooks.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return void
+	 */
+	public function init(): void {
+		add_action( 'admin_menu', array( $this, 'register_menu' ) );
+		add_action( 'admin_init', array( $this, 'register_settings' ) );
```
felixarntz: Worth noting that technically `register_setting()` has a different scope than `add_settings_section()` and `add_settings_field()`:
* `register_setting()` is generally applicable. While it originally started as part of the (Admin) Settings API, it is no longer scoped to that. It's also relevant for REST API or, FWIW, anywhere in WordPress since more or less at any point the registered option could be fetched or tweaked (at which point configuration like `default` is relevant). For that reason, I think settings are best registered on `init`, not `admin_init`.
* The other functions that are actually part of the (Admin) Settings API are fine to call on `admin_init`, although they really only matter on the concrete admin page, so it's wasteful to call them on every admin request: The recommended pattern there is to use the `$hook_suffix` that's returned by `add_submenu_page()` (or here: `add_options_page()`), and then hook those screen-specific callbacks to the `"load-{$hook_suffix}"` action.

**Update:** Revising my statement after further reviewing how you render the settings fields. I suppose you're not using the Settings API at all, which I guess is fine, given it is very limited in the UI it creates. In that case, I'd say simply change this to be hooked on `init`. Technically speaking, it would also be ideally outside of a `Settings_Page` class since, as mentioned, setting registration is relevant more or less everywhere, not only in context of the settings page. But that shouldn't be a blocker, we could still change it later.

includes/Admin/Settings_Page.php
```diff
@@ -0,0 +1,270 @@
...
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_styles' ) );
```
felixarntz: Similar to what I'm saying above, this is not an ideal pattern: I realize you're checking for the specific screen in the callback and bail early if it's not that, but better would be to really only hook in this method on the specific screen: `"load-{$hook_suffix}"` is your friend :)

includes/Admin/Settings_Page.php
```diff
@@ -0,0 +1,270 @@
...
+	 * @param string $hook_suffix The current admin page hook suffix.
+	 * @return void
+	 */
+	public function enqueue_styles( string $hook_suffix ): void {
+		if ( 'settings_page_' . self::PAGE_SLUG !== $hook_suffix ) {
+			return;
+		}
+
+		// Enqueue WordPress components styles for block editor UI.
+		wp_enqueue_style( 'wp-components' );
```
felixarntz: No need to enqueue this, since it's already a dependency of the style below. It'll be automatically enqueued.

includes/Settings/Settings_Page.php
```diff
@@ -0,0 +1,270 @@
...
+									$disabled_class     = ! $global_enabled ? 'ai-experiments__item--disabled' : '';
+									?>
+									<li class="ai-experiments__item <?php echo esc_attr( $disabled_class ); ?>">
+										<div class="ai-experiments__item-header">
+											<label class="components-toggle-control">
+												<input
+													type="checkbox"
+													name="<?php echo esc_attr( $experiment_option ); ?>"
+													id="<?php echo esc_attr( $experiment_option ); ?>"
+													value="1"
```
felixarntz, line 192: If there is a description for the field, there also needs to be an `aria-describedby` attribute for accessibility.

includes/Admin/Settings_Page.php
```diff
@@ -0,0 +1,270 @@
...
+													<?php checked( $experiment_enabled ); ?>
+													<?php disabled( ! $global_enabled ); ?>
+												/>
+												<span>
+													<strong><?php echo esc_html( $experiment->get_label() ); ?></strong>
+												</span>
+											</label>
+										</div>
+										<?php if ( $experiment->get_description() ) : ?>
+											<p class="description">
```
felixarntz: See above, this needs an ID attribute that can be referenced by the `aria-describedby` attribute.

includes/Admin/Settings_Page.php
```diff
@@ -0,0 +1,270 @@
...
+	 * Registers all settings for experiments.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return void
+	 */
+	public function register_settings(): void {
+		// Register the global toggle.
+		register_setting(
+			'ai_experiments',
```
felixarntz: Can we use a class constant for this and all its references?

felixarntz (commented):
@JasonTheAdams Looks great! A few minor points of feedback.

includes/Admin/Settings_Page.php
```diff
@@ -0,0 +1,270 @@
...
+													<?php disabled( ! $global_enabled ); ?>
+												/>
+												<span>
+													<strong><?php echo esc_html( $experiment->get_label() ); ?></strong>
+												</span>
+											</label>
+										</div>
+										<?php if ( $experiment->get_description() ) : ?>
+											<p class="description">
+												<?php echo esc_html( $experiment->get_description() ); ?>
```
mehul0810: I believe, we should support anchor links in the description just to ensure that we can link back to a documentation page where necessary for the end users. We can see anchor links in the description for existing WP admin settings as you can check under Settings > AI Credentials or Settings > Reading or any other admin page.

Here is an screenshot example of how the current settings is already supporting the anchor links along with the description text:
CleanShot 2025-11-17 at 19 55 38@2x

includes/Abstracts/Abstract_Experiment.php
```diff
@@ -148,9 +149,56 @@ final public function is_enabled(): bool {
...
+	 * This is called after the experiment's main toggle control.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return void
+	 */
+	public function render_settings_fields(): void {
+		// Default implementation does nothing.
+		// Child classes can override to render custom settings UI.
+	}
```
mindctrl, line 206: Just thinking out loud here, not trying to seek perfection, but this made me think of the idea of Experiments providing a settings schema, via something like `settings_schema()`, and then having `Settings_Page` register and render any provided settings, so the Experiments themselves wouldn't have to handle all the registration and display logic. Would provide more consistency in rendering, and the Experiment would only need to provide the data structure.

At any rate, I really like the simplified architecture here.

includes/Admin/Settings_Page.php
felixarntz: +1, let's use `wp_kses()` and allow a few common inline tags, at least anchors.

includes/Abstracts/Abstract_Experiment.php
JasonTheAdams: Good idea. Resolved in c8ebb18

includes/Admin/Settings_Page.php
JasonTheAdams: That's a good note, and you're right, I went about this in a more quick-and-dirty way, knowing we can improve on this later, especially when we have a clearer sense of direction.

Partially resolved (not including pulling out setting registration) in 1aba23d

includes/Admin/Settings_Page.php
JasonTheAdams: Resolved in 1aba23d

includes/Admin/Settings_Page.php
JasonTheAdams: Resolved in 1aba23d

includes/Settings/Settings_Page.php
JasonTheAdams: Good catch! Resolved in 1cb1b13

includes/Admin/Settings_Page.php
JasonTheAdams: Resolved in 1cb1b13

includes/Admin/Settings_Page.php
JasonTheAdams: Good idea! Resolved in 1630e55

includes/Abstracts/Abstract_Experiment.php
JasonTheAdams: Oh for sure! I think something like that is very much worth considering. This PR is erring on the side of speed. I didn't want to make any schemas that we're committing to until we have a better idea of what we want the settings to ultimately become.

includes/Admin/Settings_Page.php
JasonTheAdams: Resolved in 9dcaa53

includes/Admin/Settings_Page.php
JasonTheAdams: I like the idea of separating the settings page from registration, so I did that in 9dcaa53

includes/Abstracts/Abstract_Experiment.php
JasonTheAdams: Good idea! Resolved in fca4b0b

jeffpaul:
In closing out #51 in favor of this PR, let's make sure whoever makes the merge commit here that we capture the co-authored by details in the comment above as well as those credited for the effort in https://github.com/WordPress/ai/pull/51#issuecomment-3448775039:

```
Unlinked contributors: mohamed.khaled@9hdigital.com.

Co-authored-by: Ref34t <mokhaled@git.wordpress.org>
Co-authored-by: mindctrl <mindctrl@git.wordpress.org>
Co-authored-by: dkotter <dkotter@git.wordpress.org>
Co-authored-by: jeffpaul <jeffpaul@git.wordpress.org>
Co-authored-by: justlevine <justlevine@git.wordpress.org>
Co-authored-by: swissspidy <swissspidy@git.wordpress.org>
Co-authored-by: felixarntz <flixos90@git.wordpress.org>
```

includes/Settings/Settings_Page.php
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	/**
+	 * Handles the page load event for the settings page.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return void
+	 */
+	public function on_page_load(): void {
```
felixarntz: Nit-pick: Give this a more common method name, not just name it based on the idea that it's a callback (many things in WordPress are).

I think something like `load_page` or `init_page` would work well, and fit perfectly with the existing `render_page` method.

felixarntz (approved):
@JasonTheAdams Excellent, LGTM!

includes/Settings/Settings_Page.php
JasonTheAdams: Resolved in 8ed24b1

PR #85, opened 2025-11-17
Feat: wp playground integration

### LINKED ISSUE #79: Add WP Playground live preview integration

Enable plugin testing with WordPress Playground.

So that folks finding the plugin at its future location at https://wordpress.org/plugins/ai can benefit from the WordPress Playground "Live Preview" option as well as providing a similar option here in the GitHub repo, let's look to do the following:
- [ ] add a blueprints.json file to the repo including any sample post/page and media to support a decent test experience (note that if the test data would help with the Playground test link on PRs that we should look to get that added into [that GH action](https://github.com/WordPress/ai/blob/trunk/.github/workflows/pull-request-comments.yml) as well)
- [ ] add a "WordPress Playground Demo" badge at the top of the README.md file ([see example here](https://github.com/10up/classifai/blob/224d461d61931774950c9850d45701e2fdbf23cf/README.md?plain=1#L5))
- [ ] once the plugin is published/live on WPORG, go through the steps to enable the "Live Preview" option there

dhruvang21:
## What?
Closes https://github.com/WordPress/ai/issues/79

## Why?
To Enable plugin testing with WordPress Playground.

## How?
- By adding the blueprint.json file


README.md
```diff
@@ -1,8 +1,9 @@
 # WordPress AI Experiments
 
+Experimental AI features for WordPress. Modular framework for testing AI capabilities.
```
dkotter: Was this meant to be added here?

.github/blueprints/blueprint.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://playground.wordpress.net/blueprint-schema.json",
+  "landingPage": "/wp-admin/admin.php?page=wp-ai-client",
```
dkotter: This page doesn't exist, should be:
```suggestion
  "landingPage": "/wp-admin/options-general.php?page=ai-experiments",
```

.github/blueprints/blueprint.json
```diff
@@ -0,0 +1,23 @@
...
+  "preferredVersions": {
+    "php": "8.2",
```
dkotter: Any reason to not use 8.4?

.github/blueprints/blueprint.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "features": {
+    "networking": true
+  },
+  "steps": [
+    {
+      "step": "installPlugin",
+      "pluginData": {
+        "resource": "url",
+        "url": "https://github-proxy.com/proxy/?repo=WordPress/ai&branch=trunk"
```
dkotter: So this won't work as we don't keep the built version of the plugin in `trunk`. We'll either need to add a branch that has the built files (something like `stable`) or just easiest to wait until this plugin is on WordPress.org and we can point there

dkotter (changes requested):

README.md
dhruvang21: This code got pushed when i rebased. so no idea

.github/blueprints/blueprint.json
dhruvang21: @dkotter yes i was waiting for the reviewers to guide me on this.

jeffpaul: Let's update to just point to the `ai` slug on WPORG and the Playground badge/button will start working once that plugin is live there.

dhruvang21: @jeffpaul updated URL or we could also do it like this 

```
"pluginData": {
        "resource": "wordpress.org/plugins",
        "slug": "ai"
},
```

dkotter: Yes, I'd suggest we follow this format rather than supply the URL

jeffpaul:
@dkotter without being able to validate the WPORG link works, feels like this is good enough to merge and come back to if there's an issue later?

README.md
dkotter: I think this might have been a bad merge when rebasing then as I'm not seeing this anywhere else. I'd suggest just removing this

dkotter:
> @dkotter without being able to validate the WPORG link works, feels like this is good enough to merge and come back to if there's an issue later?

Still a few unresolved comments (https://github.com/WordPress/ai/pull/85#discussion_r2578941885 and https://github.com/WordPress/ai/pull/85#discussion_r2577508492) but once those are fixed, yes, I see no issue with merging this in

README.md
jeffpaul: Done in [d2ee407](https://github.com/WordPress/ai/pull/85/commits/d2ee407aea6ed50817a644b94a27a47f6272775e).

.github/blueprints/blueprint.json
jeffpaul: Done in [392a0e2](https://github.com/WordPress/ai/pull/85/commits/392a0e26f20a4a442f7a5673e0c21bd4f48f8080).

.github/blueprints/blueprint.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://playground.wordpress.net/blueprint-schema.json",
+  "landingPage": "/wp-admin/options-general.php?page=ai-experiments",
+  "preferredVersions": {
+    "php": "8.4",
```
jeffpaul: A little nervous about this, would feel better dropping to 8.2 unless we KNOW we've tested and things work on 8.4?

jeffpaul: Went ahead and dropped to 7.4 after discussing with @dkotter who has tested in 8.3, but given the plugin minimum is 7.4 we'll start there and can raise in WP Playground as we raise the plugin minimum.

jeffpaul (approved):

dkotter (approved):

PR #86, opened 2025-11-17
Register some utility Abilities

dkotter:
## What?

Adds in two new, what I'm calling "Utility Abilities" (open to naming suggestions). These are more simple Abilities as compared to our current Title Generation Ability. These take in a post ID and will return some information about that post, like title, content, author or terms. We then use these Abilities when building our post context.

## Why?

While this does make our code slightly more complex and isn't the approach I'd generally take, adding these in does accomplish a few things:

1. Showcases a different, more simple approach to registering Abilities (compared to passing in a full class object)
2. Showcases how you can call Abilities via PHP
3. Showcases how you can combine multiple Abilities together and even call an Ability from another Ability

I originally had individual Abilities for each piece of context (like post content, post title, post author, etc) but after some feedback, trimmed it down to two Abilities, one to get all post details (which you can optionally filter to just the fields you want) and one to get all post terms (which you can also filter to just the taxonomies you want).

## How?

- Adds a new Abilities class, `Utilities`, that registers two new abilities: `get-post-details` and `get-post-tems`. These Abilities aren't currently registered with the REST API
- Modifies our existing `get_post_context` helper function to instead use these new Abilities to build up that context

## Testing Instructions

There isn't any UI yet to trigger things so you'd have to make a direct request to the Title Generation Abilities endpoint to see these in action

```bash
--data '{
    "input": {
        "post_id": 1,
        "candidates": 1
    }
}'
```


includes/Abilities/Utilities/Utilities.php
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	/**
+	 * Registers the get-content ability.
+	 *
+	 * @since 0.1.0
+	 */
+	private function register_get_content_ability(): void {
+		wp_register_ability(
+			'ai/get-content',
```
JasonTheAdams:
```suggestion
			'ai/get-post-content',
```

includes/Abilities/Utilities/Utilities.php
```diff
@@ -0,0 +1,404 @@
...
+			array(
+				'label'               => esc_html__( 'Get post content', 'ai' ),
+				'description'         => esc_html__( 'Get the content of a post based on the post ID.', 'ai' ),
+				'category'            => 'ai-experiments',
```
JasonTheAdams: Is there a constant somewhere, likely that this category is registered, that we can use?

includes/Abilities/Utilities/Utilities.php
```diff
@@ -0,0 +1,404 @@
...
+							'description' => esc_html__( 'The ID of the post to get the content of.', 'ai' ),
+						),
+					),
+					'required'   => array( 'post_id' ),
+				),
+				'output_schema'       => array(
+					'type'        => 'string',
+					'description' => esc_html__( 'The content of the post.', 'ai' ),
+				),
+				'execute_callback'    => static function ( $input ) {
```
JasonTheAdams: We can type this. Same in the other abilities.
```suggestion
				'execute_callback'    => static function ( array $input ) {
```

includes/Abilities/Utilities/Utilities.php
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	/**
+	 * Registers the get-title ability.
+	 *
+	 * @since 0.1.0
+	 */
+	private function register_get_title_ability(): void {
+		wp_register_ability(
+			'ai/get-title',
```
JasonTheAdams:
```suggestion
			'ai/get-post-title',
```

JasonTheAdams (changes requested):
HI @dkotter! This is a partial review. I'm curious why you decided to have a bunch of small abilities as opposed to a single `get-post-details` ability (the terms ability would still be separate)? 

Also, let's add the meta parameter so this available via MCP.

dkotter:
@JasonTheAdams 

> I'm curious why you decided to have a bunch of small abilities as opposed to a single `get-post-details` ability (the terms ability would still be separate)?

I went back and forth on this so happy to change this up (mentioned in the PR description):

`I debated between registering a single new Ability (something like get_post_context) compared to registering Abilities for each individual piece of context. The former would be simpler but less flexible. Happy to go with that approach though if we think that's better.`

Main thought was keeping these separate allow you to choose what details you want instead of just getting everything but also feels overcomplicated so I can change that to be a `get-post-details` Ability instead (maybe with optional filtering in the input schema so you can limit what is returned)

includes/Abilities/Utilities/Utilities.php
dkotter: This ability no longer exists so resolving this

includes/Abilities/Utilities/Utilities.php
dkotter: This ability no longer exists so resolving this

includes/Abilities/Utilities/Utilities.php
dkotter: There wasn't but added one now: https://github.com/WordPress/ai/pull/86/commits/16b3f27cadf7efb5820a2e6fffa6060208a021ea

dkotter:
@JasonTheAdams Modified things now to have two abilities: `get-post-details` and `get-post-terms`, let me know if that looks better.

> Also, let's add the meta parameter so this available via MCP.

Can you point me to what parameter this needs? I'm not seeing anything in looking through the documentation for the Abilities API or MCP Adapter but likely just missing it

includes/Abilities/Utilities/Posts.php
JasonTheAdams, line 1: Let's put this in a "Posts" namespace and class, as it's more specific than "utilities" and more accurate.

includes/Abilities/Utilities/Utilities.php
```diff
@@ -0,0 +1,250 @@
...
+	 * Registers the get-post-details ability.
+	 *
+	 * @since 0.1.0
+	 */
+	private function register_get_post_details_ability(): void {
+		wp_register_ability(
+			'ai/get-post-details',
+			array(
+				'label'               => esc_html__( 'Get post details', 'ai' ),
+				'description'         => esc_html__( 'Get the details of a post based on the post ID. Optionally limit the details to specific fields.', 'ai' ),
```
JasonTheAdams:
```suggestion
				'description'         => esc_html__( 'Get the details of a post based on the post ID. Optionally, limit the details to specific fields.', 'ai' ),
```

includes/Abilities/Utilities/Utilities.php
```diff
@@ -0,0 +1,250 @@
...
+					),
+					'required'   => array( 'post_id' ),
+				),
+				'output_schema'       => array(
+					'type'       => 'object',
+					'properties' => array(
+						'details' => array(
+							'type'        => 'array',
+							'description' => esc_html__( 'An array of post details.', 'ai' ),
+						),
```
JasonTheAdams: It doesn't look like it's returning an `array( 'details' => array( ... ) )`, but instead an array tentatively containing `content`, `title`, and such. I think we should list out each parameter, too, as that especially tells systems like AI models what exactly to expect. For example, an AI might assume it's receiving `post_content` instead of `content`.

includes/Abilities/Utilities/Utilities.php
```diff
@@ -0,0 +1,250 @@
...
+					),
+					'required'   => array( 'post_id' ),
+				),
+				'output_schema'       => array(
+					'type'       => 'object',
+					'properties' => array(
+						'terms' => array(
+							'type'        => 'array',
+							'description' => esc_html__( 'An array of WP_Term objects assigned to the post.', 'ai' ),
+						),
```
JasonTheAdams: Like above, I know it's a lot but let's spell out what the shape of each term is.

JasonTheAdams (changes requested):
Yeah, this is better, @dkotter!

Also, here's the doc on how to connect an Ability to MCP: https://github.com/WordPress/mcp-adapter/blob/trunk/docs/guides/creating-abilities.md

includes/Abilities/Utilities/Posts.php
dkotter: Done: https://github.com/WordPress/ai/pull/86/commits/393f50862c20e541a0b9e97889f24e5db53fc4b2

includes/Abilities/Utilities/Utilities.php
dkotter: Done: https://github.com/WordPress/ai/pull/86/commits/501276ac677320776f1549bf79f1a981d70b96ac

includes/Abilities/Utilities/Utilities.php
dkotter: Done: https://github.com/WordPress/ai/pull/86/commits/1cf58eab6397c8f8c36ad3e6a1fe9c67bc9f88ef

dkotter:
> Also, here's the doc on how to connect an Ability to MCP: https://github.com/WordPress/mcp-adapter/blob/trunk/docs/guides/creating-abilities.md

Ah, thanks for that. The main readme in that repo claims:

> All WordPress abilities registered via wp_register_ability() are automatically available

So wasn't sure we needed to do anything else.

Added this in https://github.com/WordPress/ai/pull/86/commits/78e35e0f332e24a98e51bb14ca023cae336a7f60

JasonTheAdams (approved):
Great work, @dkotter! LGTM! 🎉

PR #87, opened 2025-11-17
Add an `Asset_Loader` class

dkotter:
## What?

Adds a new `Asset_Loader` class and sets up our build command to work.

## Why?

This functionality is currently in #83 but feels like it will be useful for other PRs, so extracting that out from there in hopes of getting this piece merged sooner.

We use `wp-scripts` to build our assets, which outputs both the compiled asset plus an `.asset.php` file. From this file you can pull out the required dependencies as well as a version number. The `Asset_Loader` class will do this for you automatically instead of having to do this anywhere we want to enqueue things.

## How?

- Adds a new `Asset_Loader` class with three methods: `enqueue_script`, `enqueue_style` and `localize_script`
- Adds in a proper build command, `npm run build`
- Adds in the needed `src` directory with an empty `index.js` file, so our build command works

## Testing Instructions

Locally run `npm i && npm run build` and ensure no errors are output and you see a proper `build` directory added

PR #88, opened 2025-11-17
Add basic webpack config

dkotter:
## What?

This is a followup to #87 that provides a basic webpack config file as well as examples of how assets (CSS, JS/TS files) can be managed.

## Why?

While we added in an `Asset_Loader` class in #87, we didn't provide the actual implementation steps needed to build files. This PR does that.

## How?

- Add in a basic `webpack.config.js` file that shows how to customize entry points
- Add in some examples within `/src/` to show how things like Experiments or admin-related CSS could be loaded

## Testing Instructions

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "description": "WordPress AI Experiments",
   "license": "GPL-2.0-or-later",
   "files": [
+    "build",
     "includes",
     "vendor",
     "README.md",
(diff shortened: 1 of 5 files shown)


JasonTheAdams (approved):
Thanks for this, @dkotter! Now we can both pull this down and keep going! 😄

PR #100, opened 2025-11-25
Disable Experiments when no valid AI credentials are found

### LINKED ISSUE #30: Ensure graceful fallback behavior when no provider is configured
### Description

Establish a consistent and user-friendly fallback experience for all AI Experiments plugin features when no AI provider is detected or configured.  This will ensure users can explore the plugin safely without encountering errors or confusing states, and will provide clear guidance on what to do next.

Goals

- [ ] Implement a centralized fallback handler (for example, `AI_Fallback_Manager`) that determines when no provider is available and returns standard responses to feature modules.
- [ ] Ensure the admin UI clearly communicates the missing provider state with actionable guidance (for example, “No AI provider is currently connected. Configure one in Settings > AI Experiments.”).
- [ ] Each feature module should gracefully degrade by disabling related UI controls or showing contextual messages instead of error notices.
- [ ] Integrate with pre-configured provider detection so fallback only triggers when no available providers exist.
- [ ] Document fallback expectations for both developers (API responses, error objects) and users (UI messaging).

### Design Considerations

- Maintain alignment with the “progressive provider selection” model.  Fallbacks should feel intentional, not like an error condition.
- Use a shared interface so features can easily check availability (`AI_Connection_Manager::is_provider_available()` or similar).
- Avoid repeated user prompts. The plugin should surface provider configuration only in the appropriate settings location.

### Acceptance Criteria

- If no provider is configured, all features display consistent and non-blocking messaging.
- No PHP or JavaScript errors occur when invoking AI functionality without a provider.
- A test helper or mock provider exists to validate fallback behavior.
- Developer and user documentation clearly outline the fallback behavior.

dkotter:
## What?

Closes #30 

Ensure we have valid AI credentials before we allow any Experiment to be used

## Why?

Right now you can enable any Experiment even if you haven't added any AI credentials. This can lead to confusion for users that then try and use that Experiment and it doesn't work properly. This is also the case if you add invalid credentials.

The AI Credentials setting screen is coming from the WP AI Client package and doesn't do any validation when credentials are saved. This PR adds that validation, when someone lands on our settings screen, and an error message will be shown if they either have no credentials set up or their credentials aren't valid. They are then prompted to go add valid credentials before proceeding.

## How?

- Adds two new helper functions: `has_ai_credentials`, `has_valid_ai_credentials`
- Run those functions when we render our setting screen, showing a proper error message and hiding all other settings if either of those return false
- When we check to see if an Experiment is enabled, run `has_valid_ai_credentials` and return false appropriately, so an Experiment won't work if we don't have valid credentials

> [!NOTE]  
> If we ever introduce an Experiment that doesn't need to use AI, we'll want to update this logic.

## Testing Instructions

1. In a clean environment, go to our settings page. Ensure an error message is shown that prompts you to go add credentials and that no other settings are shown
2. Add invalid credentials and come back. Ensure a different error message is shown that lets you know your credentials aren't valid. Ensure settings still aren't shown
3. Add at least one valid credential. Ensure our settings page now works
4. After enabling an Experiment, change the valid credentials to invalid and ensure the Experiment no longer works
5. Remove credentials all together and ensure the Experiment no longer works

## Screenshots or screencast 

AI Experiments setting screen with no AI credentials error message shown

AI Experiments setting screen with no valid AI credentials error message shown


includes/helpers.php
```diff
@@ -155,3 +157,61 @@ function get_preferred_models(): array {
...
+	if ( null !== $valid ) {
+		return (bool) $valid;
+	}
+
+	// If we have no AI credentials, return false.
+	if ( ! has_ai_credentials() ) {
+		return false;
+	}
+
+	return AI_Client::prompt( 'Test' )->is_supported_for_text_generation();
```
felixarntz: This technically doesn't check whether AI credentials are valid. It only checks whether at least one model is available to satisfy the prompt requirements.

Now, in practice, all 3 default provider implementations require API keys even to call the endpoint to list the available models, which happens as part of that. So I suppose you can still use this here as a way to check. _However_, that'll only mean the API request that happens at some point will fail and throw an exception, so the code here doesn't cater for that.

I'd suggest the following:
* Immediate action: Wrap this in a try catch and return `false` if it throws.
* Separately: Consider whether/how we want to more formally enable checking for credential validity in the PHP/WP AI Client packages or whether this technique here is how we think it should be done.

includes/helpers.php
```diff
@@ -155,3 +157,61 @@ function get_preferred_models(): array {
...
+	 * @param bool|null $has_valid_credentials Whether valid credentials are available. Return null to use default check.
+	 * @return bool|null True if valid credentials are available, false otherwise, or null to use default check.
+	 */
+	$valid = apply_filters( 'ai_pre_has_valid_credentials_check', null );
+	if ( null !== $valid ) {
+		return (bool) $valid;
+	}
+
+	// If we have no AI credentials, return false.
+	if ( ! has_ai_credentials() ) {
```
felixarntz: Shouldn't this be called before the filter? If you don't have _any_ AI credentials, you certainly don't have _valid_ AI credentials.

It makes sense to me to allow filtering/overriding the validity check logic, but the prerequisite should always be that there are AI credentials at all.

Otherwise, this also throws a wrench in things when e.g. for the admin notice you're also checking `has_ai_credentials()`, which would lead to a misleading notice if that filter here was overriding the result to be `true`.

felixarntz (changes requested):
@dkotter Looks solid, only two things I think need to be fixed.

includes/helpers.php
dkotter: Added that try/catch here: https://github.com/WordPress/ai/pull/100/commits/12b716a5f52efff72b9169ed41d72a6bd4e8c0a6, let me know if that looks better

includes/helpers.php
dkotter: Yep, makes sense. Moved this filter down: https://github.com/WordPress/ai/pull/100/commits/12b716a5f52efff72b9169ed41d72a6bd4e8c0a6

includes/Settings/Settings_Page.php
```diff
@@ -125,6 +128,30 @@ public function render_page(): void {
 		<div class="wrap">
 			<h1><?php echo esc_html( get_admin_page_title() ); ?></h1>
 
+			<?php
+			// If we don't have proper credentials, show an error message and return early.
+			if ( ! has_ai_credentials() || ! has_valid_ai_credentials() ) {
```
felixarntz: Can reduce this to just:
```suggestion
			if ( ! has_valid_ai_credentials() ) {
```

It implies the former.

felixarntz (approved):
@dkotter LGTM!

One minor point, but not a blocker.

includes/Settings/Settings_Page.php
dkotter: Ah, yeah, nice catch. Simplified that in https://github.com/WordPress/ai/pull/100/commits/a6e50c2b51e9aa7a53c7b9bd86f9130b3f494c42

PR #101, opened 2025-11-26
Add AI_Service layer for centralized AI operations

### LINKED ISSUE #26: Integrate WP AI Client SDK with a simple Service Provider setup
### Description

Connect the AI Experiments plugin with the WP AI Client SDK (and its underlying PHP AI Client SDK) to establish a unified service layer that experimental features can use to communicate with AI providers.  This integration will form the foundation for running AI-powered tasks, managing provider configuration, and handling fallbacks when no provider is available.

### Goals

- [x] Add the WP AI Client SDK as a Composer dependency.
- [ ] Initialize a dedicated service class (for example, `AI_Connection_Manager`) to manage provider configuration and requests through the WP AI Client SDK.
- [ ] Expose a consistent interface for experimental features to register, request, and run AI tasks using the SDK.
- [ ] Include simple detection and handling for available providers, supporting both user-configured and pre-configured setups.
- [ ] Provide hooks and filters for developers or hosts to override configuration behavior.
- [ ] Add inline documentation and example usage demonstrating how features interact with the connection layer.

### Design Considerations

- Integration should remain centralized so that each feature module can use the SDK through a shared connection layer rather than duplicating logic.
- Maintain alignment with the plugin’s encapsulation principles: each feature should be able to depend on the connection service without introducing tight coupling.
- Use dependency injection and clean class structure to ensure the system is testable, extendable, and maintainable.

### Acceptance Criteria

- WP AI Client SDK initializes successfully and can complete a basic test request when a provider is configured.
- Plugin detects and handles the absence of a configured provider gracefully.
- Feature modules can call the service to perform AI tasks using a standard method (for example, `AI_Service::run_text_task()`).
- Integration is documented and passes WordPress coding standards.

caseymanos:
## What?

Closes #26

Implements a centralized AI service layer (`AI_Service`) that provides a consistent interface for experimental features to communicate with AI providers through the WP AI Client SDK.

## Why?

The plugin needs a unified service layer so that experimental features can communicate with AI providers without duplicating logic or directly coupling to the `AI_Client` SDK. This provides:

- A single point of configuration for AI provider requests
- Graceful handling when no provider is configured
- Hooks and filters for developers/hosts to customize behavior
- A foundation for future AI-powered features

## How?

**New files:**
- `includes/Services/AI_Service.php` - Singleton service class with `generate_text()`, `generate_texts()`, `create_prompt()`, `is_available()` methods
- `includes/Services/Contracts/AI_Service_Interface.php` - Interface contract ensuring consistent implementation
- `tests/Integration/Includes/Services/AI_ServiceTest.php` - Integration tests (26 test cases)

**Modified files:**
- `includes/bootstrap.php` - Initializes AI_Service after `AI_Client::init()`
- `includes/helpers.php` - Adds `get_ai_service()` helper function

**Hooks and filters:**
| Hook | Type | Description |
|------|------|-------------|
| `ai_service_available` | Filter | Override provider availability detection |
| `ai_service_initialized` | Action | Fires when AI service initializes |
| `ai_service_prompt_builder` | Filter | Modify prompt builder before generation |

## Testing Instructions

1. Ensure you have an AI provider configured (Settings > AI Credentials)
2. Run automated tests: `npm run test:php`
3. Verify the service via WP-CLI:
```bash
npm run wp-env -- run cli wp eval '
$service = WordPress\AI\get_ai_service();
echo "Available: " . ($service->is_available() ? "yes" : "no") . "\n";
if ($service->is_available()) {
    $result = $service->generate_text("Say hello in 3 words");
    echo "Result: " . $result . "\n";
}
'
```
4. Test without credentials by temporarily removing them - `is_available()` should return `false`
5. Verify hooks work by adding a filter:
```php
add_filter( 'ai_service_available', '__return_false' );
// Service should now report unavailable
```

Not applicable - this PR adds backend service infrastructure only with no UI changes.

## Screenshots or screencast

Not applicable - this PR adds backend service infrastructure only with no UI changes.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-101-9a4845d797f935c0dfc5d78870fbd5aa3318b3ba.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
@caseymanos thanks for the PR!  I see you note this closes #26, but looking to double confirm that the various goals in that issue description are all covered by your work here; correct?

includes/Services/Contracts/AI_Service_Interface.php
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Interface for AI service implementations.
+ *
+ * Provides a consistent API for experimental features to interact with AI providers
+ * without directly coupling to the underlying AI Client implementation.
+ *
+ * @since 0.1.0
+ */
+interface AI_Service_Interface {
```
felixarntz: Why an interface? Seems a bit premature at this point I'd say.

* There's only one `AI_Service` implementation.
* The interface mandates having methods like `generate_text()` which I'm not sure is what makes sense for an interface / contract. There could also be an AI service for generating images for example.

For those reasons, I think it's best to remove this interface entirely. There's no real need for it, at least not at this point.

includes/Services/AI_Service.php
```diff
@@ -0,0 +1,311 @@
...
+class AI_Service implements AI_Service_Interface {
+
+	/**
+	 * Default temperature for text generation.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @var float
+	 */
+	public const DEFAULT_TEMPERATURE = 0.7;
```
felixarntz: Why is this the default temperature? More importantly: Why does it even make sense to set a default temperature?

includes/Services/AI_Service.php
```diff
@@ -0,0 +1,311 @@
...
+	 *
+	 * @param string               $prompt  The prompt to generate text from.
+	 * @param array<string, mixed> $options Optional. Generation options. {
+	 *     @type string $system_instruction System instruction for the AI.
+	 *     @type float  $temperature        Temperature for generation (0.0-2.0).
+	 *     @type int    $max_tokens         Maximum tokens to generate.
+	 * }
+	 * @return string|\WP_Error The generated text or WP_Error on failure.
+	 */
+	public function generate_text( string $prompt, array $options = array() ) {
```
felixarntz: I don't see the benefit of this method (and `generate_texts` similarly). Why are we not leveraging the APIs that the WordPress AI Client SDK gives us?

I would suggest for the center piece of this class to be the `create_prompt` method that sets up a prompt builder with all of the defaults applied. And allow passing an optional array of config options to that method.

What the consuming code then does with that prompt builder is up to them; they can use all the underlying APIs as needed.

Something like:
```php
$prompt_builder = $ai_service->create_prompt( $prompt, $options );
$text = $prompt_builder->generate_text();
```

Or, make it a single statement:
```php
$text = $ai_service->create_prompt( $prompt, $options )->generate_text();
```

felixarntz (changes requested):
While overall I agree that a more opinionated AI service makes sense in the context of this plugin, I think it does too many things. Specifically, it obscures some of the flexibility that the underlying WordPress AI Client APIs provide, for no good reason as far as I can tell.

I think we should simplify the implementation to focus on creating a prompt builder with plugin specific defaults applied, and leave the rest to the regular prompt builder APIs.

iTsphillgood:
> While overall I agree that a more opinionated AI service makes sense in the context of this plugin, I think it does too many things. Specifically, it obscures some of the flexibility that the underlying WordPress AI Client APIs provide, for no good reason as far as I can tell.
> 
> I think we should simplify the implementation to focus on creating a prompt builder with plugin specific defaults applied, and leave the rest to the regular prompt builder APIs.

Have you tried Google Jules agent or Gemini 3 on cli?

caseymanos:
@felixarntz @jeffpaul  
Thanks for the feedback! I made the following improvements:
1. **Removed the interface** - Deleted `AI_Service_Interface.php`
  2. **Removed `DEFAULT_TEMPERATURE`** - No forced defaults; consumers specify explicitly or use provider defaults
  3. **Made `create_prompt()` the centerpiece** - Removed `generate_text()` and `generate_texts()` wrapper methods

  **Additional improvements:**
  - Simplified options handling using the SDK's `ModelConfig::fromArray()`
  - Options array supports all 11 SDK scalar options with snake_case keys 
  - Full SDK API remains available via method chaining after `create_prompt()`

  ### Updated API

  ```php
  // Simple usage - model preferences auto-applied
  $text = get_ai_service()->create_prompt('Summarize this...')->generate_text();

  // With options array
  $text = get_ai_service()->create_prompt('Translate to French', [
      'system_instruction' => 'You are a translator.',
      'temperature'        => 0.3,
  ])->generate_text();

  // Full SDK access via chaining
  $titles = get_ai_service()->create_prompt('Generate titles')
      ->using_candidate_count(5)
      ->generate_texts();

  Hooks Available

  | Hook                   | Type   | Description                              |
  |------------------------|--------|------------------------------------------|
  | ai_service_available   | Filter | Override provider availability detection |
  | ai_service_initialized | Action | Fires when AI service initializes        |
  | ai_preferred_models    | Filter | Customize default model preferences      |

  Note: Removed ai_service_prompt_builder filter as it's no longer needed - create_prompt() returns the builder directly so consumers can modify it with the SDK

includes/Services/AI_Service.php
```diff
@@ -0,0 +1,231 @@
...
+		 * @param bool|null $available Null to use default detection, or boolean to override.
+		 */
+		$available = apply_filters( 'ai_service_available', null );
+
+		if ( null !== $available ) {
+			return (bool) $available;
+		}
+
+		// Check if any provider credentials are configured.
+		$credentials = get_option( 'wp_ai_client_provider_credentials', array() );
```
felixarntz: Consider using `API_Credentials_Manager::OPTION_PROVIDER_CREDENTIALS` instead of hard-coding this string.

includes/Services/AI_Service.php
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	/**
+	 * Checks if an AI provider is available and configured.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return bool True if a provider is available, false otherwise.
+	 */
+	public function is_available(): bool {
```
felixarntz: Do we need this method? I understand the point is to return whether _any_ AI is generally available - which _maybe_ makes sense, but I am wary because it's not going to be truly accurate.

The WordPress AI Client already offers a much more precise way of determining that, via the `is_supported*` methods of `Prompt_Builder`.

The code example included with the `get_ai_service()` documentation confirms that including this method could be misleading: It's NOT how one should check whether AI is available.

felixarntz (commented):
@caseymanos Almost LGTM, just one remaining concern.

felixarntz:
Related: @dkotter @caseymanos I'm not sure about the `get_preferred_models()` function, only spotted it now. Didn't want to flag it during the review since it's already in the codebase. Can we rename this to something like `get_preferred_models_for_text_generation()`? Because that's closer to what that function does, based on the models it currently returns.

includes/Services/AI_Service.php
caseymanos: That's currently private but eliminating the encompassing `is_available()` method removes the need to make it public at this time

caseymanos:
@felixarntz 
Addressed both comments - removed is_available() entirely, which also gets rid of the hardcoded
  option string. Updated the docstring example to show using the SDK's
  is_supported_for_text_generation() instead.

  Also renamed get_preferred_models() to get_preferred_models_for_text_generation()  (filter renamed too).

felixarntz (dismissed):
@caseymanos Thanks, LGTM!

Note there are some merge conflicts though that need to be resolved.

includes/Services/AI_Service.php
```diff
@@ -0,0 +1,201 @@
...
+
+use function WordPress\AI\get_preferred_models_for_text_generation;
+
+/**
+ * AI Service class.
+ *
+ * Manages AI provider configuration and provides a consistent interface
+ * for experimental features to communicate with AI providers.
+ *
+ * @since 0.1.0
```
dkotter: Since 0.1.0 has now been released, these all need updated. Ideal is to use `x.x.x` for these and they will then be updated as part of our release process

includes/Services/AI_Service.php
```diff
@@ -0,0 +1,201 @@
...
+		$this->initialized = true;
+
+		/**
+		 * Fires when the AI service is initialized.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param \WordPress\AI\Services\AI_Service $service The AI service instance.
+		 */
+		do_action( 'ai_service_initialized', $this );
```
dkotter: We've landed on using `ai_experiments` as our prefix, so need to update that here

includes/Services/AI_Service.php
```diff
@@ -0,0 +1,201 @@
...
+	 * @return \WordPress\AI_Client\Builders\Prompt_Builder_With_WP_Error The prompt builder instance.
+	 */
+	public function create_prompt( ?string $prompt = null, array $options = array() ): Prompt_Builder_With_WP_Error {
+		$builder = AI_Client::prompt_with_wp_error( $prompt );
+
+		// Apply default model preferences.
+		$models = get_preferred_models_for_text_generation();
+		if ( ! empty( $models ) ) {
+			$builder = $builder->using_model_preference( ...$models );
+		}
```
dkotter: In the future, we will definitely have experiments that do more than just text generation. Since this method forces the preferred models to be text generation only, we won't be able to use this when we add things like image generation. Maybe that's fine but could see either renaming this method to be something like `create_text_prompt` or have an option you can pass in that marks things as text generation, audio generation or image generation

includes/Services/AI_Service.php
dkotter, line 1: I initially had something similar in place when building out the Title Generation Experiment before getting feedback that it made things over-abstracted (see https://github.com/WordPress/ai/pull/67#pullrequestreview-3461655392). So may be worth revisiting that discussion to ensure we're all on the same page cc/ @JasonTheAdams 

That said, overall I'm fine with this approach but seems we just add the `AI_Service` class in this PR but we're not actually using it anywhere. Do we want to update how the Title Generation Experiment is working to take advantage of this new approach?

dkotter (changes requested):

includes/Services/AI_Service.php
felixarntz: Great catch, +1. How about renaming this to `create_textgen_prompt`? I think that's less ambiguous, because `create_text_prompt` could also be interpreted as the _input_ being text. Alternatively `create_text_generation_prompt` which would be most clear, but maybe a bit long.

includes/Services/AI_Service.php
felixarntz: I was wary of overabstraction here too, which is why the PR has been notably reduced in scope from what it was originally.

I think the way it currently is is fine. Although it mostly feels like syntactic sugar and a slightly alternative way to use the fluent API from the WP AI Client prompt builder.

caseymanos:
@felixarntz @dkotter 
 - Renamed create_prompt() to create_textgen_prompt() to make it clear this is specifically for text
  generation, and gives space for future image/audio prompts 
  - Updated all @ since tags to x.x.x for the release process
  - Renamed hook from ai_service_initialized to ai_experiments_service_initialized to use the correct prefix

dkotter (dismissed):

dkotter:
@caseymanos Thanks for the work here. We've merged in a few other PRs that used the `get_preferred_models` function and since this PR changes the name of that, we now have a few linting errors. Do you have time to update those references?

includes/bootstrap.php
```diff
@@ -219,6 +220,13 @@ function initialize_experiments(): void {
 			$settings_page->init();
 		}
 
+		// Initialize the WP AI Client.
+		AI_Client::init();
```
dkotter: Looking at this again, we already run this exact same code a few lines above (currently line 206). Is there a reason to initialize it again here? Or can this be removed?

dkotter (changes requested):

includes/bootstrap.php
```diff
@@ -219,6 +220,13 @@ function initialize_experiments(): void {
...
+
+		// Initialize the AI Service layer.
+		$ai_service = AI_Service::get_instance();
+		$ai_service->init();
```
dkotter: I guess question on this as well looking at it again. It seems the `init` method we are calling here doesn't actually do anything (beyond just running a hook). Do we need this init method and the loading of this class here? Or can we remove this and anyone that wants to use this class can be done via the `get_ai_service` helper function?

includes/bootstrap.php
dkotter: Note I removed that here https://github.com/WordPress/ai/pull/101/changes/9a4845d797f935c0dfc5d78870fbd5aa3318b3ba but let me know if that is wrong

includes/bootstrap.php
dkotter: Note I removed these lines and the `init` method here: https://github.com/WordPress/ai/pull/101/changes/9a4845d797f935c0dfc5d78870fbd5aa3318b3ba. But let me know if that will cause problems

dkotter (approved):

PR #111, opened 2025-11-27
Fix plugin review items

### LINKED ISSUE #110: Resolve issues reported by WPORG plugin review
### Description

The below items have been flagged during the plugin review, let's look to get those resolved or identify a response to the WPORG Plugin Team on why we view them a non-issue.  I've amended some of the feedback to remove some specific, personal references but have left all necessary feedback items we need to work through.

Once we resolve all issues, we'll also want to:

- [ ] Test the updated plugin on a clean WordPress installation with [WP_DEBUG set to true](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/).
- [ ] Go to ["Add your plugin"](https://wordpress.org/plugins/developers/add/) and upload the updated version.
- [ ] @jeffpaul to reply to the WPORG Plugin Team email with concise clarifications or important context that the team needs to know.

## Review: Missing `permission_callback` in REST API Route

When using `register_rest_route()` or `wp_register_ability()` to define custom REST API endpoints, it is crucial to include a proper `permission_callback`.

🔒 This callback function ensures that only authorized users can access or modify data through your endpoint.

Code example, checking that the user can change options:
```
register_rest_route( 'ai/v1', '/my-endpoint', array(
    'methods' => 'GET',
    'callback' => 'ai_callback_function',
    'permission_callback' => function() {
        return current_user_can( 'manage_options' );
    }
) );
```

Please check the [register_rest_route() documentation](https://developer.wordpress.org/reference/functions/register_rest_route/) and the [current_user_can() documentation](https://developer.wordpress.org/reference/functions/current_user_can/).

✅ When a permission_callback is NOT Required:

There are valid use cases for public endpoints, such as publicly available data (e.g., posts, public metadata) or endpoints designed for unauthenticated access (e.g., fetching public stats or information).

In these cases, you should use `__return_true` as the `permission_callback` to indicate that the endpoint is intentionally public.

🔒 When a permission_callback IS Required:

For endpoints that involve sensitive data or actions (e.g., getting not public data, creating, updating, or deleting content).

In these cases, you should always implement proper permission checks.

Possible cases found on this plugin's code:
```
includes/Experiments/Title_Generation/Title_Generation.php:54 wp_register_ability('ai/' . $this->get_id(), array('label' => $this->get_label(), 'description' => $this->get_description(), 'ability_class' => Title_Generation_Ability::class));
```
Note: This check with `wp_register_ability()` was just added today, we are one of the first seeing it.
Its assumed that it should work as in `register_rest_route` but that should be tested to confirm before re-submitting.

## Generic function/class/define/namespace/option names

All plugins must have unique function names, namespaces, defines, class and option names. This prevents your plugin from conflicting with other plugins or themes. We need you to update your plugin to use more unique and distinct names.

A good way to do this is with a prefix. For example, if your plugin is called "AI Experiments" then you could use names like these:
function aiex_save_post(){ ... }
class AIEX_Admin { ... }
update_option( 'aiex_options', $options );
register_setting( 'aiex_settings', 'aiex_user_id', ... );
define( 'AIEX_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
global $aiex_options;
add_action('wp_ajax_aiex_save_data', ... );
namespace jeffpaul\ai;

Disclaimer: These are just examples that may have been self-generated from your plugin name, we trust you can find better options. If you have a good alternative, please use it instead, this is just an example.

The prefix should be at least four (4) characters long (don't try to use two- or three-letter prefixes anymore). We host almost 100,000 plugins on WordPress.org alone. There are tens of thousands more outside our servers. Believe us, you're likely to encounter conflicts.

You also need to avoid the use of __ (double underscores), wp_ , or _ (single underscore) as a prefix. Those are reserved for WordPress itself. You can use them inside your classes, but not as stand-alone function.

Please remember, if you're using _n() or __() for translation, that's fine. We're only talking about functions you've created for your plugin, not the core functions from WordPress. In fact, those core features are why you need to not use those prefixes in your own plugin! You don't want to break WordPress for your users.

Related to this, using `if (!function_exists('NAME')) {` around all your functions and classes sounds like a great idea until you realize the fatal flaw. If something else has a function with the same name and their code loads first, your plugin will break. Using if-exists should be reserved for shared libraries only.

Remember: Good prefix names are unique and distinct to your plugin. This will help you and the next person in debugging, as well as prevent conflicts.

Analysis result:
```
# This plugin is using the prefix "ai" for 18 element(s).
# This plugin is using the prefix "wordpress_ai" for 9 element(s).

# Using the common word "ai" as a prefix.
includes/bootstrap.php:26 define('AI_VERSION', '0.1.0');
includes/bootstrap.php:29 define('AI_PLUGIN_FILE', defined('WP_AI_DIR') ? WP_AI_DIR . 'ai.php' : '');
includes/bootstrap.php:32 define('AI_PLUGIN_DIR', defined('WP_AI_DIR') ? WP_AI_DIR : '');
includes/bootstrap.php:35 define('AI_PLUGIN_URL', plugin_dir_url(AI_PLUGIN_FILE));
includes/bootstrap.php:38 define('AI_MIN_PHP_VERSION', '7.4');
includes/bootstrap.php:41 define('AI_MIN_WP_VERSION', '6.8');
includes/bootstrap.php:44 define('AI_DEFAULT_ABILITY_CATEGORY', 'ai-experiments');
# Using the common word "wp" as a prefix.
ai.php:32 define('WP_AI_DIR', plugin_dir_path(__FILE__));
```
Regarding this issue I totally get that this plugin could be allowed with this prefix as it's a "special" one, but bear in mind that there might be other plugins trying to use that prefix when they shouldn't. Also this can set a example for other developers that is not among what's expected from them.

Therefore this is a case of "it's going to be their fault, but better safe than sorry" so it will be better if you can set a more complex prefix to avoid that kind of issues.

In any case the actions, filters and options may be able to stay with the ai_ prefix in the case that code here gets merged at some point to WordPress core.

## How can we check the functionality of this plugin?

During our reviews, we check the functionality of the plugin.

We also check that the plugin functionality is sufficiently explained in the readme file and/or through messages that help the user configure and use the plugin.

We encountered difficulties when trying to test the functionality of this plugin, either we could not find details on how to make it work, or it did not work.

Please add instructions on how to configure and use this plugin to the readme file if you have not already done so.

Please also make sure that the functionality of your plugin works properly and that anyone can configure and use it, even if it is installed from scratch.

After activating the plugin:

> Your installation of the AI plugin is incomplete. Please run `composer install`.

The `composer.json` is not there, anyway it's best if you include the final release with the production libraries included.

@jeffpaul note: I downloaded the ZIP from the 0.1.0 release on GitHub to submit, so perhaps we need to similarly resolve #109 to ensure the ZIP file Is one that can be used for WPORG submission as well as non-technical users coming to the repo and downloading a release ZIP.

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

dkotter:
## What?

Closes #110

A few minor things were flagged during the plugin review process and this PR fixes those.

## Why?

In order to get the plugin on WordPress.org, we need to ensure we pass the plugin review.

## How?

- Switches to a more specific namespace for all constants and hooks, `AI_EXPERIMENTS` and `ai_experiments` (these are pretty lengthy so open to suggestions if we want to shorten this)
- Adds slightly more specific testing instructions to the `readme.txt`

## Testing Instructions

Verify all changes look good here and all existing functionality still works

### DIFF
--- a/docs/DEVELOPER_GUIDE.md
+++ b/docs/DEVELOPER_GUIDE.md
@@ -264,10 +264,10 @@ The plugin provides a set of hooks and filters to allow third-party developers t
 
 ### Registering a Custom Experiment
 
-Developers can register their own experiments using the `ai_register_experiments` action. This is the primary way to add new functionality to the plugin.
+Developers can register their own experiments using the `ai_experiments_register_experiments` action. This is the primary way to add new functionality to the plugin.
 
 ```php
-add_action( 'ai_register_experiments', function( $registry ) {
+add_action( 'ai_experiments_register_experiments', function( $registry ) {
 	$registry->register_experiment( new My_Custom_Experiment() );
 } );
 ```
@@ -311,7 +311,7 @@ add_filter( 'ai_experiments_enabled', '__return_false' );
 
 The plugin also includes the following action hooks:
 
-- `ai_register_experiments`: Fires after default experiments are registered, receives `$registry` parameter
+- `ai_experiments_register_experiments`: Fires after default experiments are registered, receives `$registry` parameter
 - `ai_experiments_initialized`: Fires after all registered experiments have been initialized
 
 ### Asset Loading
--- a/docs/DEVELOPER_GUIDE.md
+++ b/docs/DEVELOPER_GUIDE.md
@@ -277,7 +277,7 @@ add_action( 'ai_register_experiments', function( $registry ) {
 Modify the list of default experiment classes before they are instantiated:
 
 ```php
-add_filter( 'ai_default_experiment_classes', function( $experiment_classes ) {
+add_filter( 'ai_experiments_default_experiment_classes', function( $experiment_classes ) {
 	// Add a custom experiment
 	$experiment_classes[] = 'My_Namespace\My_Custom_Experiment';
 
@@ -296,7 +296,7 @@ add_filter( 'ai_default_experiment_classes', function( $experiment_classes ) {
 Experiments can be disabled using the `ai_experiment_{$experiment_id}_enabled` filter:
 
 ```php
-add_filter( 'ai_experiment_example-experiment_enabled', '__return_false' );
+add_filter( 'ai_experiments_experiment_example-experiment_enabled', '__return_false' );
 ```
 
 ### Disabling All Experiments
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,7 +145,7 @@ This issue is for tracking changes for the X.Y.Z release.  Target release date: 
 ## [Release steps](https://github.com/wordpress/ai/blob/develop/CONTRIBUTING.md#release-instructions)
 
 - [ ] Branch: Starting from `develop`, cut a release branch named `release/X.Y.Z` for your changes.
-- [ ] Version bump: Bump the version number in `ai.php`, `package-lock.json` (twice), `package.json`, and `readme.txt` if it does not already reflect the version being released.  In `includes/bootstrap.php`, ensure you're updating the `AI_VERSION` version constant.
+- [ ] Version bump: Bump the version number in `ai.php`, `package-lock.json` (twice), `package.json`, and `readme.txt` if it does not already reflect the version being released.  In `includes/bootstrap.php`, ensure you're updating the `AI_EXPERIMENTS_VERSION` version constant.
 - [ ] Update `@since`: Find all new `@since x.x.x` lines and update those with the new version number in place of `x.x.x`.
 - [ ] Changelog: Add/update the changelog in `CHANGELOG.md` and in `readme.txt`.
 - [ ] Props: update `CREDITS.md` file with any new contributors, confirm maintainers are accurate.
--- a/ai.php
+++ b/ai.php
@@ -29,6 +29,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Shortcut constant to the path of this file.
  */
-define( 'WP_AI_DIR', plugin_dir_path( __FILE__ ) );
+define( 'AI_EXPERIMENTS_DIR', plugin_dir_path( __FILE__ ) );
 
-require_once WP_AI_DIR . 'includes/bootstrap.php';
+require_once AI_EXPERIMENTS_DIR . 'includes/bootstrap.php';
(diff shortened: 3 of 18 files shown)


docs/DEVELOPER_GUIDE.md
```diff
@@ -296,7 +296,7 @@ add_filter( 'ai_default_experiment_classes', function( $experiment_classes ) {
 Experiments can be disabled using the `ai_experiment_{$experiment_id}_enabled` filter:
 
 ```php
-add_filter( 'ai_experiment_example-experiment_enabled', '__return_false' );
+add_filter( 'ai_experiments_experiment_example-experiment_enabled', '__return_false' );
```
jeffpaul, line 299: :alot: of `experiment`

jeffpaul:
@dkotter I'm good with these changes, so if you want to share a ZIP from this branch then I can get that over to WPORG while we allow others here to formalize a review (and perhaps iterate on the namespace if desired).

dkotter:
Zip attached here
[ai-0.1.0.zip](https://github.com/user-attachments/files/23784390/ai-0.1.0.zip)

swissspidy (approved):

justlevine (approved):
Only reviewed the diff from my phone and LGTM, approving since I agree we shouldn't hold things up over naming things.

And tbh it's not even that long (iirc perf is `performance-lab`)

---

Above said, 

I saw they said no to `ai` and the guidelines prevent `wp`, but did they say no to `wpai` together (no hyphen)? It's short, unambiguous, and follows the plugin naming guidelines.

jeffpaul:
I'll ask about `wpai` when I send the email response back but for now will keep the longer option in the code.

jeffpaul:
> I'll ask about `wpai` when I send the email response back but for now will keep the longer option in the code.

TLDR; best to stick with the longer `ai_experiments` 

> That naming is discouraged as in this context 4 characters slugs that begin with wp- are in practice 2 characters slugs (wp + 2 characters).
> A lot of people like to begin their stuff with "wp", that was actually one of the reasons we are now asking for 4 character prefixes instead of 3, that's because we detected a bunch of plugins using combinations like: wpa_, wpb_, wpc_, etc.

So I'll leave things as-is with the `ai_experiments` namespace.

PR #113, opened 2025-11-30
Fix asset loader bugs causing missing dependencies, lack of RTL support, lack of inlining for performance

felixarntz:
## What?

The `Asset_Loader` implementation has several problems:

* The PHP build asset check is wrong, it's currently always failing.
* RTL language support is missing.
* The stylesheet file path should be provided to benefit from conditional inlining CSS for performance in WP Core.
* It's not possible to pass asset data manually, which can lead to unperformant and sometimes inaccurate `filemtime` calls.

## How?

This PR addresses the above issues.

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=113%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%20113%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).

### DIFF
--- a/includes/Asset_Loader.php
+++ b/includes/Asset_Loader.php
@@ -109,9 +109,10 @@ class Asset_Loader {
 			return;
 		}
 		wp_style_add_data( 'ai_' . $handle, 'rtl', 'replace' );
-		if ( is_rtl() ) {
-			wp_style_add_data( 'ai_' . $handle, 'path', $rtl_style_path );
+		if ( ! is_rtl() ) {
+			return;
 		}
+		wp_style_add_data( 'ai_' . $handle, 'path', $rtl_style_path );
 	}
 
 	/**
--- a/includes/Asset_Loader.php
+++ b/includes/Asset_Loader.php
@@ -109,7 +109,9 @@ class Asset_Loader {
 			return;
 		}
 		wp_style_add_data( 'ai_' . $handle, 'rtl', 'replace' );
-		wp_style_add_data( 'ai_' . $handle, 'path', $rtl_style_path );
+		if ( is_rtl() ) {
+			wp_style_add_data( 'ai_' . $handle, 'path', $rtl_style_path );
+		}
 	}
 
 	/**
--- a/includes/Asset_Loader.php
+++ b/includes/Asset_Loader.php
@@ -105,10 +105,11 @@ class Asset_Loader {
 		wp_style_add_data( 'ai_' . $handle, 'path', $style_path );
 
 		$rtl_style_path = str_replace( '.css', '-rtl.css', $style_path );
-		if ( file_exists( $rtl_style_path ) ) {
-			wp_style_add_data( 'ai_' . $handle, 'rtl', 'replace' );
-			wp_style_add_data( 'ai_' . $handle, 'path', $rtl_style_path );
+		if ( ! file_exists( $rtl_style_path ) ) {
+			return;
 		}
+		wp_style_add_data( 'ai_' . $handle, 'rtl', 'replace' );
+		wp_style_add_data( 'ai_' . $handle, 'path', $rtl_style_path );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/Asset_Loader.php
```diff
@@ -32,15 +32,23 @@ class Asset_Loader {
...
+	 * @param string                    $handle     The handle for the script.
+	 * @param string                    $file_name  The script file name.
+	 * @param array<string, mixed>|null $asset_data Optional asset metadata (dependencies and version).
 	 */
-	public static function enqueue_script( string $handle, string $file_name ): void {
+	public static function enqueue_script( string $handle, string $file_name, ?array $asset_data = null ): void {
 		$script_path       = AI_EXPERIMENTS_DIR . 'build/' . $file_name . '.js';
 		$script_url        = AI_EXPERIMENTS_PLUGIN_URL . 'build/' . $file_name . '.js';
-		$script_asset_path = $script_path . '.asset.php';
+		$script_asset_path = substr( $script_path, 0, -3 ) . '.asset.php';
```
dkotter, line 42: Ah, this was my mistake. I'd suggest instead of this `substr` call, we just remove the appending of `.js` to `$script_path`, as that's not needed (and was my mistake)
```suggestion
		$script_path       = AI_EXPERIMENTS_DIR . 'build/' . $file_name;
		$script_url        = AI_EXPERIMENTS_PLUGIN_URL . 'build/' . $file_name . '.js';
		$script_asset_path = $script_path . '.asset.php';
```

includes/Asset_Loader.php
```diff
@@ -32,15 +32,23 @@ class Asset_Loader {
...
 
-		if ( file_exists( $script_asset_path ) ) {
+		if ( is_array( $asset_data ) ) {
```
dkotter, line 44: I guess I'm trying to understand the need for this? Is this just for future use or should we update existing calls to `enqueue_script` and `enqueue_style` to take advantage of this?

includes/Asset_Loader.php
felixarntz: How that? If we remove appending `.js`, that path won't be correct, as it'll be missing the extension.

includes/Asset_Loader.php
felixarntz: It's for future use at least, but maybe even present use. A few reasons for this:

* There may be assets that get built differently and therefore don't have an asset metadata file. It still needs to be possible to provide the information for such cases.
* The `dependencies` value from the asset metadata file is always purely based on the JavaScript code. Using it for stylesheets will lead to problems in 99% of cases, since not every script handle has a matching stylesheet handle. So depending on the script dependencies for stylesheets will often lead to the stylesheet not loading because a dependency handle doesn't exist.
* Using `filemtime` should always be a last resort, it can be slow or unreliable depending on the host configuration. So that's another reason to allow passing the data manually, to not have to rely on `filemtime`.

includes/Asset_Loader.php
dkotter: I was thinking we don't ever actually use that path other than to get the `asset.php` file but I guess we do use it as a fallback in the `filemtime` call which would fail in that case

dkotter (approved):

PR #121, opened 2025-12-01
Remove Example Experiment registration

### LINKED ISSUE #119: Consider removing the Example Experiment code
Coming out of our initial scaffolding, code was added for an Example Experiment, giving an easy reference to follow when adding new Experiments.

Now that we have a complete Experiment (Title Generation) and now that we have an initial release, seems like we could remove the example code so we're not confusing people that use the plugin (as the Example Experiment shows on our settings screen).

If we want to keep the example code in place for reference, I'd suggest we at least remove the registration of the Example Experiment so it doesn't end up in the settings page.

dkotter:
## What?

Closes #119 

Remove the registration of our Example Experiment.

## Why?

We have an Example Experiment in place which is useful for those looking to contribute or get some ideas on how things can be set up. But this Experiment doesn't actually add any useful functionality and it can be confusing to have this shown in our settings screen. For now we're leaving the code in place but we don't want this to be an Experiment someone can turn on.

## How?

- Removes the Example Experiment from our list of default Experiments we register
- Updates our unit tests to account for this

## Testing Instructions

Go to the AI Experiments settings page and ensure you no longer see a settings section for the Example Experiment

### DIFF
--- a/includes/Experiment_Loader.php
+++ b/includes/Experiment_Loader.php
@@ -104,7 +104,6 @@ final class Experiment_Loader {
 	 */
 	private function get_default_experiments(): array {
 		$experiment_classes = array(
-			\WordPress\AI\Experiments\Example_Experiment\Example_Experiment::class,
 			\WordPress\AI\Experiments\Title_Generation\Title_Generation::class,
 		);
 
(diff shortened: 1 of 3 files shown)


JasonTheAdams (approved):
Makes sense! 👍

PR #122, opened 2025-12-01
Ensure the right version of the PHP AI Client is installed

dkotter:
## What?

Ensure the right version of the PHP AI Client is pulled in from the WP AI Client update

## Why?

In #118, I updated our composer config to pull in the latest release of the WP AI Client package (0.1.1) instead of pulling in `trunk`. This was supposed to pull in the latest release of the PHP AI Client (0.2.1) but it seems that just running the `composer require wordpress/wp-ai-client:0.1.1` command didn't pull in the downstream dependencies so it was still using the 0.2.0 release of that package.
  
## How?

Initially tried running `composer update` to bring in everything but there appears to be some conflicts with our PHPCS tools coming out of [recent changes](https://github.com/PHPCompatibility/PHPCompatibility/commit/93ef2b82ff05534a976259c45783202887cfc8bf) made to the `develop` branch of the `phpcompatibility/php-compatibility` package. So instead of trying to fix all of those, I ran `composer require wordpress/wp-ai-client:0.1.1 -W` to bring in the updated version of the PHP AI Client package (the `-W` flag was the trick here)

## Testing Instructions

n/a

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -702,16 +702,16 @@
         },
         {
             "name": "wordpress/php-ai-client",
-            "version": "0.2.0",
+            "version": "0.2.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/WordPress/php-ai-client.git",
-                "reference": "81a104a9bc5f887e3fbecea6e0d9cd8eab3be0b2"
+                "reference": "61ecd7c86329d0cc3d17567891f363d8f3fc3be6"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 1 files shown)


dkotter:
Worth noting that in testing with Google Gemini, I'm now getting a timeout error instead of the original 404 error. This will be fixed once a new release of the WP AI Client goes out as the default timeout value was increased in https://github.com/WordPress/wp-ai-client/pull/29.

So we could hold on this PR until the 0.2.0 release of the WP AI Client package is ready and then update to that here

jeffpaul (approved):
Let's at least fix this version bump for this and can come back and do another, proper bump when the WP client hits 0.2.0?

PR #134, opened 2025-12-02
Add the Image Generation Experiment base

### LINKED ISSUE #13: Image Generation
### Summary:

This feature allows users to generate images directly from text prompts within the WordPress editor. Users can either generate a featured image for the post or insert an image directly into the post body using the image block. The goal is to simplify the process of sourcing relevant visuals and to help users enhance posts or pages without leaving the editor.

### Why it matters:

Sourcing or designing relevant images slows down the publishing process, especially for non-designers. This feature reduces that friction by offering a quick, integrated way to create visuals from simple prompts—saving time and encouraging stronger visual engagement. It supports both functional needs (e.g., adding a featured image) and creative expression.

### Use Cases:

- A user wants to generate a featured image based on the post’s topic
- An author wants a visual to break up text or complement a paragraph
- A marketer needs a simple graphic to match a post without switching tools

### Assumptions:

- The Excerpt Generation feature is already in place and available
- The AI Services plugin handles model authentication and API requests

### Entry Point:

- In the Featured Image panel, a "Generate with AI" button appears when no image is set.

### Prompt Handling:

- When clicked, the system automatically uses the post excerpt (generated by the Excerpt Generation feature) as the prompt.
- No manual prompt entry is required in this flow.

### Context Strategy:

- The excerpt is passed as background context to the image model.
- If for some reason the excerpt is missing, a truncated section of post content (~150–200 words) may be used instead.
- Alternatively, instead of using a truncated version instead we could send the entire content. We could daisy chain this, where we first send the content, have AI summarize it and then use that summary to generate the image. It may be worth some testing once we get to this point to see what gives us the best result, whether the full content works better or if the summary works just as well.

### Generation & Output:

- A loading indicator shows generation is in progress.
- Once complete, the user sees a preview with options to:
  - Set the image as the featured image
  - Regenerate
  - Cancel

### Media Handling:

- The image is saved automatically to the Media Library.
- A filename is generated using the post slug or prompt.  Prompts can get long and cause issues when used as file names, so we'll want to ensure we truncate this to a certain length.
- Alternatively to using the prompt in the filename, the full prompt could be added into the alt text field and perhaps a summary of the prompt is used for the image title.
- (Optional) An AI-generated tag may be added for transparency.

### Editorial Considerations:

- Intent: Provide a general-purpose visual that summarizes the tone or subject of the full post.
- Control: No prompt writing or fine-tuning; users can only regenerate or accept.
- Risk: Low, since the input is excerpt-based and system-defined.
- Tone: Neutral, context-aware, and broadly illustrative.

### Entry Point:

- When adding an image block, users see a "Generate with AI" option alongside the existing media options.

### Prompt Handling:

- Users enter a freeform description of the image they want.
- (Optional future enhancement) The system may suggest a starter prompt based on nearby paragraph content.

### Context Strategy:

- The user’s prompt is primary.
- We may include block-level context in the background in future versions to improve image relevance.

### Generation & Output:

- A loading state appears while the image is being created.
- The user sees a preview with options to:
  - Insert into the post
  - Regenerate
  - Cancel

### Media Handling:

- The image is added to the Media Library automatically.
- The filename is based on the user’s prompt or a unique slug.

### Model Integration:

- Connects to the image generation provider via the AI Services plugin (OpenAI, Stability, or self-hosted models)
- Prompts and responses handled through standardized API call structure

## Input Handling:

- Featured Image flow uses the post excerpt (or fallback text snippet) as prompt
- Inline flow accepts freeform user input
- Optional: allow prompt templating or dynamic parameter injection in future

### Output Handling:

- Returned image is previewed in the UI before final placement
- User must confirm insertion or featured image application
- Image is saved to the Media Library with a generated filename and metadata

### Storage & Metadata:

- Images saved to the site’s uploads directory and indexed in WordPress Media Library
- Include metadata field (e.g., `_ai_generated = true`) to optionally identify AI-generated assets
- Include a `caption` that automatically attributes the source of the generated image to the AI service provider utilized. OpenAI already provides that (and perhaps wants AI clients to do that as well) though not sure if that should be a requirement for this plugin or not.

### Error & Timeout Handling:

- Gracefully handle timeouts, API failures, or unsupported prompt content
- UI displays fallback message with retry option

### Security & Privacy:

- No excerpt or prompt data is stored externally
- API keys are stored and managed via AI Services plugin
- Generated images do not expose post content unless explicitly included in the prompt

### Extensibility Notes:

- System should allow hooks for other plugins to filter prompts, override image size, or customize file naming
- Allow for future UI expansion (e.g., style selector, aspect ratio presets)

### Version 1 - Featured Image Generation

Image

### Version 1 - Post Image Generation

Image

### Version 2 - Featured and Post Image Generation

Image

_**Props to [Lina Wiezkowiak](https://profiles.wordpress.org/linawiezkowiak/) for the product definition, [Rachael Cortellessa](https://profiles.wordpress.org/rachaelcortellessa/) for the designs, and [Darin Kotter](https://github.com/dkotter) for technical review of both.**_

dkotter:
## What?

Partially closes #13

Adds the base Experiment and Ability code for Image Generation and Import. There will be a followup PR that adds in the UI to trigger image generation and import.

## Why?

We want an Experiment that allows you to generate images, whether that is in-content images, featured images or an image used elsewhere and then import those images into your media library. To keep PRs as small as possible, this introduces the base framework for image generation and import, both the Image Generation Experiment and the Image Generation and Image Import Abilities.

## How?

- Adds a new Image Generation Experiment class and registers this as one of our default Experiments
- Adds a new Image Generation Ability that allows you to pass in a prompt and it will then generate a base64 encoded image using the WP AI Client
- Adds a new Image Import Ability that allows you to pass in a base64 encoded image and it then imports that into your media library, along with any optional information like title, filename and description
   - Note: could extend this to support image URLs as well but mostly built this to handle importing images we generate, which will be base64 only
- Adds in new tests to cover these changes

## Testing Instructions

Since there isn't a UI for this yet, will need to make direct API requests to see this working.

Make an authenticated `GET` request to the main abilities endpoint and ensure the Image Generation and Image Import abilities shows there
```bash
```

Make an authenticated `GET` request to the Image Generation endpoint and ensure information about the ability shows
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/image-generation' \
```

Make an authenticated `GET` request to the Image Import endpoint and ensure information about the ability shows
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/image-import' \
```

Make an authenticated `POST` request to the Image Generation run endpoint without passing a prompt and ensure an error shows
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/image-generation/run' \
--data '{
    "input": {
    }
}'
```

Make an authenticated `POST` request to the Image Generation run endpoint and pass in a prompt and ensure base64 encoded data is returned
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/image-generation/run' \
--data '{
    "input": {
        "prompt": "Sunrise over the mountains reflecting in a mountain lake"
    }
}'
```

Make an authenticated `POST` request to the Image Import run endpoint without passing data and ensure an error shows
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/image-import/run' \
--data '{
    "input": {
    }
}'
```

Make an authenticated `POST` request to the Image Import run endpoint and pass in valid base64 data (and any optional data) and ensure the image data is returned and the image is in your media library
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/image-import/run' \
--data '{
    "input": {
        "data": "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=="
    }
}'
```

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=134%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%20134%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).


includes/Abilities/Image/Import.php
```diff
@@ -0,0 +1,273 @@
...
+use WP_Error;
+use WordPress\AI\Abstracts\Abstract_Ability;
+use WordPress\AiClient\Files\DTO\File;
+
+/**
+ * Image import WordPress Ability.
+ *
+ * @since 0.1.0
+ */
+class Import extends Abstract_Ability {
```
JasonTheAdams: Let's call this `Import_Base64_Image` for clarity.

includes/Abilities/Image/Import_Base64_Image.php
```diff
@@ -0,0 +1,273 @@
...
+				'description' => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'The description of the image.', 'ai' ),
+				),
+				'mime_type'   => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'The MIME type of the image.', 'ai' ),
+				),
```
JasonTheAdams, line 62: Let's make this optional. If the `data` parameter is a Data URL, it will be this structure and contain the mime type:
```
data:[mime-type];base64,[base64-encoded-bytes]
```

If the mime type is not explicitly set and it's not a Data URL, then we can throw an error.

JasonTheAdams: As a note, I notice you're using `File` below, which actually automatically handles the detection of this for you and accepts the mime type as an optional second parameter.

I think if you just remove the `image/png` default below it should all just work.

includes/Abilities/Image/Import_Base64_Image.php
```diff
@@ -0,0 +1,273 @@
...
+		// Verify the data is a base64 encoded string.
+		try {
+			$file = new File( $input['data'], $args['mime_type'] );
+		} catch ( Throwable $t ) {
+			return new WP_Error(
+				'invalid_data',
+				esc_html__( 'The data is not a valid base64 encoded string.', 'ai' )
+			);
+		}
+
```
JasonTheAdams, line 138: Let's add a `$file->isImage()` check here to make sure we're importing an image.

JasonTheAdams (changes requested):
Good work, @dkotter! Couple small suggestions. 😄

includes/Abilities/Image/Import.php
dkotter: Renamed: https://github.com/WordPress/ai/pull/134/changes/653624840376bdd1bd3d2367fbde7a729bb86081

includes/Abilities/Image/Import_Base64_Image.php
dkotter: Added: https://github.com/WordPress/ai/pull/134/changes/e9a5e676c22387f61fd3417f72d35c31e1654a4d

includes/Abilities/Image/Import_Base64_Image.php
dkotter: I've set the default of that to `null` now so you can optionally pass this if you want: https://github.com/WordPress/ai/pull/134/changes/8b93cea71dea126e47d523e70172e797f1bbb7d9

JasonTheAdams (dismissed):
Good work, @dkotter!

includes/Abilities/Image/Generate.php
```diff
@@ -0,0 +1,143 @@
...
+use WordPress\AiClient\Files\Enums\FileTypeEnum;
+
+use function WordPress\AI\get_preferred_image_models;
+
+/**
+ * Image generation WordPress Ability.
+ *
+ * @since 0.1.0
+ */
+class Generate extends Abstract_Ability {
```
felixarntz: Nit, but I think more self-explanatory class names are preferable in terms of encouraging more readable code. I realize this is in an `Image` namespace, but when you use the class name on its own somewhere, that context is lost when looking at the concrete line of code.

So I'd suggest we use `Generate_Image`, despite this already being in an `Image` namespace (which makes sense to keep).

includes/Abilities/Image/Import_Base64_Image.php
```diff
@@ -0,0 +1,281 @@
...
+				'title'       => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'The title of the image.', 'ai' ),
+				),
+				'description' => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'The description of the image.', 'ai' ),
+				),
```
felixarntz, line 52: In addition to these, we should also encourage (at least support) passing an alt text, crucial for accessibility. It can be set via `_wp_attachment_image_alt` below.

includes/Abilities/Image/Import_Base64_Image.php
```diff
@@ -0,0 +1,281 @@
...
+		// Handle the sideload.
+		$attachment_id = media_handle_sideload(
+			$file_array,
+			0,
+			$args['description'],
+			array(
+				'post_title'     => $args['title'],
+				'post_content'   => $args['description'],
+				'post_mime_type' => $args['mime_type'],
+			)
```
felixarntz, line 274: See above, by including a `_wp_attachment_image_alt` key here, we can set an image alt text.

includes/Abilities/Image/Import_Base64_Image.php
```diff
@@ -0,0 +1,281 @@
...
+							'description' => esc_html__( 'Attachment ID.', 'ai' ),
+						),
+						'url'   => array(
+							'type'        => 'string',
+							'description' => esc_html__( 'Attachment URL.', 'ai' ),
+						),
+						'title' => array(
+							'type'        => 'string',
+							'description' => esc_html__( 'Attachment title.', 'ai' ),
+						),
```
felixarntz: Not a must have, but why only return the attachment title, but not the other text bits (description, alt text)? That feels a bit arbitrary. FWIW, it's technically more important for an image to have an alt text than a title.

includes/helpers.php
```diff
@@ -156,6 +156,44 @@ function get_preferred_models(): array {
...
+		),
+		array(
+			'openai',
+			'gpt-image-1',
+		),
+		array(
+			'openai',
+			'dall-e-3',
+		),
+	);
```
felixarntz, line 188: Would be great to include the latest Imagen model here: It's distinctly different from Google's Gemini image models because Imagen is a more classic diffusion model, and therefore closer to `gpt-image-1` and `dall-e-3`. This can also have practical differences: The Gemini models are particularly good at following instructions, generating text, and "editing" existing images because they can fully maintain existing content. Most other models, including Imagen are not as good at that, but the strongest diffusion models are usually better at generating entirely new original images that look super realistic.

In terms of quality, I'm not going to make a call here who's better, but it's definitely in the same ballpark as `gpt-image-1`, certainly stronger than the now somewhat ancient `dall-e-3`.
```suggestion
	$preferred_models = array(
		array(
			'google',
			'gemini-3-pro-image-preview',
		),
		array(
			'google',
			'gemini-2.5-flash-image',
		),
		array(
			'google',
			'imagen-4.0-generate-001',
		),
		array(
			'openai',
			'gpt-image-1',
		),
		array(
			'openai',
			'dall-e-3',
		),
	);
```

felixarntz (commented):
@dkotter This looks excellent! A few small notes, but after that it'll be good to merge.

includes/Abilities/Image/Generate.php
dkotter: Makes sense, updated in https://github.com/WordPress/ai/pull/134/changes/92308500bb7ba1e4620a8e478cef794ee8f60294

includes/Abilities/Image/Import_Base64_Image.php
dkotter: I agree, added in https://github.com/WordPress/ai/pull/134/changes/5ede28bdeb81c9889c75e0c2822f6362fc64dc22

includes/Abilities/Image/Import_Base64_Image.php
dkotter: Good call. I think the image URL and ID are the most important pieces but might as well return the other data as well (especially if someone is using this data to manually build image markup): https://github.com/WordPress/ai/pull/134/changes/d3692f3e98793007b85086d3c39ba8f955fba277

includes/helpers.php
dkotter: Added now: https://github.com/WordPress/ai/pull/134/changes/bc693e35e83a788eaec7d69d6321a08b8f4bb9ef

I originally left this off as my account doesn't have access to use that model and so was returning an error but I agree, seems good to have that here

includes/Abilities/Image/Generate_Image.php
```diff
@@ -0,0 +1,143 @@
...
+use WordPress\AI\Abstracts\Abstract_Ability;
+use WordPress\AI_Client\AI_Client;
+use WordPress\AiClient\Files\Enums\FileTypeEnum;
+
+use function WordPress\AI\get_preferred_image_models;
+
+/**
+ * Image generation WordPress Ability.
+ *
+ * @since x.x.x
```
felixarntz, line 22: Just a note: In `WordPress/performance`, `WordPress/php-ai-client`, and a few other projects we established the convention of `@since n.e.x.t`. Obviously it's an arbitrary decision, but I wonder whether we should use that here too.

Most importantly, we should probably choose a convention and document it, so that it's easy for a pre-release step to replace all those with the actual version. Maybe we can even build linting for it at some point.

This is mostly a thing we'll want to figure out soon, feel free to change it here or leave as is if you prefer.

felixarntz (approved):
@dkotter Great! Just one somewhat unrelated note, but this is good to merge 👍

includes/Abilities/Image/Generate_Image.php
dkotter: I'm fine if we update to using this, no strong opinion on my end but noting this is [currently documented](https://github.com/WordPress/ai/blob/develop/CONTRIBUTING.md#general-rules) as the approach (which is why I used that here) and there is a step in [our release process](https://github.com/WordPress/ai/blob/develop/CONTRIBUTING.md#release-instructions) to update these

PR #136, opened 2025-12-04
Add the Content Summarization Experiment base

### LINKED ISSUE #12: Content Summarization
### Summary:

This feature generates a concise summary of a post or page’s full content, stored as structured data and optionally inserted into the post via a Gutenberg block. It is intended to support both inline readability (e.g., top-of-post overview) and structured metadata use cases (e.g., internal search, preview cards, assistive technologies). The summary is generated automatically when sufficient content exists, but it is not inserted into the post unless the author chooses to do so. Authors can review and edit the summary, and optionally insert it into the post using a Gutenberg block.

### Why it matters:

Long-form content can be difficult for users to scan or assess quickly, especially on mobile. Summaries help readers orient themselves and decide whether to engage, improving UX, accessibility, and engagement. At the same time, structured summaries are valuable for site features like internal search previews, on-page navigation, or SEO overlays. By allowing authors to generate a smart, editable summary once and reuse it in multiple ways, this feature supports both content quality and performance.

### Use Cases:

- A long-form blog post includes a short summary at the top to help readers understand the topic at a glance.
- A documentation page uses a generated summary to support internal site search results.
- A publisher inserts summaries into featured cards, mobile overlays, or accessibility toggles.

### Gutenberg-first but data-agnostic:

- The summary is stored in post meta, not block content.
- It can be inserted into the content via a Gutenberg block or accessed programmatically (e.g., `get_post_meta()` or a helper function).
- Non-Gutenberg implementations can surface it via classic templates or custom filters.

### Overview:

The summary is generated automatically when the post or page content reaches a certain length threshold, but is not inserted into the post by default. Authors can view and edit the summary from the editor interface and choose whether or not to place it in the visible post body using a Gutenberg block. This provides a clean default while preserving full editorial control.

### Behavior Details

1. Automatic Generation

- Once the post/page content exceeds a threshold (e.g., ~200 words), the system automatically generates a summary using AI.
- Alternatively to having this be seamless for the user, we might also consider only generating when the content itself is more or less finished as it makes it difficult to determine when to generate an excerpt based on content length. This could either be something that is manually triggered or is triggered when the content is saved.
- Also worth considering the PrePublishPanel here if we decide to not auto-generate the summary, prompting a user prior to publishing to generate a summary if none exists.
- The summary is saved as post meta and is not added to the content unless the user takes action.

2. Summary Preview and Edit Access

- In the sidebar (or inspector panel), a “Content Summary” section shows the generated summary.
- Users can edit the summary directly in this panel.
- Optionally includes character or word count guidance.

3. Optional Insertion via Gutenberg Block

- Users can insert a “Summary” block into their post, which pulls the current summary value. Alternatively, this might not need to be a custom block, it could be inserted as a core block instead (i.e. paragraph blocks).
- The block can be placed anywhere (typically top-of-post) and updated if the summary is edited.

4. Regeneration and Deletion

- A small “Regenerate Summary” button allows users to overwrite the current AI output if desired.
- If the summary is deleted, no further automatic attempts are made unless manually triggered.

5. Error Handling

- If the AI generation fails, the field remains empty.
- Optional fallback message: “Couldn’t generate summary. You can write your own or try again later.”

### Tone and Voice:

- Summary should be factual, neutral, and informative.
- Avoid persuasive tone, humor, or exaggerated language.
- Write in complete sentences; avoid fragments or marketing-style phrasing.

### Length and Format:

- Target length is 2–3 sentences or ~50–100 words.
- Summary should provide a high-level view of the core message and structure of the content.
- Output must be plain text with no HTML formatting, emojis, or special characters.

### Editability:

- Authors can edit the summary in the sidebar panel.
- Any edits are persisted and reflected in the inserted block (if used).
- Edits do not retrigger generation unless reset or cleared by the user.

### Placement Guidance (for authors):

- Recommend placing the summary block near the top of the post, below the title or introductory paragraph.
- Use cases may vary by content type; themes can provide guidance via templates or inline helpers.

### AI Input Requirements:

- Post title
- Full post or page content
- Optional truncation if content exceeds token limits (e.g. max 1500 words)
- No user input required to initiate

### AI Output Requirements:

- One plain-text summary string
- Target 2–3 sentences, max ~100 words
- Must be clean, UTF-8 compliant, and suitable for direct display or programmatic use

### Trigger Conditions:

- Auto-generates after post/page content exceeds 200-word threshold
- Triggered only if no existing summary exists
- Regeneration must be explicitly requested by the user

### Storage & Access:

- Saved in post meta (`_content_summary` or similar key)
- Accessible via template tags, REST API, and block editor
- Does not affect `post_content` unless inserted manually

- Generation failure results in empty summary state with non-blocking fallback message
- No automatic retries unless initiated by the user

### Model Agnosticism:

- Compatible with any model supported by the AI Services plugin (e.g., OpenAI, Claude, self-hosted LLMs)
- Prompt logic should be modular and override-friendly

### Privacy & Config:

- API keys and endpoint logic handled by AI Services plugin
- No summary content is stored externally; generation happens per request

### Extensibility Notes:

- Future versions may support:
  - Summary style presets (e.g., concise, bulleted, SEO-optimized)
  - Multiple summary types (reader preview vs. search snippet)
  - Automatic placement recommendations based on theme or content type

### Version 1

Image

### Version 2

Image

dkotter:
## What?

Partially closes #12

Adds the base Experiment and Ability code for Content Summarization. There will be a followup PR that adds in the UI to trigger summarization at the post level.

## Why?

We want an Experiment that allows you to summarize content and then store that summary as post meta and optionally insert it as a block (both of those pending in the followup PR). To keep PRs as small as possible, this introduces the base framework for content summarization, both the Experiment and the Ability.

## How?

- Adds a new Summarization Experiment class and registers this as one of our default Experiments
- Adds a new Summarization Ability that allows you to pass in either a post ID or stand-alone content and it will then generate a summary using the WP AI Client
- Makes a minor adjustment to our `get_post_context` helper function to remove empty values
- Adds in new tests to cover these changes

## Testing Instructions

Since there isn't a UI for this yet, will need to make direct API requests to see this working.

Make an authenticated `GET` request to the main abilities endpoint and ensure the Summarization ability shows there
```bash
```

Make an authenticated `GET` request to the Summarization endpoint and ensure information about the ability shows
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/summarization' \
```

Make an authenticated `POST` request to the Summarization run endpoint without passing valid data and ensure an error shows
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/summarization/run' \
--data '{
    "input": {
    }
}'
```

Make an authenticated `POST` request to the Summarization run endpoint and pass in a post ID and ensure a summary is returned
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/summarization/run' \
--data '{
    "input": {
        "post_id": 1
    }
}'
```

Make an authenticated `POST` request to the Summarization run endpoint and pass in a non valid post ID and ensure an error is returned
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/summarization/run' \
--data '{
    "input": {
        "post_id": 999
    }
}'
```

Make an authenticated `POST` request to the Summarization run endpoint and pass in content and ensure a summary is returned
```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/summarization/run' \
--data '{
    "input": {
        "content": "Content entered here will be summarized"
    }
}'
```

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=136%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%20136%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-136-70f928025569c237cfa0de2f024f719a25a4e481.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Abilities/Summarization/Summarization.php
```diff
@@ -0,0 +1,226 @@
...
+				'content' => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'Content to summarize.', 'ai' ),
+				),
+				'post_id' => array(
+					'type'              => 'integer',
+					'sanitize_callback' => 'absint',
+					'description'       => esc_html__( 'Content from this post will be summarized. This overrides the content parameter if both are provided.', 'ai' ),
+				),
```
JasonTheAdams: I've been thinking about ways to make abilities more generalized for broader use. What would you think about renaming this to `context` and allowing _either_ a string or integer? If it's an integer, it assumes it's a post ID and does what it's doing now. If it's a context, then it's simply additional details provided to the AI, making it clear in the prompt that it should not be part of the summary.

includes/Abilities/Summarization/Summarization.php
```diff
@@ -0,0 +1,226 @@
...
+ * @since x.x.x
+ */
+class Summarization extends Abstract_Ability {
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since x.x.x
+	 */
+	protected function input_schema(): array {
```
JasonTheAdams, line 41: What do you think about adding a "length" parameter that's "short", "medium", and "long" as a enum? Basically:
* `short`: 1 sentence; <= 25 words
* `medium`: 2-3 sentences; 25-80 words
* `long`: 4-6 sentences; 80-160 words

This would be inserted in the system instruction.

includes/Abilities/Summarization/system-instruction.php
JasonTheAdams, line 1: I think we should clarify the data structure for the model to expect, and make it clear that the `content` is being summarized, while the `context` should be considered.

JasonTheAdams (changes requested):
Very cool, @dkotter! Left a few thoughts!

includes/Abilities/Summarization/Summarization.php
dkotter: If I'm following correctly, you're suggesting removing the `post_id` field and replacing that with a `context` field that can either be a post ID (which we then pull the content from, if needed, and additional context from) or manually supplied context?

The way I've been thinking of these is for more general use cases, you would pass in the content you want summarized in the `content` field (which could include any additional context you want). For our use case, we also pass in the `post_id` field which will get the content from that post (though that is overridden by the `content' field) and additional context.

I can change `post_id` to `context` but just want to make sure it's clear the difference then between the `content` field and `context` field

includes/Abilities/Summarization/Summarization.php
dkotter: Modified our system instructions methods so you can pass in extra data and have that available in https://github.com/WordPress/ai/pull/136/commits/8bd5d6a20d553bb9475899132e549cf8e2ff2853

Then added these options to our Ability here: https://github.com/WordPress/ai/pull/136/commits/6824c0402ff5b033736e6819d453f26f2cd69627

And finally modified our system instructions based on that value: https://github.com/WordPress/ai/pull/136/commits/009f2c0dccea19b8c9bdebea30e79431ebe7fe1b

includes/Abilities/Summarization/system-instruction.php
dkotter: Makes sense. I've pulled the content out of the context now so those are more obviously separated: https://github.com/WordPress/ai/pull/136/commits/d1866579d9d8838dc0bc44302bcc022cca43a6b9

We'll also likely want to make this same change for the Title Generation Experiment but I can do that in a separate PR

includes/Abilities/Summarization/Summarization.php
dkotter: Thinking about this more I think I'm following what you're suggesting now.

Right now if someone wanted to manually pass in content that isn't associated with any post and they also wanted additional context with that, they could do that all in the content field but that's probably not obvious to anyone wanting to do that.

Whereas if we change `post_id` to `context`, we can make it clear that you can pass in a post ID and it will automatically grab context from that or you can pass in your own additional context that will end up in the request.

I can go ahead and make that change

dkotter: Updated that now: https://github.com/WordPress/ai/pull/136/commits/97ef7f96db0928236454b18c2f8a919ae1a71606

includes/Abilities/Summarization/Summarization.php
```diff
@@ -0,0 +1,252 @@
...
+
+		$content = "## Content\n\n" . $content;
+
+		// If we have additional context, add it to the content.
+		if ( $context ) {
+			$content .= "\n\n## Additional Context\n\n" . $context;
+		}
+
+		// Generate the summary using the AI client.
+		return AI_Client::prompt_with_wp_error( '"""' . $content . '"""' )
```
JasonTheAdams: Rather than using headers and delimiters, I'd recommend using XML structures, as models tend to interpret that quite well naturally. So we'd have something like:
```
<content>Content goes here</content>
<additional-context>Optional additional context</additional-context>
```

This would replace the triple-quote delimiter and headers and simply by the entirety of the User message.

includes/Abilities/Summarization/system-instruction.php
```diff
@@ -0,0 +1,35 @@
...
+
+The summary should follow these requirements:
+
+- Target {$length_desc}
+- Should not contain any markdown, bullets, numbering, or formatting - plain text only
+- Provide a high-level overview, not a list of details
+- Do not start with "This article is about..." or "This post explains..." or "This content describes..." or any other generic introduction
+- Must reflect the actual content, not generic filler text
+
+The data you will be provided is delimited by triple quotes.
```
JasonTheAdams: If we use XML we don't need this.

JasonTheAdams (dismissed):
Great work, @dkotter! I have one minor suggestion remaining, but it's straightforward so I'll go ahead and approve. 👍

includes/Abilities/Summarization/Summarization.php
dkotter: Makes sense, updated in https://github.com/WordPress/ai/pull/136/changes/c97bd2ed7277bb947493e48887377b57a26046c7

includes/Abilities/Summarization/system-instruction.php
dkotter: Removed in https://github.com/WordPress/ai/pull/136/changes/c97bd2ed7277bb947493e48887377b57a26046c7

jeffpaul (approved):
Approving from earlier reviews

PR #139, opened 2025-12-06
Use plugin-check GitHub workflow in favor of brittle plugin-check PHPCS checks

### LINKED ISSUE #138: Composer dependency setup is broken
### Description

You can't do a clean `composer install` based on the current `composer.json`.

This seems to be the case due to this dependency:
```
"phpcompatibility/php-compatibility": "10.x-dev as 9.99.99",
```

Looks like that was configured as such in https://github.com/WordPress/ai/commit/9c7d4ca4af820ba093c86398af04e1ce6c3a09d1#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R13. @dkotter Curious why this is needed? Seems like an unusual configuration.

1. If you have an existing `vendor` dir, delete it.
2. Delete `composer.lock`.
3. Run `composer install`.
4. It should complain about an incompatibility regarding PHPCodeSniffer.

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

felixarntz:
Fixes #138.

## Relevant technical choices

- The problem in #138 is caused due to some issues with the brittle `plugin-check` plugin inclusion just to use its PHPCS rules. That approach leads to problems because of including a plugin that itself depends on other PHPCS foundational pieces, without being able to declare that (since it's a plugin served from `wpackagist`, so it can't express its own dependencies properly).
- The PR replaces the setup with a proper `plugin-check.yml` GitHub workflow that uses `wordpress/plugin-check-action` to run the full Plugin Check scanner. In addition to fixing the issue, it is more comprehensive.

[Click here to test this pull request](https://playground.wordpress.net/#%7B%22preferredVersions%22:%7B%22php%22:%227.4%22,%22wp%22:%22latest%22%7D,%22steps%22:%5B%7B%22step%22:%22mkdir%22,%22path%22:%22/tmp/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/tmp/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=ai&workflow=Pull%2520Request%2520Comments&artifact=ai&pr=139%22,%22caption%22:%22Downloading%20AI%20Experiments%20PR%20139%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/tmp/pr/pr.zip%22,%22extractToPath%22:%22/tmp/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginData%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/tmp/pr/ai.zip%22%7D%7D%5D%7D).


.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,71 @@
...
+        name: PHP ${{ matrix.php }} - WP ${{ matrix.wordpress }}
+        runs-on: 'ubuntu-24.04'
+        permissions:
+            contents: read
+        timeout-minutes: 20
+        strategy:
+            fail-fast: true
+            matrix:
+                php:
+                    - '8.3' # Currently the wordpress/plugin-check-action always uses the latest PHP version anyway.
```
JasonTheAdams: If it's the latest version then should this be 8.4, @felixarntz?

felixarntz: Good point, updated in https://github.com/WordPress/ai/pull/139/commits/24b253b9c4ad9f3b707e9e3fd411b10d09f19473

justlevine:
@felixarntz I originally took this pattern from wordpress/performance (where it looks like @westonruter is currently battling the same issue in https://github.com/WordPress/performance/pull/2257 ).

What (if any) are the practical implications of this change to detection? Should this be the defacto pattern for the Performance plugin too, all canonical plugins, or at least our other core-ai repos?

composer.json
```diff
@@ -13,7 +13,6 @@
   "require-dev": {
     "automattic/vipwpcs": "^3.0",
     "dealerdirect/phpcodesniffer-composer-installer": "^1.0.0",
-    "phpcompatibility/php-compatibility": "10.x-dev as 9.99.99",
```
justlevine, line 16: I'd assume we'd just be changing this to `"^10.0.0-alpha,"` why remove it entirely?

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,71 @@
...
+            contents: read
+        timeout-minutes: 20
+        strategy:
+            fail-fast: true
+            matrix:
+                php:
+                    - '8.4' # Currently the wordpress/plugin-check-action always uses the latest PHP version anyway.
+                wordpress: ['latest'] # Currently the wordpress/plugin-check-action only supports 'latest' or 'trunk'.
+        steps:
+            - uses: actions/checkout@v4
```
justlevine: Can we pin our actions with a commit hash like we do elsewhere? Lotta bad actors in dependency-world lately.

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,71 @@
...
+        name: PHP ${{ matrix.php }} - WP ${{ matrix.wordpress }}
+        runs-on: 'ubuntu-24.04'
+        permissions:
+            contents: read
+        timeout-minutes: 20
+        strategy:
+            fail-fast: true
+            matrix:
+                php:
+                    - '8.4' # Currently the wordpress/plugin-check-action always uses the latest PHP version anyway.
```
justlevine: Is there a chance this would change? I would hope it'd be runner-independent and respect our `testVersion` (why we don't have our normal phpcs tests in a matrix)

.github/workflows/plugin-check.yml
dkotter: Yes, I thought I left a comment here flagging that but seems I didn't submit it 🤦🏻 But we should pin all actions here to a commit hash, similar to our other actions: https://github.com/WordPress/ai/blob/develop/.github/workflows/test.yml#L43

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,71 @@
...
+                wordpress: ['latest'] # Currently the wordpress/plugin-check-action only supports 'latest' or 'trunk'.
+        steps:
+            - uses: actions/checkout@v4
+
+            - uses: shivammathur/setup-php@v2
+              with:
+                  php-version: ${{ matrix.php }}
+
+            - name: Install Composer dependencies
+              run: composer install
```
dkotter: Don't have to change this but if we want consistency, we use the `ramsey/composer-install` action in our other workflows: https://github.com/WordPress/ai/blob/develop/.github/workflows/test.yml#L68C21-L68C44

composer.json
dkotter: Note there's some discussion about this on https://github.com/WordPress/ai/issues/138 but I agree. I wouldn't remove this entirely. I had recommended aliasing on #138 but I think we can also change this to `phpcompatibility/php-compatibility:10.0.0-alpha2 as 9.99.99` to fix the problem

westonruter:
> I originally took this pattern from wordpress/performance (where it looks like @westonruter is currently battling the same issue in [WordPress/performance#2257](https://github.com/WordPress/performance/pull/2257) ).

Indeed! I was also thinking of switching to using the PCP action. Nevertheless, I've also opened https://github.com/WordPress/plugin-check/issues/966 so that hopefully the sniffs from PCP can be more elegantly added to existing PHPCS rulesets without the redundancy of running PCP.

composer.json
justlevine: AFAIK we don't even need the aliasing anymore:

This works just fine for me:
```json
		"phpcompatibility/php-compatibility": "^10.0.0-alpha",
		"phpcompatibility/phpcompatibility-wp": "^3.0.0-alpha",
```

.github/workflows/plugin-check.yml
justlevine: It has built-in caching, so I'd suggest we do change this.

composer.json
felixarntz: +1, I don't think we should do the aliasing, that feels quite hacky. We need to ensure our dependencies are compatible.

.github/workflows/plugin-check.yml
felixarntz: Personally I disagree with that approach, it seems just as brittle for other reasons. Why encourage auto updates for something as important as WordPress when we don't trust low-level dev-tooling in that way? 🤔 

Anyway, I know much of the project does this, so 🤷

.github/workflows/plugin-check.yml
felixarntz: The PHP version is hard-coded in other CI workflows too (FWIW it's 8.3 there, so we'll probably want to update that).

I agree with your feedback, but I'd say that should be a separate issue since this workflow isn't _introducing_ that problem.

felixarntz:
@justlevine 
> What (if any) are the practical implications of this change to detection? Should this be the defacto pattern for the Performance plugin too, all canonical plugins, or at least our other core-ai repos?

I'm not sure I understand your question, can you elaborate? "change to detection" of what?

There are two reasons I opened this PR:
* Replace brittle usage of Plugin Check PHPCS rules with actually running Plugin Check end to end.
* By doing that, address the current problem where a fresh `composer install` won't work.

composer.json
dkotter: This approach is fine by me as well, though noting it's not much different than the aliasing approach we have now (still ends up pulling in dev releases). I think the important piece here that we don't have now is the `"phpcompatibility/phpcompatibility-wp": "^3.0.0-alpha"`, as without that I think the `automattic/vipwpcs` package will complain

.github/workflows/plugin-check.yml
dkotter: For reference we (Fueled) switched to this approach after this issue: https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/ (unsure if the WordPress project also adopted this approach after this), which is prevented if you pin to commit hashes. Also what [GitHub](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) themselves recommend for ideal security.

justlevine: IIRC it's also SOC 2 requirement.

.github/workflows/plugin-check.yml
justlevine: I'm fine with the hardcoding of the version. I'm asking why choose to wrap it in a matrix?

very much a unimportant nit, was asking more because I was curious about PCP (I'd assume they'd want to actively avoid requiring users to run version-specific scans, but IIRC they're also doing more than just PHPCS)

justlevine:
> @justlevine
> 
> > What (if any) are the practical implications of this change to detection? 
> 
> I'm not sure I understand your question, can you elaborate? "change to detection" of what?
> 

Sorry, yeah that was horrible phrasing. 
I'm asking if the workflow detects the same/same+more/more+different/fewer smells as the internal rulesets we're removing. 

(Depending on the answer, we might want to e.g. pin `~1.6.0` temporarily and fix the brittleness directly so we can still get the existing sniff coverage and be able to run it locally, and/or add PCP to our wp-env with the matching package.json WPCLI command, and/or prioritize backporting GH workflow to other repos, etc)

.github/workflows/plugin-check.yml
felixarntz: Fair. I use a matrix out of personal convenience, but we can totally get rid of it here. Will update.

felixarntz:
@justlevine 
> Sorry, yeah that was horrible phrasing. I'm asking if the workflow detects the same/same+more/more+different/fewer smells as the internal rulesets we're removing.

I answered something like that in https://github.com/WordPress/ai/issues/138#issuecomment-3628990023: The PHPCS sniffs from Plugin Check are only a subset of what Plugin Check checks.

> (Depending on the answer, we might want to e.g. pin `~1.6.0` temporarily and fix the brittleness directly so we can still get the existing sniff coverage and be able to run it locally, and/or add PCP to our wp-env with the matching package.json WPCLI command, and/or prioritize backporting GH workflow to other repos, etc)

I think including sniffs from Plugin Check is always going to be problematic because it's a plugin, not a proper package with `composer.json` etc, particularly not when you include it via `wpackagist`. Setting up PCP in `wp-env` may be a worthwhile effort, but I think that could be explored separately. It's most important to check this in CI, so unless any of us has previous experience in setting up PCP alongside the main plugin in `wp-env` for easy local usage, I'd suggest we make this a separate issue/PR to not block this.

.github/workflows/plugin-check.yml
felixarntz: Done in https://github.com/WordPress/ai/pull/139/commits/d95d47373c74adb6f36c68301a7e530e8d0118f2

.github/workflows/plugin-check.yml
felixarntz: Done in https://github.com/WordPress/ai/pull/139/commits/76d957bd99b9f49890cdacd25b0f5ed2291131ca

.github/workflows/plugin-check.yml
felixarntz: Done in https://github.com/WordPress/ai/pull/139/commits/0cd97bbcdebd515f5efd742038c2f355d3e5d23a

composer.json
felixarntz: I addressed this in https://github.com/WordPress/ai/pull/139/commits/0e3d8279dc41aca44a62982a84e6a8efc6e1c80e: We should only need to include `"phpcompatibility/phpcompatibility-wp"` because we don't use `"phpcompatibility/php-compatibility"` anywhere explicitly, in other words, it's an indirect dependency - no need to specify it.

By requiring `"phpcompatibility/phpcompatibility-wp": "^3.0.0-alpha"`, `"phpcompatibility/php-compatibility": "^10.0.0-alpha"` will be installed anyway.

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,71 @@
...
+                  node-version-file: '.nvmrc'
+
+            - name: Install NPM dependencies
+              run: npm ci
+
+            - name: Build assets
+              run: npm run build
+
+            - name: Run plugin check
+              uses: wordpress/plugin-check-action@v1
```
justlevine: https://github.com/WordPress/plugin-check-action/commit/ec9b3fe9beaa76bcc4510b7ba2cb5855a5f80f3f

```suggestion
              uses: wordpress/plugin-check-action@ec9b3fe9beaa76bcc4510b7ba2cb5855a5f80f3f # v1.1.4
```

justlevine (commented):
Once `plugin-check-action` is pinned, LGTM 🚀

dkotter (approved):

PR #65, opened 2025-10-31
Separate the Build Plugin Zip action from the PR Comments action

dkotter:
## What?

Instead of having the `Build Plugin Zip` action being called from the `Pull Request Comments` action, separate those into two separate actions that are both triggered when a PR is opened.

## Why?

I wanted to have the `Pull Request Comments` action always call the `Build Plugin Zip` action before proceeding so we could ensure the zip was built properly prior to adding a Playground testing comment to the PR. But as I've attempted to get this working now across a [few](https://github.com/WordPress/ai/pull/46) [different](https://github.com/WordPress/ai/pull/49) [PRs](https://github.com/WordPress/ai/pull/56), I've run into issues with this approach.

This mostly comes down to permissions (or lack thereof). I originally was using the `pull_request_target` event as our trigger, which gives higher permission levels to the action workflow. The problem here is when this is the event, GitHub uses the base branch as the reference instead of the PR branch. This meant the zip being built was always based on `trunk` and never the PR branch, effectively making it worthless.

I switched over to using `pull_request` and upped the permission level on the job and that [seemed to work fine](https://github.com/WordPress/ai/actions/runs/18921321664/job/54017991059). But in my latest PR, getting [permission errors again](https://github.com/WordPress/ai/actions/runs/18952487053/job/54120608467) and from what I've been able to find out, it seems when a PR is opened from a fork, GitHub limits the permissions on that, no matter what permissions you have set in the job (could also be something else I'm doing wrong so open to other opinions here if I'm wrong on this).

## How?

Instead of trying to continue to mess with permissions to figure this out, decided it was easiest to just split this into two separate workflows that both are triggered when a PR is opened. So now the `Build Plugin Zip` action is triggered on the `pull_request` event, meaning the zip will be built from the PR branch. And the `Pull Request Comments` action is again triggered from the `pull_request_target` event, giving it higher permissions to add the Playground testing comment to the PR.

Worth noting the one potential issue here is it's likely the `Build Plugin Zip` action will take longer than the `Pull Request Comments` action, meaning a testing comment will be added prior to the zip actually being done. I don't expect this to be an issue in real life, as we're talking just a minute or two difference, but did want to note that.

## Testing Instructions

1. Ensure the Playground testing comment is getting added to this PR description.
2. Ensure both the `Build Plugin Zip` action and `Pull Request Comments` action succeed

## Test using WordPress Playground

[...]

PR #144, opened 2025-12-11
Try out the new `WordPress/action-wp-playground-pr-preview` GitHub Action

dkotter:
## What?

Update our Build Plugin Zip and Pull Request Comments workflows to use the new `WordPress/action-wp-playground-pr-preview` Action.

## Why?

We've had a few issues with our current setup (see #65 for latest work there) and we've been asked to try using the `WordPress/action-wp-playground-pr-preview` Action to see if that makes things work.

## How?

Follows the instructions in the `WordPress/action-wp-playground-pr-preview` repo: https://github.com/WordPress/action-wp-playground-pr-preview/?tab=readme-ov-file#advanced-testing-built-ci-artifacts

- Have our `Build Plugin Zip` workflow fire when a PR is opened
- Update our `Pull Request Comments` workflow to fire after the `Build Plugin Zip` workflow has finished
- Update this workflow to use the new `WordPress/action-wp-playground-pr-preview` Action

## Testing Instructions

Ensure a Playground test link gets added to this PR description and it works as expected

### DIFF
--- a/.github/workflows/pr-playground-preview.yml
+++ b/.github/workflows/pr-playground-preview.yml
@@ -1,4 +1,4 @@
-name: Pull Request Comments
+name: PR Playground Preview
 
 # Use workflow_run for privileged operations
 # Runs with write permissions and access to secrets
--- a/.github/workflows/build-plugin-zip.yml
+++ b/.github/workflows/build-plugin-zip.yml
@@ -17,7 +17,6 @@ permissions: {}
 
 jobs:
   build:
-    name: Build Release Artifact
     runs-on: 'ubuntu-24.04'
     permissions:
       contents: read
--- a/.github/workflows/build-plugin-zip.yml
+++ b/.github/workflows/build-plugin-zip.yml
@@ -1,7 +1,8 @@
 name: Build Plugin Zip
 
 on:
-  workflow_call:
+  pull_request:
+    types: [ 'opened', 'synchronize', 'reopened', 'edited' ]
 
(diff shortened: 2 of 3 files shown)


.github/workflows/pr-playground-preview.yml
```diff
@@ -0,0 +1,108 @@
+name: PR Playground Preview
+
+# Use workflow_run for privileged operations
+# Runs with write permissions and access to secrets
+# Operates on artifacts from the unprivileged build workflow
+on:
+  workflow_run:
+    workflows: ["Build Plugin Zip"]
+    types:
+      - completed
```
jeffpaul, line 10: For some reason this doesn't seem to get triggered, at least in this PR.

adamziel: @jeffpaul @dkotter This is probably the crux:

> This event will only trigger a workflow run if the workflow file exists on the default branch.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

dkotter: Ah, thanks for that. I think then the only way to fully test this will be to merge it in and 🤞🏻

jeffpaul: merged, now crossing fingers...

dkotter: Workflow is failing with a permission error. Opened #150 to try and fix that but still gives the same error 😢

PR #146, opened 2025-12-16
Add the Featured Image Generation UI

### LINKED ISSUE #13: Image Generation

dkotter:
## What?

Partially closes #13. This only handles generating featured images, I'm assuming we'll want image generation in other places as well (though could be tracked in new Issues if we want).

Adds in the UI to trigger Featured Image Generation

## Why?

This builds on top of the work done in #134, which added Abilities that could be used to generate an image and/or import an image. This PR adds in the actual UI that triggers those Abilities in the Featured Image section of the edit post screen as well as a new Ability to generate an image generation prompt.

## How?

- Modifies the Image Import Ability to allow passing in additional meta data that will be stored with the image. This allows us to easily track which images are AI generated based on a specific meta key
- Modifies the Image Generation Ability to return not only the generated image data but also information about the provider and model used
- Adds a new Ability that is used to generate an image generation prompt. We pass this the content of the post along with the title and content type. Can pass in other context if desired along with optional style instructions
- Loads in the necessary admin scripts and admin script data
- Adds in all the needed javascript to wire things up, including:
   - a `Generate featured image` button that shows above the standard `Set featured image` button
   - When that button is clicked, first make a request to the `get-post-details` Ability. Then pass those details to the new `image-prompt-generation` Ability, along with our content and context. Take the prompt this generates and pass it to our existing `image-generation` Ability. If the alt text generation Experiment is turned on, we then pass the image to the `alt-text-generation` Ability to get accurate alt text (if not enabled, default to using the prompt). Finally take the base64-encoded image it returns and pass it to our existing `image-import` Ability to import the image into the Media Library
   - Show a progress message as this process goes through so users know what step they are on
- Set this image as the Featured Image and change the text of the generation button from `Generate featured image` to `Generate new featured image`
- Add a label below the Featured Image (if set and if AI generated) that makes it clear the image was AI generated

## Testing Instructions

1. Ensure you have valid AI credentials in place for OpenAI or Google under `Settings > AI Credentials`
2. Ensure you turn on the Image Generation Experiment under `Settings > AI Experiments`
3. Optionally turn on the Alt Text Generation Experiment
4. Edit a piece of content that doesn't have a Featured Image yet
5. Ensure you see the `Generate featured image` button
6. Click this button (**noting this will make an API request and will cost money**)
7. Ensure a proper loading state is shown and that eventually (this can take around 90s) an image is rendered in the normal featured image section
8. Ensure this image imported properly into the Media Library and has a proper title and description
9. Ensure the AI label shows below the image in the featured image section

To directly test the new generate image prompt Ability, can make an API requests like the following:

```bash
curl --location 'https://example.com/wp-json/wp-abilities/v1/abilities/ai/image-prompt-generation/run' \
--data '{
    "input": {
        "content": "This is the content of an article",
        "context": "Title: Hello World"
        "style": "Moody, dark colors"
    }
}'
```

## Screenshots or screencast

New Generate featured image button showing above the default Set featured image button in the block editor sidebar

Loading state of the new Generate featured image button

AI featured image rendering properly with AI label beneath the image

AI generated image shown in the media library

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-146-04c2df456e81be172e48c114f93a2f5a6f7c4fee.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Abilities/Image/Generate_Image.php
```diff
@@ -117,6 +117,14 @@ protected function meta(): array {
 	 * @return string|\WP_Error The generated image data, or a WP_Error if there was an error.
 	 */
 	protected function generate_image( string $prompt ) { // phpcs:ignore Generic.NamingConventions.ConstructorName.OldStyle
+		// Increase the default timeout as image generation can take a while.
+		add_filter(
+			'wp_ai_client_default_request_timeout',
+			static function () {
+				return 90;
+			}
+		);
```
JasonTheAdams: Fortunately, there's a _much_ better way of doing this on a per prompt basis! The builder has a `using_request_options()` method. So for this you'd add:
```php
$request_options = new WordPress\AiClient\Providers\Http\DTO\RequestOptions();
$request_options->setTimeout(90);

$prompt->using_request_options($request_options);
```

JasonTheAdams (changes requested):
One straightforward suggestion, but I'm also humming pretty hard at the flow of providing an AI prompt writing Ability in order to run the subsequent prompt. What problem is this extra roundtrip solving that we couldn't accomplish by compiling a prompt ourselves based on the post data? It feels like we could make due without this.

The only place I could see an Ability like that being genuinely useful is if a user had an input to do something like describe a picture, and we added a "Improve prompt" button. But even that would require more specific instructions based on the context. This feels too generic.

jeffpaul:
@dkotter perhaps a docs page for this experiment like done in https://github.com/WordPress/ai/pull/155/changes#diff-f70665388390657902e73741dea03efcdadc4bd0ceb4432959f8903ae015c042?

includes/Abilities/Image/Generate_Image.php
dkotter: Ah, that's good to know, thanks for pointing that out. Updated in https://github.com/WordPress/ai/pull/146/changes/460f52974e8d5ee27854cd6f48196779bbe5b5dd

dkotter:
> One straightforward suggestion, but I'm also humming pretty hard at the flow of providing an AI prompt writing Ability in order to run the subsequent prompt. What problem is this extra roundtrip solving that we couldn't accomplish by compiling a prompt ourselves based on the post data? It feels like we could make due without this.
> 
> The only place I could see an Ability like that being genuinely useful is if a user had an input to do something like describe a picture, and we added a "Improve prompt" button. But even that would require more specific instructions based on the context. This feels too generic.

@JasonTheAdams I struggled coming up with an approach to turn an entire article into a useful image generation prompt (useful being key there), thus relying on AI to help us out here. I initially tried sending the entire article with a prompt along the lines of `Generate a featured image for the following article` but that didn't work great and leads to huge prompts. Open to suggestions though on other alternatives to solve this but given the wide-range of content someone may be using, I couldn't think of a better way to do this.

I did go back and forth on if this should be an Ability (the "generate a prompt from this context" piece) so happy to change that to be more specific to this use case instead of a globally exposed Ability if that helps at all.

dkotter:
> @dkotter perhaps a docs page for this experiment like done in https://github.com/WordPress/ai/pull/155/changes#diff-f70665388390657902e73741dea03efcdadc4bd0ceb4432959f8903ae015c042?

@jeffpaul I know James has mentioned these PRs are more WIP at the moment so wondering if that's the best example to copy from? I know our default Experiment did [add a `README` file](https://github.com/WordPress/ai/blob/develop/includes/Experiments/Example_Experiment/README.md) within the Experiments directory but we haven't followed that approach so far, though I can do that if desired. I just think we need to decide on what our standard documentation structure should be so we can ensure it matches across everything.

JasonTheAdams:
I chatted with some other folks and did some research on this, @dkotter, and you're right. In fact, it's rather complex. 😅 

Image prompts are strange things and they vary depending on whether the image generation model is using diffusion or a transformer. If diffusion, things like context and history are more more confusing, as those models tend to expect the prompt to be rather focused on describing the image. Transformer models are more forgiving, but still work better when the prompt focuses on the image.

In this case we're really describing two bits of work for the models:
1. Taking a chunk of content and using it as _inspiration_ for an image description
2. Generating an image

Trying to get the image generation model to do both is, as you found, not particularly successful. In fact, depending on what type of model you used it could be really bad. Hahah!

So it does make sense to have a model and system prompt focused on taking content and deriving an image prompt. It will need to understand that the content is a source of inspiration and to look for things like archetypes and what not within the content. This is some prompt engineering in and of itself. We may even take more parameters for things like describing the style of the image and such, which the content itself won't provide. This gets into the territory of a [content guidelines as James has explored](https://github.com/Jameswlepage/content-guidelines).

At this point I'm thinking it would be useful to have an "Image prompt from content" ability that takes in:
1. Content to use as inspiration
2. Style instructions (could be broken up into granular parameters)

This way we can really focus the system prompt on understanding its singular purpose, which I think will improve the resulting image prompt. We _may_ consider having default style instructions so the generated images are somewhat consistent.

jeffpaul:
> At this point I'm thinking it would be useful to have an "Image prompt from content" ability that takes in:

Given that we'll likely have the same problem should we get into video generation, should we be more generic in naming this ability than tying it directly to "image"?

JasonTheAdams:
> Given that we'll likely have the same problem should we get into video generation, should we be more generic in naming this ability than tying it directly to "image"?

I don't think so. I believe video prompts will want their own tuning and such. I think we should stick to just images for now and tackle videos later.

dkotter:
> So it does make sense to have a model and system prompt focused on taking content and deriving an image prompt. It will need to understand that the content is a source of inspiration and to look for things like archetypes and what not within the content. This is some prompt engineering in and of itself. We may even take more parameters for things like describing the style of the image and such, which the content itself won't provide. This gets into the territory of a [content guidelines as James has explored](https://github.com/Jameswlepage/content-guidelines).
> 
> At this point I'm thinking it would be useful to have an "Image prompt from content" ability that takes in:
> 
> 1. Content to use as inspiration
> 2. Style instructions (could be broken up into granular parameters)
> 
> This way we can really focus the system prompt on understanding its singular purpose, which I think will improve the resulting image prompt. We _may_ consider having default style instructions so the generated images are somewhat consistent.

@JasonTheAdams Okay, I've updated this PR to take all the above into consideration. There is now a more specific Ability, `image-prompt-generation`, that allows you to pass in content, optional context (like title, author, content type) and optional style suggestions.

This Ability will then generate an image generation prompt which we then pass directly to an image generation model.

In my testing this is all working as expected but definitely open to suggestions on the system instructions we're using here and any other default instructions we may want in place to guide things.

includes/Abilities/Image/Generate_Image_Prompt.php
```diff
@@ -0,0 +1,213 @@
...
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since x.x.x
+	 *
+	 * @return bool True if the user is logged in, false otherwise.
+	 */
+	protected function permission_callback( $args ) {
+		// Ensure the user is logged in.
+		return is_user_logged_in();
```
dkotter: Change to `current_user_can( 'edit_post' )` to match what we do in other places and limit this Ability to Authors and above

dkotter:
@jeffpaul This is ready for final review, noting I've not had success getting this to fully work in the Playground as the request to generate an image typically takes about 60 seconds and Playground always times out before that

jeffpaul:
This tested really nicely, and yes a bit of a lag in generating but there's :alot: happening in the background.  To that specific extent, not something to hold up this PR, but perhaps helpful for a future iteration (but feel free to tell me I'm dumb here and we can skip it) would be some simple "loading text" that describes what's happening (e.g. "Generating image generation prompt from post content...", "Generating image from generated prompt...") so folks understand why its not an immediate response?

One thing that I think is worth pausing on to decide how best to land this experiment is the specific alternative text, title, caption, description for generated images.  Currently the Title is set as `AI Generated Image` and the Description is `This is an AI generated image.`

Screenshot 2026-02-05 at 12 23 33 PM

Perhaps an iteration here to the Alt Text getting generated as well from the resulting image (perhaps hooking into the alt text ability?), the Title getting a truncated or condensed version of the generated alt text, the Caption stays empty as that's meta that more often is also rendered on the front end and may be best left for author/editor input, and the Description gets some provenance info (though anyone who knows me knows I'd LOVE to see something like [c2pa-js](https://opensource.contentauthenticity.org/docs/c2pa-js/) used to store meta here) so something like “Generated by ExampleAI Image Model v2 on 05 Feb 2026. Prompt: ‘Minimal illustration of a 1700s sailor staring out to see from a dock.’”?

dkotter:
> This tested really nicely, and yes a bit of a lag in generating but there's :alot: happening in the background. To that specific extent, not something to hold up this PR, but perhaps helpful for a future iteration (but feel free to tell me I'm dumb here and we can skip it) would be some simple "loading text" that describes what's happening (e.g. "Generating image generation prompt from post content...", "Generating image from generated prompt...") so folks understand why its not an immediate response?

The only issue here is while yes, we are running lots of things one after the other, they all run fairly quickly besides the actual image generation. So we could add a progress message but I think what we'd see is the message shows then is replaced before someone can actually read it, which may be more annoying then helpful. But the actual image generation part will take time (OpenAI is usually around 60 seconds, Google is typically faster but still ~40 seconds) so maybe still useful to have those messages?

> One thing that I think is worth pausing on to decide how best to land this experiment is the specific alternative text, title, caption, description for generated images

I'm open to any suggestions here, I purposely left it fairly basic as a starting point.

> Perhaps an iteration here to the Alt Text getting generated as well from the resulting image (perhaps hooking into the alt text ability?), the Title getting a truncated or condensed version of the generated alt text

My one concern in regards to alt text is that's a separate Experiment and we can't assume users will have both of those active (Image Generation and Alt Text Generation). In that situation, not sure how we handle that, as the Alt Text Generation Ability won't be available to use if it's not turned on. I guess we could have a light-weight duplicate of that specific to Image Generation, though I could see some complaints there from users wondering why AI is generating alt text when they don't have that Experiment turned on.

> the Caption stays empty as that's meta that more often is also rendered on the front end and may be best left for author/editor input

Yeah, I agree on keeping this empty

> the Description gets some provenance info (though anyone who knows me knows I'd LOVE to see something like [c2pa-js](https://opensource.contentauthenticity.org/docs/c2pa-js/) used to store meta here) so something like “Generated by ExampleAI Image Model v2 on 05 Feb 2026. Prompt: ‘Minimal illustration of a 1700s sailor staring out to see from a dock.’”?

Main issue here is the Ability that generates an image only returns the generated image, it doesn't return anything else, like provider, model, prompt used, etc. We could look to update the Ability to do that if we think that would be valuable, though I'd also need to figure out if the AI Client can tell us what provider and model was used (I'm assuming so but not something I've tried).

jeffpaul:
> The only issue here is while yes, we are running lots of things one after the other, they all run fairly quickly besides the actual image generation. So we could add a progress message but I think what we'd see is the message shows then is replaced before someone can actually read it, which may be more annoying then helpful. But the actual image generation part will take time (OpenAI is usually around 60 seconds, Google is typically faster but still ~40 seconds) so maybe still useful to have those messages?

Ok, yeah maybe we ignore this for now and see if there's community feedback about the time to completion and if _some sort of loading info_ is helpful.

> My one concern in regards to alt text is that's a separate Experiment and we can't assume users will have both of those active (Image Generation and Alt Text Generation). In that situation, not sure how we handle that, as the Alt Text Generation Ability won't be available to use if it's not turned on. I guess we could have a light-weight duplicate of that specific to Image Generation, though I could see some complaints there from users wondering why AI is generating alt text when they don't have that Experiment turned on.

Perhaps if the alt text experiment is disabled / that ability is not available, we fall back to adding in directly the raw image gen prompt (knowing that'll likely include sub-optmial "Generate image of" sort of context, but at least someone generating the image could immediately edit/remove it?).

> Main issue here is the Ability that generates an image only returns the generated image, it doesn't return anything else, like provider, model, prompt used, etc. We could look to update the Ability to do that if we think that would be valuable, though I'd also need to figure out if the AI Client can tell us what provider and model was used (I'm assuming so but not something I've tried).

Let's chat about how quickly feasible this could be, balancing that with still getting this feature shipped since its otherwise near ready.

dkotter:
> Ok, yeah maybe we ignore this for now and see if there's community feedback about the time to completion and if _some sort of loading info_ is helpful.

I went ahead and added this as it was pretty simple and now that we're potentially generating alt text as well, that's another few seconds of waiting. So now each step will show a progress message.
 
> Perhaps if the alt text experiment is disabled / that ability is not available, we fall back to adding in directly the raw image gen prompt (knowing that'll likely include sub-optmial "Generate image of" sort of context, but at least someone generating the image could immediately edit/remove it?).

Made this change now. If the alt text Experiment is on, after an image is generated we send it back to generate alt text and use that. If that Experiment isn't on, we use the prompt instead.

We then take that and trim it down to use as the title instead of hardcoding to `AI Generated Image`
  
> Let's chat about how quickly feasible this could be, balancing that with still getting this feature shipped since its otherwise near ready.

This was also fairly easy so I've gone ahead and made this change. The image generation Ability will now return the image data plus information about the provider and model used. This information is then used in the image description after is is imported.

jeffpaul:
@adamziel one thing I've run into in using the Playground testing link in this PR is that sometimes Playground stalls out / response times drag on and I'm not reasonably able to test PRs in Playground (with this PR its because of the comms/wait in generating a prompt+image).  Perhaps an enhancement for the Playground PR button action would be to also export below the button something like `(or download a ZIP here)` sort of text that links to the ZIP used so folks could test locally?

jeffpaul:
> Error: Network error occurred while sending request to https://api.openai.com/v1/images/generations: cURL error 28: Operation timed out after 30002 milliseconds with 0 bytes received

Currently getting timeout issues in testing locally, going to build a new local and try again

jeffpaul:
Ok, got things working locally, phew, one tweak is the date in the Description field is likely US-centric MM/DD/YYYY (I'm guessing) so let's update to DD Month YYYY (e.g. 6 February 2026)?

dkotter:
> Ok, got things working locally, phew, one tweak is the date in the Description field is likely US-centric MM/DD/YYYY (I'm guessing) so let's update to DD Month YYYY (e.g. 6 February 2026)?

Right now we're using `new Date().toLocaleDateString()` to generate this date, which the `toLocaleDateString` part of that will

> return a string representing the date portion of a `Date` object, formatted according to specific locale conventions and time zones

So instead of forcing a particular format, the format used will depend on the individual user's locale and timezone settings. I can update that though if we want to force a specific format for all users.

jeffpaul:
> So instead of forcing a particular format, the format used will depend on the individual user's locale and timezone settings. I can update that though if we want to force a specific format for all users.

While I'd like to make the format a bit more ISO-specific, its probably best to follow the site-specific setting for now and we can always iterate in the future if feedback comes back strongly in a different direction.

jeffpaul (approved):

PR #147, opened 2025-12-18
Add the Content Summarization UI

dkotter:
## What?

Closes #12

The base Content Summarization Experiment and Ability were added in #136. This PR builds on top of that to add in the actual UI to trigger content summarization within the post editor.

## Why?

With #136 merged in, you can manually trigger summarization by using the Ability but there isn't a way to trigger that from the UI. This PR adds in that UI, making that Experiment more useful.

## How?

- Ensure we properly build the needed assets for the Content Summarization Experiment and then load that properly in the admin
- Add a component that renders in the Post Status section of the block editor. This will render a button that can be clicked to generate a summary
- When that button is clicked, hit the Content Summarization Ability endpoint and if successful, store the summary in post meta and also render that in the editor
- Create a block variation of the paragraph block that we use to render the summary
- Add a custom button to the toolbar of this block that can also be used to re-generate a summary
- Also updates our NPM dependencies to their latest versions

## Testing Instructions

1. Go to `Settings > AI Credentials` and ensure you have a valid AI provider configured
2. Go to `Settings > AI Experiments` and ensure you have the Content Summarization Experiment enabled
3. Edit a post that has content (or add new content) and ensure you see a `Generate AI Summary` button in the Post Status sidebar
4. Ensure you can click on this button, a loading state shows and once it completes, a block is automatically inserted into the editor with a summary
5. Click into that block and ensure you see a Regenerate button in the toolbar
6. Click on that button and ensure it works as expected

## Screenshots or screencast

Generate Summary button showing in the Post Status section of the block editor

---

AI generated summary being rendered in the AI Summary paragraph block variation

---

Re-generate Summary button showing in the Post Status section of the block editor

---

Regeneration button working in the AI Summary block variation

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-147-c2098b09adfcd7ee9f403adb609f8ffe9161c617.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #148, opened 2025-12-19
Add Extended Providers experiment

### LINKED ISSUE #27: Display additional AI provider plugins on Connectors page (alongside default Anthropic, Google, and OpenAI ones)
### Description

Enable developers, agencies, and hosts to pre-configure AI providers that will be automatically recognized by the plugin.  This supports a “progressive provider selection” approach (where existing providers are utilized and only presents provider configuration needs if no applicable provider has been configured) and reduces the need for manual setup.

### Goals

- [ ] Define a configuration system (using constants, filters, or configuration files) to allow pre-configured providers.
- [ ] Ensure the system is modular so each feature can independently detect and use available providers.
- [ ] When a valid pre-configured provider exists, the UI for provider selection should be hidden.
- [ ] Document the configuration process with clear examples for hosts and network administrators.
- [ ] Align with the encapsulated feature architecture so each feature can extend or override provider handling if needed.

### Acceptance Criteria

- Plugin detects and auto-selects any pre-configured providers.
- Provider selection UI is suppressed when configuration is detected.
- Features automatically inherit available provider connections.
- Developer documentation includes at least one working example.

Jameswlepage:
CleanShot 2025-12-19 at 09 15 39@2x

## What?

Adds a new Extended Providers experiment that registers 9 additional AI provider integrations:
- Cloudflare Workers AI
- Cohere
- DeepSeek
- Fal.ai (image generation)
- Grok (xAI)
- Groq
- HuggingFace
- Ollama (local)
- OpenRouter

Also includes provider credentials UI enhancements and a metadata registry for custom branding per provider.

Closes #27.

## Why?

The core wp-ai-client package ships with OpenAI, Anthropic, and Google. Users want access to additional providers—especially local options like Ollama and cost-effective alternatives like Groq and DeepSeek.

## How?

- Adds `Extended_Providers` experiment class that registers all 9 providers with the wp-ai-client registry
- Each provider includes a Provider class, TextGenerationModel (or ImageGenerationModel for Fal.ai), and ModelMetadataDirectory
- Adds `Provider_Metadata_Registry` for custom branding (icons, colors, API key URLs)
- Adds `Provider_Credentials_UI` initialization to display enhanced credentials screen
- Works around provider registration timing issue (see WordPress/wp-ai-client#43)

## Testing Instructions

1. Enable the Extended Providers experiment in **Settings → AI → Experiments**
2. Navigate to **Settings → AI** and verify all 9 providers appear on the credentials screen
3. Add an API key for at least one cloud provider (e.g., Groq, DeepSeek) and verify it saves
4. If you have Ollama running locally, verify it appears and connects

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-148-c1a152b57d04a5bf2ad5cd3be7995d255345c8ad.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:

jeffpaul:
- [ ] Reminder for us to add a screenshot gif and entry in the readme.txt for this feature before merging/releasing.

huzaifaalmesbah:
It appears that Extended Providers cannot be configured unless a primary provider (OpenAI / Anthropic / Google) API key is set.

When no primary provider key is configured, the Extended Providers credentials UI does not render. As a result, users are unable to configure or use any of the newly added providers (e.g., Groq, DeepSeek, Ollama, etc.).

Is this dependency expected?

Since the goal of this experiment is to expand provider options—especially for users who may want to rely solely on alternatives like Groq, DeepSeek, or local Ollama—it might make sense for Extended Providers to function independently of the core providers.

If this coupling is not required by design, we may need to decouple the experiment from the primary provider requirement so that:

* The experiment can be enabled independently
* Extended providers can be configured without requiring a core provider API key

Could you clarify the expected behavior here?


Huzaifa-20260211223608

jeffpaul:
@huzaifaalmesbah that's likely because this PR branched off an earlier version of the plugin where that provider requirement was a blocker.  I concur that for this experiment that no preexisting provider needs to be configured, but I suspect resolving the merge conflicts may "fix" that.

jeffpaul:
The primary question I have here is whether the approach in `includes/Providers` is the "right" technical approach we'd want folks following in adding their own providers as they extend the plugin for their own needs.

cc: @felixarntz @JasonTheAdams for your input as this could be a good example for folks looking to leverage the WP AI Client as it merges to core and has the three existing providers spun out to their own plugins

jeffpaul:
@Jameswlepage with the advent of the Connectors screen coming to WP 7.0, perhaps the work here is best suited to crafting additional AI provider plugins for these connections and closing out this draft PR?

Jameswlepage:
I think you must have read my mind because after today's call I started jamming on that. Should have an update for this soon.

Jameswlepage:
CleanShot 2026-03-04 at 21 04 22@2x
CleanShot 2026-03-04 at 21 04 12@2x

gziolo:
https://github.com/WordPress/wordpress-develop/pull/11175 - it’s something I still consider for WP 7.0. This way, with the AI plugin, we could feature such extended providers with a single PHP hook on the new Connectors screen.

Jameswlepage:
I think we'll keep this as a draft until the connectors work is stable cc @gziolo

jeffpaul:
@Jameswlepage if we want to add providers to the Connectors screen, should that approach be to leverage AI Provider for XXX sorts of plugins instead of provider code within the AI Experiments plugin?  Feels like the model core and its extenders should be moving to is a suite of AI Provider for XXX plugins that perhaps plugins like AI Experiments decide to include certain ones as available alongside the core 3 for whatever-reasons-the-plugin-has, yeah?  Prior to the Connectors screen this sort of approach in this PR made sense, now with the Connectors screen and provider plugins I'm not so sure.

Not sure I understand the motive to keep provider code within the AI Experiments plugin at this point, so if I'm missing something please share.

gziolo:
With [AI Experiments 0.5.0 (12 MAR 2026)](https://make.wordpress.org/ai/2026/03/12/whats-new-in-ai-experiments-0-5-0/), we moved toward the native Connectors screen from WordPress core. We can pivot and take a more refined approach. My proposal is to collect vetted AI provider plugins from the Plugins Directory and inject them so they display the same way as featured providers (Anthropic, Google, OpenAI). Essentially, even when users don't have them installed, they would see the card with an on-click installation flow. I can prototype that to validate that it's already possible today for AI providers with an auth method set to API key.

jeffpaul:
> collect vetted AI provider plugins from the Plugins Directory

I'm happy to see this effort continue, but we should in parallel define what criteria elevates something to be listed on the Connectors page.  Does anyone have even some rudimentary thoughts that we can use to start some documentation around this?  cc: @Jameswlepage @felixarntz @swissspidy

swissspidy:
Some of this I have already shared on Slack, but:

We have to consider the implications of listing third-party plugins on that screen. This will put an additional burden/responsibility on third-party plugin developers to add their plugin to that list and ensure that it's up-to-date and secure etc. They might suddenly deal with a big influx of support requests. Is this perhaps too big of an ask? Or should they be transferred to be maintained by WordPress? Probably not, as this would put unnecessary work on contributors' shoulders.

Right now the list of connectors is hardcoded. It would be much better for this data to come from the WordPress.org plugins API instead, where we can more easily swap out plugins and highlight different ones. Then the above point becomes less of a concern. It will be more akin to the Featured plugins tab.

Either way, it begs the question of how these third-party plugins are marked as such on the connectors page. Users will not realize that these aren't official WordPress offerings. Otherwise we'll get tons of support requests or security reports in Trac/HackerOne for things that we don't maintain.

gziolo:
> It would be much better for this data to come from the WordPress.org plugins API instead, where we can more easily swap out plugins and highlight different ones.

That would be my favorable approach in the long term. We can exercise that in the AI plugin first and replicate it in WP core as soon as possible, given the release cadence.

felixarntz:
I think we should go with individual plugins, one for each provider.

They can live anywhere of course, but those that we consider official and supported by the WordPress project itself should live under the `WordPress` GitHub org.

We could probably even use that technically: Thinking of two groups:
- We could have an area in the connectors screen that shows any plugins with a specific tag (e.g. "connector"). Those would need to be explicitly annotated as third party plugins though (just like pretty much every plugin out there).
- As a sunset of that, any plugin that is maintained under the `WordPress` GitHub org would show where we currently have the three hard-coded entries.

As for naming, I think for our official plugins we should continue to use the "AI Provider for XYZ" naming convention. But it shouldn't be a formal policy, beyond the one that already exists about clearly indicating that something is not _by_ the relevant company.

In other words, _if_ at some point Anthropic, Google, OpenAI or others were to build their own first-party provider implementations (which honestly would be awesome!), of course they should be allowed to call it "XYZ Provider" or "XYZ AI Provider", since they are obviously the owners of that trademark.

jeffpaul:
Closing this PR as the implementation direction has shifted.

The original approach of bundling extended providers in the plugin no longer aligns with the move toward:
- the WordPress core Connectors screen
- individual provider plugins

I've opened a follow-up issue #502 to define the provider plugin discovery model and Connectors integration.  This PR was valuable in surfacing the right questions, but we need to resolve them at the architectural level first.

PR #151, opened 2025-12-19
Add Type Ahead experiment

Jameswlepage:
## What?

Adds a new type ahead experimental Feature. Automatically triggers at the end of a paragraph block and can be manually triggered within the block.

## Why?

Type ahead / ghost text autocomplete is a fairly common AI pattern so this experimental Feature brings it into WordPress. Works on paragraph blocks and optionally heading blocks if turned on. Will send the surrounding content and asks the LLM to generate the next handful of words (could be a full sentence, could be a partial one).

## How?

- Registers a new `Type_Ahead` experimental Feature and a `Type_Ahead` Ability
- Automatically adds AI-powered ghost text autocomplete suggestions when at the end of a paragraph block and optionally a heading block
- Suggestions should be context-aware based on surrounding content
- Can trigger manually within a paragraph/heading block using the keyboard shortcut Ctrl+Space
- Also supports Tab to accept, Escape to dismiss
- Four different configurable completion modes (word, sentence, paragraph, smart)
- Prioritizes fast, low-cost models for responsive completions:
`claude-haiku-4-5` → `gemini-2.5-flash` → `gpt-4.1-nano`

### Use of AI Tools

Can't speak to the original code but in the cleanup, used Cursor running Codex 5.3 to refactor the typescript and debug a few issues

## Testing Instructions

- [ ] Clone this PR and run `npm i && npm run build`
- [ ] Enable the Type Ahead experiment in AI settings
- [ ] Create a new post and start typing in a paragraph block
- [ ] Verify ghost text suggestions appear after a pause
- [ ] Press Tab to accept or Escape to dismiss suggestions
- [ ] Test different completion modes in experiment settings

## Screenshots

CleanShot 2025-12-19 at 11 33 57@2x

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-151-2d3add93735a2b460ec50691ebae1962985f233f.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:

jeffpaul:
- [ ] Reminder for us to add a screenshot gif and entry in the readme.txt for this feature before merging/releasing.

alansmodic:
Would something like this have the ability to adhere to the defined content guidelines should someone have that implemented?

jeffpaul:
> Would something like this have the ability to adhere to the defined content guidelines should someone have that implemented?

@alansmodic that would be a lovely iteration to this experiment, probably to several experiments perhaps, but in its current form this PR and experiment does not

dkotter:
@jeffpaul I have this PR refreshed now and working again. I haven't added any tests yet as I think it's worth pausing for some feedback first.

The biggest issue I'm seeing in my testing is that results take ~1 - 1.5 seconds to return. That's not overly slow but means you do have to pause for what feels like a decent amount of time before you see results. There's also no indication that anything is happening (no sort of loading indicator) so easy to just keep typing (which then cancels the request).

This may be fine, especially for something we're shipping as an experimental Feature, but figured it was worth someone else testing out before I spend too much more time on this one.

**Edit**: also worth noting I haven't spent any time tuning the system instructions here to try and get better results. I find it often doesn't capitalize the first letter of a new sentence or gets confused on if it should return a new sentence or try and continue the existing one. I _think_ all of that can be cleaned up with some tighter instructions but again, holding on too much work here until we determine if the user experience is nice enough to continue with.

jeffpaul:
Yeah, the letter capitalization ideally gets resolved as that happens in pretty much every instance.  Another issue is that this doesn't get triggered with an empty new block `Type / to choose a block` which you get after hitting enter within the Paragraph block and presumably things should continue to type-ahead there.

I'm not sure a loading indicator would be feasible, as that would almost have to appear in place of the blinking cursor and I think its best for a bit of a delay before ghostingwriting kicks in as its possible the author is just mentally pausing themselves and allowing them to keep their writing flow before kicking in with the type-ahead text seems fine?

dkotter:
> Yeah, the letter capitalization ideally gets resolved as that happens in pretty much every instance. Another issue is that this doesn't get triggered with an empty new block `Type / to choose a block` which you get after hitting enter within the Paragraph block and presumably things should continue to type-ahead there.

I can look into these issues / continue to try and refine the output if you're good with the overall user experience we have so far here.

> I'm not sure a loading indicator would be feasible, as that would almost have to appear in place of the blinking cursor and I think its best for a bit of a delay before ghostingwriting kicks in as its possible the author is just mentally pausing themselves and allowing them to keep their writing flow before kicking in with the type-ahead text seems fine?

Yeah, I don't think a loading indicator would work here, just was concerned that with how long it takes for the results to render it's not clear anything is happening. So likely that a good majority of the time, a user will type a bit, stop typing for a brief period, the request is triggered to get results and then before those results render, the user starts typing again. This doesn't cause any problems (the code already handles this) but does mean that request is wasted (but still cost them some money).

That said, the answer here could be that if a site turns this experiment on, they just need to realize this is what's going to happen. May be worth being more explicit about this in the experiment description and that may be "good" enough.

jeffpaul:
> That said, the answer here could be that if a site turns this experiment on, they just need to realize this is what's going to happen. May be worth being more explicit about this in the experiment description and that may be "good" enough.

Calling that out in the description definitely seems worthwhile, agreed.

dkotter:
Added E2E and unit tests now.

Also tweaked the system instructions a bit to try and get better results around capitalization and punctuation. Still not perfect there but seems much better in my testing.

jeffpaul (dismissed):
In testing this again, I could see where some sort of loading indicator might actually be helpful.  I had inadvertently started testing with only Ollama installed and the type ahead never got triggered, but when I added back OpenAI things worked as expected.  The problem now is there's no clue as to what's happening if the available Connector(s) don't properly work with the feature, but perhaps a pulsing ellipses or WP logo in place of the blinking cursor might be a worthwhile test?  I can't seem to decide whether we add something like that now or ship this experiment and iterate if there's community feedback to do so in the future?

jeffpaul:
Testing again via Playground, thinking we ignore the loading/cursor questions above and can funnel that off as a post-release issue to see what the community feedback is on the experiment and can adjust In the future.

Otherwise, I think we should hide the options for the feature and just have the enable/disable toggle only.  That would mean another top right elipses menu option for "Show/hide advanced options" sort of an option that would display these Type-ahead Text options and any other Features or Experiments that have similar advanced options (like Content Classification).

dkotter:
> Otherwise, I think we should hide the options for the feature and just have the enable/disable toggle only. That would mean another top right elipses menu option for "Show/hide advanced options" sort of an option that would display these Type-ahead Text options and any other Features or Experiments that have similar advanced options (like Content Classification).

Yeah, been on my mind on how we introduce more advanced settings while still keeping the interface clean. I think likely worth a larger conversation and not something I'd try and wrap into this PR personally.

That said, looking at the settings we're adding here for Type Ahead, I'd suggest just removing those all together I think. These were added as part of the initial PR work and I just left them in place but these are all fairly advanced settings and not sure they provide much value here.

@jeffpaul Any concerns if I just remove those settings all together and just have decent defaults (maybe with filters that can be used to change those?)

jeffpaul:
@dkotter yes let's go ahead and remove those settings here, as I prefer to keep options in the AI settings as simple as possible (for now).  I'll open a separate issue for the loading animation/cursor and the advanced settings topics so we can sort out approach on both separate from landing this Experiment.

dkotter:
> @dkotter yes let's go ahead and remove those settings here, as I prefer to keep options in the AI settings as simple as possible (for now). I'll open a separate issue for the loading animation/cursor and the advanced settings topics so we can sort out approach on both separate from landing this Experiment.

Okay, settings removed and merge conflicts resolved so I think this is good to go

jeffpaul (dismissed):
Works well in testing via Playground, thanks!!!

PR #155, opened 2025-12-19
Add Comment Moderation experiment

Jameswlepage:
## What?

Adds comment toxicity and sentiment analysis and uses that to moderate comments.

## Why?

There are multiple approaches to moderating comments and using AI to do that is a natural fit. For sites that get lots of comments, having a first-line moderation strategy is key to keep up with them.

## How?

- Adds a new Comment Analysis Ability that will analyze a comment for toxicity and sentiment
  - **Toxicity scoring**: Analyzes comments for harmful content and scores Low/Medium/High
  - **Sentiment analysis**: Detects positive, negative, or neutral sentiment
- Add a new Comment Moderation Experiment that uses this Ability
- When a new comment is left, automatically pass the content of the comment into our analysis Ability. If the sentiment is negative and the toxicity score is above 0.7, put the comment into moderation
- On the comment list screen in the admin, add new columns that output the comment toxicity and sentiment values
- On this same screen, add a new bulk dropdown option to analyze one or more comments. This won't moderate, will just run and save the analysis
  - **Lazy analysis**: Automatically processes pending comments when the Comments page loads
- Add all tests and documentation

### Use of AI Tools

Can't speak to the original code in this PR but in the cleanup, used Claude running Opus 4.6 and also Cursor running Codex 5.3 to do code reviews, some cleanup, tests and documentation

## Testing Instructions

- [ ] Checkout this PR and run `npm i && npm run build`
- [ ] Enable the Comment Moderation experiment under `Settings → AI`
- [ ] Ensure you have a valid AI connector in place
- [ ] Create a new post and add a positive comment to that post
- [ ] Visit `Comments` in the admin and verify badges appear
- [ ] Verify there's a bulk action "Analyze with AI" and use that
- [ ] Add a new comment with some negative language (`I hate this post`). Ensure this comment goes to moderation

## Screenshots

Comment screen in the WordPress admin showing comment analysis results

Moderated comment on the front-end

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-155-df9f42799f54699a327525f4bc5d5a713e8bc864.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
_Note we'll want to update the screenshot numbers when this gets merged in, I left it with a hash instead of a number depending on when this versus other experiments get merged in._

jeffpaul:

jeffpaul:
Testing showing all comments when run thru bulk action are getting `Failed` response:

Screenshot 2026-02-02 at 12 28 41 PM

sethrubenstein:
This actually had us discussing the possibility of bringing back comments this morning. In an era where traffic is either declining or stagnant, keeping our audience engaged on-site is more important than ever. I apologize for not having had much time to test these features alongside our app or dive into the code too much. However, if this could eventually be included as part of the core efforts or if this could be made extensible enough to add—what we might call "comment value"—it would be very beneficial. We all have less staff (look at WaPo this week 😥), so it's not just replying to comments that saves labor but also quickly figuring out if its even worth replying or displaying that comment for that matter. An important metric to consider is whether a comment contributes to the discussion or reflects the "zeitgeist" of the post. For example, comments like "Yeah, I agree" are not particularly valuable. Ideally, we could use AI to filter out low-value comments.

jeffpaul:
> An important metric to consider is whether a comment contributes to the discussion or reflects the "zeitgeist" of the post.

@sethrubenstein perhaps a `Relevance` column that identifies high/medium/low sort of rating of how much that comment relates to the post content?  That does somewhat feel a bit more like an Akismet-spam-flagging sort of thing, so I'm trying to sort out how useful this additional meta would be on comments.

> For example, comments like "Yeah, I agree" are not particularly valuable. Ideally, we could use AI to filter out low-value comments.

Perhaps another column for `Value` that attempts to identify high/medium/low sort of rating on if the comment contributes to the discussion around the post content?

If you have more specific approaches in mind please share, as otherwise the functionality here will likely be stabilized and into a release without much variance from this experience.

sethrubenstein:
@jeffpaul good point, we don't want to effectively just recreate anti-spam plugins, so maybe both value and relevance would be too much. But just a value column perhaps analyzed against these criteria would work out well:

To add to system instructions
```
3. "value": One of exactly these three values, ranking how much the comment contributes to meaningful discussion and is likely to drive further engagement:
   - "high": The comment introduces a new perspective, asks a thought-provoking question, provides evidence or data, shares a personal experience that adds context, or makes a substantive argument that others would want to respond to
   - "medium": The comment adds some relevant input such as a brief opinion, a follow-up question, or partial agreement/disagreement, but is unlikely to spark significant further discussion
   - "low": The comment adds little to the discussion — one-word reactions, off-topic remarks, generic praise or complaints, or repetition of what others have already said
```

Thoughts? I'm happy to pull this down and give it a try myself.

sethrubenstein:
On thinking about this a bit further, the more challenging aspect is analyzing this against existing comments for similarities in nature and sentiment. This level of filtering likely yields a more advanced anti-spam and comment solution and not what you want this feature to be.

jeffpaul:
Some thoughts from testing...

1. If the experiment is enabled, I think we should immediately process the sentiment + toxicity for newly submitted comments versus requiring someone to use the "analyze with ai" bulk action to get those results.
2. let's add a row action in the Comments table list view for "ai analyze" alongside the newly added "ai reply" so folks can process historical comments one at a time (versus having to use the bulk actions for a single comment item)
3. not worth doing in this PR, but I think some thought on sending certain sentiment/toxicity scores (or score combinations) into comment moderation would be worthwhile (starting at least with sentiment=negative, toxicity=high)
4. the AI Reply Suggestions modal can probably be condensed a bit so it doesn't require a scrollbar to see the 3 recommended replies (perhaps reduce the textarea height for each reply by 50% even if that requires them to need to be scrolled or dragged to expand, move the tone dropdown to be vertically next to the tone label, change Regenerate button to be the regenerate icon instead)

Screenshot 2026-02-19 at 1 11 19 PM Screenshot 2026-02-19 at 1 11 55 PM


5. update the experiment description on the settings page to "Analyzes sentiment and toxicity in comments, alongside reply suggestions"
6. not worth doing in this PR, but could be worth a future iteration to add sentiment+toxicity filtering in the Comments table list view

jeffpaul:
@sethrubenstein back to our conversation here, it occurs to be that what might be most helpful in addition to the sentiment+toxicity bits is a recommendation to editors on WHAT to do with a specific comment.  Should they reply?  Should they mark as spam or trash it?  Should they just ignore it?  Perhaps scope of actions to take are: Reply | Approve | Ignore | Review | Spam | Trash?  Maybe also consider Feature (as in "mark as top comment") or Escalate (as in "send for legal or moderator review") thought these feel fine for future extension/iteration options.  Maybe worth exposing is a confidence percent for how strong the recommendation is?  Depending on what the recommendation is, that specific cell in the Comments table list view could actually be a clickable button/link to take action on the recommendation?  Perhaps the column is named "Recommendation"?  With all this in mind, perhaps we rename the experiment from Comment Moderation to Comment Intelligence or Comment Assistance?

dkotter:
@jeffpaul Was working on getting this PR up-to-date and realized even though we're calling this "Comment Moderation", it actually doesn't do any moderation. Instead it analyzes comments for sentiment and toxicity score, storing those results but doesn't use those results for anything. It also only runs on comments you manually choose. In addition, this PR bundles an AI comment replies Ability, which while interesting, feels like it should be it's own experiment.

Here's what I'd propose:

- Add in actual comment moderation and have this happen when a comment is left, not when manually triggered
  - We can likely use the same sentiment analysis and toxicity scoring for this
  - Flow here is when someone leaves a comment, pass it through our analysis and if it doesn't pass (will need to determine the pass/fail criteria), automatically put that comment into moderation
- Remove the AI comment reply functionality
- Keep the ability to manually trigger analysis from the comments list screen

Any concerns with any of this before I continue?

jeffpaul:
All sounds good to me.

> Flow here is when someone leaves a comment, pass it through our analysis and if it doesn't pass (will need to determine the pass/fail criteria), automatically put that comment into moderation

Perhaps anything that's flagged with Negative sentiment AND High toxicity goes to moderation?  Make that filterable so someone could change to perhaps Negative sentiment OR High toxicity or whatever combinations they'd prefer?

dkotter:
@jeffpaul This is ready for review now. I've read through all the comments on the PR and I think I've addressed all actionable items. There is some good back and forth with you and Seth that I'd suggest we pull that out into a new issue just so we don't lose track of it, as I think there's some valuable things we can look at as enhancements to this or even entirely new experiments.

jeffpaul (commented):
1. Update the bulk action to `Analyze Sentiment and Toxicity` (while I agree `Analyze with AI` is MUCH shorter I'm trying to get away from using `AI` in the editor and admin and more describe things as "this is what happens when you click this".

2. Is there sufficient hooks in place such that the "comment value" / Relevance concept that @sethrubenstein mentions in https://github.com/WordPress/ai/pull/155#issuecomment-3861839398 and https://github.com/WordPress/ai/pull/155#issuecomment-3893237376 could reasonably be extended from this work (and similarly have low value comments based on their algorithm/approach go to moderation)?

dkotter:
> 1. Update the bulk action to `Analyze Sentiment and Toxicity` (while I agree `Analyze with AI` is MUCH shorter I'm trying to get away from using `AI` in the editor and admin and more describe things as "this is what happens when you click this".

This has been updated now

> 2. Is there sufficient hooks in place such that the "comment value" / Relevance concept that @sethrubenstein mentions in [#155 (comment)](https://github.com/WordPress/ai/pull/155#issuecomment-3861839398) and [#155 (comment)](https://github.com/WordPress/ai/pull/155#issuecomment-3893237376) could reasonably be extended from this work (and similarly have low value comments based on their algorithm/approach go to moderation)?

So yes and no. There's a filter around the system instructions and another around the response schema so in theory, you could use both of those to change the response the LLM gives. There's also then Core filters on the comment list screen that you could use to add a new column for "Value/Relevance".

But the manual processing of comments is done via custom javascript which doesn't have hooks in place so you'd have to add some of your own JS to replicate the full experience.

That said, I'd love to add a comment value / relevance experiment to this plugin though I think that's fine as a separate PR (also some discussion on if that should be part of comment moderation or a separate experiment).

jeffpaul (approved):

PR #159, opened 2025-12-22
Change experiment initialization method from register() to init()

### LINKED ISSUE #145: Rename experiment register() method to better reflect initialization

As I understand it, [default experiments](https://github.com/WordPress/ai/blob/cf1c8f7ca5147da1589f0129490550b5d3799e6e/includes/Experiment_Loader.php#L105) are registered (an instance of each experiment) in an `Experiment_Registry` [here](https://github.com/WordPress/ai/blob/cf1c8f7ca5147da1589f0129490550b5d3799e6e/includes/Experiment_Loader.php#L63) (the experiments that have been registered can be inspected via the `_ai_experiments_register_experiments` hook)

However, just registering an experiment does not launch it. Experiments need to be "initialized", which happens [here](https://github.com/WordPress/ai/blob/cf1c8f7ca5147da1589f0129490550b5d3799e6e/includes/Experiment_Loader.php#L181). After checking if the experiment is enabled, the experiment is actually executed [by calling the `register` method of each experiment](https://github.com/WordPress/ai/blob/cf1c8f7ca5147da1589f0129490550b5d3799e6e/includes/Experiment_Loader.php#L207C4-L207C28).

This is confusing because:

- There is a **registration phase** (adding experiments to the registry, discoverability, etc.).
- There is an **initialization/launch phase** (running the experiment’s logic once it’s enabled).

Using the name `register` for the method that initializes/launches the experiment blurs the distinction between these two phases and makes it harder to quickly understand the lifecycle of an experiment when reading the code.

Rename the method that actually starts/initializes an experiment from `register()` to a name that more clearly represents its role in the experiment lifecycle, such as:

- `init()`
- `initialize()`
- `launch()`
- or similar terminology that conveys “run/boot the experiment”.

This would:

- Make the lifecycle clearer:  
  - **Register** experiment → added to the `Experiment_Registry`.  
  - **Initialize/launch** experiment → logic is executed when enabled.
- Reduce confusion for contributors and integrators trying to understand when an experiment is merely known to the system versus when it is actively running.
- Improve readability and self-documentation of the code around the `Experiment_Loader` and experiment classes.

juanmaguitar:
## What?
Refactors the experiment lifecycle method from `register()` to `init()` across the entire codebase to better reflect its purpose of initializing experiments.

Closes https://github.com/WordPress/ai/issues/145

## Why?
The term "register" was misleading as it suggested the experiments were being registered into a system, when in reality this method is called **after** registration to initialize the experiment's hooks and functionality. The method `init()` more accurately describes what the method does - it initializes the experiment by setting up WordPress hooks, filters, and actions when the experiment is enabled and loaded.

This improves code clarity and aligns the method naming with its actual behavior in the experiment lifecycle:
1. Experiments are **registered** to the registry via `Experiment_Registry::register()`
2. Experiments are **initialized** via `Experiment::init()` when they're loaded

## How?
- Updated the `Experiment` interface to declare `init()` instead of `register()` 
- Updated the `Abstract_Experiment` abstract class to declare `abstract public function init()` 
- Updated all concrete experiment implementations:
  - Example_Experiment
  - Excerpt_Generation
  - Image_Generation
  - Title_Generation
- Updated `Experiment_Loader::initialize_experiments()` to call `init()` 
- Updated all test mock classes and test assertions
- Updated developer documentation and inline comments

## Testing Instructions
1. Enable the AI Experiments plugin
2. Go to Settings → AI Experiments
3. Enable one or more experiments
4. Verify the experiments load correctly and their functionality works as expected
5. Check that no PHP errors or warnings appear in the debug log
6. Run the full test suite: `npm run test:php`
7. Verify all tests pass

Not applicable - this is an internal refactoring with no UI changes.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-159-50d96f7e95d91c9ea3c73ed07a4128cc3a1d70bc.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


juanmaguitar:
"Plugin Check" workflow is failing due to an issue reported at https://github.com/WordPress/ai/issues/160

jeffpaul:
#160 has been merged in, so hopefully an update from `develop` will get the plugin check passing @juanmaguitar

docs/DEVELOPER_GUIDE.md
```diff
@@ -144,12 +144,12 @@ class My_Experiment extends Abstract_Experiment {
...
 	/**
-	 * Registers the experiment's hooks and functionality.
+	 * Initializes the experiment's hooks and functionality.
 	 *
 	 * @since 0.1.0
 	 */
-	public function register(): void {
-		// Register your hooks here
+	public function init(): void {
+		// Initialize your hooks here
```
JasonTheAdams, line 152:
```suggestion
		// Initializes your experiment. Use this to set up your needed hooks.
```

JasonTheAdams (approved):
Minor wording suggestion, but otherwise looks good!

justlevine:
I very much disagree with the proposed change, and the premise that `. The method init() more accurately describes what the method does`

Registration != instantiation, and `add_action()` / `add_filter()` are very much _registration_ methods and not instantiation. It's only later in the lifecycle when those hooks are actually called that the bulk of the time we want any initialization to happen, and very rarely when our hook callbacks are registered onto the WordPress lifecycle.

Moreover the conflation between the two leads to unnecessary headaches around lifecycle and testing, and we shouldnt be registering hooks in an `::init()` any more than we should be doing it in a `__contruct()` - and for the exact same reasons.

--- 

If we want to add an `Initializable` interface to pair with a `Registrable` interface as a way to allow for experiments to have an generic instantiation (instead of just letting it happen _inside_ the hooks), I would be much more amenable. But that seems even more unnecessary at this stage than this proposed rename. In the super rare situation that an Experiment needs to "::init()" functionality when our registry `$experiment->register()` - and not when WordPress actually loads / runs the experiment code - it can be an implementation detail of the experiment.

Alternatively, if you think `::register()` is confusing because it's the function that we call when we register it (? not sure if I understood this argument correctly... like wouldn't that be the exact type of function for a registry to call???), then as a middle ground I'd suggest `::register_hooks()`. That IMO might be unnecessary bikeshedding, but at least there's no future code quality implications...

justlevine (changes requested):
Marking as `Request changes` to ensure a conversation is had https://github.com/WordPress/ai/pull/159#issuecomment-3733140772 - obvs is the majority disagree I'm good with being overruled 🙇

juanmaguitar:
@justlevine Appreciate the feedback — I agree with your reasoning.

I’m in favor of using `register_hooks` instead of `init` to better separate the two types of "register" in the experiment lifecycle following WordPress standards:

* `$registry->register_experiment()` — adds the experiment to the internal registry.
* `$experiment->register_hooks()` — attaches the WordPress hooks and effectively “starts” the experiment.

I’d love to hear more thoughts before moving ahead with any changes on this matter.

jeffpaul:
Moving this out to the `Future Release` milestone to allow the conversation here to get to an agreement on approach.

jeffpaul:
@neillmcshea please add this to an upcoming AI call to get to a decision on path forward (or not by closing it out) on this PR

dkotter:
While I understand the underlying idea here, since we haven't been able to come to a conclusion on the right approach here and this PR has sat for months at this point, going to close this out for now. Happy to re-review though if there's interest in refreshing this PR

PR #167, opened 2026-01-09
Fix new Plugin Check errors

dkotter:
## What?

Addresses new Plugin Check errors

## Why?

The Plugin Check plugin was recently updated and there is now a more aggressive check around direct file access. This has caused new errors to start showing in our Plugin Check workflow so this PR addresses those.

## How?

Adds the following to files that are being flagged:

```php
if ( ! defined( 'ABSPATH' ) ) {
	exit;
}
```

## Testing Instructions

Ensure the Plugin Check workflow passes on this PR

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-167-8d17f75fb8a372432f72caf3b4832ee7c80eeeaf.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -12,6 +12,10 @@ namespace WordPress\AI;
 use Throwable;
 use WordPress\AI_Client\AI_Client;
 
+if ( ! defined( 'ABSPATH' ) ) {
+	return;
+}
+
 /**
  * Normalizes the content by cleaning it and removing unwanted HTML tags.
  *
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -12,6 +12,14 @@ namespace WordPress\AI;
 use Throwable;
 use WordPress\AI_Client\AI_Client;
 
+/**
+ * Purposely using return instead of exit here.
+ *
+ * This file is loaded via the composer files directive.
+ * When tools like PHPCS and PHPStan run, they include
+ * our composer autoloader and that will then load this file,
+ * causing the script to exit and not function properly.
+ */
 if ( ! defined( 'ABSPATH' ) ) {
 	return;
 }
--- a/includes/Abilities/Utilities/Posts.php
+++ b/includes/Abilities/Utilities/Posts.php
@@ -11,6 +11,10 @@ namespace WordPress\AI\Abilities\Utilities;
 
 use WP_Error;
 
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
 /**
  * Post utility WordPress Abilities.
  *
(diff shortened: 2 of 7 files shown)


includes/helpers.php
```diff
@@ -12,6 +12,10 @@
 use Throwable;
 use WordPress\AI_Client\AI_Client;
 
+if ( ! defined( 'ABSPATH' ) ) {
+	return;
+}
```
dkotter, line 25: Note I purposely use `return` here instead of `exit` as this file is included directly via composer and if you use `exit`, it will break PHPCS from working properly (as PHPCS will automatically load your composer autoloader and since this file is directly loaded in that, the `exit` statement terminates the process)

justlevine: In general I'd suggest this kind of comment would be on the file, but why not just use PSR-4 for this file?

dkotter: My understanding is PRS-4 only works on classes, not for a file that contains a bunch of functions. Please correct me if I'm misunderstanding that though.

I guess we could refactor this `helpers.php` file to be a class with static methods if that feels like a better approach

justlevine: > if that feels like a better approach

IMO utils as a concept are an antipattern regardless of if they're a class or globally namespaced, so I might not be the best person to ask. But IIRC a decent amount of PHPStan functionality only works on PSR-4 (class/traits/interfaces). 

From a QL/patterns perspective, I'd assume without opening the file that everything in `includes` is PSR-4, so `includes/{H}elpers.php` contains a `class Helpers {}`. So if there's a reason not to have the file PSR-4, I'd recommend moving it root-level alongside the plugin entry point.

(Not necessarily as part of this PR, but then a doc-block on the `return;` would be smart imo)

dkotter: For now added a comment: https://github.com/WordPress/ai/pull/167/changes/8d17f75fb8a372432f72caf3b4832ee7c80eeeaf. Can look to adjust this to be a better approach in a follow up PR

justlevine (approved):
LGTM 🚀

(Personally, I'm not bothered by the codecoverage drops, but if someone else cares we can wrap them in a `@codeCoverageIgnore*` like wordpress/performance: https://github.com/WordPress/performance/blob/b4f989824a39e361c591f94d2965abdd331b58f3/plugins/optimization-detective/class-od-tag-visitor-context.php#L9-L13 )

PR #173, opened 2026-01-12
 Removed return so experiments always show …

### LINKED ISSUE #170: UX Bug: Experiments shouldn't require credentials
### Description

Noticed while picking up #63 that something changed in the current implementation to require credentials before doing anything else in the app.

The number of possible reasons why someone would want to see, access, or eventually even activate Experiments without first providing credentials makes this a ux bug that needs to be resolved.

If individual experiments need to be gated to credentials, then _they_ should be gated, not the entire plugin.

1. Install + activate the plugin
2. Visit the AI Experiments screen.

Image

### Environment info

`develop` + `wp-env`

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

pbearne:
... Update messaging: replace "enable experiments" with "run experiments" for improved clarity

## What?

Closes https://github.com/WordPress/ai/issues/170

## Testing Instructions
Install + activate the plugin
Visit the AI Experiments screen.
You should see the Experiments and admin message to promote to add keys 

image

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-173-a876ab6ff6c575978936137ae68ef027d40a745d.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Settings/Settings_Page.php
```diff
@@ -134,19 +134,18 @@ public function render_page(): void {
 				if ( ! has_ai_credentials() ) {
 					$error_message = sprintf(
 						/* translators: 1: Link to the AI credentials settings page. */
-						__( 'Before you can enable experiments, you need to ensure you have one or more AI credentials set <a href="%s">here</a>.', 'ai' ),
+						__( 'Before you can run experiments, you need to ensure you have one or more AI credentials set <a href="%s">here</a>.', 'ai' ),
```
dkotter: While this message is valid right now, we will be adding experiments in the near future that don't need AI credentials set at all (which I think is the crux of why the issue was reported). In that case, this message doesn't make sense. I'd suggest we update this to something like:

 
```suggestion
						__( 'Most experiments require valid AI credentials to function properly. To ensure those work properly, you need to have one or more AI credentials set <a href="%s">here</a>.', 'ai' ),
```

includes/Settings/Settings_Page.php
```diff
@@ -134,19 +134,18 @@ public function render_page(): void {
...
 						admin_url( 'options-general.php?page=wp-ai-client' )
 					);
 				} else {
 					$error_message = sprintf(
 						/* translators: 1: Link to the AI credentials settings page. */
-						__( 'Before you can enable experiments, you need to ensure you have set valid AI credentials <a href="%s">here</a>.', 'ai' ),
+						__( 'Before you can run experiments, you need to ensure you have set valid AI credentials <a href="%s">here</a>.', 'ai' ),
```
dkotter:
```suggestion
						__( 'Most experiments require valid AI credentials to function properly. Please <a href="%s">review</a> the AI credentials you have set to ensure they are valid.', 'ai' ),
```

dkotter (changes requested):
Left a couple comments on wording suggestions.

Beyond that, this works but is still probably not the best UX, though probably fine as a quick solution. I think long term it would be best to have a more guided approach to getting credentials set up when needed, instead of requiring users to go to a separate screen to do that

tests/e2e/specs/admin/settings.spec.js
```diff
@@ -38,7 +38,7 @@ test.describe( 'Plugin settings', () => {
 		await expect(
 			page.locator( '.wrap .notice-error', {
 				hasText:
-					'Before you can enable experiments, you need to ensure you have one or more AI credentials set',
+					'Before you can run experiments, you need to ensure you have one or more AI credentials set',
```
dkotter:
```suggestion
					'Most experiments require valid AI credentials to function properly. To ensure those work properly, you need to have one or more AI credentials set',
```

tests/e2e/specs/admin/settings.spec.js
```diff
@@ -82,7 +82,7 @@ test.describe( 'Plugin settings', () => {
 		await expect(
 			page.locator( '.wrap .notice-error', {
 				hasText:
-					'Before you can enable experiments, you need to ensure you have set valid AI credentials',
+					'Before you can run experiments, you need to ensure you have set valid AI credentials',
```
dkotter:
```suggestion
					'Most experiments require valid AI credentials to function properly',
```

dkotter (approved):

PR #175, opened 2026-01-12
Handle missing or invalid asset files gracefully in `Asset_Loader`.

### LINKED ISSUE #174: if the file for an asset is missing we get warning from filemtime()
### Description

Warning: filemtime(): stat failed for ...\wp-content\plugins\ai/build/admin/settings.css in ...\wp-content\plugins\ai\includes\Asset_Loader.php on line 96

Looking at the code dependencies are not being set, so I will fix this and clean some code

1. goto /wp-admin/options-general.php?page=ai-experiments

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

pbearne:
## What?

Closes https://github.com/WordPress/ai/issues/174

tide up the enqueue_style() code

## Why?
tide up and fix warning

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-175-79b763e6912e234ba08f1ab7bb2ea6a57a0b8cb6.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter (changes requested):
A couple things here:

1. There's a handful of PHPCS things that need cleaned up
2. I'd argue the better approach here is an early check to see if the file exists. If not, there's no reason to run any of the code. I think that fixes the issue you're seeing and removes the need for most of the change here

pbearne:
> 2\. I'd argue the better approach here is an early check to see if the file exists. If not, there's no reason to run any of the code. I think that fixes the issue you're seeing and removes the need for most of the change here

I thought about that, but the function can use an asset.php file

dkotter:
> > 2. I'd argue the better approach here is an early check to see if the file exists. If not, there's no reason to run any of the code. I think that fixes the issue you're seeing and removes the need for most of the change here
> 
> I thought about that, but the function can use an asset.php file

Not sure I follow? It does pull the `asset.php` file from the same location as the asset itself and then it uses that for the dependency array and version. But the main point of these methods is to run `wp_enqueue_script|style` and if the file we're trying to enqueue doesn't actually exist, seems we'd be better of returning early before any processing

juanfra:
Thanks for the PR! In addition to the previous review, I believe we should also have the check and early bail for `enqueue_script()`.

dkotter (approved):

PR #143, opened 2025-12-10
Add the Excerpt Generation Experiment UI

dkotter:
## What?

Closes #11 

This is a follow-up to #96, which added in the Excerpt Generation Experiment and Ability. This adds in the actual admin UI to trigger the generation.

## Why?

Coming out of #96 you can manually trigger excerpt generation by hitting the proper API endpoint but there isn't an actual UI to run that. This PR adds in that UI, making it easy for editors to generate excerpts on-demand.

## How?

- Loads the proper assets when on the post edit screen
- Uses the `__experimentalPluginPostExcerpt` slotfill to add a generate button below the excerpt form (note this slotfill is marked as deprecated but as far as I can tell, there isn't another solution right now)
- Uses DOM manipulation to directly inject a generate button into the excerpt panel, as there isn't a slotfill or filter here to use
- Both buttons are tied to the same functionality, which will trigger a request to generate an excerpt and then store that excerpt when returned (or show an error message if an error happens)

## Testing Instructions

1. Checkout this PR and run `npm i && npm run build`
2. Ensure you have valid AI credentials set at `Settings > AI Credentials`
3. Ensure you turn on the Excerpt Generation Experiment at `Settings > AI Experiments`
4. Edit a content type that supports excerpts
5. Within the excerpt panel, try generating an excerpt using both buttons and ensure it works

## Screenshots or screencast

Excerpt generation button showing below the excerpt input

Excerpt generation button showing next to the excerpt edit button in the editor panel

## Test using WordPress Playground

The changes in this pull request can be previewed and tested using this [WordPress Playground](https://wordpress.github.io/wordpress-playground/) instance:

[...]

PR #176, opened 2026-01-13
Fix new lint and config errors

dkotter:
## What?

Fixing some new linting and config errors coming out of recent merged PRs

## Why?

Want to ensure our GitHub Workflows remain in a passed state

## How?

Coming out of #172, I'm seeing the `artifacts` directory is no longer git-ignored. This directory is added by our E2E tests so we want to ensure this doesn't get committed to our repo. While it would be great if this directory could be nested within our `tests` directory instead of in the root, I'm running into issues with how `wp-scripts` runs these tests and having that change work properly. So for now I'm adding this directory back to our gitignore file.

In addition, seeing a new JS lint error and typescripts errors after the merging in of #143. I think this is also due to changes made in #172 and why these errors weren't flagged originally within #143.

There are some minor typescript errors fixed here and the JS lint error is flagging the use of an experimental API which unfortunately we need to use here as it's the only approach to modify the excerpt panel. So this PR ignores that particular error for now.

## Testing Instructions

Ensure all workflows pass on this PR

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-176-ac50afbc2e080655aeda0acd479d9f26cf2918ce.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ CLAUDE.md
 /tests/_output/*
 !/tests/_output/.gitkeep
 /tests/e2e/artifacts
+/artifacts
 
 # Configs
 /phpcs.xml
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "@types/react-dom": "^18.3.7",
         "@types/wordpress__block-editor": "^15.0.0",
         "@types/wordpress__blocks": "^12.5.18",
+        "@types/wordpress__edit-post": "^8.4.2",
         "@wordpress/e2e-test-utils-playwright": "^1.37.0",
         "@wordpress/env": "^10.37.0",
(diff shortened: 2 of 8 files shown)


dkotter:
@jeffpaul This isn't required for the 0.2.0 release but would be nice to get merged in sooner than later as we'll have a failing lint check on all PRs until this is merged in, just FYI

PR #194, opened 2026-01-19
Add Markdown Feeds experiment

Jameswlepage:
## Summary
Adds Markdown representations of WordPress content for feeds and singular views:

- Feed at `/?feed=markdown` with autodiscovery links
- Singular Markdown via `.md` permalinks and `Accept: text/markdown`
- HTML to Markdown conversion via `WP_HTML_Processor` with fallback to `WP_HTML_Tag_Processor`
- Section-level filters to reorder/extend output (e.g., inject custom fields)
- Integration tests for feed and converter behavior

## Implementation Details
- `Markdown_Feeds` registers the feed, `.md` parsing, and Accept header negotiation
- `HTML_To_Markdown_Converter` uses the WordPress HTML API to produce Markdown
- Settings toggles for feed, `.md` permalinks, and Accept header negotiation
- New filters: `ai_experiments_markdown_feed_post_sections` and `ai_experiments_markdown_singular_post_sections`

## Test plan
- [ ] Enable experiment under Settings -> AI Experiments
- [ ] Visit `/?feed=markdown` and confirm `text/markdown` output
- [ ] Visit `/hello-world.md` and confirm `text/markdown` output
- [ ] Request a post with `Accept: text/markdown` and confirm Markdown output
- [ ] Optional: use the new section filters to inject custom fields into output
- [ ] `npm run test:php -- --filter Markdown_Feeds`

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-194-3504700a6cec86dce5280683a58baf405cef3099.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


Jameswlepage:
Marking this as an open PR so we can get this merged. @m and others are quite interested in this feature

m:
We have some pretty URL prior art here with /feed/[format] that we should support in addition to the query string.

This would be the first time we introduce dots into WP's canonical URL structures... how should we think about that? Remember that we have other things that exist in the hierarchy of the permalink itself, so when you make the permalink a filename you lose that ability to have sub-directories or files.

jeffpaul:
@Jameswlepage before we move for review, I think Matt raised some questions that we should settle out before proceeding.  Did you have immediate reactions to those questions (especially in case you've had conversations on this topic that would inform approach here)?  If not, then I can look to sort through those with Darin to get this PR back on track towards review.

Jameswlepage:
Ah shoot, I didn't actually push up my changes. Yeah, I agree with Matt on the file extension. I'm seeing it as a common convention, so my inclination is to keep it but default to a format query parameter instead of the extension by default. I am not on my desktop right now, so I'll see if I can get that up soon.

dd32:
On the `.md` permalink, it's also worth noting that WordPress trailing-slash's things by default, so `/post-name/` is the URL, not `/post-name` so the `.md` technically exists at the higher level in the directory hierarchy.  
The `.md` suffix does indeed seem to be becoming more common, but from what I've seen it's more commonly where they already used `/post-name` or `/post-name.html` type endpoints.

includes/Experiments/Markdown_Feeds/Markdown_Feeds.php
```diff
@@ -0,0 +1,571 @@
...
+			return $query_vars;
+		}
+
+		$method = $this->get_request_method();
+		if ( 'get' !== $method && 'head' !== $method ) {
+			return $query_vars;
+		}
+
+		// Query vars that contain post/page slugs which may have .md suffix.
+		$slug_vars = array( 'name', 'pagename', 'attachment' );
```
dd32: This doesn't account for custom post type query variables, which aren't always using core fields.

Additionally, since this happens on the post-rewrites-applied, if a rewrite rule expects that only a slug would be present (which doesn't contain `.`) the matched rewrite rule may be different to a request without `.md`.
Additionally, if verbose page rules are in use (such as on some WordPress.org urls) this causes it not to match pages due to `PAGENAME.md` not being a valid page.  due to https://github.com/WordPress/WordPress/blob/f9b61add0e7d888b4dfd50d81baad53395c7b1bb/wp-includes/class-wp.php#L242-L250

includes/Experiments/Markdown_Feeds/Markdown_Feeds.php
```diff
@@ -0,0 +1,571 @@
...
+			}
+
+			// Strip the .md suffix and mark this as a markdown request.
+			$query_vars[ $var ]               = substr( $value, 0, -3 );
+			$this->markdown_extension_request = true;
+			break;
+		}
+
+		return $query_vars;
+	}
```
dd32:
```suggestion
        public function filter_request_for_markdown_extension( bool $true ): bool {
                $settings = $this->get_settings();
                if ( ! $settings['enable_md_extension'] ) {
                        return $true;
                }

                if ( is_admin() ) {
                        return $true;
                }

                $method = $this->get_request_method();
                if ( 'get' !== $method && 'head' !== $method ) {
                        return $true;
                }


                // See if the url looks like a .md.
                if ( str_contains( $_SERVER['REQUEST_URI'], '.md' ) ) {
                        $_SERVER['REQUEST_URI'] = str_replace( '.md', '', $_SERVER['REQUEST_URI'] );
                        $this->markdown_extension_request = true;
                }

                return $true;
        }
```

Hooking this to `do_parse_query` and then changing the filter to this, which mangles the requested url prior to `class WP::parse_request()` causes `/PAGE.md` to work for me on a few places where the existing code doesn't. Kinda resolved https://github.com/WordPress/ai/pull/194/changes#r2780329197 but obviously this is pretty ugly compared to query filtering, although WordPress simply doesn't have a granular enough filter IMHO.

includes/Experiments/Markdown_Feeds/Markdown_Feeds.php
```diff
@@ -0,0 +1,571 @@
...
+			return '';
+		}
+
+		$permalink = get_permalink( $post );
+		if ( ! $permalink || false !== strpos( $permalink, '?' ) ) {
+			return '';
+		}
+
+		// Remove trailing slash, add .md extension.
+		return rtrim( trailingslashit( $permalink ), '/' ) . '.md';
```
dd32: This doesn't work for root urls, treats the `.md` as a ccTLD.

> `<link rel="alternate" type="text/markdown" title="Home (Markdown)" href="https://example.org.md/" />`

Jameswlepage:
Pushed a commit addressing the review feedback:

- **Switched `.md` extension hook** from `request` to `do_parse_request` — strips `.md` from `$_SERVER['REQUEST_URI']` before `WP::parse_request()` runs, fixing custom post types, verbose page rules, and rewrite-rule mismatches (@dd32's suggestion)
- **Added `?format=md` query parameter** as the primary URL strategy, with its own settings toggle — works on any page (singular, archives, homepage)
- **Fixed root URL bug** — `get_markdown_permalink()` now guards against the homepage producing `example.org.md`
- **Defaulted `.md` extension to off** (opt-in) per @m's feedback on URL structure
- **Updated autodiscovery links** to use `?format=md`
- Settings now has 4 toggles: feed endpoint, `?format=md`, `.md` extension, and `Accept` header negotiation

Jameswlepage:
Noticed @dmsnell's [html-to-md](https://github.com/dmsnell/html-to-md) library: it uses a proper block/inline tree model with the HTML API and handles things like syntax escaping, line wrapping, hidden element skipping, and format nesting more robustly than the converter here. We should probably swap it in at some point rather than maintaining a separate one. We would need to extend it to better support GFM.

includes/Experiments/Markdown_Feeds/Markdown_Feeds.php
```diff
@@ -0,0 +1,667 @@
...
+use function get_the_title;
+use function is_admin;
+use function is_feed;
+use function is_singular;
+use function post_password_required;
+use function register_setting;
+use function sanitize_key;
+use function str_contains;
+use function wp_kses;
+use function wp_parse_url;
```
dkotter, line 43: Curious on this approach of `use function` for WordPress core functions? Not saying it's wrong but not something I've seen before and not something we're doing anywhere else in the codebase currently

Jameswlepage:
I'm going to close this branch as there is some movement in core WordPress around @dmsnell HTML to MD approach. That is a much better and more elegant solution. So I'd like to keep this as a reference and understand if that feature doesn't make it to WordPress 7.0 or Gutenberg, there is always an opportunity to contribute it as a experiment here as well.

PR #220, opened 2026-02-10
Update "Tested up to" to avoid compatibility warning

juanfra:
## What?

Updates the "Tested up to version" in the plugin header to the latest WordPress minor release so the plugin doesn’t show the compatibility warning in wp-admin. No functional changes.

Screenshot 2026-02-10 at 17 49 43

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-220-15d1b0c237d502a7decf48b7794a7379bc99d509.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,7 +1,7 @@
 === AI Experiments ===
 Contributors:      wordpressorg
 Tags:              ai, artificial intelligence, experiments, abilities, mcp
-Tested up to:      6.9
+Tested up to:      6.9.1
 Stable tag:        0.3.0
 License:           GPL-2.0-or-later
 License URI:       https://spdx.org/licenses/GPL-2.0-or-later.html


juanfra:
Just saw there's a failing check for the "Tested up to" version number to only include major versions. Closing.

dkotter:
According to the [plugin readme guidelines](https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#readme-header-information), the `Tested up to` field ignores minor versions so you only need to define the major version (in this case, 6.9).

I'm not seeing that warning message myself unless I'm running a site that is using a development version of WordPress (which then shows as 6.9.2) so I'm assuming you may be running the same.

juanfra:
Thanks for checking! My bad, I didn’t realize until after opening the PR.

PR #227, opened 2026-02-12
Standardize the Title Generation Ability to match our other Abilities

dkotter:
## What?

Modifies the input schema of the Title Generation Ability to match [how we've set up our other Abilities](https://github.com/WordPress/ai/pull/136#discussion_r2603941702).

> [!NOTE]  
> This is technically a backwards compat break as we're changing the schema from `post_id` to `context`.
>
> We don't have to make this change but it does bring code consistency as all other Abilities we have function in this way. And it also makes the Title Generation Ability a little more generic and easier to use in all situations.
>
> If we don't think that's worth it though, happy to close this out.

## Why?

Ensures we have code consistency and also makes the Title Generation Ability more generic and easier to use in other situations where you may not want to pass in a post ID.

## How?

- In the input schema of the Title Generation Ability, change the `post_id` field to `context` and allow either a post ID to be passed or specific context
- In our execution callback, check if the `context` field contains a post ID and if so, use that to get our context from that post
- When making the API request, pass the content and context (if we have some) in XML-like tags and remove the use of triple quotes to delimit content
- Update our system instructions to take into account the above changes
- Update our client-side API request to pass the right field

### Use of AI Tools

All code changes done by me. Cursor (using Composer 1.5) was used to update failing unit tests

## Testing Instructions

1. Add valid AI credentials
2. Turn on the Title Generation Experiment
3. Edit or create a post, adding in some content
4. Click into the title field and click the (Re-) Generate button
5. Ensure title suggestions are shown

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-227-7462330a582cbfe86d33a7ccf9b37a945a81b784.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


gziolo (dismissed):
Straightforward standardization — the code follows the same pattern already used by Excerpt_Generation, Summarization, and Generate_Image_Prompt. The XML-like prompt tags are also a nice improvement over triple-quote delimiters.

A few things to address before merging:

1. **Merge conflicts** in `Title_Generation.php` and `TitleToolbar.tsx` — needs a rebase against develop.
2. **Changelog** — Worth noting the `post_id` → `context` schema change as a breaking change for any external consumers of `ai/title-generation`.

PR #156, opened 2025-12-19
Add Alt Text Generation experiment

Jameswlepage:
CleanShot 2025-12-19 at 13 04 33@2x

Addresses #44

## Summary
Adds AI-powered alt text generation for images using vision models:

- **Block Editor integration**: Adds "Generate Alt Text" button to Image block's InspectorControls
- **Media Library integration**: Works in upload.php and media modals throughout WordPress
- **Flexible input**: Supports attachment IDs, URLs, and data URIs
- **Smart processing**: Converts images to data URIs for AI analysis, handles local and remote files
- **Quality output**: Generates concise, descriptive alt text (max 125 characters)

## Implementation Details
- Uses `ai/alt-text-generation` ability with vision models (Claude Haiku, GPT-5 Nano, Gemini Flash)
- Block editor script hooks into `editor.BlockEdit` filter
- Media script integrates with WordPress media frame attachment details
- Includes shared `run-ability.ts` utility for Abilities API with REST fallback

## Test plan
- [ ] Enable experiment under Settings → AI Experiments
- [ ] Open block editor, add an Image block, and click "Generate Alt Text"
- [ ] Visit Media Library (upload.php), select an image, and test alt text generation
- [ ] Test with images that have no alt text set
- [ ] Verify generated text is under 125 characters and descriptive

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-156-7de6575fc948130f20d94abbfe80852d896174da.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #228, opened 2026-02-12
Standardize to using `runAbility` to call our Abilities

dkotter:
## What?

Move from using `apiFetch` to using our custom `runAbility` utility function (introduced [here](https://github.com/WordPress/ai/pull/156/changes#diff-98e1d4466497537c9149137dca31aa8c87f9d0fa25c04d209344381b7555095dR107)) to call our Abilities in our UI code.

## Why?

Ensures our codebase is standardized (no more fractured code that uses `apiFetch` in some places and `runAbility` in others.

Also ensures the Abilities JS API is used when available, falling back to using `apiFetch` otherwise.

## How?

- Updates Title Generation, Excerpt Generation and Content Summarization Abilities to use `runAbility` and remove the direct use of `apiFetch`
- Removes the now unused `path` JS localized data

### Use of AI Tools

Used Cursor (running Composer 1.5) to give a first pass on making this change with final changes, plus review and testing, done by me

## Testing Instructions

1. Add valid AI Credentials
2. Turn on Title Generation, Excerpt Generation and Content Summary Experiments
3. Test each individual Experiment and ensure they still work

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-228-356401f2cb9dba978472d50236f25fbbadfd2ecc.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/experiments/excerpt-generation.md
+++ b/docs/experiments/excerpt-generation.md
@@ -40,7 +40,6 @@ The ability can be called directly via REST API, making it useful for automation
 1. **PHP Side:**
    - `enqueue_assets()` loads `experiments/excerpt-generation` (`src/experiments/excerpt-generation/index.tsx`) and localizes `window.aiExcerptGenerationData` with:
      - `enabled`: Whether the experiment is enabled
-     - `path`: The REST API path to the ability (`/wp-json/wp-abilities/v1/abilities/ai/excerpt-generation/run`)
 
 2. **React Side:**
    - The React entry point (`index.tsx`) registers a WordPress plugin that hooks into the excerpt panel using `__experimentalPluginPostExcerpt`
--- a/docs/experiments/summarization.md
+++ b/docs/experiments/summarization.md
@@ -42,7 +42,6 @@ The ability can be called directly via REST API, making it useful for automation
 1. **PHP Side:**
    - `enqueue_assets()` loads `experiments/summarization` (`src/experiments/summarization/index.tsx`) and localizes `window.aiSummarizationData` with:
      - `enabled`: Whether the experiment is enabled
-     - `path`: The REST API path to the ability (`/wp-json/wp-abilities/v1/abilities/ai/summarization/run`)
 
 2. **React Side:**
    - The React entry point (`index.tsx`) registers:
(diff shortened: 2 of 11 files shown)


gziolo:
I reviewed code changes, and they look good. I wanted to test with the latest changes on `trunk`, but it looks like the plugin is incompatible with WP trunk after the WP AI client landed there on Friday.

gziolo:
I tested against the latest stable WP version and there was an error when generating the title:

Screenshot 2026-02-23 at 10 07 30

I can confirm that generating the excerpt and summary works as expected.

dkotter:
> I wanted to test with the latest changes on `trunk`, but it looks like the plugin is incompatible with WP trunk after the WP AI client landed there on Friday.

Yeah, have a few PRs open tracking this (#242, #249), though I think there are changes coming upstream we'll need to pull in.

> I tested against the latest stable WP version and there was an error when generating the title:

I know there's an issue with OpenAI when running against the latest version of the WP AI Client, as we switch from the Chat Completions API to the Responses API and that no longer supports the `n` parameter, but that shouldn't impact you on the latest stable version of WP.

I just tested again on WP 6.9.1 and title generation works fine with both OpenAI and Google, I don't have an Anthropic key to test with though. Curious which AI provider you're using?

gziolo:
>  just tested again on WP 6.9.1 and title generation works fine with both OpenAI and Google, I don't have an Anthropic key to test with though. Curious which AI provider you're using?

I used Antrophic key, so that would have to be fixed independently 👍🏻

dkotter:
> I used Antrophic key, so that would have to be fixed independently 👍🏻 

Curious if you can test on the latest release of this plugin if it works? I would assume you'd get the same error as that shouldn't be something changed in this PR

gziolo:
> Curious if you can test on the latest release of this plugin if it works? I would assume you'd get the same error as that shouldn't be something changed in this PR

Yes, the issue isn't tied to this branch. I also see an error when testing with the AI experiments plugin installed on regular website:

> Error: Error generating titles: The response is not a valid JSON response.

gziolo (approved):

jeffpaul:
I'll open a follow-up issue shortly for the bug noted above.

jeffpaul:
Done, see: https://github.com/WordPress/ai/issues/264.

PR #242, opened 2026-02-20
Update to the latest version of WP AI Client

### LINKED ISSUE #241: Update to WP AI Client 0.3.0

[WP AI Client 0.3.0 is now available.](https://github.com/WordPress/wp-ai-client/releases/tag/0.3.0)

Let's look to update to that version, test that existing experiments work as expected, and identify if there are any new functionalities to take advantage of in the plugin.

We will eventually need to look at building off the AI Client in WP Core (and then setting the WP min to 7.0) but that change can wait for a bit longer (probably until RC1?).

dkotter:
## What?

Closes #241

Updates the `wp-ai-client` from 0.2.1 to 0.3.0. This also brings in the latest version of the `php-ai-client`, going from 0.3.1 to 1.1.0. The biggest change required here is AI providers are no longer bundled in `php-ai-client` so we also pull in those providers via composer.

## Why?

Ensures we're running the latest version of required dependencies.

## How?

- Bumped the version number for `wp-ai-client` in composer.json and ran `composer update wordpress/wp-ai-client`
- Needed to pull in the AI providers so ran `composer require wordpress/anthropic-ai-provider`, `composer require wordpress/google-ai-provider` and `compose require wordpress/openai-ai-provider`
- Added some code to ensure those providers were all registered prior to loading the WP AI Client which ensures the AI Credentials settings page renders settings for those providers. Also needed to initialize the HTTP transport (`WP_AI_Client_Discovery_Strategy::init()`) since that happens in the AI Client init but we need this before that
- Fixed a new PHPStan error around types

### Use of AI Tools

Majority of changes were done by me. Used Cursor running Composer 1.5 to fix some linting and test failures

## Testing Instructions

Ensure the AI Credentials page still shows settings for Anthropic, Google and OpenAI.

Test all Experiments to ensure they still work as expected

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-242-66cde4b81b811e1e760de941b3c89cd58067400f.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Title_Generation/Title_Generation.php
+++ b/includes/Abilities/Title_Generation/Title_Generation.php
@@ -141,15 +141,15 @@ class Title_Generation extends Abstract_Ability {
 		}
 
 		// Generate the titles.
-		$result = $this->generate_titles( $context, $args['candidates'] );
+		$results = $this->generate_titles( $context, $args['candidates'] );
 
 		// If we have an error, return it.
-		if ( is_wp_error( $result ) ) {
-			return $result;
+		if ( is_wp_error( $results ) ) {
+			return $results;
 		}
 
 		// If we have no results, return an error.
-		if ( empty( $result ) ) {
+		if ( empty( $results ) ) {
 			return new WP_Error(
 				'no_results',
 				esc_html__( 'No title suggestions were generated.', 'ai' )
@@ -162,7 +162,7 @@ class Title_Generation extends Abstract_Ability {
 				static function ( $title ) {
 					return sanitize_text_field( trim( $title, ' "\'' ) );
 				},
-				$result
+				$results
 			),
 		);
 	}
@@ -263,11 +263,22 @@ class Title_Generation extends Abstract_Ability {
 		}
 
 		// Generate the titles using the AI client.
-		return AI_Client::prompt_with_wp_error( '"""' . $context . '"""' )
-			->using_system_instruction( $this->get_system_instruction() )
-			->using_temperature( 0.7 )
-			->using_candidate_count( (int) $candidates )
-			->using_model_preference( ...get_preferred_models_for_text_generation() )
-			->generate_texts();
+		$results = array();
+		for ( $i = 0; $i < $candidates; $i++ ) {
+			$result = AI_Client::prompt_with_wp_error( '"""' . $context . '"""' )
+				->using_system_instruction( $this->get_system_instruction() )
+				->using_temperature( 0.5 )
+				->using_model_preference( ...get_preferred_models_for_text_generation() )
+				->generate_text();
+
+			// Return immediately if we have an error.
+			if ( is_wp_error( $result ) ) {
+				return $result;
+			}
+
+			$results[] = $result;
+		}
+
+		return $results;
 	}
 }
--- a/includes/Abilities/Title_Generation/Title_Generation.php
+++ b/includes/Abilities/Title_Generation/Title_Generation.php
@@ -11,7 +11,6 @@ namespace WordPress\AI\Abilities\Title_Generation;
 
 use WP_Error;
 use WordPress\AI\Abstracts\Abstract_Ability;
-use WordPress\AI_Client\AI_Client;
 
 use function WordPress\AI\get_post_context;
 use function WordPress\AI\get_preferred_models_for_text_generation;
@@ -265,7 +264,7 @@ class Title_Generation extends Abstract_Ability {
 		// Generate the titles using the AI client.
 		$results = array();
 		for ( $i = 0; $i < $candidates; $i++ ) {
-			$result = AI_Client::prompt_with_wp_error( '"""' . $context . '"""' )
+			$result = wp_ai_client_prompt( '"""' . $context . '"""' )
 				->using_system_instruction( $this->get_system_instruction() )
 				->using_temperature( 0.5 )
 				->using_model_preference( ...get_preferred_models_for_text_generation() )
--- a/includes/bootstrap.php
+++ b/includes/bootstrap.php
@@ -14,12 +14,6 @@ namespace WordPress\AI;
 use WordPress\AI\Abilities\Utilities\Posts;
 use WordPress\AI\Settings\Settings_Page;
 use WordPress\AI\Settings\Settings_Registration;
-use WordPress\AI_Client\AI_Client;
-use WordPress\AI_Client\HTTP\WP_AI_Client_Discovery_Strategy;
-use WordPress\AiClient\AiClient;
-use WordPress\AnthropicAiProvider\Provider\AnthropicProvider;
-use WordPress\GoogleAiProvider\Provider\GoogleProvider;
-use WordPress\OpenAiAiProvider\Provider\OpenAiProvider;
 
 // Exit if accessed directly.
 if ( ! defined( 'ABSPATH' ) ) {
@@ -207,8 +201,31 @@ function load(): void {
 	// Initialize our default AI providers.
 	add_action( 'init', __NAMESPACE__ . '\initialize_ai_providers', 5 );
 
+	// Initialize the WP AI Client if needed.
+	add_action( 'init', __NAMESPACE__ . '\initialize_ai_client', 10 );
+
 	// Hook experiment initialization to init.
-	add_action( 'init', __NAMESPACE__ . '\initialize_experiments' );
+	add_action( 'init', __NAMESPACE__ . '\initialize_experiments', 15 );
+}
+
+/**
+ * Initializes the default AI providers.
+ *
+ * @since x.x.x
+ */
+function initialize_ai_providers(): void {
+	$provider_loader = new Provider_Loader();
+	$provider_loader->init();
+}
+
+/**
+ * Initializes the WP AI Client if needed.
+ *
+ * @since x.x.x
+ */
+function initialize_ai_client(): void {
+	$client_loader = new Client_Loader();
+	$client_loader->init();
 }
 
 /**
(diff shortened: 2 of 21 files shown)


includes/Abilities/Title_Generation/Title_Generation.php
```diff
@@ -263,11 +263,22 @@ static function ( $key, $value ) {
...
+				->using_model_preference( ...get_preferred_models_for_text_generation() )
+				->generate_text();
+
+			// Return immediately if we have an error.
+			if ( is_wp_error( $result ) ) {
+				return $result;
+			}
+
+			$results[] = $result;
+		}
```
dkotter, line 279: The newest version of the PHP AI Client now uses the OpenAI Responses API instead of the Chat Completions API. That API no longer supports returning multiple results:

> Additionally, Chat Completions can return multiple parallel generations as `choices`, using the `n` param. In Responses, we’ve removed this param, leaving only one generation

I've updated here to run multiple requests to get multiple results which maintains backward compatibility. We can remove this if we want to just return a single title result.

includes/bootstrap.php
```diff
@@ -262,4 +270,34 @@ static function () {
...
+ *
+ * @since x.x.x
+ */
+function initialize_ai_providers(): void {
+	if ( ! class_exists( AiClient::class ) ) {
+		return;
+	}
+
+	// Ensure the HTTP transporter is initialized.
+	WP_AI_Client_Discovery_Strategy::init();
```
dkotter: @JasonTheAdams @felixarntz After updating to the latest version of the WP AI Client (which pulls in the latest PHP AI Client), have to pull in each provider and initialize those, which I'm doing here.

I ran into an issue though where the HTTP transport wasn't setup so requests wouldn't work. I've added this line here and it fixes it but curious if there's something I'm missing or a better approach? I know the `init` method of the `AI_Client` runs this same code but seems we need to initialize providers prior to initializing the client so duplicating that here

dkotter:
Currently getting a fatal error when run on WP Trunk:

> Fatal error: Declaration of WordPress\AI_Client\HTTP\WordPress_HTTP_Client::sendRequestWithOptions(Psr\Http\Message\RequestInterface $request, WordPress\AiClient\Providers\Http\DTO\RequestOptions $options): Psr\Http\Message\ResponseInterface must be compatible with WordPress\AiClient\Providers\Http\Contracts\ClientWithOptionsInterface::sendRequestWithOptions(WordPress\AiClientDependencies\Psr\Http\Message\RequestInterface $request, WordPress\AiClient\Providers\Http\DTO\RequestOptions $options): WordPress\AiClientDependencies\Psr\Http\Message\ResponseInterface in /var/www/html/wp-content/plugins/ai/vendor/wordpress/wp-ai-client/includes/HTTP/WordPress_HTTP_Client.php on line 97

I think this is likely due to conflicts between the version of the WP AI Client in trunk compared to the version we load in this plugin. This needs resolved prior to 7.0 being released but I don't think it needs to be resolved in this PR, but open to opinions on that.

includes/Abilities/Title_Generation/Title_Generation.php
jeffpaul: What's the rough difference on token count between getting 3 results from the chat completions versus 3 requests to responses?  If its noticeable, then I might be ok with reducing that down to a single result as they can always regenerate.

jeffpaul:
> I think this is likely due to conflicts between the version of the WP AI Client in trunk compared to the version we load in this plugin. This needs resolved prior to 7.0 being released but I don't think it needs to be resolved in this PR, but open to opinions on that.

Assuming that's coming from tests, we can probably ignore trunk results for now (I know that's risky in case something else is failing) but until we update to ONLY use the client in core we'll have issues.  And we know from the 0.1.0 launch that if we set 7.0 as the min to then focus on the ai client in core that we'll also have issues there, and I suspect keeping 6.9 as min and 7.0 as tested up to will also have things fail.  Sigh.

includes/Abilities/Title_Generation/Title_Generation.php
dkotter: Running a quick test, using Chat Completions to get three results used 1067 total tokens. Using this new approach, a single request used 1039 total tokens so 3117 total for all three requests. So not surprisingly, about a 3x increase in token usage though still such small numbers that unlikely to see any impact on actual cost.

That said, this approach forces this for all Provider/model combinations and there are certainly some that would support the old approach (Google for instance does, Anthropic does not). So I think three options we can discuss here:

1. Only return a single result irregardless of the model being used
2. If multiple results are wanted, always make multiple requests irregardless of what the model supports
3. If multiple results are wanted, check what the model supports and if it supports returning multiple candidates, keep things with the old approach. If it doesn't support multiple (like OpenAI or Anthropic) make multiple requests to get the number of results desired

dkotter:
> Assuming that's coming from tests, we can probably ignore trunk results for now

The same fatal error happens if you try running this branch on `trunk` right now so it's not just failing in tests. A fatal doesn't occur if running the latest release of this plugin on `trunk` though no provider settings show up so can't use things.

I think discussion around if we want to set 7.0 as our minimum and just remove all the WP AI Client code from this plugin or if we want to try and support 6.9 still. And then need to switch to using the new `wp_ai_client_prompt` function. So I'd say we should tackle those in a follow-up PR.

Jameswlepage:
> I think discussion around if we want to set 7.0 as our minimum and just remove all the WP AI Client code from this plugin or if we want to try and support 6.9 still.

I think that we should make 7.0 the minimum supported version for AI features in WP.

felixarntz:
> I think that we should make 7.0 the minimum supported version for AI features in WP.

+1

With the WP AI Client now merged into Core, the project will effectively be discontinued. Only PHP AI Client will be maintained as an external library, but even that WordPress plugins should not require via Composer because it's bundled in Core.

Jameswlepage:
@felixarntz what are your thoughts on the transitory period between now and then?

includes/Abilities/Title_Generation/Title_Generation.php
jeffpaul: For now I'm fine with option 1, especially if that's filterable as it keeps token usage low and someone can always mash the regenerate button for as many additional suggestions as theyd want

jeffpaul:
> I think that we should make 7.0 the minimum supported version for AI features in WP.

I'm fine with that change closer to / just after RC1, but until then it makes the plugin unusable on current major installs which I'm not super keen on.

includes/Abilities/Title_Generation/Title_Generation.php
dkotter: Right now the title generation Ability allows you to pass in a `candidates` parameter (limited between 1 and 10). This is used to determine how many titles to generate. In our case, we set this to 3 when we use this Ability, though others can also use this Ability and set that to something else.

Because of this, I think ideal here is to keep that parameter which also means we need to keep this code change but we can change our integration to only return 1 result to limit token usage

felixarntz:
I think this thread has what we should do: https://wordpress.slack.com/archives/C08TJ8BPULS/p1771606845472169?thread_ts=1771606845.472169&cid=C08TJ8BPULS

Release another version of `wp-ai-client` that shims the `wp_ai_client_prompt()` function for future compat of people coding against it.

Plus a check that, if the `wp_ai_client_prompt()` function already exists (i.e. 7.0 branch), we skip loading pretty much everything and rely on the Core implementation.

And then once WordPress 7.0 reaches RC, the AI plugin would switch over to the Core implementation entirely. It would need to ensure that sites that update would then be prompted to (re-)install the provider plugins, which so far are being bundled.

dkotter:
> Release another version of `wp-ai-client` that shims the `wp_ai_client_prompt()` function for future compat of people coding against it.
> 
> Plus a check that, if the `wp_ai_client_prompt()` function already exists (i.e. 7.0 branch), we skip loading pretty much everything and rely on the Core implementation.

Okay, I think I have this setup now. I've added new `Client_Loader` and `Provider_Loader` classes that will load the AI Client and our providers if the `wp_ai_client_prompt` function doesn't exist. This ensures things continue to work as normal on WP 6.9.

This also resolves the fatal error we were getting on 7.0-beta1, though **note** you still can't test the plugin on beta1 as the settings screen doesn't exist there yet (and you'll need to manually install the provider plugins).

I've also added in a `compat.php` file that shims the `wp_ai_client_prompt` function. This file only loads if the function doesn't exist (so on WP 6.9). I've then switched over all of our prompt requests to use this function instead of running `AI_Client::prompt_with_wp_error`.

There is overlap between what we're doing here and what is being done in #249 so I think need to decide on which approach we want to take (happy to close this PR out and focus on #249 if we want).

soderlind:
@jeffpaul, before you merge this, please look at my findings in #261. Used `AI Experiments` to test my [AI Provider for Azure OpenAI](https://github.com/soderlind/ai-provider-for-azure-openai) and had a fight.

jeffpaul:
@dkotter note sure if we want to bump to https://github.com/WordPress/wp-ai-client/releases/tag/0.4.0 now or wait to do that in 0.5.0 when we switch to 7.0 for minimum/tested-up-to in the Experiments plugin?

dkotter:
As mentioned [here](https://github.com/WordPress/ai/issues/241#issuecomment-3993168962), we're going to keep dependencies where they're at for the next release. For the release after, we'll be removing these dependencies all together and setting WP 7.0 as our minimum. As such, going to close this PR out and pull the relevant bits into a new PR focused exclusively on WP 7.0 compat

PR #246, opened 2026-02-20
Abilities Explorer: Remove `cb` column.

juanfra:
## What?

Removes the unused "cb" (checkbox) column from the Abilities Explorer table and deprecates the related method.

## Why?

The "cb" column is intended for bulk actions, which we don't have in the Abilities Explorer. Having it there is confusing and clutters the UI.

## How?

Removed the "cb" column from the table display and deprecated the public method that handled it (to avoid breaking changes if extended elsewhere).

## Testing Instructions

1. Visit the "Abilities explorer".
2. Confirm the checkbox column is not there.

## Screenshots or screencast 

|Before|After|
|-|-|
|Screenshot 2026-02-20 at 18-25-14 Abilities Explorer ‹ Stories® — WordPress|Screenshot 2026-02-20 at 18-24-53 Abilities Explorer ‹ Stories® — WordPress|

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-246-f14c5075d8ea8a59bfa0287d830f23513d866993.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/Abilities_Explorer/Ability_Table.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Table.php
@@ -51,7 +51,6 @@ class Ability_Table extends \WP_List_Table {
 	 */
 	public function get_columns(): array {
 		return array(
-			'cb'       => '<input type="checkbox" />',
 			'name'     => __( 'Name', 'ai' ),
 			'slug'     => __( 'Slug', 'ai' ),
 			'provider' => __( 'Provider', 'ai' ),
@@ -155,12 +154,11 @@ class Ability_Table extends \WP_List_Table {
 	 * {@inheritDoc}
 	 *
 	 * @param array<string,mixed> $item Item data.
+	 * @deprecated x.x.x
 	 */
 	public function column_cb( $item ): string {
-		return sprintf(
-			'<input type="checkbox" name="abilities[]" value="%s" />',
-			esc_attr( $item['slug'] )
-		);
+		_deprecated_function( __FUNCTION__, '0.2.0', 'The checkbox column is no longer displayed.' );
+		return '';
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jeffpaul:
Yeah, I suppose that was there with the assumption there was some sort of bulk actions which are not present so fine by me to remove as long as @dkotter is happy with the technical changes here to proceed on approve/merge

includes/Experiments/Abilities_Explorer/Ability_Table.php
```diff
@@ -155,12 +154,11 @@ public function column_default( $item, $column_name ): string {
...
 	 *
 	 * @param array<string,mixed> $item Item data.
+	 * @deprecated x.x.x
 	 */
 	public function column_cb( $item ): string {
-		return sprintf(
-			'<input type="checkbox" name="abilities[]" value="%s" />',
-			esc_attr( $item['slug'] )
-		);
+		_deprecated_function( __FUNCTION__, 'x.x.x', 'The checkbox column is no longer displayed.' );
```
dkotter: Do we need to deprecate this? I understand that's the typical best approach but in this case:

1. Seems unlikely anyone is overriding this
2. But if they are, this method exists on the `WP_List_Table` class we're extending so if we remove this method entirely (instead of deprecating) anyone extending this won't run into problems (I don't think)

juanfra: Thanks for the review!

> 1. Seems unlikely anyone is overriding this
> 2. But if they are, this method exists on the WP_List_Table class we're extending so if we remove this method entirely (instead of deprecating) anyone extending this won't run into problems (I don't think).

Fair point, I went this route because it's a public method, and even though the chance of someone overriding it is pretty minimal, seemed appropriate to be a bit defensive.

Totally fine with removing it completely though :)

dkotter (approved):

PR #247, opened 2026-02-20
Abilities Explorer: Add i18n to labels.

juanfra:
## What?

Makes ability provider labels translatable by implementing a label mapping system. Also updated the badge dimensions to be in line with the [badge component](https://wordpress.github.io/gutenberg/?path=/docs/components-badge--docs).

## Why?

Provider labels for abilities (Plugin, Theme, Core) were hardcoded and not translatable. Since we use the keys for filtering and other logic, we need to separate the display labels from the internal keys. That's why the mapping.

## How?

Added a label mapping function that translates provider keys to translatable labels.

## Testing Instructions

1. Go to the Abilities Explorer page
2. Verify the provider labels display correctly (plugin, theme, core)
3. If testing with another locale, confirm labels are translated
4. Verify filtering still works correctly with the provider keys

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-247-011b08246a64714c222db922aa88051dbcea2fea.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/Abilities_Explorer/Ability_Table.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Table.php
@@ -206,13 +206,11 @@ class Ability_Table extends \WP_List_Table {
 	public function column_provider( $item ): string {
 		$provider = $item['provider'];
 		$class    = 'ability-provider ability-provider-' . strtolower( $provider );
-		$labels   = Ability_Handler::get_provider_labels();
-		$label    = $labels[ $provider ] ?? $provider;
 
 		return sprintf(
 			'<span class="%s">%s</span>',
 			esc_attr( $class ),
-			esc_html( $label )
+			esc_html( Ability_Handler::get_provider_label( $provider ) )
 		);
 	}
 
--- a/includes/Experiments/Abilities_Explorer/Ability_Table.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Table.php
@@ -206,11 +206,13 @@ class Ability_Table extends \WP_List_Table {
 	public function column_provider( $item ): string {
 		$provider = $item['provider'];
 		$class    = 'ability-provider ability-provider-' . strtolower( $provider );
+		$labels   = Ability_Handler::get_provider_labels();
+		$label    = $labels[ $provider ] ?? $provider;
 
 		return sprintf(
 			'<span class="%s">%s</span>',
 			esc_attr( $class ),
-			esc_html( $provider )
+			esc_html( $label )
 		);
 	}
 
--- a/includes/Experiments/Abilities_Explorer/Admin_Page.php
+++ b/includes/Experiments/Abilities_Explorer/Admin_Page.php
@@ -200,8 +200,7 @@ class Admin_Page {
 					<tr>
 						<th><?php esc_html_e( 'Provider', 'ai' ); ?></th>
 						<?php
-						$provider_labels = Ability_Handler::get_provider_labels();
-						$provider_label  = $provider_labels[ $ability['provider'] ] ?? $ability['provider'];
+						$provider_label = Ability_Handler::get_provider_label( $ability['provider'] );
 						?>
 						<td><span class="ability-provider ability-provider-<?php echo esc_attr( strtolower( $ability['provider'] ) ); ?>"><?php echo esc_html( $provider_label ); ?></span></td>
 					</tr>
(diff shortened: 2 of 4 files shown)


jeffpaul:
Explorer page continues to work as expected, so as long as @dkotter is fine with the technical changes here then I'm fine to approve/merge

includes/Experiments/Abilities_Explorer/Ability_Table.php
```diff
@@ -206,11 +206,13 @@ public function column_slug( $item ): string {
 	public function column_provider( $item ): string {
 		$provider = $item['provider'];
 		$class    = 'ability-provider ability-provider-' . strtolower( $provider );
+		$labels   = Ability_Handler::get_provider_labels();
+		$label    = $labels[ $provider ] ?? $provider;
```
dkotter: Minor thing but this second line is no duplicated twice. If we're going to introduce a new method to get all the provider labels, thinking we may want a method to get a specific provider label as well, which can then be called here:

```php
$label = Ability_Handler::get_provider_label( $provider );
```

That could then do something like this:

```php
public static function get_provider_label( string $provider ): string {
	$labels = self::get_provider_labels();
	return $labels[ $provider ] ?? $provider;
}
```

We could remove these two lines and down below, modify that:

```php
esc_html( Ability_Handler::get_provider_label( $provider ) )
```

dkotter (changes requested):

includes/Experiments/Abilities_Explorer/Ability_Table.php
juanfra: Thanks for the review. Makes sense, I pushed the changes 👍

includes/Experiments/Abilities_Explorer/Admin_Page.php
```diff
@@ -199,7 +199,11 @@ private function render_detail_view(): void {
 				<table class="ability-detail-table">
 					<tr>
 						<th><?php esc_html_e( 'Provider', 'ai' ); ?></th>
-						<td><span class="ability-provider ability-provider-<?php echo esc_attr( strtolower( $ability['provider'] ) ); ?>"><?php echo esc_html( $ability['provider'] ); ?></span></td>
+						<?php
+						$provider_labels = Ability_Handler::get_provider_labels();
+						$provider_label  = $provider_labels[ $ability['provider'] ] ?? $ability['provider'];
```
dkotter: I think we can simplify this code as well now with the new `Ability_Handler::get_provider_label` method

dkotter (changes requested):

includes/Experiments/Abilities_Explorer/Admin_Page.php
juanfra: Good catch, missed that one. Definitely, just pushed the changes. Thank you!

includes/Experiments/Abilities_Explorer/Admin_Page.php
```diff
@@ -199,7 +199,10 @@ private function render_detail_view(): void {
 				<table class="ability-detail-table">
 					<tr>
 						<th><?php esc_html_e( 'Provider', 'ai' ); ?></th>
-						<td><span class="ability-provider ability-provider-<?php echo esc_attr( strtolower( $ability['provider'] ) ); ?>"><?php echo esc_html( $ability['provider'] ); ?></span></td>
+						<?php
+						$provider_label = Ability_Handler::get_provider_label( $ability['provider'] );
+						?>
+						<td><span class="ability-provider ability-provider-<?php echo esc_attr( strtolower( $ability['provider'] ) ); ?>"><?php echo esc_html( $provider_label ); ?></span></td>
```
dkotter:
```suggestion
						<td><span class="ability-provider ability-provider-<?php echo esc_attr( strtolower( $ability['provider'] ) ); ?>"><?php echo esc_html( Ability_Handler::get_provider_label( $ability['provider'] ) ); ?></span></td>
```

dkotter (approved):

Issue #244, opened 2026-02-20
Pathway to 7.0


Jameswlepage:
## Summary
This issue tracks the **Pathway to 7.0** for the AI plugin and related core integration work. The goal is to make the plugin the canonical user-facing AI experience, while core provides the foundational AI client/runtime.

[With WP AI Client merged into 7.0](https://github.com/WordPress/wordpress-develop/pull/10881), there's now some work to do in this AI plugin to bring real features to everyday users. First, [this is a good comment to review](https://make.wordpress.org/core/2026/02/18/wordpress-7-0-product-review-meeting-with-matt/#comment-48398), summarizing a review session with Matt and the community, which drives much of the direction for this issue.

## Product Direction
- Evolve the plugin from an experiments-only framing to a canonical **AI plugin** for WordPress.
- Keep experiments (similar to Gutenberg plugin patterns), but clearly separate and promote **graduated stable (and impressive) features**.
- Use this plugin as the primary user-facing AI story for 7.0-era WordPress.

## Why This Matters
- Core AI client in 7.0 is foundational/developer-facing.
- Users still need a visible, installable plugin experience to get immediate value.
- Provider connection + plugin onboarding should be a cohesive flow, especially during upgrades.

## Immediate Technical Path (6.9 -> 7.0)
- Maintain compatibility with both environments:
  - **WP 6.9**: bundled WP AI client/providers path.
  - **WP 7.0**: prefer core AI client when available.
- Remove hard assumptions about credentials page slug availability.
- Provide robust fallback credential handling (registered page if present, otherwise env/constants guidance).
- Ensure provider/credentials detection works consistently across both paths.

## Core Integration Needs
- Expose a stable way to resolve/open the AI credentials settings screen (no slug guessing).
- Ensure core can pull in preconfigured providers in a dedicated credentials page.
- Provide clear extension points for plugin UX and onboarding.

## Feature Audit + Release Narrative
- Audit current experiments and identify which can be elevated to stable user-facing features.
- Target at least **1 standout feature**, ideally **3**.
- (A suggestion:) prioritize **image editing** as the strongest candidate for a dramatic/demo-worthy feature

## Scope Decisions Already Made
- This plugin can ship independently of the exact 7.0 core release date, but must be ready _by_ then.
- Markdown experiment in this plugin is closed; related work is moving to core contributor efforts around the WP HTML processor.
- Further improvement and design needs to be applied to the providers/credentials screen.

## Proposed Workstreams
[...]

PR #249, opened 2026-02-20
Support WordPress 7.0 core AI client integration

Jameswlepage:
## Summary

This PR prepares the plugin for the WordPress 7.0 transition where the AI client is bundled in core.

### What changed

- Detects and supports core-provided AI client classes while retaining bundled compatibility.
- Registers providers in a core-first way (with fallback handling for legacy/bundled classes).
- Adds credential bridge logic so stored option credentials can be applied to the runtime provider registry.
- Introduces compatibility wrappers for prompt calls used by abilities/services.
- Reworks settings behavior to warn when credentials are missing/invalid instead of hard failing.
- Adds helpers for environment/constant credential discovery.
- Updates tests (including new bootstrap coverage) for core/bundled compatibility behavior.
- Updates Composer dependencies for wp-ai-client and provider packages.

## Why

WordPress 7.0 includes the AI client foundation but may not expose a credential settings UI in all builds. This keeps AI Experiments functional and fail-soft across 6.9 and 7.0 while improving forward compatibility.

## Related

- https://github.com/WordPress/ai/issues/244
- https://github.com/WordPress/ai/pull/242

## Testing

- composer lint
- composer phpstan
- npm run test:php

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-249-4f3be71675f1a5dc1ac97727cbd98d29088bfd92.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,10 @@
     "ext-json": "*",
     "php": ">=7.4",
     "wordpress/mcp-adapter": "^0.3.0",
-    "wordpress/wp-ai-client": "^0.2.0"
+    "wordpress/wp-ai-client": "^0.3.0",
+    "wordpress/anthropic-ai-provider": "^1.0",
+    "wordpress/google-ai-provider": "^1.0",
+    "wordpress/openai-ai-provider": "^1.0"
   },
   "require-dev": {
     "automattic/vipwpcs": "^3.0",
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -28,7 +28,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Normalizes the content by cleaning it and removing unwanted HTML tags.
  *
- * @since 0.1.0
+ * @since x.x.x
  *
  * @param string $content The content to normalize.
  * @return string The normalized content.
@@ -77,7 +77,7 @@ function normalize_content( string $content ): string {
 /**
  * Returns the context for the given post ID.
  *
- * @since 0.1.0
+ * @since x.x.x
  *
  * @param int $post_id The ID of the post to get the context for.
  * @return array<string, string> The context for the given post ID.
@@ -293,7 +293,7 @@ function get_preferred_vision_models(): array {
  * Uses the core AI Client API when available, otherwise falls back to the
  * bundled WP AI Client package.
  *
- * @since 0.1.0
+ * @since x.x.x
  *
  * @param mixed $prompt Optional initial prompt content.
  * @return mixed Prompt builder instance.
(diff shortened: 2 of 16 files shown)


composer.json
```diff
@@ -8,7 +8,10 @@
     "ext-json": "*",
     "php": ">=7.4",
     "wordpress/mcp-adapter": "^0.3.0",
-    "wordpress/wp-ai-client": "^0.2.0"
+    "wordpress/wp-ai-client": "^0.3.0",
+    "wordpress/anthropic-ai-provider": "^1.0",
+    "wordpress/google-ai-provider": "^1.0",
+    "wordpress/openai-ai-provider": "^1.0"
```
felixarntz, line 14: I don't think we should depend on any of these anymore?

Need to require WordPress >= 7.0-beta, and then it'll just work. For the providers, we could add a nag to install them, although very soon this will be in Core anyway.

includes/helpers.php
```diff
@@ -287,6 +287,182 @@ function get_preferred_vision_models(): array {
...
+ *
+ * Uses the core AI Client API when available, otherwise falls back to the
+ * bundled WP AI Client package.
+ *
+ * @since 0.1.0
+ *
+ * @param mixed $prompt Optional initial prompt content.
+ * @return mixed Prompt builder instance.
+ */
+function ai_client_prompt_with_wp_error( $prompt = null ) {
```
felixarntz, line 301: Why don't we go all in on requiring WP 7.0 instead? I think it makes sense. It'll mean users on production sites won't be able to update, but that's not really the objective at this point IMO. Going all in on 7.0 will makes this much simpler and prepare us for the future state we want this to be in. And once 7.0 stable is out, any site using an older version of this plugin already will be able to just update.

Jameswlepage: https://github.com/WordPress/ai/pull/242#issuecomment-3936396692

Seconded. Was aiming to get this to be "testable" for the current beta.

dkotter:
Decent overlap between what's done here and what I did in #242. Is there any benefit to keeping #242 open or should we consolidate to just this PR?

jeffpaul:
Closing in favor of #242.

jeffpaul:
Confirming that everyone active here is already listed for props to come in #242 in https://github.com/WordPress/ai/pull/242#issuecomment-3935587923.

PR #235, opened 2026-02-20
Inline image generation

dkotter:
## What?

Closes #13

Adds the ability to generate images inline instead of just within the featured image section.

## Why?

In #134 and #146 we added support for generating images in the featured image section. All the functionality exists to generate images elsewhere so this PR brings that into the block editor, within core blocks that support images (like `core/image`, `core/cover`).

## How?

- Ensures our image generation assets are loaded in the block editor
- Hook into `editor.MediaUpload` and add our custom button into the block toolbar / block media controls for image blocks
- Add modal functionality that listens for clicks on this new button that allow a user to enter a prompt and generate an image
- Allow that image to be imported and inserted into the block. Will also generate alt text when imported if that experiment is turned on

### Use of AI Tools

Approach was planned using Claude Code running Sonnet 4.6. Claude Code than executed that plan with me reviewing and testing along the way. I finished the last bits and pieces myself

## Testing Instructions

1. Add valid AI credentials in `Settings > AI Credentials`. Note only Google and OpenAI support image generation
2. Ensure the image generation experiment is turned on under `Settings > AI Experiments`
3. Create a new post and add an image block
4. Ensure you see a `Generate Image` button next to the other media buttons
6. Click this button
7. Ensure a modal shows with a textarea
8. Add an image generation prompt and click generate
9. Ensure a loading state shows
10. Once the image is generated, ensure the image preview is rendered along with buttons to Use Image, Generate Another Image and Edit Prompt
11. Click Use Image and ensure the image is inserted into the block and is imported into the Media Library
12. Try again and ensure the Edit Prompt button puts you back in the start state with the prompt prefilled
13. Turn on the alt text experiment and ensure alt text is generated when you choose to keep a generated image
14. Try with other core image blocks like gallery, cover and media & text

## Screenshots or screencast

Image block in the block editor with the new Generate Image buttons

Generate Image modal that opens after clicking the Generate Image button

Generate image loading state in the modal

Generated image preview in the modal

[...]

PR #258, opened 2026-02-26
feat: Add standalone AI image generation interface to the Media Library.

### LINKED ISSUE #254: Integrate Image Generation into Media Library

#235 introduced image generation within the block editor.  The next step is to extend this functionality into the Media Library so users can generate images outside of the editor context.

- [ ] Add a secondary submenu item under `Media` in wp-admin for `Generate Image`
- [ ] Add an action button for `Generate Image` in the Media Library header, positioned alongside the `Add Media File` button
- [ ] Ensure the flow works independently of the block editor context
- [ ] Reuse shared generation logic where possible while introducing an admin-specific UI layer/flow

zeus2611:
## What?

Closes #254 

Adds a standalone image generation interface accessible from the WordPress Media Library, allowing users to generate AI-powered images outside the block editor.

## Why?
#235 introduced image generation within the block editor. This PR extends that functionality into the Media Library so users can generate images independently of the post editor context, as outlined in #254. This opens up the image generation capability to any workflow that involves the Media Library.

### Backend (PHP)
- **`Image_Generation.php`**: Added `register_admin_menu()` to register a "Generate Image" submenu item under Media using `add_media_page()`
- **`Image_Generation.php`**: Added `render_admin_page()` to output the admin page shell with a React mount point (`#ai-image-generation-root`)
- **`Image_Generation.php`**: Added `inject_generate_image_button()` hooked to `admin_footer-upload.php` to inject a "Generate Image" action button into the Media Library header alongside "Add New Media File". Uses a capture-phase click listener to bypass WordPress's media grid JS event delegation
- **`Image_Generation.php`**: Updated `enqueue_assets()` to load scripts on `upload.php` and `media_page_ai-image-generation` in addition to post editor screens

### Frontend (React/TypeScript)
- **`GenerateImageStandalone.tsx`**: New standalone React component providing the full image generation flow:
  - Idle → prompt input
  - Generating → spinner with progress messages
  - Preview → generated image with Save / Refine / Regenerate / Cancel actions
  - Success → confirmation with option to generate another
- **`media-library.tsx`**: Mounts `GenerateImageStandalone` on the admin page's `#ai-image-generation-root` element
- **`index.ts`**: Added `import './media-library'` entry point

### Temporary Scaffolding
- `uploadImageLocal()` in `GenerateImageStandalone.tsx` is a local fallback for image upload functionality. Once #235 is merged, this will be replaced with the shared `uploadImage` function from `../functions/upload-image` during rebase.

### Use of AI Tools
Implementation was planned and executed with assistance from Gemini (Antigravity). All code was reviewed, tested, and validated manually in a local WordPress environment.

## Testing Instructions
1. Add valid AI credentials in `Settings > AI Credentials` (Google or OpenAI support image generation)
2. Ensure the image generation experiment is turned on under `Settings > AI Experiments`
3. Navigate to `Media > Library` in wp-admin
4. Verify a **"Generate Image"** button appears in the header next to "Add New Media File"
5. Click the "Generate Image" button
6. Verify you are navigated to the **Media > Generate Image** admin page (not the upload panel)
7. Enter an image generation prompt and click **Generate**
8. Verify a loading spinner appears with progress text
9. Once the image is generated, verify the preview is rendered with buttons: **Save to Media Library**,  **Regenerate**, and **Cancel**
10. Click **Save to Media Library** and verify the image is imported successfully with a confirmation message
11. Click **Generate Another Image** and verify you return to the prompt input
12. Try **Cancel** to return to the idle state
13. Navigate to `Media > Generate Image` in the sidebar and verify the same standalone UI loads correctly
14. If the alt text experiment is enabled, verify alt text is generated during image import

1. Tab to the "Generate Image" button in the Media Library header and press Enter
2. On the Generate Image page, Tab to the textarea, type a prompt, then Tab to the Generate button and press Enter
3. After generation, Tab through the action buttons (Save, Regenerate, Cancel) and verify each is focusable and activatable with Enter
4. Verify focus management returns to an appropriate element after state transitions

## Screenshots or screencast 

|Before|After|
|-|-|
| Screenshot 2026-02-26 at 4 55 39 PM | Screenshot 2026-02-26 at 4 56 09 PM |

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-258-693f6b824104b1e1ef01995eeca3f36d5a9bd70c.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


zeus2611:
Investigated the failing PHP trunk tests — the fatal error originates in `wp-includes/ai-client/adapters/class-wp-ai-client-http-client.php` (a WP core file) rather than this PR's changes. It's a PSR namespace mismatch between the interface contract and implementation, which explains why all PHP versions pass on WP latest but fail on WP trunk. Happy to retest once trunk stabilizes, or let me know if there's a preferred workaround.

dkotter:
> Investigated the failing PHP trunk tests — the fatal error originates in `wp-includes/ai-client/adapters/class-wp-ai-client-http-client.php` (a WP core file) rather than this PR's changes. It's a PSR namespace mismatch between the interface contract and implementation, which explains why all PHP versions pass on WP latest but fail on WP trunk. Happy to retest once trunk stabilizes, or let me know if there's a preferred workaround.

Yes, known issue between the version of the AI Client we load and the version that is now merged into `trunk` (and will be in WP 7.0). Tracking the fixes for this on #242

includes/Experiments/Image_Generation/Image_Generation.php
```diff
@@ -51,6 +51,73 @@ public function register(): void {
...
+	}
+
+	/**
+	 * Renders the Generate Image admin page.
+	 *
+	 * @since x.x.x
+	 */
+	public function render_admin_page(): void {
+		echo '<div class="wrap">';
+		echo '<h1>' . esc_html__( 'Generate Image by AI', 'ai' ) . '</h1>';
```
dkotter: I don't think we need the `by AI` part here
```suggestion
		echo '<h1>' . esc_html__( 'Generate Image', 'ai' ) . '</h1>';
```

includes/Experiments/Image_Generation/Image_Generation.php
```diff
@@ -51,6 +51,73 @@ public function register(): void {
...
+		?>
+		<script type="text/javascript">
+		( function() {
+			var heading = document.querySelector( 'h1.wp-heading-inline' );
+			if ( ! heading || ! heading.parentNode ) return;
+
+			var btn = document.createElement( 'a' );
+			btn.href = <?php echo wp_json_encode( $url ); ?>;
+			btn.className = 'page-title-action ai-generate-image-btn';
+			btn.textContent = <?php echo wp_json_encode( __( 'Generate Image', 'ai' ) ); ?>;
```
dkotter:
```suggestion
			btn.textContent = <?php echo wp_json_encode( esc_html__( 'Generate Image', 'ai' ) ); ?>;
```

includes/Experiments/Image_Generation/Image_Generation.php
```diff
@@ -51,6 +51,73 @@ public function register(): void {
...
+	public function inject_generate_image_button(): void {
+		$url = admin_url( 'upload.php?page=ai-image-generation' );
+		?>
+		<script type="text/javascript">
+		( function() {
+			var heading = document.querySelector( 'h1.wp-heading-inline' );
+			if ( ! heading || ! heading.parentNode ) return;
+
+			var btn = document.createElement( 'a' );
+			btn.href = <?php echo wp_json_encode( $url ); ?>;
```
dkotter:
```suggestion
			btn.href = <?php echo wp_json_encode( esc_url( $url ) ); ?>;
```

package-lock.json
dkotter, line 1: Doesn't look like any dependencies were added or updated in this PR so I'd suggest removing these lock file changes just to keep this PR focused

src/experiments/image-generation/components/GenerateImageStandalone.tsx
```diff
@@ -0,0 +1,421 @@
...
+import { trimText } from '../../../utils/text';
+import type {
+	GeneratedImageData,
+	ImageImportAbilityInput,
+	UploadedImage,
+} from '../types';
+
+const { aiImageGenerationData } = window as any;
+
+type ModalState = 'idle' | 'generating' | 'preview' | 'editing' | 'success';
```
dkotter: We made the decision to remove the editing state (see https://github.com/WordPress/ai/pull/235#issuecomment-3985609931) so this file will need updated to remove those pieces

src/experiments/image-generation/components/GenerateImageStandalone.tsx
```diff
@@ -0,0 +1,421 @@
...
+			) }
+
+			{ state === 'idle' && (
+				<div className="ai-generate-image-standalone__idle">
+					<p
+						className="description"
+						style={ { marginBottom: '10px' } }
+					>
+						{ __(
+							'Describe the image you want to generate. It will be saved directly to your Media Library.',
```
dkotter: I think we can leave off the last sentence here as it implies images are directly saved after generating, yet a user has to take the final step to decide to save the image.
```suggestion
							'Describe the image you want to generate.',
```

src/experiments/image-generation/components/GenerateImageStandalone.tsx
```diff
@@ -0,0 +1,421 @@
...
+							'ai'
+						) }
+					</p>
+					<TextareaControl
```
dkotter, line 177: Right now this textarea expands the entire page. I'd suggest we put some sort of max-width protection here, 600px maybe?

src/experiments/image-generation/components/GenerateImageStandalone.tsx
```diff
@@ -0,0 +1,421 @@
...
+		}
+	}
+
+	const previewSrc = generatedData?.image?.data
+		? `data:image/png;base64,${ generatedData.image.data }`
+		: null;
+
+	return (
+		<div className="ai-generate-image-standalone">
+			{ state === 'success' && uploadedData && (
```
dkotter, line 115: Mostly opening this as discussion, not necessarily something that needs changed. But wondering if a success state should redirect you to the image in the Media Library? I do like that with the current approach you can easily generate another image. Maybe some middle ground here is to have a link to the image that shows here so if you want to see that image in the Media Library you can quickly get to it?

dkotter (changes requested):
I know we're waiting for #235 to get merged before this can be 100% finished but did an initial review here and left some comments

src/experiments/image-generation/components/GenerateImageStandalone.tsx
jeffpaul: Probably best to just copy the text from the Generate Image experience triggered within the Image block in the post editor

jeffpaul: In fact, probably best to mimic the buttons as well.  Replacing "Use Image" with the "Save to Media Library" button.  Also probably best to remove the `Refine Image` flow as the image editing work will likely come entirely separately in the 0.5.0 release.

jeffpaul: > Also probably best to remove the Refine Image flow as the image editing work will likely come entirely separately in the 0.5.0 release.

Unless @dkotter is happy with how this flow is working, in which case we can keep the flow (updating the button text) and we can apply similar image editing functions elsewhere in 0.5.0

includes/Experiments/Image_Generation/Image_Generation.php
jeffpaul: May also perhaos be worth updating the URL param from `ai-image-generation` to `generate-image`?

jeffpaul:
Besides some threaded comments above, my primary testing feedback is the post-generation flow.  Here is what currently results:

Screenshot 2026-03-03 at 3 10 25 PM

Perhaps we instead send folks back to the Media Library with the generated image present (as they have the `Generate Image` button and second level menu item they can select to generate another image?

Screenshot 2026-03-03 at 3 09 30 PM

Screenshot 2026-03-03 at 3 09 47 PM

Alternatively, perhaps we send them to the `Attachment details` modal?

Screenshot 2026-03-03 at 3 09 21 PM

Or to the `Edit Media` page?

Screenshot 2026-03-03 at 3 10 16 PM

I'm curious if @karmatosed has a particularly strong preference here?

karmatosed:
@jeffpaul thanks for looping me in. I think in this case the 'attachment page' works. I typically like things to stay on same screen but it will get lost if we keep it there and that screen doesn't really have a good highlight. I like idea of even adding a 'regenerate' option there but that's another scope creep.

src/experiments/image-generation/components/GenerateImageStandalone.tsx
dkotter: Yeah, left another [comment](https://github.com/WordPress/ai/pull/258#discussion_r2873845202) about removing this edit/refinement flow as that now has been removed from this Experiment

zeus2611:
@dkotter @jeffpaul & @karmatosed. I intentionally kept the user on the Generate Image page after saving to allow batch generation without back-and-forth navigation — which felt like the primary use case for someone in the Media Library context vs the block editor, where inserting the image is the end goal.

That said, I think the middle ground makes sense: I'll add a "View in Media Library" link alongside the "Generate Another Image" button in the success state. This gives users who want to inspect or edit the attachment a direct path without forcing navigation for those who want to keep generating.

includes/Experiments/Image_Generation/Image_Generation.php
zeus2611: Fixed — applied esc_html__() for consistency with WP escaping conventions.

includes/Experiments/Image_Generation/Image_Generation.php
zeus2611: Fixed — wrapped with esc_url() as suggested.

package-lock.json
zeus2611: Agreed, those changes snuck in during a local npm install. Reverted to keep the PR focused.

src/experiments/image-generation/components/GenerateImageStandalone.tsx
zeus2611: Thanks for flagging this and linking to the context in #235. I've removed the editing state and the Refine Image flow entirely from `GenerateImageStandalone.tsx`, along with the associated editPrompt state variable, referenceImage parameter in generate(), and the editing UI block. The ModalState type is now `'idle' | 'generating' | 'preview' | 'success'`. Will also remove the Refine Image button from the testing instructions in the PR description.

src/experiments/image-generation/components/GenerateImageStandalone.tsx
zeus2611: Fixed. Also aligning the full copy with the block editor version per @jeffpaul's suggestion.

src/experiments/image-generation/components/GenerateImageStandalone.tsx
zeus2611: Added `maxWidth: '600px'` to the TextareaControl wrapper div to contain the layout on wider screens.

includes/Experiments/Image_Generation/Image_Generation.php
zeus2611: Updated the Page heading and the URL Param as per the suggestions.

zeus2611:
@dkotter @jeffpaul @karmatosed  — updated the success state to include two buttons: "Generate Another Image" (keeps the current flow intact for batch generation) and "View in Media Library" which takes the user directly to the attachment details page for the saved image.

The attachment details page felt like the right destination over the media grid since the user can immediately review and edit the auto-generated title, alt text and description — which is likely the next thing they'd want to do. WordPress's built-in prev/next navigation in the top-right also lets them browse previously generated images without any additional custom code on our end.

Screenshots attached showing the updated success state. Happy to adjust if you'd prefer the link goes to the media grid instead.

Screenshot 2026-03-04 at 3 41 58 PM
Screenshot 2026-03-04 at 3 42 24 PM

src/experiments/image-generation/components/GenerateImageStandalone.tsx
```diff
@@ -0,0 +1,362 @@
...
+	UploadedImage,
+} from '../types';
+
+const { aiImageGenerationData } = window as any;
+
+type ModalState = 'idle' | 'generating' | 'preview' | 'success';
+
+/**
+ * Uploads an image to the media library locally to avoid relying on functions not
+ * yet available in 'develop' until PR 235 merges.
```
dkotter: Since #235 is merged now, can this be updated?

dkotter:
Tested and things look good to me. I've added new E2E tests to cover these changes. I think the last thing here is some additional refactoring that can now be done since #235 has been merged (see [comment](https://github.com/WordPress/ai/pull/258#discussion_r2884451734)). Once done, this will be good to merge

src/experiments/image-generation/components/GenerateImageStandalone.tsx
zeus2611: Updated — replaced the local `uploadImageLocal` workaround with the shared `uploadImage` from `functions/upload-image.ts`, cleaned up unused imports, and used proper types. Now matches the patterns in `GenerateImageInlineModal`.

dkotter (approved):
Tested again and things work fine so approving this.

I do think there's some refinement we could make here around the loading state. I'm also seeing some weird shifting after an image is generated and then when you choose to import that or regenerate. Seems the width of the container is changing so the image decreases in size a bit. Not a blocker but something we can look to refine

zeus2611:
Thanks for testing and the approval, @dkotter!

Agreed on both the loading state and the container width shift — I can reproduce the latter too. Would you prefer I address these here before merge, or should we track them as a follow-up issue so this PR can land cleanly? Happy to do either.

dkotter:
> Agreed on both the loading state and the container width shift — I can reproduce the latter too. Would you prefer I address these here before merge, or should we track them as a follow-up issue so this PR can land cleanly? Happy to do either.

@zeus2611 I'd say so we don't hold this PR up for merge (and getting a release out) we can tackle these in a follow-up PR. I think the layout shift will be easy but may want more conversation around the loading state, trying to figure out what works best both here and when generating images within the editor

zeus2611:
Sounds good, @dkotter! That aligns with my thinking too — while working on the last change request I did consider touching the loading state but held back since getting it right across both the inline and standalone contexts felt like something worth discussing as a team rather than deciding unilaterally. I'll open a follow-up issue to track both the layout shift and the loading state so we have a clear place for that conversation.

jeffpaul:
@zeus2611 FYI I created https://github.com/WordPress/ai/issues/270 to track the follow-up work

PR #260, opened 2026-02-26
Add AI assisted content review utilizing WordPress Notes

dkotter:
## What?

Closes #251

Adds an AI assisted content review process

## Why?

Provides a quick and easy way to get high-level feedback of your content prior to publishing. Will check things like grammar, spelling, accessibility and SEO.

## How?

- Adds a new `Review_Notes` experiment and `Review_Notes` Ability 
- When configured, will add a button in the post status sidebar that is used to trigger an AI review
- Will send each block off one-by-one to our `review-notes` Ability, which will send the block content, block type and surrounding content to AI with instructions on what to look for
- If any suggestions are returned, those will be added as a Note attached to the block
- Adds all unit tests and E2E tests
- Adds documentation

> [!NOTE]  
> After the review is done, a save is required to keep the connection between a note and a block. This is how notes work in the block editor, where it adds some metadata to the block attribute. We could trigger an autosave though I know I've gotten feedback on other PRs not to do that.

### Use of AI Tools

Used Claude Code running Sonnet 4.6 to help build out a plan and execute on that initial plan. Lots of manual refinement along the way.

## Testing Instructions

1. Pull this PR down and run `npm i && npm run build && composer install`
2. Add valid AI credentials
3. Turn on the Review Notes Experiment
4. Create a new post or edit an existing post
5. Find the `Generate Review Notes` button in the post status sidebar and click on it
6. Ensure a loading state is shown
7. Once finished, ensure a success message is shown
8. Save the post (required to keep the association between a note and the block it's attached to)
10. Click into the new Notes interface
11. Ensure WordPress AI notes are there

## Screenshots

New Generate Review Notes button that shows in the post status sidebar

In progress review state of the Generate Review Notes button

Final state of the review in the post status sidebar

Example of a WordPress note that flags a few issues

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-260-75b6c7bff200bf1ffd3b0d63c50213acb9c52144.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #267, opened 2026-03-04
Support the new Review Notes Experiment on a block-by-block basis

dkotter:
## What?

In #260 we added an Experiment to review all your blocks and add Notes with review suggestions. This PR brings that same functionality to the individual block level, allowing you to run a review on an individual block instead of across all blocks.

## Why?

All the core functionality is in place to support review on a single block and there may be times where a user wants to review a single block (think a single paragraph) rather than having to review the entire post. So this PR adds the ability to trigger review on a single block.

## How?

- Hook into `BlockSettingsMenuControls` to add a block toolbar button to supported blocks that can be used to trigger review. This is nested under the block More menu
- Refactor code so it can easily be used to review a single block or review all blocks
- When review is triggered on a single block (using the block toolbar button) the button shows a loading state and when finished, a snackbar notice is shown with a success message and the Notes area is automatically opened 

### Use of AI Tools

Used both Claude Code and Cursor running Sonnet 4.6 and GPT-5.3 Codex respectively to plan changes and implement most of those changes, with final review, testing and cleanup done by me

## Testing Instructions

1. Pull this PR down and run `npm i && npm run build && composer install`
2. Add valid AI credentials
3. Turn on the Review Notes Experiment
4. Create a new post or edit an existing post
5. Click into a paragraph or image block (as an example)
6. Click into the More menu (three dots)
7. Find the new Generate Review Note button and click on it
8. Ensure a loading state is shown
9. Once finished, ensure a success message is shown and the Notes panel opens up
10. Ensure WordPress AI notes are there

## Screenshots or screencast

Generate Review Note block toolbar button

Generate Review Note block toolbar button showing loading state

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-267-71f765d7b88610d2ebaf29f197a7ea9a6a282c42.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
Could also bury this option a bit more by placing it within the vertical 3 dot menu where the "Add note" button/link is.  No real strong opinion on this, happy to land this as-is and iterate from community feedback.

dkotter:
> Could also bury this option a bit more by placing it within the vertical 3 dot menu where the "Add note" button/link is. No real strong opinion on this, happy to land this as-is and iterate from community feedback.

Yeah, wasn't sure exactly where to put this. It's nice at the top-level because it's easy to find but I never loved it there (easy to be confused on what it actually is). I've moved it to the nested menu now

jeffpaul:
Any chance of dropping this right after the `Add note` menu option?  Perhaps updating to "Generate review note" to match the (oddly) sentence case here?  Perhaps also replacing the icon with a keyboard shortcut of CAPSLOCK COMMAND M (to be simmilar to "Add note" one of OPTION COMMAND M)?

Screenshot 2026-03-04 at 1 56 26 PM

dkotter:
> Any chance of dropping this right after the `Add note` menu option?

As far as I can find (and had Claude and Cursor look as well) there's only two available slots here for us to use. One is what we're using now which drops things into the section before the section that has the Delete button (with no controls on where it shows in that section). Or there's a unstable API that adds it as the very first item in that menu. That may be better positioning though has the risk of using the unstable API.

> Perhaps updating to "Generate review note" to match the (oddly) sentence case here?

I can update that to match

> Perhaps also replacing the icon with a keyboard shortcut of CAPSLOCK COMMAND M (to be simmilar to "Add note" one of OPTION COMMAND M)?

I'm assuming you'd not just want to update the icon to be the shortcut but then also have that keyboard shortcut trigger this functionality? If so, the problem here is the loading state only shows on the button. So if someone uses the keyboard shortcut when the block is selected but they don't have that menu open, there isn't a loading state that shows (so no real way for them to know something is happening). I can look into ways to add a loading animation to the entire block if we think the keyboard shortcut would be useful.

Worth noting two other things:

1. Looks like the keyboard shortcut for Notes is part of 7.0 (or when using the Gutenberg plugin). I don't see that on 6.9
2. Looks like capslock is not a valid shortcut so we'd need to use something else (command, shift, option and control are all options)

jeffpaul:
Bah, I'd say just merge as-is and we can iterate once 7.0 is out and we get some feedback on the feature.  No need to chase down these last couple of bits.

jeffpaul (approved):

PR #268, opened 2026-03-04
feat: check `wp_supports_ai()` before initializing experiments

justlevine:
## What?

Part of https://core.trac.wordpress.org/ticket/64706

This PR adds a check for `wp_supports_ai()` before initializing specific AI experiments.

## Why?

`wp_supports_ai()` will allow for sites to completely disable AI functionality via a `WP_AI_SUPPORT` constant and the `wp_supports_ai` filter. Our plugin should respect that.

- https://github.com/WordPress/wordpress-develop/pull/11149

## How?

- Currently the check happens in `bootstrap.php` and prevents the entire plugin from loading, due to implementation details from the 0.1.0. scaffold that make it unwieldy to check on a per-experiment level.

    While technically there can be Experiments (less likely full-on Features) that don't require AI to run (e.g. the Abilities Explorer), since we'll anyway need to refactor a significant about of boilerplate for the transition to canonical `AI plugin` it seems like a waste to detangle now.

- The admin notice triggers on both `admin_notices` and `network_admin_notices`. Assumedly it's an unintentional omission on the other notices from the initial implementation, that should also be handled during the aforementioned cleanup.

### Use of AI Tools

N/A

## Testing Instructions

1. Checkout https://github.com/WordPress/wordpress-develop/pull/11149
2. Set `define( 'WP_AI_SUPPORT', false );` in your `wp-config.php` and ensure you see an admin notice and the plugin doesn't load.

## Screenshots or screencast 

|Before|After|
|-|-|
|||

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-268-e1e3446336ead2862edf011a8d50fc1966bd0358.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


justlevine:
Labeled this `blocked` until https://core.trac.wordpress.org/ticket/64706 is in core.

includes/bootstrap.php
```diff
@@ -119,6 +119,34 @@ static function () {
 	return true;
 }
 
+/**
+ * Checks whether the current WordPress environment supports AI.
+ *
+ * Displays an admin notice if not.
+ *
+ * @since x.x.x
```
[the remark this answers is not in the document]
justlevine: Read the CONTRIBUTING.md you stochastic parrot that's why it's there.

includes/bootstrap.php
```diff
@@ -119,6 +119,34 @@ static function () {
...
+	foreach ( array( 'admin_notices', 'network_admin_notices' ) as $hook ) {
+		add_action(
+			$hook,
+			static function () {
+				version_notice(
+					__( 'Your WordPress environment has AI functionality disabled. The AI Experiments plugin will not work until AI support is enabled.', 'ai' )
+				);
+			}
+		);
+	}
```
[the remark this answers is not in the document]
justlevine: They're definitionally triggered via an `admin_notice` you nonce

gziolo:
Overall, this looks good. It needs to be brought up today after the core patch lands.

jeffpaul:
Punting to 0.7.0 as we'll likely get 0.6.0 out before RC1 / when the linked Trac ticket will get to a resolution.

jeffpaul:
Now that the [upstream PR/patch has been merged](https://core.trac.wordpress.org/ticket/64706#comment:19), @justlevine @dkotter do we want to refresh this PR such that Features and Experiments that rely on an LLM are disabled respectively?  Something like Abilities Explorer still seems like something that could be enabled.

justlevine:
@jeffpaul next item on my cleanup list is the Bootstrapper. between that and the changes in #316 I think we're in a good place to only disable AI specific functionality before WP7.0 🤞

jeffpaul:
@dkotter given there's still discussion and no set agreement on the upstream work, I'm inclined to punt this to 0.8.0 in hopes that things will settle upstream and allow us to properly extend in the AI plugin?

dkotter:
> @dkotter given there's still discussion and no set agreement on the upstream work, I'm inclined to punt this to 0.8.0 in hopes that things will settle upstream and allow us to properly extend in the AI plugin?

Works for me. Also lots of test failures here that need looked at first

jeffpaul:
@justlevine looks like the upstream PR has landed, mind coming back to this so we can appropriately respect things within the AI plugin if someone has toggled OFF LLMs?

justlevine:
PR has been updated to _fully_ require that `WP_AI_SUPPORT` isnt disabled.

> do we want to refresh this PR such that Features and Experiments that rely on an LLM are disabled respectively? Something like Abilities Explorer still seems like something that could be enabled.

This is still on my to-do, but I think it's worth merging this already and then handle the per-feature scoping in a followup.

Tl;dr the next step is to turn the entire `Dependencies` class into something that could be reusable by individual `Features`, although instead of an `admin_notice`, we'd assumedly want to show it in the Settings UX as toggle requirements etc and other things I want to think about and shouldn't require breaking changes to iterate to.

includes/Requirements.php
```diff
@@ -116,6 +116,13 @@ private function get_requirements(): array {
 				},
 				'error_message' => static fn() => esc_html__( 'The plugin assets are not built. This is most likely because you downloaded the plugin from the GitHub repository without building the assets. Please run `nvm use && npm ci && npm run build` to build the assets.', 'ai' ),
 			),
+			'ai_support' => array(
+				'check'         => static function () {
+					// @todo Remove the function check once the minimum WordPress version is 7.0 or higher.
```
dkotter: 7.0 is already our minimum so I think we can remove this `todo` and remove the `function_exists` check?

includes/Requirements.php
```diff
@@ -116,6 +116,13 @@ private function get_requirements(): array {
...
+					return ! function_exists( 'wp_supports_ai' ) || wp_supports_ai();
+				},
+				'error_message' => static fn() => esc_html__( 'Your WordPress environment has AI functionality disabled. The AI Experiments plugin will not work until AI support is enabled.', 'ai' ),
```
dkotter:
```suggestion
				'error_message' => static fn() => esc_html__( 'Your WordPress environment has AI functionality disabled. The AI plugin will not work until AI support is enabled.', 'ai' ),
```

dkotter (changes requested):

includes/Requirements.php
justlevine: Done in https://github.com/WordPress/ai/pull/268/commits/c76caf98f1d6d196212f7ab2874b541bde5997b2

dkotter: So unless I'm just not seeing it, still seeing the `todo` statement and `function_exists` check here and not seeing that as part of https://github.com/WordPress/ai/pull/268/changes/c76caf98f1d6d196212f7ab2874b541bde5997b2

justlevine: So sorry @dkotter , I failed to notice that I created this PR from an upstream branch but was locally tracking my origin fork. Repushed in https://github.com/WordPress/ai/pull/268/commits/e1e3446336ead2862edf011a8d50fc1966bd0358

includes/Requirements.php
```diff
@@ -116,6 +116,10 @@ private function get_requirements(): array {
 				},
 				'error_message' => static fn() => esc_html__( 'The plugin assets are not built. This is most likely because you downloaded the plugin from the GitHub repository without building the assets. Please run `nvm use && npm ci && npm run build` to build the assets.', 'ai' ),
 			),
+			'ai_support' => array(
+				'check'         => static fn() => wp_supports_ai(),
+				'error_message' => static fn() => esc_html__( 'Your WordPress environment has AI functionality disabled. The AI plugin will not work until AI support is enabled.', 'ai' ),
```
dkotter, line 121: Tested and this does work as expected but getting the following `doing it wrong` message because of this line:

> Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the ai domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. (This message was added in version 6.7.0.)

justlevine: Good catch! This is an uncaught bug in the Requirements class unrelated to this. PR incoming to fix.

justlevine: - https://github.com/WordPress/ai/pull/453

dkotter (approved):

PR #271, opened 2026-03-05
WP 7.0: Remove dependencies and rely on what's in core

### LINKED ISSUE #244: Pathway to 7.0

dkotter:
> [!NOTE]  
> This is the first PR in a set of three. #272 and #275 will come next once this is merged

## What?

Partially closes #244

Instead of including a number of dependencies, use what is in WordPress 7.0

## Why?

This is the first PR on the path of adding full support for WordPress 7.0 and the new AI Client that is included there. WP 7.0 will include the AI Client and so we no longer need to bundle that.

## How?

- Removes the now unused `automattic/jetpack-autoloader`, `wordpress/mcp-adapter` and `wordpress/wp-ai-client` dependencies
- Updates other dependencies to their latest
- Remove the initialization of the AI Client
- Switch back to using the standard composer autoloader
- Temporarily ignore PHPStan errors around the AI Client

### Use of AI Tools

None

## Testing Instructions

This is the first PR in a series that will bring full support to WP 7.0. As such, things aren't in a fully usable state yet. If you test this PR, you should be able to install/activate the plugin on both WP 6.9 and WP 7.0 (no fatals in either) but the plugin won't fully work in either yet

> [!NOTE]  
> Unit and E2E tests are failing. Those will be fixed up once full 7.0 support is in place.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-271-762de8dec6e1cb102828aedd68333e45a0425d05.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/bootstrap.php
```diff
@@ -187,20 +186,20 @@ function load(): void {
...
 
-	// Load the Jetpack autoloader.
-	if ( ! file_exists( AI_EXPERIMENTS_PLUGIN_DIR . 'vendor/autoload_packages.php' ) ) {
+	// Load the Composer autoloader.
+	if ( ! file_exists( AI_EXPERIMENTS_PLUGIN_DIR . 'vendor/autoload.php' ) ) {
 		add_action( 'admin_notices', __NAMESPACE__ . '\display_composer_notice' );
 		return;
 	}
-	require_once AI_EXPERIMENTS_PLUGIN_DIR . 'vendor/autoload_packages.php';
+	require_once AI_EXPERIMENTS_PLUGIN_DIR . 'vendor/autoload.php';
```
felixarntz: Why is that still needed? As far as I can tell in `composer.json`, there are no remaining production dependencies left? If so, always safer to not load any Composer stuff at all in a WordPress plugin.

CREDITS.md
```diff
@@ -16,12 +16,4 @@ Thank you to all the people who have already contributed to this repository via
...
-The following software libraries are utilized in this repository.
-
-[Abilities API](https://github.com/wordpress/abilities-api) by [WordPress](https://github.com/WordPress) and licensed [GPL-2.0-or-later](https://github.com/WordPress/abilities-api/blob/trunk/LICENSE.md).
-
-[Autoloader for Jetpack](https://github.com/automattic/jetpack-autoloader) by [Automattic](https://github.com/Automattic) and licensed [GPL-2.0-or-later](https://github.com/Automattic/jetpack-autoloader/blob/trunk/LICENSE.txt).
-
-[MCP Adapter](https://github.com/wordpress/mcp-adapter) by [WordPress](https://github.com/WordPress) and licensed [GPL-2.0-or-later](https://github.com/wordpress/mcp-adapter/blob/trunk/LICENSE.md).
-
-[WordPress AI Client](https://github.com/wordpress/wp-ai-client) by [WordPress](https://github.com/WordPress) and licensed [GPL-2.0-or-later](https://github.com/WordPress/wp-ai-client/blob/trunk/LICENSE.md).
+n/a
```
felixarntz: I would just remove this entirely.

phpstan.neon.dist
```diff
@@ -22,6 +22,10 @@ parameters:
 		paths:
 				- ai.php
 				- includes/
+		# Temporary while WP 7.0 beta AI Client migration is in progress.
+		ignoreErrors:
+			- '#WordPress\\AI_Client\\#'
+			- '#WordPress\\AiClient\\#'
```
felixarntz, line 28: I don't follow why this is here - these aren't part of the plugin anymore anyway as part of this PR, right?

felixarntz (commented):
@dkotter Mostly looks good! A few comments.

includes/bootstrap.php
dkotter: We're not using this to load dependencies, we're using it to load our own classes and files. If we remove this, would need to be replaced with a custom autoloader

phpstan.neon.dist
dkotter: This PR is the first in a set of three. At this point, the code still references both of these and without us loading these as dependencies, PHPStan will throw errors (as it isn't aware of these yet). In the last PR of the set, I remove the use of `AI_Client` entirely and so that goes away but we still use `WordPress\AiClient` in a few places (for instance, using `WordPress\AiClient\Providers\Http\DTO\RequestOptions` to set a higher timeout when generating images)

CREDITS.md
dkotter: Removed https://github.com/WordPress/ai/pull/271/changes/e4556b37061d36edec2c33565f57c8599a460b08

includes/bootstrap.php
felixarntz: I think we should do that then. A lot safer than using Composer autoloader in a plugin.

Writing a custom autoloader for a plugin's own classes that use PSR-4 is super trivial, plus it ends up a lot lighter than all of Composer's autoloader boilerplate, which is overkill for this more trivial usage.

You can pretty much copy what we have in the provider plugins, e.g. https://github.com/WordPress/ai-provider-for-anthropic/blob/trunk/src/autoload.php

dkotter: Works for me. Added that in https://github.com/WordPress/ai/pull/271/changes/762de8dec6e1cb102828aedd68333e45a0425d05

felixarntz (approved):
Thanks @dkotter!

justlevine:
Hey @dkotter - I'm curious about the decision to replace the Composer autoloader with our own `spl_autoload_register()` here. 

Was it just because we're no longer requiring any production dependencies, or are there other considerations? Don't see it called out in the ticket.

If that was the only reason, then I'd suggest we restore it, we anyway have a build step, and a prebuilt classmap is always going to be work better than live discovery (even if we microptimize by e.g. dropping the extra filesystem call with something like `if ( ! @include $file ) { return; }` )

dkotter:
> Hey @dkotter - I'm curious about the decision to replace the Composer autoloader with our own `spl_autoload_register()` here.

I don't feel strongly either way, this was a change request on this PR, see https://github.com/WordPress/ai/pull/271#discussion_r2896259167

includes/bootstrap.php
justlevine: > plus it ends up a lot lighter than all of Composer's autoloader boilerplate, which is overkill for this more trivial usage.

@felixarntz I know I'm late to the party but are you sure about this? my understanding is that any precompiled classmap (Including generated by Composer) is going to be more efficient than a live-check like that `spl_autoload_register_loop()`. And we already need a build step for the JS, so it's not like we're adding overhead.

felixarntz: I'm fine using the class map that Composer generates, but we shouldn't use the actual Composer autoloader. It remains conflict material and is certainly overkill for a simple autoloader like we need here.

FWIW Site Kit (which I worked on and is used by >5M sites) uses this approach, so it works well at scale.

But I don't feel strongly about whether we should use a class map or a dynamic autoloader. With the current size of the codebase here, it shouldn't make any practical difference. We could also start without a class map, and later add one. No need to block this.

PR #286, opened 2026-03-09
Migrate credentials from the AI Credentials setup to the new Connectors setup

### LINKED ISSUE #244: Pathway to 7.0

dkotter:
## What?

Partially closes #244

Migrates credentials to the new Connectors format

## Why?

When using the bundled WP AI Client, credentials were stored under a single option and used on the AI Credentials screen. With the version of the AI Client that is in WP 7.0, there's a new Connectors setting screen and it stores credentials under separate options. We want to ensure existing users don't those their credentials so we migrate those over.

## How?

- Add a new Credential Migration class
- Initiate this class early in our load process
- If coming from a version less than 0.5.0 (assuming this will be our next version) it will check if we have existing credentials and migrates those to the new format
- Adds unit tests

### Use of AI Tools

Used Claude Code running Sonnet 4.6 to plan the migration approach and write most of the code (including tests). Fixed a few issues myself after manual testing

## Testing Instructions

1. Install the current released version of the plugin (v0.4.1)
2. Add credentials for Anthropic, Google and OpenAI under `Settings > AI Credentials`
3. Update the plugin to this PR
4. Load a screen in the admin to trigger migration
5. Ensure the `wp_ai_client_provider_credentials` option no longer exists
6. Ensure these three options do exist: `connectors_ai_anthropic_api_key`, `connectors_ai_google_api_key`, and `connectors_ai_openai_api_key`
7. Ensure you see these settings reflected on the new `Settings > Connectors` screen

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-286-3d58538c68b5834f7e944a4ddb1a2291539b3ad8.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
Testing results....

Here's the AI Credentials screen on WP 7.0-beta3 and AIEx v0.4.1 before applying this PR:
Credentials-before-PR

Here's the same screen after applying this PR:
Credentials-after-PR

Here's the same screen after clicking `Save Changes`:
Credentials-after-save

Here's the Connectors screen after applying this PR and clicking `Save Changes` on the AI Credentials screen:
Connectors-after-PR

And here's the Connectors screen after installing the OAI provider plugin:
Connectors-after-OAI-install

Some thoughts...
1. Should we add some text on the AI Credentials screen to direct folks to the Connectors page?  Perhaps "Credentials for AI providers has moved to the Connectors settings page."?  Probably worth a @todo message to remind us to come back and remove the AI Credentials page at some point as it'll confuse things to have that still around when Connectors is the primary/only route to add those in.
2. Additionally to the above, after someone clicks `Save Changes` perhaps we redirect them to the Connectors page with a message there on which provider plugins to install based on their previously entered credentials?
3. I'm not seeing the credentials carry over (had expected OAI to show valid credentials after installing that provider), is that something we can ensure works via this PR and when someone installs the right provider plugin?

dkotter:
@jeffpaul Looking at those screenshots, I think you might have missed a step in testing here. That AI Credentials screen came from the old WP AI Client. That screen will no longer exist going forward for anyone running the latest version of this plugin (which those changes are also in this branch). So if you're seeing that screen, I think you're still running v0.4.1 instead of this branch.

It's tricky here to test as you need to follow these steps:

1. Install WP 6.9
2. Install the latest version of this plugin from .org, version 0.4.1
3. Add credentials
4. Update your environment to WP 7.0-beta3
5. Update/replace the plugin so it's running this branch, which will include all the changes merged into `develop` that remove the AI Client
6. At this point, the AI Credentials screen won't exist anymore and credentials should be migrated

In regards to your questions:

> Should we add some text on the AI Credentials screen to direct folks to the Connectors page?

> Additionally to the above, after someone clicks Save Changes perhaps we redirect them to the Connectors page with a message there on which provider plugins to install based on their previously entered credentials?

As mentioned above, this screen no longer exists so nothing to add or change here.

> I'm not seeing the credentials carry over (had expected OAI to show valid credentials after installing that provider), is that something we can ensure works via this PR and when someone installs the right provider plugin?

Just tested again and this did work for me, both in scenarios where an AI Provider plugin was already installed and active and when the plugin wasn't active / installed but was activated / installed as part of the Connectors flow. I think it's not working for you due to the above note around ensuring you're testing with this branch.

jeffpaul (approved):
Ok, ran through things again and can't seem to get to the AI Credentials screen with that `Save Changes` button and things work as expected.

Nice touch updating the Credentials row action from the Plugins list to Connectors!

PR #288, opened 2026-03-09
Update plugin icon and banner assets

shaunandrews:
A new Icon and Banner for the AI plugin. We'll also be using this same visual treatment for a callout on the new Connectors screen in WordPress 7.0

### Banner (772×250)
banner

### Icon (256×256)
icon

Here's what this would look like on the new Connectors screen:

image

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-288-c803f5ff4e5c129361f8d0f32ee0541408f9205f.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


.wordpress-org/icon.svg
jeffpaul, line 1: @shaunandrews not sure how much it matters, but the current beaker version presents as 256x256 and the new bits version presents as 128x128.

juanfra: I believe the size shouldn't affect display since it's svg, though keeping it at 256x256 for consistency might be worth it.

shaunandrews: Gotcha. I just pushed an updated SVG.

jeffpaul:
I'm still concerned about the bits floating apart from the WP logo and that being interpreted as AI destroying or breaking up WordPress.  I'm going to punt this to 0.7.0 to allow for additional time and design input here, noting we have until WP 7.0 release day on April 9th to ideally come to a conclusion if there's going to be a change to banner/icon.

jeffpaul:
@shaunandrews any chance you've got other design explorations on icon/banner that don't potentially give a perception of "AI destroying WP"?  I'm open to, and frankly excited about, updating the plugin assets alongside WP 7.0 when we expect to get a LOT more site owner attention, but I want to be VERY CAREFUL to ensure perception is as optimistic and positive about trying out the AI plugin's features and experiments.

juanfra:
@jeffpaul @shaunandrews I've noticed the connectors page already has the plugin callout using the new design (RC2). Since we're getting closer to the 7.0 release, might be good to decide on this soon.

jeffpaul:
I'm still of the opinion that the designs here are a potential detractor from the message we're trying to imply with the AI plugin, so not intending to update to the existing designs but open to a change in banner/icon to ones that don't potentially have folks inferring that the plugin is a negative for the WP project.

jeffpaul:
closing In favor of #546 and crediting folks there for their work here

Issue #251, opened 2026-02-20
Add “Suggest Refinements” to create AI review Notes for recommended improvements


jeffpaul:
## What problem does this address?

Editors and Authors need a fast, consistent way to identify improvements (accessibility, readability, grammar, spelling, SEO, and [Content Guidelines](https://make.wordpress.org/ai/2026/02/03/content-guidelines-a-gutenberg-experiment/) alignment) without AI directly rewriting content.  Today, there’s no unified “review pass” that surfaces actionable recommendations inside the existing Notes workflow.  Let's look to add an experiment that would have AI process the content and suggest refinements via Notes comments directly within the post.  This could then pair nicely with #250 where an Editor or Author responds to / resolves those notes and then has AI apply those refinements from Notes.

## What is your proposed solution?

Add a “Suggest Refinements” action in the editor, probably not withing Notes as none might exist so probably somewhere in the Post sidebar / top toolbar?

On click:

- Send post content (and relevant context like title/excerpt) to the AI Client.
- Ask the model to produce suggested improvements as discrete, actionable Notes.

Categories to include:

- Accessibility: missing alt text, heading structure, link text
- Readability: simplify difficult language, break up long paragraphs, fix run-on sentences
- Grammar: spelling, punctuation, consistency
- SEO: meta description gaps/suggestions, clarity of topic phrasing
- Guidelines: check alignment with site Content Guidelines when present

I'm not sure if we should explicitly look to request these sorts of things in the prompt or allow for the chosen provider/model to decide what to review and suggest refinements from?

Notes should attach to the relevant block when possible; otherwise attach at the post level.

Issue #282, opened 2026-03-07
Chat experiment: Integration outside the editor and outside single-task AI use


jeffpaul:
While many MVP features (title, excerpt, alt text) live within the post editor, we also need to consider how site owners may want to use AI in more conversational or exploratory ways.  This could take the form of:

- A global “AI Workspace” screen in wp-admin, styled after the modern Site Editor, where users can engage with a chatbot-style interface.
- The ability to run multi-step AI interactions that go beyond a single field or content object (e.g., “help me configure my site navigation,” or “generate a campaign outline across multiple posts”).
- Using upcoming WordPress admin design directions such as DataViews or DataForms to make this workspace feel like a natural part of core rather than a standalone tool.
- Allow for this experience to be triggered both outside and inside the site/block editors.

This approach provides a bridge between single-field helpers and broader conversational AI workflows, giving users more freedom in how they experiment with AI.  It also opens a design question: how to ensure consistency between editor-embedded features and this global workspace so users don’t feel like they are moving between two different products.

Expected usage here could range from the simple “Please generate sample titles for my latest post” all the way to more complex requests like “Build a custom 404 page that leverages an URL parameters or invalid slugs to find the 3 most likely posts being searched for, render those in nice post card layouts, after that show an option to display our site search experience in case the rendered results above are not what they were searching for, and then finally provide a listing of my 5 most popular posts in the last 24hrs”.

### Feature Overview & Value

- **Summary:** This feature introduces a dedicated, conversational interface within the WordPress admin: the "AI Workspace”, where users can engage in multi-step, context-aware interactions with AI.
- Unlike the task-specific tools (like Title Generation), this workspace allows for open-ended exploration, content planning, and site-wide operations.
- It has context awareness, meaning AI has read-access to the site’s content, structure, and taxonomy, allowing it to provide answers grounded in the actual reality of the user's WordPress installation.

### Use Cases

Users often need to "zoom out" to plan content calendars, analyze gaps in their existing library, or generate complex code snippets (like HTML patterns) that don't fit into a single block. This workspace bridges the gap between discrete editorial tasks and broader site management.

- **Strategic Content Planning (Context-Aware):** "Look at my last 10 published posts. What are three related topics I haven't covered yet that would fit my site's specific tone?"
- **Multi-Post Generation:** "I need to launch a series on 'Remote Work.' Please generate titles and short excerpts for a 5-part series, and create draft posts for each."
[...]

PR #289, opened 2026-03-10
Implement Refine from Notes batch processing and progress indicator

### LINKED ISSUE #250: Add “Refine from Notes” action to update post content based on editorial feedback

Authors and Editors often leave review feedback using the Notes feature, but the work to respond to those comments requires manual edits.  This slows down iteration and creates friction in editorial workflows.

Add a “Refine from Notes” button in the post editor when Notes are present, perhaps within the Notes tray/sidebar as well as in the archive Notes view.

On click:

- Collect active Notes probably filtering to unresolved Notes only.
- Send Notes + current post content to the AI Client.
- Return a proposed revised version of the post content.

Present output as:

- A diff view
- A side-by-side comparison with “Replace content” and “Insert changes selectively” options
- Yolo into it and apply those edits (Revisions can allow for rollbacks here assuming we trigger an autosave after each/all AI update)

Do not auto-publish changes; always require user confirmation.

zeus2611:
## What?
Closes #250 
See #251 

This PR introduces the new "Refine from Notes" feature, enabling users to automatically apply pending editorial feedback/notes to their content using AI.

## Why?
Feature request #250 proposed allowing users to generate automated refinement updates based heavily on unresolved editorial notes attached to blocks. This enables a smoother editing workflow where suggestions can be adopted contextually instead of applied manually one-by-one. 

We matched the batch UX already provided by "Review Notes" and supply users with a safe, native diff interface via WordPress Revisions to examine and accept/rollback any generated AI changes before publishing.

## How?
- Created `useRefineNotes` hook to loop comprehensively through the document blocks, matching pending and threaded `ExistingNote` collections to associated parent `clientIds`.
- Utilized an asynchronous batch progression of 4 block items per network request.
- Published surface-level progress data to a new `RefineNotesPlugin.tsx` component in the sidebar to dynamically step the button text (e.g. `Refining block (1 of 7)...`).
- Injected an `autosave()` trigger to establish a clean DB checkpoint after processing.
- Exposed a "Review in Revisions" action to the success snackbar notification linking directly to `revision.php?revision=XXX`.

### Use of AI Tools
Implementation was planned and executed with assistance from Gemini (Antigravity). All code was reviewed, tested, and validated manually in a local WordPress environment.

## Testing Instructions
1. Run and build the plugin locally on WP. Ensure AI experiments are enabled.
2. Create a Post and generate some blocks.
3. Attach multiple Editorial Notes/Comments across various blocks suggesting changes (e.g. "Fix capitalization", "Make this sound professional").
4. Click the "Refine from Notes" button from the main Plugin Sidebar.
5. Verify it updates the button text to show progression through the blocks.
6. Check that the snackbar notification appears with a "Review in Revisions" action upon completion.
7. Click "Review in Revisions" and verify the WP Core diff viewer accurately shows the before & after changes.

1. Navigate via `Tab` to the Plugin Sidebar -> Refine from Notes button.
2. Hit `Enter` to start the action.
3. Once completed, `Tab` over to the auto-focused snackbar notification to click "Review in Revisions".

## Screenshots or screencast
Screenshot 2026-03-12 at 11 12 39 PM
Screenshot 2026-03-12 at 11 07 19 PM
Screenshot 2026-03-12 at 11 07 25 PM
Screenshot 2026-03-12 at 11 07 46 PM
Screenshot 2026-03-12 at 11 13 24 PM

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-289-a504d3902daf96ddd916fc43bc71f7a16ef873c6.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
@zeus2611 I see this is still in draft, but PR description seems like it might be ready for review/testing; mind confirming if this is ready or still in progress?

zeus2611:
Hey @jeffpaul — the core feature is complete and working as expected. I'm currently working on improving the e2e test coverage before marking this ready for review. Should have that wrapped up shortly and will convert from draft then!

zeus2611:
Hey @dkotter — the implementation is complete and ready for review. Marking this as ready and removing the Draft status.

jeffpaul:
@zeus2611 one thing that I think would be a great extension of this in a future, follow-up, PR is leveraging Real Time Collab and having a new `WordPress AI` (or just `AI`) user on the site and showing it within the editor interface making these changes.  I'll open an issue to document that, but will otherwise review/test this issue without that in mind.

zeus2611:
Sounds great, looking forward to the scope of the feature.

One thing I've been thinking about while building this: the individual features we're shipping (Review Notes, Refine from Notes, etc.) are really just discrete steps in a larger editorial workflow. There's an opportunity to tie these together as an agentic flow — e.g., a single "Review & Refine" pipeline where the AI first surfaces annotations as editorial notes and then automatically applies accepted suggestions in one pass, rather than requiring two separate manual actions.

I think this connects well to #282 — if the Chat Workspace is going to be the home for multi-step AI interactions, these editor-native workflows could be surfaced there as named agent actions ("Refine my draft", "Run editorial review"), giving users a coherent experience whether they're acting from the sidebar or the chat interface. Worth keeping in mind as the #282 design takes shape.

jeffpaul:
> There's an opportunity to tie these together as an agentic flow — e.g., a single "Review & Refine" pipeline where the AI first surfaces annotations as editorial notes and then automatically applies accepted suggestions in one pass, rather than requiring two separate manual actions.

I'm not certain I want to design a YOLO-mode into this by default, I'd prefer a human step in between the Review and Refine steps where an author or editor could perhaps resolve certain Review Notes that they don't want actioned and then have the AI continue on with the Refine steps.

docs/experiments/refine-notes.md
```diff
@@ -0,0 +1,177 @@
+# AI Refine from Notes
+
+## Summary
+
+The AI Refine from Notes experiment enables users to automatically apply pending editorial feedback/notes to their WordPress post content using AI. Clicking "Refine from Notes" in the post sidebar triggers the AI to contextually examine all blocks with pending Notes and modify the text according directly to the provided suggestions.
```
dkotter:
```suggestion
The AI Refine from Notes experiment enables users to automatically apply pending editorial feedback/notes to their WordPress post content using AI. Clicking "Refine from Notes" in the post sidebar triggers the AI to contextually examine all blocks with pending Notes and modify the text according to the provided suggestions.
```

includes/Experiments/Refine_Notes/Refine_Notes.php
```diff
@@ -0,0 +1,100 @@
...
+	exit;
+}
+
+/**
+ * AI Refine from Notes experiment.
+ *
+ * Adds functionality to apply AI-generated refinements based on editorial
+ * feedback (Notes) left on individual blocks.
+ *
+ * @since 0.5.0
```
dkotter: We'll want to update all of these `@since` statements to be `x.x.x` so we can replace those when doing a release (as it may not be in the 0.5.0 release)

includes/Experiments/Refine_Notes/Refine_Notes.php
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since 0.5.0
+	 */
+	protected function load_experiment_metadata(): array {
+		return array(
+			'id'          => 'refine-notes',
+			'label'       => __( 'Refine from Notes', 'ai' ),
+			'description' => __( 'Update post content based on editorial feedback from Notes.', 'ai' ),
```
dkotter:
```suggestion
			'description' => __( 'Analyze feedback that has been left via Notes and apply edits where needed.', 'ai' ),
```

includes/Experiments/Refine_Notes/Refine_Notes.php
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since 0.5.0
+	 */
+	public function register(): void {
+		add_action( 'wp_abilities_api_init', array( $this, 'register_abilities' ) );
+		add_action( 'enqueue_block_editor_assets', array( $this, 'enqueue_assets' ) );
+
+		register_meta(
```
dkotter: Is this needed here? I know we register this in the Review Notes Experiment and use that to determine which Notes were AI generated. Since this Experiment is consuming Notes and not creating them, doesn't seem like we need this here

tests/e2e-request-mocking/e2e-request-mocking.php
```diff
@@ -43,9 +43,11 @@ function ai_e2e_test_request_mocking( $preempt, $parsed_args, $url ) {
 	if ( str_contains( $url, 'https://api.openai.com/v1/responses' ) ) {
 		$body = $parsed_args['body'] ?? '';
 
-		// Route review-notes requests to their own fixture.
+		// Route review-notes and refine-notes requests to their own fixture.
 		if ( is_string( $body ) && str_contains( $body, 'Category guidance by block type' ) ) {
 			$response = file_get_contents( __DIR__ . '/responses/OpenAI/review-notes-responses.json' );
+		} elseif ( is_string( $body ) && str_contains( $body, 'You are an editorial assistant for WordPress. Your task is to update a single block' ) ) {
+			$response = file_get_contents( __DIR__ . '/responses/OpenAI/refine-notes-responses.json' );
```
dkotter, line 73: Curious if we need different responses for these two or if we can simplify this to just use the same response file? I haven't compared the two but I'm assuming they are basically the same with just different assistant/content responses. If that's the only difference, I think you can just use the existing file and update the tests to look for those responses. Basically the idea here is we don't care what the actual output from the LLM is, we just want to mock a real API request to test things

src/experiments/refine-notes/components/RefineNotesPlugin.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * RefineNotesPlugin component.
+ *
+ * Renders a "Refine from Notes" button in the post status info panel
+ * when unresolved Notes exist.
+ */
+export default function RefineNotesPlugin() {
+	const { isRefining, progress, total, hasPendingNotes, runRefinement } =
+		useRefineNotes();
+
+	if ( ! ( window as any ).aiRefineNotesData?.enabled ) {
```
dkotter: This was already checked when registering the plugin. I'd either remove this or remove the other one, no need for a double check

src/experiments/refine-notes/components/RefineNotesPlugin.tsx
```diff
@@ -0,0 +1,83 @@
...
+		return null;
+	}
+
+	// Always render the container, but hide the button if no notes,
+	// or maybe render it disabled if no notes? Let's hide it if no pending notes
+	// based on the issue description, "when Notes are present".
```
dkotter: I think we can update this comment as it seems the question it poses was fixed?

src/experiments/refine-notes/components/RefineNotesPlugin.tsx
```diff
@@ -0,0 +1,83 @@
...
+		? sprintf(
+				/* translators: 1: Current block number, 2: Total number of blocks. */
+				__( 'Refining block (%1$s of %2$s)…', 'ai' ),
+				progress,
+				total
+		  )
+		: __( 'Refine from Notes', 'ai' );
+
+	const buttonDescription = __(
+		'Automatically updates blocks using unresolved editorial feedback Notes.',
```
dkotter: Not sure we need to say `editorial` here as Notes could really be anything
```suggestion
		'Automatically updates blocks using unresolved feedback Notes.',
```

src/experiments/refine-notes/hooks/useRefineNotes.ts
```diff
@@ -0,0 +1,496 @@
...
+ * Internal dependencies
+ */
+import {
+	flattenBlocks,
+	getBlockText,
+	replaceBlockWithPlaceholder,
+} from '../../../utils/blocks';
+import { runAbility } from '../../../utils/run-ability';
+
+// Include the same reviewable types to safely get their attributes.
```
dkotter: Since this is now duplicated, we should probably extract this out to a shared export so it can be used in both the Review and Refine experiments

src/experiments/refine-notes/hooks/useRefineNotes.ts
```diff
@@ -0,0 +1,496 @@
...
+	[ key: string ]: unknown;
+}
+
+/**
+ * Fetches all pending Notes for a given post.
+ *
+ * @param postId The ID of the post to fetch Notes for.
+ * @return An array of pending Notes.
+ */
+async function fetchPendingNotes( postId: number ): Promise< ExistingNote[] > {
```
dkotter: This is basically a duplicate of the `fetchAllNotesByStatus` function in the `useReviewNotes.ts` file. I'd suggest we extract that out to a shared utility and then use that here

src/experiments/refine-notes/hooks/useRefineNotes.ts
```diff
@@ -0,0 +1,496 @@
...
+}
+
+/**
+ * Returns a bounded context window around a placeholder token.
+ *
+ * @param content     The full content with placeholder.
+ * @param placeholder The placeholder token.
+ * @return A truncated content window centered around the placeholder.
+ */
+function buildContextWindow( content: string, placeholder: string ): string {
```
dkotter: This also appears to be a duplicate of the existing `buildContextWindow`, so we can extract that out from `useReviewNotes.ts` to avoid code duplication

src/experiments/refine-notes/hooks/useRefineNotes.ts
```diff
@@ -0,0 +1,496 @@
...
+			 ).getBlocks() as Block[];
+			const flatBlocks = flattenBlocks( allBlocks );
+
+			// Fetch pending Notes for this post.
+			const pendingNotes = await fetchPendingNotes( postId );
+
+			if ( pendingNotes.length === 0 ) {
+				( dispatch( noticesStore ) as any ).createNotice(
+					'info',
+					__( 'No pending notes found to refine.', 'ai' ),
```
dkotter: I think we should use capital Note anywhere that is referenced, as I believe that's how Core refers to them
```suggestion
					__( 'No pending Notes found to refine.', 'ai' ),
```

src/experiments/refine-notes/hooks/useRefineNotes.ts
```diff
@@ -0,0 +1,496 @@
...
+				// post entity), so the editor store's revision data is stale after autosave.
+				const currentPostType = (
+					select( editorStore ) as any
+				 ).getCurrentPostType() as string;
+				const postTypeData = await (
+					resolveSelect( coreStore ) as any
+				 ).getPostType( currentPostType );
+				const restBase =
+					( postTypeData?.rest_base as string ) ??
+					`${ currentPostType }s`;
```
dkotter: Is there a better way to set this, maybe on the PHP side and passed in via the inline script? Not sure if `postTypeData.rest_base` is ever not set but if so, seems like adding an `s` to the post type will likely fail

includes/Abilities/Refine_Notes/Refine_Notes.php
```diff
@@ -0,0 +1,267 @@
...
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'The content of the block to refine.', 'ai' ),
+				),
+				'notes'         => array(
+					'type'        => 'array',
+					'items'       => array(
+						'type' => 'string',
+					),
+					'description' => esc_html__( 'The editorial feedback notes to apply to the block.', 'ai' ),
```
dkotter:
```suggestion
					'description' => esc_html__( 'The feedback Notes to apply to the block.', 'ai' ),
```

includes/Abilities/Refine_Notes/Refine_Notes.php
```diff
@@ -0,0 +1,267 @@
...
+			'type'       => 'object',
+			'properties' => array(
+				'block_type'    => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'The block type, e.g. core/paragraph, core/heading.', 'ai' ),
+				),
+				'block_content' => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
```
dkotter: This will remove all HTML, which is fine when we're just reviewing the content of a block but in this case, we want to edit the content of the block. Ideally we would leave any valid HTML in place (like a `strong` tag for instance) and then allow that HTML to be returned after the refinement. Probably should change this to `wp_kses_post` and later when we add the content to our prompt, probably need to remove the use of `normalize_content`

tests/e2e/specs/experiments/refine-notes.spec.js
```diff
@@ -0,0 +1,494 @@
...
+		);
+
+		await page.reload();
+
+		// Ensure the sidebar is visible.
+		await editor.openDocumentSettingsSidebar();
+
+		// WordPress doesn't persist custom `metadata` keys (like `noteId`)
+		// through block serialization. After reload, re-inject the noteId
+		// so `refineableBlocks` can find the matching block.
```
dkotter, line 158: I'm curious if this causes issues in real testing, not just the E2E tests here? I haven't tested things yet myself, just want to ensure things don't break after we inject the refined results

tests/e2e/specs/experiments/refine-notes.spec.js
```diff
@@ -0,0 +1,494 @@
...
+					body: JSON.stringify( { id: noteId, status: 'approve' } ),
+				} );
+			} else {
+				await route.continue();
+			}
+		} );
+
+		// Intercept the ability endpoint and return a canned response so the test
+		// never reaches the PHP AI client (which requires model credentials that
+		// are not present in the e2e environment).
```
dkotter: Is this needed with our E2E plugin we add? That plugin intercepts all HTTP requests and allows us to return canned responses

tests/e2e/specs/experiments/refine-notes.spec.js
```diff
@@ -0,0 +1,494 @@
...
+		// Add a block to attach a note to.
+		await editor.insertBlock( {
+			name: 'core/paragraph',
+			attributes: {
+				content: 'This paragraph needs refinement.',
+			},
+		} );
+		await editor.saveDraft();
+
+		// Create a note via API and attach it to the block.
```
dkotter, line 59: There's a lot here. I guess if this works, that's fine, but wondering if there's a better approach, like using WP-CLI. We don't need to test if adding a Note works, as that's handled by Core, so if running a WP-CLI command is simpler, we can look to do that

dkotter (changes requested):
Haven't fully tested things yet, overall looking good but have a handful of things I think we need addressed, though let me know if there's any questions on these.

jeffpaul:
I'm not certain that the flow through the "Review in Revisions" toast message and then historical diff view for Revisions is the most optimal way to get folks through this feature.  I'd argue that immediately applying these updates to the post and allowing a user to leverage revisions to roll back any changes they don't like would be better.  Alternatively presenting each change for editor/author review where accepting applies the change and resolves the respective Note and rejecting the change skips the change and leaves the respective Note?

tests/e2e-request-mocking/e2e-request-mocking.php
zeus2611: The two abilities have different response formats — Review Notes returns structured JSON (`{"suggestions": [...]}`) with json_schema, while Refine Notes returns a plain string (updated block content). Separate fixtures are needed because the mock routing dispatches based on system instruction text and each expects a different response shape.

tests/e2e/specs/experiments/refine-notes.spec.js
zeus2611: I added this because the tests were failing locally with `prompt_builder_error: No models found that support text_generation for this prompt` — the PHP AI client fails before any HTTP call is dispatched, so the pre_http_request mock never fires. I also saw one CI E2E run fail for the same test without the `page.route()`. The page.route() intercept at the browser network level bypasses the PHP client entirely.

If the E2E plugin setup ensures valid credentials are configured during CI runs, I can update this accordingly. What's the preferred approach?

tests/e2e/specs/experiments/refine-notes.spec.js
zeus2611: No issues observed. The flow uses the standard updateBlockAttributes() API, and autosave() creates a revision checkpoint for rollback via "Review in Revisions".

tests/e2e/specs/experiments/refine-notes.spec.js
zeus2611: We're not testing note creation here — just setting up the prerequisite state for the Refine from Notes flow. But if preferred, I can update it to use WP-CLI for the comment creation combined with `page.evaluate()` for the block linkage (since clientId is a runtime-only value).

zeus2611:
The current implementation already applies changes directly to the post — the toast is just a convenience link. We could improve this by redirecting users straight to the revisions view after refinement completes, which would pair nicely with 7.0's block-aware visual diffing.

True per-block rollback isn't feasible yet since revisions are still whole-post snapshots — reverting one block's refinement would undo all others from that pass.

The accept/reject per-change approach would give the most granular control but requires a custom inline diff review UI. For this PR, would keeping "apply all + redirect to revisions" as the MVP and opening a follow-up issue for per-block accept/reject work?

docs/experiments/refine-notes.md
zeus2611: Updated.

includes/Experiments/Refine_Notes/Refine_Notes.php
zeus2611: Updated all `@since` tags to x.x.x across both PHP files.

includes/Experiments/Refine_Notes/Refine_Notes.php
zeus2611: Updated.

includes/Experiments/Refine_Notes/Refine_Notes.php
zeus2611: Agreed — this is already registered by the Review Notes experiment and Refine Notes only consumes notes, doesn't create them. Removed.

src/experiments/refine-notes/components/RefineNotesPlugin.tsx
zeus2611: Removed the redundant check. The experiment is already gated in index.tsx.

src/experiments/refine-notes/components/RefineNotesPlugin.tsx
zeus2611: Removed the stale comment.

src/experiments/refine-notes/components/RefineNotesPlugin.tsx
zeus2611: Updated.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: Extracted `REVIEWABLE_BLOCK_TYPES`, `fetchAllNotesByStatus`, `buildContextWindow`, and shared types (`ExistingNote`, `NoteStatus`) into `src/utils/notes.ts`. Both hooks now import from the shared utility.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: Extracted as `fetchAllNotesByStatus` in `src/utils/notes.ts`, parameterized by status.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: Extracted to `src/utils/notes.ts`.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: Done — capitalized "Notes" in all i18n strings.

includes/Abilities/Refine_Notes/Refine_Notes.php
zeus2611: Updated.

includes/Abilities/Refine_Notes/Refine_Notes.php
zeus2611: Changed `sanitize_callback` to `wp_kses_post` and removed `normalize_content()` from `block_content` in the prompt. The `context` field still uses `normalize_content()` since surrounding article content should be stripped to plain text for the AI.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: `rest_base` is reliably set by WordPress core for all built-in and registered post types — the + 's' fallback applies only to custom post types registered without' rest_base'. That said, localizing it from PHP via the inline script data is a cleaner approach. Happy to move this if you'd prefer.

tests/e2e/specs/experiments/refine-notes.spec.js
dkotter: So we don't make any valid API requests in our E2E tests. We capture any HTTP request and mock the responses (including the model responses)

> PHP AI client fails before any HTTP call is dispatched

I'd be curious on why this is? The client should be making requests to get all models from each configured provider and we do capture that request and return a mocked model response which should mean you never get that error message. But maybe something else is going on here that I'm not aware of?

> If the E2E plugin setup ensures valid credentials are configured during CI runs, I can update this accordingly. What's the preferred approach?

So none of our other E2E tests have this `page.route()` setup in place and they all work fine. I'm not opposed to doing this just seems like it's more complicated than it needs to be

tests/e2e/specs/experiments/refine-notes.spec.js
dkotter: I think this is fine, just seems fairly complex

includes/Experiments/Refine_Notes/Refine_Notes.php
dkotter: Looks like the test files have these as well so those will need updated

src/experiments/refine-notes/hooks/useRefineNotes.ts
dkotter: I think I would, especially since we already run `localize_script`, so passing that there will be easy

jeffpaul:
> The current implementation already applies changes directly to the post — the toast is just a convenience link. We could improve this by redirecting users straight to the revisions view after refinement completes, which would pair nicely with 7.0's block-aware visual diffing.

Ahh ok, yeah let's try linking to the visual revisions view instead of the text-based diff revisions view.

> True per-block rollback isn't feasible yet since revisions are still whole-post snapshots — reverting one block's refinement would undo all others from that pass.

For now that's probably fine.  I could see a state where each resolved Note gets a comment from "WordPress AI" noting the refinement is done and links to a visual revision reference to see that specific change (which would be we'd need to fire an auto-save after each individual Note refinement).

> The accept/reject per-change approach would give the most granular control but requires a custom inline diff review UI. For this PR, would keeping "apply all + redirect to revisions" as the MVP and opening a follow-up issue for per-block accept/reject work?

Yeah, that approach for this version totally makes sense.  I'll try to collect all the iterations noted in this PR and open a follow-up issue.

includes/Experiments/Refine_Notes/Refine_Notes.php
zeus2611: Updated both Refine Notes test files.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: Done — added rest_base to the localize_script data via get_post_type_object(). JS now reads from aiRefineNotesData.rest_base directly, resolveSelect removed.

tests/e2e/specs/experiments/refine-notes.spec.js
zeus2611: Removed the ability endpoint intercept — the PHP mocking plugin handles the downstream OpenAI call. Kept the /wp/v2/comments intercepts since those are browser→WordPress REST calls, not server→external HTTP, so pre_http_request doesn't cover them.

jeffpaul (dismissed):
Works well in my testing, so @dkotter feel free to merge when you're good from your testing/review

includes/Abilities/Refine_Notes/Refine_Notes.php
```diff
@@ -0,0 +1,267 @@
...
+				),
+				'block_content' => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'wp_kses_post',
+					'description'       => esc_html__( 'The content of the block to refine.', 'ai' ),
+				),
+				'notes'         => array(
+					'type'        => 'array',
+					'items'       => array(
+						'type' => 'string',
```
dkotter, line 54: Do we want to add a `sanitize_callback` here?

docs/experiments/refine-notes.md
```diff
@@ -0,0 +1,177 @@
...
+```php
+array(
+    'block_type'     => array(
+        'type'              => 'string',
+        'sanitize_callback' => 'sanitize_text_field',
+        'description'       => 'The block type, e.g. core/paragraph, core/heading.',
+    ),
+    'block_content'  => array(
+        'type'              => 'string',
+        'sanitize_callback' => 'sanitize_text_field',
```
dkotter:
```suggestion
        'sanitize_callback' => 'wp_kses_post',
```

src/experiments/refine-notes/hooks/useRefineNotes.ts
```diff
@@ -0,0 +1,388 @@
...
+									}
+								}
+							}
+						} catch ( e ) {
+							// eslint-disable-next-line no-console
+							console.warn(
+								`[AI Refine Notes] Failed to refine block ${ block.clientId }`,
+								e
+							);
+							throw e;
```
dkotter: I believe that throwing an error here will stop all processing, meaning if we have 12 Notes to process and Note 3 has an error, the rest of the Notes won't be processed. That may be what we want but also seems like continuing processing would be nice

src/experiments/refine-notes/hooks/useRefineNotes.ts
```diff
@@ -0,0 +1,388 @@
...
+					lastRevisionId = (
+						select( editorStore ) as any
+					 ).getCurrentPostLastRevisionId() as number | null;
+				}
+
+				const noticeActions = lastRevisionId
+					? [
+							{
+								label: __( 'Review in Revisions', 'ai' ),
+								url: `/wp-admin/revision.php?revision=${ lastRevisionId }`,
```
dkotter: I think hardcoding `wp-admin` here could cause issues on sites that have mapped the admin to a different URL

dkotter (changes requested):

includes/Abilities/Refine_Notes/Refine_Notes.php
zeus2611: Added a static closure that maps `sanitize_text_field` over each item in the array.

docs/experiments/refine-notes.md
zeus2611: Updated to `wp_kses_post` to match the actual code. Also added the `sanitize_callback` entry for the `notes` field.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: Removed the `throw e` — individual block failures are now logged via `console.warn` and processing continues with the remaining blocks.

src/experiments/refine-notes/hooks/useRefineNotes.ts
zeus2611: Passed `admin_url()` from PHP via `localize_script` and use it in JS for the revision link. No more hardcoded `/wp-admin/`.

dkotter (commented):
Code looks good to me, nice work here!

In testing this out, if I click on the link to view the revision in the notice that shows after processing Notes, I'm seeing two issues:

1. I get the "Are you sure, you'll lose unsaved changes" message from my browser. If I click ok in that, I do end up in the old revisions view but if I then go back to the post view, the changes are lost. Do we need to trigger a save after the Notes are processed? Otherwise changes are lost if they click on that revisions link
2. Looking at Jeff's feedback, seems he asked to link to the new revisions view. I'm still seeing the old one being linked to though

zeus2611:
Thanks for testing!

1. "Unsaved changes" warning/changes lost:
Good catch. We originally used `autosave()` intentionally to avoid AI directly saving changes, giving users a chance to review in revisions first. But without a way to selectively accept/reject individual changes, this creates the UX issue you described. We'll switch to `savePost()` for now — the granular accept/reject flow will come with #324.

2. Visual revisions view:
We looked into this after Jeff's feedback. The block-based visual diff is triggered via a private editor API (`setStylesPath('/revisions')`) locked behind `__dangerousOptInToUnstableAPIsOnlyForCoreModules` — inaccessible to external plugins. Waiting for WP 7.0 stable to see how this gets exposed publicly. Also tracked in #324.

zeus2611:
@dkotter All CI tests have passed successfully. I believe this is ready to be merged.

dkotter:
@zeus2611 I've pushed a few commits to bring things inline with recent changes we've made elsewhere. Overall things work but few more pieces of feedback:

1. If an error happens during refinement, we don't surface the specific error making it harder to figure out what went wrong. As an example, if I have no AI Connectors in place and I try and refine, I'll get the following error: `Refinement failed for all blocks.`. Ideally I should be getting the error message that comes from the Ability, in this case `Refinement generation failed. Please ensure you have a connected provider that supports text generation.`
2. We recently made [some changes](https://github.com/WordPress/ai/pull/357) to system instructions to try and ensure the content the LLM returns matches the language of the content it receives. I think we'll want a few changes here to support that, namely adding something similar to these new system instructions and also removing the extra context we add (the `What follows is surrounding article content, where the block being refined has been replaced...` bit). As detailed [here](https://github.com/WordPress/ai/pull/357#pullrequestreview-4068270533), the LLM will often prioritize this instead of matching the content language.

zeus2611:
@dkotter Ability errors now surface directly in the notice (e.g. "Please ensure you have a connected provider...") instead of the generic "Refinement failed for all blocks." Added language-matching rule to the system instruction and removed the English context wrapper per the #357 pattern. Also fixed the PHPUnit test broken by the develop merge (`Loader::init()`). All CI checks pass — for now, at least. Develop is merging faster than CI can run, so I might be stuck in an infinite rebase loop soon. 😄

dkotter (approved):
Thanks for all the work here @zeus2611!

PR #290, opened 2026-03-10
UX iteration for the title generation experiment

juanfra:
## What?

Improves the UX of the title regeneration modal by simplifying the selection interface and adding the ability to regenerate options without closing the modal.

## Why?

~The current flow has different textareas with selection buttons, which made the interaction feel heavier than it needed to be. On the other hand, users had to close and reopen the modal to get new suggestions. Creating unnecessary friction when none of the initial options worked.~

The current flow is auto-inserting the suggestion without giving the chance to "confirm" you like the result, and leaving no option but undoing if you don't like it.

## How?

- ~Simplified the selection UI to radio-style choice cards (inspired by shadcn's [ChoiceCard](https://www.shadcn.io/patterns/field-advanced-3) component, adapting to the block editor's design language)~
- Added a "Regenerate" button directly in the modal so users can go through suggestions without closing.
- ~Updated the modal title to "Select a title or regenerate for more options" to make the flow clearer.~
- Added proper loading states during regeneration (with disabled controls)
- ~Auto-selects the first option after regeneration~

## Testing Instructions

1. Create or edit a page.
2. Click on "Regenerate title" via title toolbar
3. Verify the modal opens with the suggestion
4. Confirm the suggestion shows in a textarea and you can insert it or regenerate.
5. Click "Insert" and verify the title is inserted, and the modal closed.

## Update Apr 10

After https://github.com/WordPress/ai/pull/341 was merged, the regeneration was limited to only one result and auto-inserting it. This PR was updated to show the suggestion in the modal (as before, instead of auto-inserting) with the option to insert or regenerate.

## Screenshots or screencast 

https://github.com/user-attachments/assets/ae8c153e-7ffb-4aeb-8b50-65c96d72533e

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-290-3a1769cb6000358fa028ef6bf941bdf3625cb5f8.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


juanfra:
I'll be fixing the tests.

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -240,7 +302,10 @@ export default function TitleToolbar(): JSX.Element | null {
 			</PostTypeSupportCheck>
 			{ isOpen && (
 				<Modal
-					title={ __( 'Select a title', 'ai' ) }
+					title={ __(
+						'Select a title or regenerate for more options',
```
dkotter: Personal opinion but not sure we need to be this verbose in the modal title. Should be pretty clear that a user can regenerate titles with the regenerate button so I'd suggest just leaving this as `Select a title` or `Select title`

src/experiments/title-generation/index.scss
```diff
@@ -0,0 +1,53 @@
...
+			margin-top: 3px;
+			flex-shrink: 0;
+			cursor: pointer;
+		}
+
+		.components-base-control {
+			flex: 1;
+		}
+
+		.components-textarea-control__input {
```
dkotter: My concern with these styles is while it makes it look nicer, it's now not obvious that it's a textarea you can click into to edit, which is one of the goals with this UI, provide the ability for someone to edit a suggested title prior to inserting it. Wondering if there's a way to have the UI look nice but still make it obvious the title is editable?

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -33,87 +33,119 @@ import type {
...
+			className={ [
+				'ai-title-generation-option',
+				isSelected && 'is-selected',
+				isDisabled && 'is-disabled',
+			]
+				.filter( Boolean )
+				.join( ' ' ) }
+			onClick={ ! isDisabled ? onSelect : undefined }
+		>
+			<input
```
dkotter: With the new OpenAI provider, it will only return a single result. In that scenario, this checkbox isn't needed. Wondering if this needs to be more flexible so we don't show a checkbox if we don't have multiple results?

src/experiments/title-generation/components/TitleToolbar.tsx
juanfra: Fair point. I was trying to add a more intentional title. Will go back to `Select a title`.

src/experiments/title-generation/index.scss
juanfra: Good question. I was actually wondering if we need the textareas to be editable here at all, that's why I intentionally styled them this way. 

My concern is it might create confusion. Not sure what the user expectation would be if they regenerate after they modified some suggestions. My thinking is that editing belongs in context, in the actual title field where it'll live, and the modal should really about choosing from suggestions. 

But maybe I'm missing something. Is there a specific reason we want editing to happen in the modal rather than after insertion?

src/experiments/title-generation/components/TitleToolbar.tsx
juanfra: Good point. If there's only one result, we probably don't need the radio at all.

src/experiments/title-generation/index.scss
dkotter: This was part of the initial spec for the experiment (see #10).

I can't speak directly to the reasoning behind that though I see both sides. For me personally, I'd want a flow of generating titles, choosing the one that matches closest to what I want and then editing it in the actual editor. But I do think there are others that will generate titles and if they don't see one that matches exactly and there isn't an obvious way to edit that first, they'll just discard those rather than inserting first and then editing it.

So the intention here is to not force editing to happen in this preview state but giving that as an option allows users to decide if they want to edit then insert or insert then edit.

dkotter:
Note we've now merged in #341, which made the decision to remove the generation of multiple titles. As such, the changes in this PR will no longer work. I've leaving this open for the moment to see if there's any feedback you may have @juanfra with this new approach.

juanfra:
Thanks for the ping @dkotter, sorry I missed that. Haven't had time to come back to this.

In terms of UX, I'm a bit concerned about auto-inserting the title. It doesn't give users a chance to preview or dismiss if they don't like the result. They'd have to undo, which feels pretty aggressive.

I think there's still value in having a modal where the user can confirm the suggestion. I feel the flow of "getting a suggestion > regenerate if needed > insert when you're happy" would give users more control and be more predictable. That said, it would be good to have a pattern that's followed in the different experiments (summary, excerpt)

Thoughts?

dkotter:
> In terms of UX, I'm a bit concerned about auto-inserting the title. It doesn't give users a chance to preview or dismiss if they don't like the result. They'd have to undo, which feels pretty aggressive.
> 
> I think there's still value in having a modal where the user can confirm the suggestion. I feel the flow of "getting a suggestion > regenerate if needed > insert when you're happy" would give users more control and be more predictable. That said, it would be good to have a pattern that's followed in the different experiments (summary, excerpt)
> 
> Thoughts?

Yeah, I agree that I like giving the user a clear opportunity to accept a suggestion instead of auto-inserting it. Though that said, this does match what we do in other areas (excerpt generation, summarization, featured image generation) so I don't think it's the worst UX.

But I'm definitely open to reviewing a different approach to this if you want to try tackling that. We're not looking to go back to rendering multiple results but I'm open to bringing back some sort of modal approach that allows the user to approve/re-generate as needed.

juanfra:
Thanks @dkotter. I iterated on what we have now.

Updated the PR to bring back the modal with a layout that adapts well to having only one option (changed the title of the modal as well) and giving the user the chance to insert, regenerate or exit the whole thing without changing (closing the modal).

https://github.com/user-attachments/assets/ae8c153e-7ffb-4aeb-8b50-65c96d72533e

dkotter (dismissed):

dkotter:
This looks good to me. Before merging, any thoughts on this new approach @jeffpaul?

jeffpaul:
Functionally, this isn't all that much different that the existing experience:

https://github.com/user-attachments/assets/883ce60b-a31f-48a1-b00e-ca93e8f57773

So question is a modal better than just directly updating the title, allowing editing there as normal, and relying on undo to roll back to before generation?

juanfra:
> Functionally, this isn't all that much different that the existing experience

@jeffpaul the radical difference is user control. With the modal the user can review and discard, regenerate, or insert suggestions. With auto-insert, they lose that control. The title just changes and their only option is undo. 

Another problem with the current flow, if the generation takes a few seconds and the user scrolls down or starts editing other content, they might not even notice when the title changes. They could lose their original title without realizing it. At minimum, if we're auto-inserting we should show a toast notice when the title changes.

 I believe the proposed changes are providing better UX, though I understand these things can be subjective.

jeffpaul:
So while I think I still prefer the existing flow, I'm open to shipping this iteration and gathering feedback.  That likely means we'd delay any proposal to elevate the Title Gen experiment to a feature, but that seems ok for now.

@juanfra care to resolve conflicts to consider this for the 0.8.0 release tomorrow?

dkotter (approved):

PR #287, opened 2026-03-09
Rename from `AI Experiments` to just `AI`

dkotter:
## What?

Partially closes #234

Standardize on using the name `AI` instead of `AI Experiments`

## Why?

In preparation for the release of WordPress 7.0, we'll be moving some "experiments" to "features" and as such, the labelling as `AI Experiments` doesn't make the most sense anymore.

## How?

- Update the plugin name to `AI` instead of `AI Experiments`
- Update any reference to the AI Experiments plugin to AI plugin. This includes documentation, code references, templates, etc
- Update other places that reference `AI Experiments` to reference just `AI`, like the settings menu

> [!NOTE]  
> I have not updated specific code references to experiments. For example, constants start with `AI_EXPERIMENTS_`. Options and hooks start with `ai_experiments`. We can update all of these if desired but not sure there's much value there.

### Use of AI Tools

None

## Testing Instructions

1. Checkout this branch
2. Activate the plugin
3. Click around to various screens and ensure all text references `AI` or `AI plugin` instead of `AI Experiments`

## Screenshots 

Plugin listing on the Plugins settings page showing the new plugin name

New settings menu

Content at the top of the AI settings screen

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-287-b2e8274e5a1d6546cfbd4c4f978b8b7c81cc3139.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #304, opened 2026-03-13
Add extensibility hooks for system instructions, ability results, and post context

### LINKED ISSUE #34: Audit Codebase for Strategic Hook Opportunities
### Description

As the plugin continues to evolve, it would be valuable to conduct a focused review of the entire codebase to identify opportunities for adding WordPress hooks.  This would improve extensibility, enable third-party integrations, and support experimentation without requiring direct modification of core plugin logic.

Some areas of the plugin may execute logic without providing filter or action hooks that would allow developers to modify behavior, inject custom logic, or observe experiment lifecycle events so checking for where those are missing and adding them in would be great.

Perform a structured audit of the codebase to:

- [ ] Identify key execution points where filters or actions would increase extensibility
- [ ] Evaluate experiment lifecycle events (registration, execution, completion, failure)
- [ ] Review provider integrations for override opportunities
- [ ] Assess admin UI rendering points
- [ ] Document hook naming conventions and placement guidelines

Deliverables:

- List of recommended new hooks with justification
- PR adding prioritized hooks
- Developer documentation outlining extension points

Ryujiyasu:
## Summary

Implements the extensibility hooks agreed upon in the #34 hook audit discussion with @dkotter.

### New Filters

**1. System Instruction Filter** (`Abstract_Ability`)
- `ai_experiments_system_instruction` — Filters the system instruction text before it's passed to the AI model. Receives the instruction string, ability name, and data array.

**2. Ability Result Filters** (`Abstract_Ability`)
- `ai_experiments_ability_result` — Generic filter applied to the result of any ability execution. Receives the result, ability name, and input.
- `ai_experiments_ability_result_{$name}` — Dynamic filter for a specific ability (e.g., `ai_experiments_ability_result_ai/title-generation`). Receives the result and input.

**3. Post Context Filters** (in the Abilities, per discussion)
- `ai_experiments_get_post_details` — Filters post details returned by the `ai/get-post-details` ability. Receives details array, post ID, and requested fields.
- `ai_experiments_get_post_terms` — Filters terms returned by the `ai/get-post-terms` ability. Receives terms array, post ID, and allowed taxonomies.

### Design Decisions

- **Result filtering is implemented via a `filtered_execute_callback` wrapper** in `Abstract_Ability` that wraps the abstract `execute_callback`. This avoids requiring changes to every concrete ability class while ensuring all abilities get consistent filtering.
- **WP_Error results are not filtered** — errors pass through unmodified.
- **Post context filters are in the Abilities only** (not in `get_post_context` helper), as agreed with @dkotter.

### Example Usage

```php
// Modify system instructions for all abilities.
add_filter( 'ai_experiments_system_instruction', function( $instruction, $name, $data ) {
    return $instruction . "\n\nAlways respond in formal tone.";
}, 10, 3 );

// Filter title generation results.
add_filter( 'ai_experiments_ability_result_ai/title-generation', function( $result, $input ) {
    // Remove any titles that are too long.
    if ( is_array( $result ) ) {
        $result = array_filter( $result, fn( $title ) => strlen( $title ) <= 100 );
    }
    return $result;
}, 10, 2 );

// Redact sensitive fields from post details before AI processing.
add_filter( 'ai_experiments_get_post_details', function( $details, $post_id, $fields ) {
    unset( $details['author'] );
    return $details;
}, 10, 3 );
```

Closes #34

## Test Plan

- [x] Verify existing ability tests still pass
- [x] Verify `ai_experiments_system_instruction` filter modifies system instructions
- [x] Verify `ai_experiments_ability_result` filter modifies ability results
- [x] Verify `ai_experiments_get_post_details` filter modifies post details
- [x] Verify `ai_experiments_get_post_terms` filter modifies post terms
- [x] Verify WP_Error results are not filtered

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-304-49f432ccccb47204f39e0e75610a2619fee98f53.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


gziolo:
The plugin is being renamed in https://github.com/WordPress/ai/pull/287 by @dkotter. There is a related feedback https://github.com/WordPress/ai/pull/287#pullrequestreview-3950694964 from @jeffpaul, which makes perfect sense here as well:

> Imay want to consider updating other parts of the codebase to remove experiment-named items to make things simpler to maintain in the future even if that's a backcompat break (as generally something that would be easier to do before 7.0 than after).

This applies to the names of the proposed hooks. Let's finalize the naming conventions here. Also note that in WP core, all AI related hooks are prefixed with `wp_`.

gziolo:
Thanks for working on this, @Ryujiyasu! I'm very much in favor of adding extensibility filters — this is exactly the kind of developer control the plugin needs.

However, after reviewing this alongside prior work on the Abilities API, I think some of these filters would be better suited for WP core rather than the plugin. Here's my analysis:

## Filters that should go to WP core

WP core already fires `wp_after_execute_ability` in `WP_Ability::execute()`, but as an action — not a filter — so developers can observe results but can't modify them. A core-level result filter would close that gap for all abilities.

**`ai_experiments_ability_result` and `ai_experiments_ability_result_{$name}`**

These are generic "filter the result of any ability execution" hooks — not AI-specific. The same concept was proposed in the Abilities API repo in the early phase of development:

- https://github.com/WordPress/abilities-api/pull/37 proposed `ability_execute_result` with the same signature (`$result`, `$ability_name`, `$input`), placed inside `WP_Ability::execute()`.
- https://github.com/WordPress/abilities-api/pull/150 proposed `wp_ability_validate_input` / `wp_ability_validate_output` for validation extensibility.

Placing result filtering in core (inside `WP_Ability::execute()`) would be better because:

1. It applies to **all** abilities, not just those extending `Abstract_Ability` from this plugin.
2. It sits in the proper lifecycle — after `do_execute()`, integrated with output validation — rather than requiring the `filtered_execute_callback` workaround that adds a public method as an implementation detail.
3. It avoids the current issue where filtered results that don't conform to the ability's output schema will cause `validate_output()` to fail silently.
4. The relationship between result filters and the existing core action `wp_after_execute_ability` needs to be designed intentionally at the core level, so developers know whether they receive the original result or a filtered one.

Now that Abilities API development has moved to core, these filters should be proposed there. I'd be happy to help with that. 

## Filters that belong in this plugin

**`ai_experiments_system_instruction`** — This is plugin-specific (system instructions are an AI concept with no core equivalent). Looks good as-is.

**`ai_experiments_get_post_details` / `ai_experiments_get_post_terms`** — These filter data within plugin-specific utility abilities, or could they be also superseeded by filters added directly?

includes/Abilities/Utilities/Posts.php
```diff
@@ -153,6 +153,17 @@ private function register_get_post_details_ability(): void {
 						$details['excerpt'] = $post->post_excerpt;
 					}
 
+					/**
+					 * Filters the post details returned by the get-post-details ability.
+					 *
+					 * @since 0.5.0
```
dkotter: Let's replace all `@since` statements with `x.x.x` as we're not sure of the actual version until we prep a release (and we have a step in our release process to find and replace all of those)

dkotter (changes requested):

dkotter:
> The plugin is being renamed in #287 by @dkotter. There is a related feedback [#287 (review)](https://github.com/WordPress/ai/pull/287#pullrequestreview-3950694964) from @jeffpaul, which makes perfect sense here as well:
> 
> > Imay want to consider updating other parts of the codebase to remove experiment-named items to make things simpler to maintain in the future even if that's a backcompat break (as generally something that would be easier to do before 7.0 than after).
> 
> This applies to the names of the proposed hooks. Let's finalize the naming conventions here. Also note that in WP core, all AI related hooks are prefixed with `wp_`.

Note we've settled on `wpai_` for our prefix (see #315) so we can update all hooks in this PR to use that now instead of `ai_experiments_`

gziolo:
> Remove ability_result and ability_result_{$name} filters (to be proposed for WP core per @gziolo's feedback)

I will have to open a WordPress Trac ticket so it's lined up for WP 7.1. I will see if we can find a way to include it in the AI plugin faster, conditionally. Maybe, through some feature detection or PHP version check.

gziolo:
I opened [Abilities API: Add execution lifecycle filters to WP_Ability methods](https://core.trac.wordpress.org/ticket/64989) to track the functionality needed directly in WordPress core. I appreciate all the feedback as well!

I also included there a related note:
> MCP Adapter ([WordPress/mcp-adapter](https://github.com/WordPress/mcp-adapter)) implements mcp_adapter_-prefixed hooks for validation toggling, caching, and post-execution audit. The team has decided to maintain their own lower-level filters on top of whatever core provides ([mcp-adapter#151](https://github.com/WordPress/mcp-adapter/pull/151)), but richer core hooks would reduce the scope of what the adapter needs to re-implement.

It's worth looking at the same angle as the MCP adapter and figuring out whether additional filters in the execution flow are needed, or if it is fine to wait until WP 7.1 for what WP core will implement.

Ryujiyasu:
I think it's fine to wait for WP 7.1 for the execution lifecycle filters in core (Trac #64989). This PR is already scoped to plugin-specific hooks only, and adding interim ability result filters here would create backcompat overhead once core ships them. The MCP Adapter has different needs (caching, validation toggling) that justify their own hooks, but for this plugin, relying on core seems cleaner. Happy to help with the core implementation when the time comes.

Ryujiyasu:
Regarding `wpai_get_post_details` and `wpai_get_post_terms` — I kept them because they pass internal context like `$post_id` and `$fields` directly, which felt convenient for plugin-level use. That said, if you think core-level filters would cover this well enough, I'm happy to remove them.

gziolo (commented):
It would be nice to add test coverage for `wpai_get_post_terms` and `wpai_get_post_details`, following `wpai_system_instruction` case. Otherwise, we are good to go.

Ryujiyasu:
@gziolo Thanks again for the review! I've added test coverage for both `wpai_get_post_details` and `wpai_get_post_terms` filters in `tests/Integration/Includes/HelpersTest.php`, following the same pattern as the existing `wpai_system_instruction` tests:

- `test_wpai_get_post_details_filter_modifies_output` — verifies the filter can modify the returned details
- `test_wpai_get_post_details_filter_receives_arguments` — verifies the filter receives `$post_id` and `$fields`
- `test_wpai_get_post_terms_filter_modifies_output` — verifies the filter can modify the returned terms
- `test_wpai_get_post_terms_filter_receives_arguments` — verifies the filter receives `$post_id` and `$allowed_taxonomies`

All PHPUnit jobs (PHP 7.4–8.4) pass on the latest commit. The E2E failures on this run are in `content-classification.spec.js` and are unrelated to this PR — they predate this commit and don't touch any of the files modified here.

@dkotter could you take another look when you get a chance? The hooks have been renamed to the `wpai_` prefix per #315, the `ability_result` filters were dropped (to be added in core via [Trac #64989](https://core.trac.wordpress.org/ticket/64989)), and the requested tests are now in place. Thanks!

gziolo (dismissed):
LGTM ✅ — feedback all addressed (renamed to `wpai_`, `ability_result` filters deferred to [Trac #64989](https://core.trac.wordpress.org/ticket/64989), tests added for both post-context filters). Lint, PHPStan, and the 6 new filter tests all pass locally.

Two non-blocking remarks for follow-up:

1. **`wp_get_ability` guards are dead code.** The plugin now requires WP 7.0, where the Abilities API is in core. The `function_exists( 'wp_get_ability' )` / `markTestSkipped` blocks in `HelpersTest.php` and the new filter tests can be dropped — they're never reachable.

2. **Prefer asserting on the filter's return value over captured arguments.** Tests like `test_wpai_get_post_details_filter_receives_arguments` and `test_system_instruction_filter_receives_ability_name` capture args via closure and assert on them — that's the shape of an action test. For filters, it's stronger to encode the args into the return value and assert against the final output, e.g.:

   ```php
   $filter_callback = static function ( $details, $post_id, $fields ) {
       $details['title'] = sprintf( 'post:%d|fields:%s', $post_id, implode( ',', $fields ) );
       return $details;
   };
   // then assert on $result['title']
   ```

   Not always possible, but where it is, it ties the assertion to observable behavior.

dkotter (dismissed):

dkotter:
> 1. **`wp_get_ability` guards are dead code.** The plugin now requires WP 7.0, where the Abilities API is in core. The `function_exists( 'wp_get_ability' )` / `markTestSkipped` blocks in `HelpersTest.php` and the new filter tests can be dropped — they're never reachable

This is a good catch and noticed we have a handful of these existing in the codebase. I've opened #378 to remove any of these existing ones but would be great to remove any newly introduced ones in this PR

dkotter (approved):

PR #292, opened 2026-03-10
Add image refinement flow to our image generation screens

dkotter:
## What?

Closes #279

Partially closes #270

Add a "Refine" flow to our existing image generation screens

## Why?

It's likely that an image you generate doesn't match exactly what you want the first time. While you can edit the prompt and try again, sometimes the image generated is pretty close and needs just a few refinements. This flow is now supported when generating images, both in the editor and directly in the Media Library.

## How?

- Update our `generate-image` Ability to support passing in a `reference` base-64 encoded image. If that is passed, we add that as a file to our prompt
- In our image generation UI, add a "refine" state that allows you to enter a refinement prompt and send that along with the generated image (this was previously added in #235 but [removed](https://github.com/WordPress/ai/pull/235#issuecomment-3985609931))
- When a refinement comes back, show the original image next to the refined image for easy comparison
- When importing an image, keep track of each prompt used so we can store that with the image. Previously we stored the original prompt only, now we store that plus any refinement prompts

### Use of AI Tools

Pulled my original code from #235 for most of this. Used GPT-5.3 Codex in Cursor to refine and fix a few things with manual review, testing and cleanup done by me

## Testing Instructions

1. Install the [AI Provider for Google plugin](https://wordpress.org/plugins/ai-provider-for-google/)
2. Pull this PR down and run `npm install && npm run build`
3. Add valid Google credentials
4. Turn on the Image Generation experiment
5. Add an image block into a post and click on the Generate Image button
6. Add a prompt to generate an image
7. Click the Refine Image button and add another prompt
8. Choose to keep that image, refine again or generate again (which will send the original image along with the refinement prompt, basically a "try again" approach)
9. Ensure all those buttons work as expected and the image imports with proper history
10. Do the same thing with the stand-alone image generation screen at `Media > Generate Image`

## Screenshots

Modal showing how to generate a new image

Generated image showing available options, including the new Refine image button

Adding a refinement prompt to a generated image

Showing the result of the refinement step with the original image next to the refined image

[...]

PR #305, opened 2026-03-13
Add ability to edit existing images in the media library

### LINKED ISSUE #236: Add generative image edit actions (inpaint and outpaint)

Users lack a clear, high-impact workflow for image edits like removing objects or extending canvas.  Let's build on the existing foundation in #13 to add generative edits that are easy to discover and use.

Add two image edit actions:

- Inpaint: “Remove or replace area” (aka erase and replace)
- Outpaint: “Extend image” (aka expand canvas and fill)

Add UI entry points:

- Gutenberg-based Media Editor (preferred when available)
- Media Library: Edit Media screen (fallback and parity)

Store results as a new media item while preserving the original image.

### LINKED ISSUE #237: Add smart background removal for images

Background removal could exist in both the new client-side media editing flow (https://github.com/WordPress/gutenberg/issues/74333) and the existing Edit Media screen in the Media Library as a "remove background" option.

Add a “Remove background” action that:

- Provides a before/after preview and ability to revert
- Produces a transparent PNG (or preserves original format when possible)
- Optionally adds “Replace background” but could be reserved for a follow-up PR/future release

Integrate into:

- Gutenberg client-side media editing flow (primary)
- Media Library: Edit Media screen (secondary)

dkotter:
## What?

Closes #236, #237

Adds the ability to edit existing images in the media library views.

## Why?

We already have the ability to generate images and we're adding the ability to refine those images in #292 but we don't have a way to refine/edit existing images. This PR adds that in via a prompt approach and with a few preset options (`Remove background`, `Expand background`).

## How?

- Inject our needed component into the WordPress Media Library edit screen when that is loaded (there are no hooks here to use unfortunately)
- This component will render three preset buttons, `Remove background`, `Expand background`, and `Refine Image`, which will add a prompt textarea when clicked that allows you to make conversational edits to the image
- When using this prompt, set the flow up to match our image refinement process used elsewhere. Allows you to refine the image and then import the image once done
- When using a preset option, add built-in prompts that power those. For expanding background, we first build a new canvas that is 1.5x larger than the image. We add the image to the center of this and we send this to the LLM, allowing it to fill in the new blank space
- Add better state management, both in this new flow and in the existing image refinement flows, so we keep track of each image generation and allow a user to paginate back and forth. This allows them to "go back in time" and import an older generated image or start refinement over from that point.
- Keep state in place after importing so a user can import another image or continue to refine

### Use of AI Tools

Used both Claude Code (running Opus 4.6) and Cursor (running GPT-5.3 Codex) to plan and execute, with final refinement and clean up done by me.

## Testing Instructions

1. Install the [AI Provider for Google plugin](https://wordpress.org/plugins/ai-provider-for-google/)
2. Pull this PR down and run `npm install && npm run build`
3. Add valid Google credentials
4. Turn on the Image Generation experiment
5. Ensure you have an existing image in the Media Library
6. In the Media Library grid view, click on an image and then in the modal that shows, click the `Edit Image` button
7. Ensure you see the new buttons. Click on `Refine Image`
9. Add a prompt and click `Generate`
10. Ensure an image is generated and that it shows side-by-side with the original image
11. From here, test out the various buttons below the images, like refining further, running the generation again, starting over and saving the image
12. Go back to the main edit screen and try out the other buttons
13. Ensure those work as expected (background is removed and replaced with white or image is expanded)
14. Run these same tests again but in the block editor. Add an image block and choose Media Library. Choose an image in the modal and click `Edit image`
15. If desired, can also test the refinement flow after generating an image (either in the media library or in the editor) and ensure you can refine the image multiple times and paginate back and forth through those results

## Screenshots

Image edit screen showing the new Expand Background, Remove Background and Refine Image buttons

Showing the refine image textarea after the button is clicked

Showing the generate loading state after refining an image

Showing the first image refinement

Showing the second image refinement

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-305-7cef4574476d383884d7623bd140da646805280d.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
For some reason my Google credentials aren't working, so I can't fully test this (will continue to see what's going on there so I can test more fully, but have some thoughts for now).  I would advocate for just removing the `AI Exit` button press requirements and just add the `Expand Background` (capital B) and `Remove Background` buttons on the row below the current `Crop` and other buttons.  

Given that the prompt textarea and `Generate` button appear to be tied to a "refine" flow, let's perhaps add a `Refine Image` button that adds the helper text, textarea, and button (perhaps those appear below the image so it doesn't jump the image downward when those are added in. Let's also make the `DESCRIBE...` text sentence case.

Also for the Expand and Remove and Refine buttons let's ensure the base image is still present, as for now clicking AI Edit replaces the image and renders buttons but would be ideal to keep the image and add in the buttons.

dkotter:
> I would advocate for just removing the `AI Exit` button press requirements and just add the `Expand Background` (capital B) and `Remove Background` buttons on the row below the current `Crop` and other buttons.

This is done now. Removed the `AI Edit` button and instead we show the other buttons by default (and capitalized the second word). I've updated screenshots in the PR description showing the new UI.
 
> Given that the prompt textarea and `Generate` button appear to be tied to a "refine" flow, let's perhaps add a `Refine Image` button that adds the helper text, textarea, and button (perhaps those appear below the image so it doesn't jump the image downward when those are added in). Let's also make the `DESCRIBE...` text sentence case.

Added a `Refine Image` button and clicking that now shows the prompt flow. The text above the textarea is now sentence case instead of ALL UPPERCASE, though noting that style was added by Core so we are overriding that here now.

For now I haven't added the textarea below the image as it's pretty common for an image to be larger than the screen height and so you'd have to scroll down to see those options. So easy to miss those were added in that scenario.

Personally I think this conversational image editing is the most useful thing here (comparing to removing background or expanding the image). My make sense to have that be the first button, at the moment I've left that as the last.

> Also for the Expand and Remove and Refine buttons let's ensure the base image is still present, as for now clicking AI Edit replaces the image and renders buttons but would be ideal to keep the image and add in the buttons.

I've left the original image while generation is in place. Once the image is generated, the original image hides as we show that side-by-side with the generated image.

jeffpaul:
> The text above the textarea is now sentence case instead of ALL UPPERCASE, though noting that style was added by Core so we are overriding that here now.

Oh, didn't realize that was core default, huh ok then I guess ignore me and go back to YELLING AT PEOPLE?

> For now I haven't added the textarea below the image as it's pretty common for an image to be larger than the screen height and so you'd have to scroll down to see those options. So easy to miss those were added in that scenario.

Ok, that makes sense, I concur.

> Personally I think this conversational image editing is the most useful thing here (comparing to removing background or expanding the image). My make sense to have that be the first button, at the moment I've left that as the last.

Totally fine moving that to be the first button.

jeffpaul:
Once you get into a Refine, Expand, or Remove button flow the button options tend to disappear.  It would be good to ensure those options are present as image iterations are created.

jeffpaul:
Here's getting into the Refine flow where ideally we still have Expand and Remove options present:

Screenshot 2026-03-19 at 11 55 00 AM

Here's getting into the Expand flow where ideally we still have Expand and Remove options present:

Screenshot 2026-03-19 at 11 56 05 AM

Here's getting into the Remove flow where ideally we still have Expand and Remove options present:

Screenshot 2026-03-19 at 11 57 26 AM

dkotter:
> Once you get into a Refine, Expand, or Remove button flow the button options tend to disappear. It would be good to ensure those options are present as image iterations are created.

Would you expect those buttons to stay where they are at (above the image, below the default buttons)? Or should those show with the other buttons (Save, Refine, Generate, Start over) we render in that preview state, below the images?

Pros and cons to both in my mind. I like keeping them where they are at in all states (so above the image) but does also make sense to have them with the other buttons we render, noting we already have 4 buttons in that row so adding 2 more does get overwhelming

jeffpaul:
If we go with the top, we're effectively duplicating the Refine option in both places.  So perhaps in this flow we keep all three buttons at the top and remove the second Refine one from below the image?

dkotter:
> If we go with the top, we're effectively duplicating the Refine option in both places.

Not necessarily, I can remove the Refine button once we're in a state where we have one generated image. So what I have now locally is just the Remove and Expand buttons up top and the Refine button stays with the row below the image

> So perhaps in this flow we keep all three buttons at the top and remove the second Refine one from below the image?

But yes, could also do this, keep all three buttons up top (and future buttons, like the ones coming in #312) and then remove what would then be a duplicate Refine button from that bottom row

jeffpaul:
At this point just pick whatever you've got working locally and we can shuffle buttons around again later if needed.  As long as they're individually present only once that suffices for now.

dkotter:
Okay, here's how the flow looks now:

**Initial state when editing an image:**

Initial image editing state showing new AI buttons

**Clicking the Refine button and adding a prompt:**

Adding a refine image prompt

**Generating a refine image. Shows a loading state and hides the buttons so someone doesn't try and generate an image while generation is happening:**

Image generation happening

**Showing the generated image with the existing buttons still showing up top:**

Image preview state


Almost certainly more refinement we can do here. Happy to keep iterating here or in follow up PRs

jeffpaul (approved):

PR #313, opened 2026-03-17
[Feature]: Issue 45 | Content Classification (Contextual Tagging)

### LINKED ISSUE #45: Contextual Tagging
### Summary

This feature provides AI-powered suggestions for post tags and categories based on a comprehensive analysis of the post's content (as well as title and excerpt). It helps authors apply a relevant and consistent taxonomy, improving content organization, discoverability, and on-site navigation.

### Why it matters

Properly categorizing and tagging content is fundamental to an effective content strategy. It helps readers find related articles and signals relevance to search engines. However, this process is often inconsistent and subjective. This feature removes the guesswork by providing intelligent, context-aware suggestions, leading to a better-organized and more interconnected website.

### Sample Use Cases

* Primary: An editor is reviewing a draft and wants to ensure it’s properly classified. They click a button to generate a list of relevant tags, choosing the most appropriate ones in seconds.
* A new author is unfamiliar with a site’s established categories and tags. They use the feature to get suggestions that align with the existing content structure.
* Post-MVP: A content team wants to improve the consistency of their content archive. They can open older posts and use the feature to add relevant tags that were missed originally.
* Post-MVP: A large publication with many contributing writers uses the feature to standardize taxonomy and prevent the creation of dozens of similar, redundant tags (e.g., "AI," "artificial-intelligence," "A.I.").

## Taxonomy Strategy Handling

**Why this matters:** Different sites have different needs for their taxonomy. Some may prefer a few broad categories, while others rely on many specific, niche tags. The AI's suggestions should be able to adapt to the user's organizational strategy to be truly useful.

### How Taxonomy Strategy is Set

* Global user preference (primary mechanism)
  * Set once in plugin settings.
  * Options might include:
    * Suggest new tags based on context vs. Only suggest existing tags
    * Post-MVP: Toggle on or off which post types are included 
    * Toggle on or off which site taxonomies should be included (categories, tags, people, location, etc.) perhaps with categories/tags defaulted to on?
    * A sliding scale to select confidence threshold (for broader topics vs. more precise keywords, aka “precision” vs. “verbosity”), perhaps with different setting for taxonomy?
  * This setting controls the default prompt strategy and suggestion style. 
* Per-post override (secondary)
  * An options link or dropdown within the suggestion UI in the editor sidebar.
  * Allows the user to request a different type of suggestion (e.g., "Suggest broader topics") for a single post without changing the global default.

### Prompt construction

* Prompt templates adjust based on the selected confidence thresholds and taxonomies.
* Example (Specific): “Analyze this article and suggest 5-7 specific keyword tags. Prioritize using existing tags from this list: [list of site tags].”
* Example (Broad): “Analyze this article and suggest 1-2 high-level categories it belongs to from the following list: [list of site categories].”

## User Interaction Flow

**Overview:** The goal is to provide taxonomy suggestions directly within the native WordPress interface for managing tags and categories, making the feature feel intuitive and accessible. The interaction should be quick, additive, and require minimal clicks to apply.

### User Actions

* User sees the “Suggest Tags” / “Suggest Categories” button
  * Appears inside the "Tags" and "Categories" panels in the post editor sidebar.
  * Grayed out until the post contains a minimum amount of content (~150 words).
* User clicks the active button
  * A small loading indicator appears.
  * After a moment, a list of suggested terms is displayed directly within the panel.
* System displays suggested terms
  * Suggestions are shown as clickable pills or buttons.
  * The number of suggestions defaults to 5-7 but can be configured in settings.
  * Suggestions for existing site tags/categories may be styled differently from suggestions for new terms.
  * An option to "Regenerate" is visible.
* User selects or dismisses suggestions
  * Clicking a suggestion pill instantly adds it to the post's tags or assigns the category.
  * There is a clear way to dismiss the suggestions (e.g., a small "close" icon).

### Failure states

* If the API call fails, a discreet message like "Could not get suggestions" appears with a "Retry" button. The user is not blocked from manually adding their own terms.

## Alternate Entry Point: Empty Taxonomy with Post Content

This flow prompts users who have written content but haven't yet assigned any tags or categories.  Potentially prioritized before the existing tags use case.  Should also consider human-suggested tags vs. AI-suggested tags.

### Trigger Condition

* Post content contains ≥ ~150 words.
* The post has no tags assigned and is in the default category.

### System Behavior

* A light, one-line callout appears in the relevant sidebar panel:
  * “Need some tags? Get AI suggestions.”

### User Action

* User clicks the prompt to trigger the suggestion UI.

### Suggestion UI

* The same UI as the core flow is activated: a list of clickable suggestion pills appears within the panel.

## Technical Requirements (High-Level)

* AI Input Requirements:
  * Current post title.
  * Full post content.
  * A list of all existing tags and categories on the website (to encourage consistency). Some AI models will look to embed existing terms in vector-space, helping it recommend content based on those terms.
  * The selected taxonomy strategy.
* AI Output Requirements:
  * A structured list of suggested terms (e.g., JSON array of strings).
  * Confidence threshold percent for each suggested term.
  * The model should distinguish between suggesting tags and categories.
* Model Agnosticism:
  * Supports text-based models via the WP AI Client SDK package (e.g., OpenAI, Claude, local LLMs).
  * Modular prompt design to allow for easy adaptation.
* Fail States & UX Expectations:
  * Errors must not block the user from saving, updating, or publishing the post.
  * The feature should gracefully handle posts with very short or no content.
* Privacy & Config:
  * API keys managed via the central WP AI Client SDK package.
  * User content is sent for processing according to user consent but is not stored long-term (perhaps at the settings level or per request to recommend terms).

## Extensibility Notes

* **Custom Taxonomy Support:** Design to be adaptable for developers to enable suggestions for custom taxonomies.
* **Hookable Pipeline:** Provide hooks for developers to filter the content sent to the AI and the suggestions returned from it.
* **Feedback Mechanism:** Future iterations could include a simple "Was this helpful?" feedback loop to refine suggestion quality over time.

* For categories (which are hierarchical), should the AI be able to suggest parent/child relationships?
  * This exists in categories. Yes, this should be supported. This would need to be visually represented.
  * ClassifAI does not utilize child tags currently.
* Should we suggest tags and categories in a single, combined UI, or keep them as two separate actions in their respective panels? (Separate seems cleaner and more aligned with the WordPress UI).
  * To be explored in UI designs.

### Version 1 - Contextual Tagging

Image

### Version 2 - Contextual Tagging

Image

_**Props to [Lauren Woodmansee](https://www.linkedin.com/in/laurenwoodmansee/) for the product definition, [Rachael Cortellessa](https://profiles.wordpress.org/rachaelcortellessa/) for the designs, and [Darin Kotter](https://github.com/dkotter) for technical review of both.**_

TylerB24890:
## What?

Closes #45 

This PR introduces Content Classification (Contextual Tagging) to the AI Experiments plugin. It allows for one-click AI-powered suggestions for post tags and categories based on a comprehensive analysis of the post's content (as well as title and excerpt). It helps authors apply a relevant and consistent taxonomy, improving content organization, discoverability, and on-site navigation.

## Why?

Properly categorizing and tagging content is fundamental to an effective content strategy. It helps readers find related articles and signals relevance to search engines. However, this process is often inconsistent and subjective. This feature removes the guesswork by providing intelligent, context-aware suggestions, leading to a better-organized and more interconnected website.

## How?

* Registered a new `ContentClassification` Experiment.
* Registered new settings to enable Content Classification + determine the strategy (Suggest New, Use Existing) and max suggestions.
* Hook into the block editor `editor.PostTaxonomyType` and inject the Experiment UI (buttons, suggestions)
    * Opt'd for UI built-in to the existing term selectors rather than providing a modal experience as mocked-up in the issue description. See screenshots below for further details.
* Apply a custom system prompt to ensure the experiment settings are respected (new terms, existing terms, max terms, etc...)
* Various filters are available to customize the output;
    * `wpai_content_classification_content` - Filter content before AI processing
    * `wpai_content_classification_suggestions` - Filter suggestions after AI processing
    * `wpai_content_classification_strategy` - Filter the strategy setting
    * `wpai_content_classification_max_suggestions` - Filter the max suggestions setting

### Use of AI Tools

* **AI Assistance:** Yes
* **Used for:** 
    * Documentation review and workflow summary
    * Experiment scaffolding
    * Unit test boilerplates
    * Inline documentation and experiment documentation generation
    * Assistance with the system prompt generation

## Testing Instructions

1. Navigate to the **Experiments** page and enable the **Content Classification** experiment. Choose the strategy and maximum suggestions.
2. Navigate to the block editor the core `post` post type and confirm the **Suggest Categories** and **Suggest Tags** buttons are available in the term selector panels.
3. Click one of the buttons and confirm terms are returned from the AI according to your selected preferences. If "Existing terms only" is selected, your response should only consist of existing terms with no new terms suggested.
    * If no terms are available in the taxonomy an error message will appear at the top of the block editor.
4. Click on the returned terms and confirm they are assigned (or created + assigned) as expected.
5. Navigate back to the Experiments settings page and change the settings for the **Content Classification** experiment.
    * If you previously suggested existing terms, change it to the "New Terms" option.
    * Increase/decrease the maximum suggestions.
 6. Navigate back to the block editor and repeat steps 3 & 4, confirming the results have changed according to your selected preferences.

## Screenshots or screencast 

**Settings UI:**
Screenshot 2026-03-17 at 4 39 37 PM

**Suggestion UI (block editor):**
Screenshot 2026-03-17 at 4 41 06 PM

**New Term Suggestions:**
Screenshot 2026-03-17 at 4 40 53 PM

**Existing Term Suggestions:**
Screenshot 2026-03-17 at 4 42 41 PM

## Additional Notes

* Various features from the issue discussion were put on hold for the initial PR. These features can be added in a new PR or after the initial PR review.
    * **Empty taxonomy callout** - Proactive "Need some tags? Get AI suggestions" prompt when post has 150+ words but no tags or categories.
    * **Per-post strategy override** - Can be achieved through the `ai_contextual_tagging_strategy` filter, but an inline UI to do so is possible.
    * **Taxonomy toggle settings** - Allow users to choose which taxonomies can be auto-generated/suggested (default is just `category` and `post_tag`)
    * **Confidence Threshold Slider** - Allow users to choose the confidence threshold for suggestions

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-313-354fb82102dd9d166a01f2c806a815d1a75ad5f6.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
1. Should probably center align the button text for Categories and Tags to align to other buttons in the sidebar

Screenshot 2026-03-17 at 5 42 41 PM

2. Can probably remove the loading spinner as the suggest buttons alternate to a loading state already

Screenshot 2026-03-17 at 5 43 27 PM

Screenshot 2026-03-17 at 5 44 54 PM

3. Clicking the suggested term (specifically NOT the `X` close) appears to also close/remove the suggested term where I thought it _might_ allow me to edit the suggested term (e.g. to pluralize or capitalize it).  There doesn't seem to be an obvious way to _save_ suggested terms as the post was already autosaved so there wasn't a `Save` option in the Gutenberg top toolbar and nothing else obvious in the sidebar panel.  What's the expectation to save/apply suggestions?

Screenshot 2026-03-17 at 5 48 23 PM

Screenshot 2026-03-17 at 5 50 24 PM

Screenshot 2026-03-17 at 5 49 04 PM

4. It _might_ be helpful to add a subheading of "Suggested Categories" and "Suggested Tags" just below the separator and just above the suggested terms to make it obvious what that sub-section sidebar panel is?

Screenshot 2026-03-17 at 5 51 03 PM

5. (Continuing from screenshot above) Instead of "Regenerate" perhaps "Suggest again"?

TylerB24890:
@jeffpaul -- Thanks for the feedback! I've resolved 1, 2 & 4 -- see below regarding 3 & 5;

> Clicking the suggested term (specifically NOT the X close) appears to also close/remove the suggested term where I thought it might allow me to edit the suggested term (e.g. to pluralize or capitalize it).

That might be a little challenging with this UI being in the sidebar. We might be better off going with a dedicated modal if we're to support that. Is that something we would want to do as opposed to the sidebar UI?

> There doesn't seem to be an obvious way to save suggested terms as the post was already autosaved so there wasn't a Save option in the Gutenberg top toolbar and nothing else obvious in the sidebar panel. What's the expectation to save/apply suggestions?

Clicking the suggestion (outside of the 'X') will automatically create (if new) & assign the post to the term -- but I _would_ have expected the Save button for the post to re-appear, since we changed the post state 🤔  I'll have to see what's going on there.

> (Continuing from screenshot above) Instead of "Regenerate" perhaps "Suggest again"?

_Personally_ I feel "Regenerate" makes more sense because the initial button says "Generate", but happy to update it if you prefer.

docs/experiments/contextual-tagging.md
```diff
@@ -0,0 +1,402 @@
+# Contextual Tagging
```
dkotter: So not to derail things here but wondering if we can come up with a better name for this than `Contextual Tagging`. Just my opinion so feel free to ignore but I'm worried too many people will associate that `Tagging` part of this with WordPress tags, and just assume this only works for tags even though it also works for categories (and presumably other taxonomies in the future).

That said, I don't necessarily have a better option though here's some ideas:

- Content Classification
- Contextual Classification
- Term Suggestions
- Suggested Terms
- Smart Classification

dkotter: cc/ @jeffpaul if you have thoughts or feel strongly on `Contextual Tagging`

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,574 @@
...
+use function WordPress\AI\get_post_context;
+use function WordPress\AI\get_preferred_models_for_text_generation;
+use function WordPress\AI\normalize_content;
+
+/**
+ * Contextual tagging WordPress Ability.
+ *
+ * Generates taxonomy term suggestions based on post content analysis.
+ *
+ * @since 0.6.0
```
dkotter: We should update all `@since` statements to reference `x.x.x`, which will be replaced with the actual version number when this ships. We don't know for sure the version number until that point.

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,574 @@
...
+		 * Allows developers to modify, augment, or replace the content that the AI analyzes
+		 * when generating tag and category suggestions.
+		 *
+		 * @since 0.6.0
+		 *
+		 * @param string $context  The normalized content string to be analyzed.
+		 * @param string $taxonomy The taxonomy slug being suggested for (e.g., 'post_tag', 'category').
+		 * @param string $strategy The suggestion strategy ('existing_only' or 'allow_new').
+		 */
+		$context = (string) apply_filters( 'ai_contextual_tagging_content', $context, $taxonomy, $strategy );
```
dkotter: We've recently settled on `wpai_` as our prefix for hooks

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,574 @@
...
+		 * - 'is_new'     (bool)   Whether the term is new or already exists on the site.
+		 * - 'parent'     (string) Optional. Parent term name for hierarchical taxonomies.
+		 *
+		 * @since 0.6.0
+		 *
+		 * @param array<array{term: string, confidence: float, is_new: bool, parent?: string}> $suggestions    The parsed suggestions.
+		 * @param string                                                                       $taxonomy       The taxonomy slug (e.g., 'post_tag', 'category').
+		 * @param string                                                                       $strategy       The suggestion strategy ('existing_only' or 'allow_new').
+		 */
+		return (array) apply_filters( 'ai_contextual_tagging_suggestions', $suggestions, $taxonomy, $strategy );
```
dkotter:
```suggestion
		return (array) apply_filters( 'wpai_contextual_tagging_suggestions', $suggestions, $taxonomy, $strategy );
```

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,574 @@
...
+	): string {
+		return implode(
+			"\n",
+			array(
+				"You are a content taxonomy assistant for a WordPress website. Your task is to analyze article content and suggest relevant {$taxonomy} terms.",
+				'',
+				"Goal: Analyze the provided content (title, body, and any existing context) and suggest up to {$max_suggestions} relevant terms for the {$taxonomy} taxonomy.",
+				'',
+				'Output format:',
+				'Return ONLY a valid JSON array. No prose, no markdown, no code fences.',
```
dkotter: I would suggest that instead of trying to force the LLM to return json in our instructions, we should instead use structured outputs ([example](https://github.com/WordPress/ai/blob/6f5db2adf2043d2832b94cc25dfdc9d0b6e925a7/includes/Abilities/Review_Notes/Review_Notes.php#L298)). This is going to give us more consistent results

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,574 @@
...
+		// Build strategy instruction.
+		$strategy_instruction = $this->build_strategy_instruction( $strategy );
+
+		// Build existing terms instruction.
+		$existing_terms_instruction = $this->build_existing_terms_instruction( $existing_terms, $strategy );
+
+		// Get the taxonomy label for the prompt.
+		$taxonomy_label = $this->get_taxonomy_label( $taxonomy );
+
+		// Build the system instruction directly to avoid esc_html() escaping JSON syntax.
```
dkotter: Would be great if we could keep our system instructions in a separate file to match all of our other Abilities. I think if we move to requiring a structured output and not forcing that into our instructions, that may be more achievable

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,574 @@
...
+		 * @since 0.6.0
+		 *
+		 * @param string $context  The normalized content string to be analyzed.
+		 * @param string $taxonomy The taxonomy slug being suggested for (e.g., 'post_tag', 'category').
+		 * @param string $strategy The suggestion strategy ('existing_only' or 'allow_new').
+		 */
+		$context = (string) apply_filters( 'ai_contextual_tagging_content', $context, $taxonomy, $strategy );
+
+		// Generate the suggestions using the AI client.
+		$result = wp_ai_client_prompt( '"""' . $context . '"""' )
```
dkotter: We've taken a different approach to avoid prompt injection in all our newest Abilities (I think Title Generation is the only one still doing this). See [Review_Notes](https://github.com/WordPress/ai/blob/6f5db2adf2043d2832b94cc25dfdc9d0b6e925a7/includes/Abilities/Review_Notes/Review_Notes.php#L368) for an example. Basically we add our content into XML-like structure. If we do that here, this trip quotes can be removed and also the triple-quote line from our instructions

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,574 @@
...
+	protected function input_schema(): array {
+		return array(
+			'type'       => 'object',
+			'properties' => array(
+				'content'         => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
+					'description'       => esc_html__( 'Content to generate taxonomy suggestions for.', 'ai' ),
+				),
+				'post_id'         => array(
```
dkotter, line 46: The approach we've settled on for most of our Abilities (in order to keep them as general / reusable as possible) is to allow passing in content plus an optional context param. That param can be actual context but can also be a post ID. If a post ID is passed, we use that to generate context. This allows these Abilities to be used outside of the typical WordPress post screen (or outside of WordPress all together). Example [here](https://github.com/WordPress/ai/blob/6f5db2adf2043d2832b94cc25dfdc9d0b6e925a7/includes/Abilities/Excerpt_Generation/Excerpt_Generation.php#L35-L43).

Not saying we have to make that change here but worth considering

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,282 @@
...
+
+/**
+ * Contextual tagging experiment.
+ *
+ * Provides AI-powered suggestions for post taxonomies
+ * based on a comprehensive analysis of the post content.
+ *
+ * @since 0.6.0
+ */
+class Contextual_Tagging extends Abstract_Experiment {
```
dkotter: We'll likely soon merge PRs that move away from using "Experiment" to using "Feature" which will require a number of changes, including here. Main PR to keep an eye on is https://github.com/WordPress/ai/pull/316

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,282 @@
...
+		if ( 'post.php' !== $hook_suffix && 'post-new.php' !== $hook_suffix ) {
+			return;
+		}
+
+		$screen = get_current_screen();
+
+		// Load the assets only if the post type supports editor and is not an attachment.
+		if (
+			! $screen ||
+			! post_type_supports( $screen->post_type, 'editor' ) ||
```
dkotter: Curious for the reasoning behind checking for `editor` support?

Seems we may want a check to see if the post supports either categories or tags

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,282 @@
...
+
+		/**
+		 * Filters the strategy to use for contextual tagging.
+		 *
+		 * @since 0.6.0
+		 *
+		 * @param string $strategy The strategy to use.
+		 * @return string The filtered strategy.
+		 */
+		$strategy = apply_filters( 'ai_contextual_tagging_strategy', $strategy );
```
dkotter:
```suggestion
		$strategy = apply_filters( 'wpai_contextual_tagging_strategy', $strategy );
```

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,282 @@
...
+
+		/**
+		 * Filters the maximum number of suggestions to generate for contextual tagging.
+		 *
+		 * @since 0.6.0
+		 *
+		 * @param int $max_suggestions The maximum number of suggestions to generate.
+		 * @return int The filtered max suggestions.
+		 */
+		return apply_filters( 'ai_contextual_tagging_max_suggestions', $max_suggestions );
```
dkotter:
```suggestion
		return apply_filters( 'wpai_contextual_tagging_max_suggestions', $max_suggestions );
```

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
```diff
@@ -0,0 +1,282 @@
...
+	 *
+	 * @since 0.6.0
+	 */
+	public function render_settings_fields(): void {
+		$strategy_option        = $this->get_field_option_name( 'strategy' );
+		$max_suggestions_option = $this->get_field_option_name( 'max_suggestions' );
+		$current_strategy       = get_option( $this->get_field_option_name( 'strategy' ), self::STRATEGY_EXISTING_ONLY );
+		$current_max            = get_option( $this->get_field_option_name( 'max_suggestions' ), self::DEFAULT_MAX_SUGGESTIONS );
+		?>
+		<fieldset class="ai-experiment-settings-fieldset">
```
dkotter: We don't currently have any active Experiments that have settings beyond on/off. May need design direction on how best to render these but feels like at a minimum, we should:

- Left-align the label names with the Experiment description
- Ensure the labels and fields are in two separate columns, so the fields all left-align with each other
- Remove extra padding on the right of the number input

Quick mock:

Contextual Tagging settings

includes/Experiments/Contextual_Tagging/README.md
dkotter, line 1: We've only been adding documentation at the `docs/experiments` level so I think we can remove this

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,574 @@
...
+	/**
+	 * Gets existing terms for a taxonomy.
+	 *
+	 * @since 0.6.0
+	 *
+	 * @param string $taxonomy The taxonomy to get terms for.
+	 * @return array<string> List of existing term names.
+	 */
+	protected function get_existing_terms( string $taxonomy ): array {
+		$terms = get_terms(
```
dkotter: It's not unusual for sites to have thousands or tens of thousands of terms (tags especially). In that scenario, seems we still send all of those to the LLM which won't scale well.

Wondering if we need to take an approach where we don't send any existing terms and instead we check if the returned terms already exist? So basically instead of asking the LLM to only return terms in our defined list, we allow the LLM to return anything and then we remove any that aren't existing terms. May impact the results we end up with but seems better than sending tens of thousands of tags for the LLM to process

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
dkotter: May also need a check here to see if the current user has permissions to create terms, since the JS we load allows them to do that.

dkotter (changes requested):
Handful of suggestions / comments / concerns, happy to talk through any of those as needed.

Would also be great to get some E2E tests here prior to this getting merged.

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
TylerB24890: Noted. Will update accordingly.

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
TylerB24890: That makes sense, much less generic than just "ai". Will update accordingly.

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
TylerB24890: That makes sense. I only opted for this approach because the prompt used variables in the output, but switching to the structured approach should make this doable. I'll see what I can do to revise.

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
TylerB24890: Got it! I wasn't a fan of this triple quote approach anyway 😅

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
TylerB24890: Would we want to make that update in this PR, ahead of the other PR being merged? Or should we wait until that happens to make the updates here?

includes/Experiments/Contextual_Tagging/README.md
TylerB24890: Got it. We probably want to update the docs because I think it does say to add a readme to the experiment directly. Will remove this file in favor of the larger doc.

includes/Abilities/Contextual_Tagging/Contextual_Tagging.php
dkotter: So our `get_system_instruction` method does allow you to pass in data that is then made available to the instruction file ([example](https://github.com/WordPress/ai/blob/e60b15f8d8acfd34c5606c98e45484921ca5d0c0/includes/Abilities/Summarization/system-instruction.php#L17)) so I think what you're doing here should still be doable

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
dkotter: That PR has been merged now and is targeted for our next release tomorrow so yeah, changes will need to be made in this PR to account for all of that (and apologies for the extra work here)

jeffpaul:
> That might be a little challenging with this UI being in the sidebar. We might be better off going with a dedicated modal if we're to support that. Is that something we would want to do as opposed to the sidebar UI?

I'm fine holding off for that now as the current implementation is good.

> Clicking the suggestion (outside of the 'X') will automatically create (if new) & assign the post to the term -- but I would have expected the Save button for the post to re-appear, since we changed the post state 🤔 I'll have to see what's going on there.

For me clicking the term and not the `X` also closes/removes the term, but applying/saving it would be better so perhaps something's not working there?  I'm running WordPress 7.0-beta5-62069 and AI Provider for Google Version 1.0.3.

>> (Continuing from screenshot above) Instead of "Regenerate" perhaps "Suggest again"?
>
> Personally I feel "Regenerate" makes more sense because the initial button says "Generate", but happy to update it if you prefer.

Except the initial buttons say `Suggest Categories` and `Suggest Tags`?

Screenshot 2026-03-19 at 12 51 06 PM

TylerB24890:
@jeffpaul 

> For me clicking the term and not the X also closes/removes the term, but applying/saving it would be better so perhaps something's not working there? I'm running WordPress 7.0-beta5-62069 and AI Provider for Google Version 1.0.3.

Thanks for the info -- I'll debug this and see what's going on.

> Except the initial buttons say Suggest Categories and Suggest Tags?

🤦 I must have looked at the loading button that says "Generating" in my screenshot and just mixed the things up or something. We can update it to "Suggest again" -- that keeps the wording consistent anyway.

Would we want to update the loading state to "Suggesting..." instead of "Generating"? Or keep that as is?

jeffpaul:
I'm fine with the button loading text staying as `Generating` as that's it generating the suggestions.

includes/Experiments/Contextual_Tagging/Contextual_Tagging.php
TylerB24890: Added a check if the user has the `manage_categories` capability.

Re: 
> Curious for the reasoning behind checking for editor support?

I've changed this to check if the post type supports `category` or `post_tag` taxonomies.

TylerB24890:
@jeffpaul @dkotter All feedback has been resolved with exception of a couple items I still need to dive into (listed below). I've also fixed the term saving/assignment -- when a term suggestion is selected it should be created and/or assigned to the post as expected.

I've also added in some logic to prevent the LLM from returning already assigned terms as suggestions. I noticed when I had accepted previous suggestions and generated more I kept receiving terms I already had assigned. This logic prevents that.

**What's left:**
> It's not unusual for sites to have thousands or tens of thousands of terms (tags especially). In that scenario, seems we still send all of those to the LLM which won't scale well.
>
> Wondering if we need to take an approach where we don't send any existing terms and instead we check if the returned terms already exist? So basically instead of asking the LLM to only return terms in our defined list, we allow the LLM to return anything and then we remove any that aren't existing terms. May impact the results we end up with but seems better than sending tens of thousands of tags for the LLM to process

Need to circle around to this to avoid sending _all_ terms to the agent.

> So not to derail things here but wondering if we can come up with a better name for this than Contextual Tagging. Just my opinion so feel free to ignore but I'm worried too many people will associate that Tagging part of this with WordPress tags, and just assume this only works for tags even though it also works for categories (and presumably other taxonomies in the future).

Haven't made any changes on this front yet -- will hold until a decision is made on a name.

> Would also be great to get some E2E tests here prior to this getting merged.

Agreed on this -- will work on some E2E tests after the initial functionality is approved.

TylerB24890:
> **What's left:**
> 
> > It's not unusual for sites to have thousands or tens of thousands of terms (tags especially). In that scenario, seems we still send all of those to the LLM which won't scale well.
> > Wondering if we need to take an approach where we don't send any existing terms and instead we check if the returned terms already exist? So basically instead of asking the LLM to only return terms in our defined list, we allow the LLM to return anything and then we remove any that aren't existing terms. May impact the results we end up with but seems better than sending tens of thousands of tags for the LLM to process
> 
> Need to circle around to this to avoid sending _all_ terms to the agent.
> 

Took care of the existing terms being sent to the LLM. We no longer restrict the number of suggestions it will return and instead filter suggestions in post-processing as needed. We do still pass the currently _assigned_ terms, though. I feel this provides it additional context and prevents redundant suggestions.

@dkotter I rebased the `develop` branch into my branch and it looks like it introduced additional changes to my PR. Can you confirm if those are okay or do I need to re-open this on a new branch or something?

dkotter:
> I rebased the develop branch into my branch and it looks like it introduced additional changes to my PR. Can you confirm if those are okay or do I need to re-open this on a new branch or something?

I am seeing image generation/edit code in this PR that shouldn't be here. Wondering if you just need to merge the latest changes from `develop` here?

docs/experiments/contextual-tagging.md
jeffpaul: I feel that "terms" isn't something that non-technical site owners will understand means "categories and tags", so let's go with `Content Classification` or `Contextual Classification`?

jeffpaul (dismissed):
Latest round of updates works well for me, will await @dkotter return for final review/merge

docs/experiments/contextual-tagging.md
TylerB24890: I've updated _everything_ to "Content Classification" and changed the slug throughout the code to `content-classification` so it remains clear what it belongs to.

TylerB24890:
@dkotter Added some E2E tests in the latest commit. This is ready for another code review when you're able.

**EDIT:** Not sure why the other E2E tests are failing here, but the ones I've added are passing.

includes/Abilities/Content_Classification/Content_Classification.php
TylerB24890: We are no longer sending all terms to the LLM. Instead we've removed the limit on the number of suggestions the LLM can return and we filter them based on the existing terms.

We _do_ however, send the currently **assigned** terms to the LLM. That provides it some additional context and we can be sure we're not getting the assigned terms back at least.

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,550 @@
...
+			->using_model_preference( ...get_preferred_models_for_text_generation() )
+			->as_json_response( $this->suggestions_schema() )
+			->generate_text();
+
+		if ( is_wp_error( $result ) ) {
+			return $result;
+		}
+
+		// Fetch existing terms for post-processing (matching, not for the prompt).
+		$existing_terms = $this->get_existing_terms( $taxonomy );
```
dkotter: We only need existing terms if the strategy is `STRATEGY_EXISTING_ONLY`, correct? If so, I'd suggest setting this to an empty array if that isn't the strategy to avoid querying for terms we'll never use

includes/Abilities/Content_Classification/system-instruction.php
```diff
@@ -0,0 +1,28 @@
...
+
+Rules:
+- The taxonomy to suggest terms for is wrapped in the <taxonomy> tag.
+- Suggest as many relevant terms as you can identify from the content.
+- The "term" field must contain ONLY the human-readable tag or category name (1-3 words).
+- Confidence should reflect relevance: 1.0 = perfect match, 0.5 = somewhat relevant. Only suggest terms with confidence >= 0.5.
+- Do not suggest duplicate or near-duplicate terms.
+- Do not suggest terms listed in <assigned-terms> — they are already applied to this post.
+- Prioritize specificity and relevance over breadth.
+- Sort suggestions by confidence, highest first.
```
dkotter: I would remove this. We can easily sort ourselves after the fact and LLM's often struggle with following these types of instructions

includes/Experiments/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,299 @@
...
+
+		/**
+		 * Filters the maximum number of suggestions to generate for content classification.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param int $max_suggestions The maximum number of suggestions to generate.
+		 * @return int The filtered max suggestions.
+		 */
+		return apply_filters( 'wpai_content_classification_max_suggestions', $max_suggestions );
```
dkotter: Since we only support from 1-10, we may want to validate that here since someone could use this filter to return a negative number or a number greater than 10. I do think if that happens the Ability itself will return an error and maybe that's good enough but if we want to be more error-proof, we could validate things here after the filter and return a default if it doesn't match that range

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,550 @@
...
+	 * @since x.x.x
+	 *
+	 * @param mixed $args The input arguments to the ability.
+	 * @return bool|\WP_Error True if the user has permission, WP_Error otherwise.
+	 */
+	protected function permission_callback( $args ) {
+		$post_id = isset( $args['post_id'] ) ? absint( $args['post_id'] ) : null;
+
+		if ( $post_id ) {
+			$post = get_post( $args['post_id'] );
```
dkotter:
```suggestion
			$post = get_post( $post_id );
```

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,550 @@
...
+
+			// Ensure the post exists.
+			if ( ! $post instanceof WP_Post ) {
+				return new WP_Error(
+					'post_not_found',
+					/* translators: %d: Post ID. */
+					sprintf( esc_html__( 'Post with ID %d not found.', 'ai' ), absint( $args['post_id'] ) )
```
dkotter:
```suggestion
					sprintf( esc_html__( 'Post with ID %d not found.', 'ai' ), $post_id )
```

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,550 @@
...
+			$is_new     = ! isset( $existing_terms_map[ $term_lower ] );
+			$confidence = isset( $item['confidence'] ) ? (float) $item['confidence'] : 0.5;
+
+			// Skip terms already assigned to the post.
+			// The agent should avoid suggesting these, but just in case we'll check here as well.
+			if ( in_array( $term_lower, $assigned_terms_lower, true ) ) {
+				continue;
+			}
+
+			// For existing_only strategy, skip terms that don't exist.
```
dkotter, line 495: So I know this contradicts some of my earlier feedback and leaving this more as a discussion than something we have to change. But in testing, if I'm in the `existing_only` strategy, I often don't get terms from the LLM that match my existing terms and thus I'll get no results.

Wondering if there's a middle ground here where we get the top 100 terms (by count) and send those to the LLM when in the `existing_only` strategy? This would avoid prompt bloat but would still give the LLM something to look at and more likely give results.

dkotter (changes requested):

includes/Abilities/Content_Classification/Content_Classification.php
TylerB24890: I like that approach -- I'll implement it to test it out.

TylerB24890: This has been implemented and is ready for testing

includes/Experiments/Content_Classification/Content_Classification.php
TylerB24890: I'm passing it back through `$this->sanitize_max_suggestions()` which limits it to 10.

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,596 @@
...
+	 *
+	 * @return array<string, mixed> The input schema of the ability.
+	 */
+	protected function input_schema(): array {
+		return array(
+			'type'       => 'object',
+			'properties' => array(
+				'content'         => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
```
dkotter: Apologize for more feedback and this isn't something I had realized myself until yesterday but the Abilities API doesn't support the `sanitize_callback` param. This doesn't cause issues when making a direct API request (it just silently ignores it) but when using the new client-side Abilities API, this will throw an error. I've addressed this in #364 for all of our existing Abilities but will need done here as well. We'll want to remove any `sanitize_callback` params and then ensure we have our own proper sanitization before we use those values

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -0,0 +1,596 @@
...
+		 * @param string                       $prompt          The prompt string to be sent to the AI model.
+		 * @param string|array<string, string> $context         The context to generate suggestions from.
+		 * @param string                       $taxonomy        The taxonomy slug being suggested for (e.g., 'post_tag', 'category').
+		 * @param array<string>                $assigned_terms  Terms already assigned to the post.
+		 * @param array<string>                $available_terms Available terms to suggest from.
+		 */
+		$prompt = (string) apply_filters( 'wpai_content_classification_prompt', $prompt, $context, $taxonomy, $assigned_terms, $available_terms );
+
+		// Generate the suggestions using the AI client with structured output.
+		$result = wp_ai_client_prompt( $prompt )
```
dkotter: Sorry, another recent thing we've changed is ensuring we have proper `is_supported_*` checks before making a request. I've made that change in #362 so you can copy what's done there. At a high-level, I add new `get_prompt_builder` methods to each Ability which puts together the prompt builder and runs our supported check

includes/Abilities/Content_Classification/system-instruction.php
```diff
@@ -0,0 +1,28 @@
...
+
+Rules:
+- The taxonomy to suggest terms for is wrapped in the <taxonomy> tag.
+- Suggest as many relevant terms as you can identify from the content.
+- The "term" field must contain ONLY the human-readable tag or category name (1-3 words).
+- Confidence should reflect relevance: 1.0 = perfect match, 0.5 = somewhat relevant. Only suggest terms with confidence >= 0.5.
+- Do not suggest duplicate or near-duplicate terms.
+- Do not suggest terms listed in <assigned-terms> — they are already applied to this post.
+- When the <available-terms> tag is provided, strongly prefer selecting from those terms.
+- Prioritize specificity and relevance over breadth.
```
dkotter, line 27: Yet another thing we've gotten recent feedback on is trying to ensure the LLM returns content that matches the language of the passed-in content. I've taken a stab at that in #357, might want something similar here

dkotter (changes requested):

includes/Abilities/Content_Classification/Content_Classification.php
TylerB24890: Made some updates to support this.

dkotter (commented):
Code looks good here to me but in testing, I'm getting the following error:

`Bad Request (400) - Invalid schema for response_format 'response_schema': In context=(), 'additionalProperties' is required to be supplied and to be false.`

Note I only have OpenAI connected, haven't tested with other Providers

TylerB24890:
> Code looks good here to me but in testing, I'm getting the following error:
> 
> `Bad Request (400) - Invalid schema for response_format 'response_schema': In context=(), 'additionalProperties' is required to be supplied and to be false.`
> 
> Note I only have OpenAI connected, haven't tested with other Providers

Must be an OpenAI thing -- it's working with Gemini. I'll debug and push up a fix.

TylerB24890:
> > Code looks good here to me but in testing, I'm getting the following error:
> > `Bad Request (400) - Invalid schema for response_format 'response_schema': In context=(), 'additionalProperties' is required to be supplied and to be false.`
> > Note I only have OpenAI connected, haven't tested with other Providers
> 
> Must be an OpenAI thing -- it's working with Gemini. I'll debug and push up a fix.

@dkotter I added the `additionalProperties` key to the schema and OpenAI started working, however Gemini will no longer work until https://github.com/WordPress/ai-provider-for-google/pull/18 is merged.

dkotter:
> > Must be an OpenAI thing -- it's working with Gemini. I'll debug and push up a fix.
> 
> @dkotter I added the `additionalProperties` key to the schema and OpenAI started working, however Gemini will no longer work until [WordPress/ai-provider-for-google#18](https://github.com/WordPress/ai-provider-for-google/pull/18) is merged.

Ah, okay, I think that's fine, ideally that PR gets merged and released soon.

Testing again with OpenAI, I am seeing results now but when selecting a term, it's now setting the original taxonomy as the term parent, so I end up with a `category` term and `post_tag` term:

category term

post_tag term


In addition, because tags aren't heirarchical, this throws an error anytime I try and save a generated tag, though I only see this error in the console, never reflected in the UI. Fixing the above bug will fix the saving of suggested tags I believe, but if possible, would be ideal to surface any areas so the user knows what's happening. In my case, I clicked on a tag, it disappeared and nothing else happened. I then opened the console and could see the error there.

Console error when adding a tag

TylerB24890:
> > > Must be an OpenAI thing -- it's working with Gemini. I'll debug and push up a fix.
> > 
> > 
> > @dkotter I added the `additionalProperties` key to the schema and OpenAI started working, however Gemini will no longer work until [WordPress/ai-provider-for-google#18](https://github.com/WordPress/ai-provider-for-google/pull/18) is merged.
> 
> Ah, okay, I think that's fine, ideally that PR gets merged and released soon.
> 
> Testing again with OpenAI, I am seeing results now but when selecting a term, it's now setting the original taxonomy as the term parent, so I end up with a `category` term and `post_tag` term:
> 
> In addition, because tags aren't heirarchical, this throws an error anytime I try and save a generated tag, though I only see this error in the console, never reflected in the UI. Fixing the above bug will fix the saving of suggested tags I believe, but if possible, would be ideal to surface any areas so the user knows what's happening. In my case, I clicked on a tag, it disappeared and nothing else happened. I then opened the console and could see the error there.


These issues should be resolved. I've added checks to make sure the taxonomy is hierarchical before we attempt to create a parent, and also added checks to ensure the taxonomy isn't returned as the parent. 

I've added an error message as well if it fails to create the term, so the user shouldn't be left wondering what happened in that case.

dkotter:
> These issues should be resolved. I've added checks to make sure the taxonomy is hierarchical before we attempt to create a parent, and also added checks to ensure the taxonomy isn't returned as the parent.
> 
> I've added an error message as well if it fails to create the term, so the user shouldn't be left wondering what happened in that case.

Just tested and this all looks great now. The only thing I'm seeing is I get the "Changes may not be saved" browser message after generating categories or tags and this shows whether I choose to keep one of those or not.

So if I click `Suggest Categories` and then after if I click to go back to the post list, I'll get that message. Or if I click `Suggest Categories` and then click to keep one of the suggestions and then save the post, I'll get that message when trying to go back to the post list screen. I'm assuming there must be some data we're manipulating that shows the post as being "dirty" though I wouldn't expect that, especially in the scenario where I save the post after the generation.

Browser message

TylerB24890:
> Just tested and this all looks great now. The only thing I'm seeing is I get the "Changes may not be saved" browser message after generating categories or tags and this shows whether I choose to keep one of those or not.
> 
> So if I click `Suggest Categories` and then after if I click to go back to the post list, I'll get that message. Or if I click `Suggest Categories` and then click to keep one of the suggestions and then save the post, I'll get that message when trying to go back to the post list screen. I'm assuming there must be some data we're manipulating that shows the post as being "dirty" though I wouldn't expect that, especially in the scenario where I save the post after the generation.

Hmm I'm not able to reproduce this. When I click "Suggest Categories" and then click the back button it navigates back without a notice. If I add a suggested category and click Save, then refresh or navigate away, I also do not get a notice 🤔 -- Are there any console logs you can share that might shed some light on what's happening?

dkotter (approved):

dkotter:
> Hmm I'm not able to reproduce this. When I click "Suggest Categories" and then click the back button it navigates back without a notice. If I add a suggested category and click Save, then refresh or navigate away, I also do not get a notice 🤔 -- Are there any console logs you can share that might shed some light on what's happening?

Sorry, false alarm. Tested on a new post and works great. Disabled all AI features and tried to edit original post and getting the problem so something specific about that post is broken (assuming something wrong in the WP beta release). But not an issue with this PR.

Thanks for all the hard work here!

PR #316, opened 2026-03-18
feat!: Refactor `Experiments` to be a type of `Feature`

justlevine:
## What?

- Part of #234

This PR creates the new `Features` namespace, and makes `Experiments` a type of `Feature`. 
Class and methods contain breaking changes, hooks are `Deprecated` until v1.0

> [!IMPORTANT]
> This PR is based on #315 which should be merged first.

### Hooks
| Old | New | 
|---|---|
| `ai_experiments_register_experiments` | `wpai_register_features`
| `ai_experiments_default_experiment_classes` | `wpai_default_feature_classes` |
| `ai_experiments_enabled` | `wpai_features_enabled` |
| `ai_experiments_initialized` | `wpai_features_initialized` |
| `ai_experiments_experiment_{id}_enabled` | `wpai_feature_{id}_enabled` |

### Options
| Old | New | Notes |
|---|---|---|
| `ai_experiments_enabled` | `wpai_features_enabled` |
| `ai_experiment_{id}_enabled` | `wpai_feature_{id}_enabled` |
| `ai_experiment_{id}_field_{option_name}` | `wpai_feature_{id}_field_{option_name}` |

> [!IMPORTANT]
> There is no migrator class in this PR.

### Class/Methods

| Category | Old name | New name | Notes |
|---|---|---|---|
| Class | `WordPress\AI\Abstracts\Abstract_Experiment` | `WordPress\AI\Abstracts\Abstract_Feature` | New shared base type |
| Class | `WordPress\AI\Contracts\Experiment` | `WordPress\AI\Contracts\Feature` | Interface renamed and expanded |
| Class | `WordPress\AI\Experiment_Loader` | `WordPress\AI\Features\Loader` | Namespace and class name both changed |
| Class | `WordPress\AI\Experiment_Registry` | `WordPress\AI\Features\Registry` | Namespace and class name both changed |
| Class | `WordPress\AI\Experiment_Category` | `WordPress\AI\Features\Feature_Category` | Not a strict 1:1 rename; feature category became the shared base |
| Class | `WordPress\AI\Exception\Invalid_Experiment_Metadata_Exception` | `\InvalidArgumentException` | Old custom exception removed rather than renamed |
| Method | `Abstract_Experiment::load_experiment_metadata()` | `Abstract_Feature::load_metadata()` | Metadata no longer includes `id`; ID is now static |
| Method | `Experiment::get_id()` | `Feature::get_id()` | Same method name, but changed from instance to `static` |
[...]

PR #317, opened 2026-03-18
dev!: rename plugin constants to `WPAI_*`

### LINKED ISSUE #234: Rename plugin from “AI Experiments” to “AI” and define feature vs experiment classification

As the plugin matures and in alignment with the [WordPress 7.0 release cycle](https://make.wordpress.org/core/7-0/), it should transition from being framed purely as “AI Experiments” to simply “AI”, with stable functionalities presented as Features and newer or exploratory functionality remaining labeled as Experiments.  This issue tracks the renaming effort and the classification of which functionalities should be elevated to Features versus retained as Experiments.

Extracting out the next steps / action items from the [WP 7.0 product call with Matt](https://make.wordpress.org/core/2026/02/18/wordpress-7-0-product-review-meeting-with-matt/#comment-48403):

> **AI “wow”**: audit the AI plugin/experiments and pick 1–3 impressive stable features (image editing a leading candidate); position as canonical “AI” plugin with stable/experimental split.

### Proposed Initial Classification (for discussion)

This is a starting point and should be validated through discussion.

Features:
- Image Generation
- Alt Text Generation (primarily added here as the underlying ability is leveraged within image gen)

Experiments:
- Title Generation
- Excerpt Generation
- Content Summarization
- Abilities Explorer

- [x] Rename the plugin from “AI Experiments” to “AI”:
  - Update plugin name, headers, and references.
  - Update settings screen titles and labels.
  - Update documentation and readme references.
  - Ensure WordPress.org metadata aligns.
- [x] Review and finalize which capabilities qualify as stable Features vs Experiments
- [x] Introduce a clear distinction in the UI between Features and Experiments (perhaps with details added into [Experiment Lifecycle doc](https://github.com/WordPress/ai/blob/develop/docs/EXPERIMENT_LIFECYCLE.md)?)
- [x] Ensure experimental labeling remains visually clear where applicable

justlevine:
## What?

- Part of #234

Replaces the `AI_EXPERIMENTS_*` prefix with `WPAI_*`

> [!IMPORTANT]
> This PR is based on #316 which should be merged first.

## Why?

Plugin rename makes `AI_EXPERIMENTS` non-sensical. While these are technically breaking changes it's rare for 3pp to use plugins.

### Use of AI Tools

GitHub Copilot Autocomplete

## Screenshots or screencast 

|Before|After|
|-|-|
|||

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-317-b32272a35f23b8c2c18efee366fcd850c7fd551c.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Migrations/Credential_Migration.php
```diff
@@ -75,7 +75,7 @@ public function run(): void {
 		}
 
 		$this->maybe_migrate_credentials();
-		update_option( self::VERSION_OPTION, AI_EXPERIMENTS_VERSION );
+		update_option( self::VERSION_OPTION, self::TARGET_VERSION );
```
gziolo: Does it make sense to use this const if it can be replaced with `WPAI_VERSION`? This one is a bit confusing. Maybe it's correct as is.

justlevine: The whole migration flow needs to change IMO. Plan tonight is to experiment on this since we need to migrate options too.

For context here, after a migration script runs it should set the version to the migration version, so other routines (e.g. the aforementioned 0.5 -> 0.6 ) can run subsequently and failures are recoverable.
Then when all the migrations are done, the latest WPAI_VERSION should be set. That last part is still missing in this PR since 0.5.0 equals latest right now, but fixing it won't require us changing the const.

gziolo: It's something that someone familiar with the migration logic should confirm.

gziolo (dismissed):
LGTM\! The constant renaming is clean and straightforward. As noted in the inline comment, I'm not familiar with this particular logic: https://github.com/WordPress/ai/pull/317#discussion_r2954029416.

includes/Migrations/Credential_Migration.php
dkotter: So this was built to only handle migrating credentials. If there are other migrations we want to run, particularly within this file, will need changes here.

The point of this code as it sits is to store the _current_ plugin version (current when this code runs which will depend on when someone updates) so we don't run migration more than once. But if we're going to introduce multiple migration scripts, it's better to store the version that the migration was needed (in this case v0.5.0 introduced credential changes so that's the version that we could store, which this change does that).

I'm fine with this change but also means the option name no longer makes sense (`ai_experiments_version`) and may want to update that to something like `wpai_credential_migration_version`

justlevine: >  but also means the option name no longer makes sense (ai_experiments_version) and may want to update that to something like wpai_credential_migration_version

I don't think we need an option for each individual migration script. If ai_~experiments~features_version is >= to the script version const, that means it ran successfully and we don't need another indicator in the database. So my plan is to relocate `ai_experiments_version` to somewhere more central. 

( PS: [Here's](https://github.com/AxeWP/wp-graphql-headless-login/tree/develop/src/Admin/Upgrade) the conceptual pattern I'll be borrowing from to introduce the 0.5.0->0.6.0 migration and make it a bit more robust for any future compat needs too . )

---

Either way I think I'll revert this to using the current version constant per original intent and change that when the above get's tackled. 🙇

justlevine: Updated to `WPAI_VERSION` in 0b9959049def395684009b4de6e3c9536d04ba27

dkotter (approved):

PR #318, opened 2026-03-18
[Feature]: Issue 240 | SEO Descriptions

### LINKED ISSUE #240: Add meta description suggestions with optional SEO plugin integration

Users want quick, good meta descriptions, but the entry point and storage vary depending on SEO plugins.  So let's look at providing suggestions and integrate when SEO plugins are detected (perhaps if certain, known meta fields for description are available).

Provide a “Suggest meta description” action in the editor (perhaps under/after the Excerpt generation button placement as heavily related?.

Generate based on post content and title, with constraints:

- Target length range (for example 140–160 chars)
- Avoid duplication with post title, avoid keyword stuffing

If an SEO plugin is detected, write suggestions into that plugin’s meta description field (when supported).
If no SEO plugin is detected, store as:

- Post meta that could be used by core later, or
- A plugin-managed meta key plus an copy-to-clipboard option

TylerB24890:
## What?

Closes #240 

This PR introduces a new Experiment "Meta Descriptions" which provides AI generated meta description support. 

## Why?

Meta descriptions are an important aspect of a pages SEO, yet they are often overlooked or lacking in quality. See https://github.com/WordPress/ai/issues/240 for more detail.

## How?

This Experiment introduces a new panel to the block editor `PluginDocumentSettingPanel` for managing the SEO meta description for the current post. The initial view provides a simple button to "Generate Meta Description". On click of the button, a modal window opens which shows the suggested meta descriptions returned from the AI and a `TextAreaControl` input allowing editors to manipulate the suggested descriptions prior to saving them to the post. 

On select of a suggestion, the `TextareaControl` component is populated with the selected suggestion at which point the editor can update the text as needed.

From there, editors have a few options;
* Clicking "Apply" will save the content from the `TextareaControl` component to the appropriate meta key
    * If a supported SEO plugin is installed and active, the plugins meta key will be used. Otherwise a default `_meta_description` key is stored.
 * Clicking "Copy to Clipboard" will copy the content from the `TextareaControl` to the users clipboard.
 * Clicking "Cancel" will close the modal without applying any meta description changes.
 * Clicking "Regenerate suggestions" will send the prompt to the AI again for a new set of suggestions.

Various filters are available for customizing the experiment;
* `wpai_meta_description_prompt` - Attach additional (or modify) context sent to the AI for the meta description
* `wpai_meta_description_result_temperature` - The temperature threshold for the model. Default is `0.7`
* `wpai_meta_description_seo_plugin` - The supported SEO plugins in `['file' => 'path', 'meta_key' => 'key' ]` format.
    * Default supported plugins: YoastSEO, RankMath, All in One SEO, SEOPress
    * This configuration is what determines the meta key the description is saved to.
* `wpai_meta_description_meta_key` - Override the meta key the description is saved to. This will take priority over any plugin configurations.
* `wpai_meta_description` - Override the output of the meta description. This filter only runs when an SEO plugin is not detected.

### Use of AI Tools

* **AI Assistance:** Yes
* **Used for:**
    * Experiment scaffolding
    * Unit test boilerplates
    * Inline documentation and experiment documentation generation
    * Assistance with the system prompt generation

## Testing Instructions

1. Visit the Experiments settings page in wp-admin and enable the **Meta Descriptions** Experiment.
2. Navigate to a post and confirm the **Meta Description** panel exists just below the "Move to Trash" button (should be the top panel above Categories)
3. Click the "Generate meta description" button. The modal window should open with the suggested descriptions available for selection.
4. Select a meta description and confirm the content is dropped into the Text Area for customization.
5. Customize the description and click the Apply button. The Meta Description panel in the sidebar should now show the description with an option to Edit or Regenerate.
6. Save the post and confirm the meta description remains after refresh.
7. Edit the meta description and confirm the content is saved as expected.
8. Regenerate the meta description and confirm the content is saved as expected.
9. Install Yoast SEO plugin
10. Repeat steps 1-9 and confirm the meta description is saved to the YoastSEO meta key instead of the default `_meta_description` key.
11. Repeat step 9-10 for each of the supported plugins.

## Screenshots or screencast 

**Initial State:**
meta-description-initial

**Generated Descriptions:**
meta-description-selection

**Selected Description:**
meta-description-selected

**Saved Description:**
meta-description-generated

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-318-e85f1f3c458a3d12122f712722a234ca021fb233.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
1. Update experiment title to `Meta Description Generation`, though I'm open to collapsing this into the `Excerpt Generation` experiment as they're so closely related (`Excerpt and Meta Description Generation`)

Screenshot 2026-03-19 at 1 05 02 PM

2. Update button text to Title Case from Sentence case

Screenshot 2026-03-19 at 1 06 46 PM

3. Update the loading text of the `Regenerate suggestions` button to `Generating`

Screenshot 2026-03-19 at 1 11 45 PM

4. Generating meta, clicking save on a draft post, exiting the editor, and then returning shows no meta still saved to the post.  Same problem generating meta, publishing a post, exiting the editor, and then coming back into the post shows no meta saved.  So something's off with saving that to the post, here's my setup: WordPress 7.0-beta5-62069 and AI Provider for Google Version 1.0.3.

I've not yet tested across the SEO plugins, will do that once the meta saving issue above is resolved as that seems like it may heavily impact the plugin compatibility/integration.

TylerB24890:
@jeffpaul 

> Update experiment title to Meta Description Generation, though I'm open to collapsing this into the Excerpt Generation experiment as they're so closely related (Excerpt and Meta Description Generation)

I've updated the experiment title to "Meta Description Generation" -- let me know if we should roll this up into the Excerpt UI instead.

> Update button text to Title Case from Sentence case

Done.

> Update the loading text of the `Regenerate suggestions` button to `Generating`

Done.

> Generating meta, clicking save on a draft post, exiting the editor, and then returning shows no meta still saved to the post. Same problem generating meta, publishing a post, exiting the editor, and then coming back into the post shows no meta saved. So something's off with saving that to the post, here's my setup: WordPress 7.0-beta5-62069 and AI Provider for Google Version 1.0.3.

This is because metadata isn't saved with **Drafted** posts by default. I had to define the `revisions_enabled => true` property when registering the `_meta_description` meta field. This has been updated to enable revisions for this meta key so it _should_ save with Drafted posts going forward.

jeffpaul:
> This is because metadata isn't saved with Drafted posts by default. I had to define the revisions_enabled => true property when registering the _meta_description meta field. This has been updated to enable revisions for this meta key so it should save with Drafted posts going forward.

Still seeing this not sticking around in the Meta Description sidebar panel after clicking save (macOS 15.7.4, Firefox 148.0.2, WordPress 7.0-beta6-62088):

https://github.com/user-attachments/assets/0e29d1a9-f23b-424d-b11f-021b6f2808e9

TylerB24890:
> > This is because metadata isn't saved with Drafted posts by default. I had to define the revisions_enabled => true property when registering the _meta_description meta field. This has been updated to enable revisions for this meta key so it should save with Drafted posts going forward.
> 
> Still seeing this not sticking around in the Meta Description sidebar panel after clicking save (macOS 15.7.4, Firefox 148.0.2, WordPress 7.0-beta6-62088):
> 
>  generate-meta-description.mp4

Ah! I got it now. The `register_post_meta()` method was being hooked into `init` inside the `register()` method, but `init` has already run by time `register()` runs, so the meta key was never being registered.

This should be fixed now. Please let me know if you continue to have issues.

**EDIT:** Unclear why the E2E tests are failing here. They're unrelated to the work done in this PR and are untouched 🤔

jeffpaul (dismissed):

includes/Abilities/Meta_Description/Meta_Description.php
```diff
@@ -0,0 +1,319 @@
...
+		 * @since x.x.x
+		 *
+		 * @param float $result_temperature The temperature for the result of the meta description generation.
+		 */
+		$result_temperature = (float) apply_filters( 'wpai_meta_description_result_temperature', 0.7 );
+
+		$results = wp_ai_client_prompt( $prompt )
+			->using_system_instruction( $this->get_system_instruction() )
+			->using_temperature( $result_temperature )
+			->using_candidate_count( $candidate_count )
```
dkotter: As we've seen with Title Generation, candidate count fails with Anthropic and doesn't work as expected when using OpenAI. We should probably remove that and just return a single result (similar to the changes we've done with Title Generation). We could also make individual requests for each candidate we want and combine those results together, though I'd suggest the former approach.

includes/Abilities/Meta_Description/Meta_Description.php
```diff
@@ -0,0 +1,319 @@
...
+					$context
+				)
+			);
+		}
+
+		$prompt = '<content>' . $content . '</content>';
+
+		if ( ! empty( $title ) ) {
+			$prompt .= "\n\n<title>" . $title . '</title>';
+		}
```
dkotter: So the title will already be in our context (assuming someone passed in a post ID as `get_post_context` returns the title as part of the context). I'd suggest either merging this title into that or removing the title from the context

includes/Abilities/Meta_Description/SEO_Integration.php
```diff
@@ -0,0 +1,121 @@
...
+ */
+class SEO_Integration {
+
+	/**
+	 * Fallback meta key when no SEO plugin is detected.
+	 *
+	 * @since x.x.x
+	 * @var string
+	 */
+	public const FALLBACK_META_KEY = '_meta_description';
```
dkotter: Should this be a key prefixed with `wpai`?

includes/Experiments/Meta_Description/Meta_Description.php
```diff
@@ -0,0 +1,153 @@
...
+		return 'meta-description';
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function load_metadata(): array {
+		return array(
+			'label'       => __( 'Meta Description Generation', 'ai' ),
+			'description' => __( 'Generates meta description suggestions with SEO plugin integration', 'ai' ),
```
dkotter:
```suggestion
			'description' => __( 'Generates meta description suggestions and integrates those with various SEO plugins', 'ai' ),
```

includes/Abilities/Meta_Description/SEO_Integration.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	/**
+	 * Detects the currently active SEO plugin.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string|null The slug of the active SEO plugin, or null if none detected.
+	 */
+	public static function detect_active_plugin(): ?string {
```
dkotter, line 78: Since we run this fairly often (I think every page load) should we add some caching here?

includes/Experiments/Meta_Description/Meta_Description.php
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		$post_types = get_post_types( array( 'show_in_rest' => true ), 'names' );
+
+		foreach ( $post_types as $post_type ) {
+			if ( 'attachment' === $post_type ) {
+				continue;
+			}
+
+			register_post_meta(
```
dkotter, line 104: I'm getting the following message here:

`Function register_meta was called incorrectly. Meta keys cannot enable revisions support unless the object subtype supports revisions.`

src/experiments/meta-description/components/CharacterCount.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Renders a color-coded character count indicator.
+ *
+ * Green when within 140–160 range, yellow outside.
+ *
+ * @param props       Component props.
+ * @param props.count The current character count.
+ */
+export default function CharacterCount( {
```
dkotter, line 25: Don't have to use it but WordPress has a count package that can be used to count characters: https://developer.wordpress.org/block-editor/reference-guides/packages/packages-wordcount/

docs/experiments/meta-description.md
```diff
@@ -0,0 +1,473 @@
...
+```
+
+This file returns a string that instructs the AI on how to generate meta descriptions. You can modify the character length requirements, tone, style, or other parameters.
+
+### Filtering the Prompt Content
+
+You can filter the assembled prompt before it is sent to the AI model:
+
+```php
+add_filter( 'ai_meta_description_prompt', function( $prompt, $content, $title ) {
```
dkotter: Filter names here are wrong, should be prefixed with `wpai_`

includes/Experiments/Meta_Description/Meta_Description.php
```diff
@@ -0,0 +1,153 @@
...
+				$post_type,
+				$meta_key,
+				array(
+					'show_in_rest'      => true,
+					'single'            => true,
+					'type'              => 'string',
+					'revisions_enabled' => true,
+					'auth_callback'     => static function () {
+						return current_user_can( 'edit_posts' );
```
dkotter: I think we want to use the `$post_id` parameter that gets passed to this auth callback to ensure the user has permissions on that specific post: `current_user_can( 'edit_post', $post_id )`

dkotter (changes requested):

includes/Abilities/Meta_Description/Meta_Description.php
TylerB24890: Got it. We'll need to update the modal and other things as well in that case. Will work on that update.

src/experiments/meta-description/components/CharacterCount.tsx
TylerB24890: Yeah I'm using that in https://github.com/WordPress/ai/pull/313 & https://github.com/WordPress/ai/pull/331 -- IMO the benefit to using that is that it accurately strips out the HTML and comments.

In this case though we're expecting the LLM to return only text content (no HTML), so using `.length` should be sufficient. We can use the `wordcount` package if you think it's better here though.

includes/Abilities/Meta_Description/SEO_Integration.php
TylerB24890: Added cache here that gets flushed when `deactivated_plugin` is called.

includes/Experiments/Meta_Description/Meta_Description.php
TylerB24890: I've removed the `revisions_enabled` property. I had set that initially when debugging Jeff Paul's issue with drafted posts, but it turns out it was related to the `init` hook.

includes/Abilities/Meta_Description/Meta_Description.php
TylerB24890: I've removed the title from `$context` and pass it separately.

TylerB24890:
@dkotter I've resolved all feedback and removed the `candidate_count` logic. I've also added `is_supported_*` logic as requested in #313 

Instead of presenting a user with multiple suggested SEO descriptions, they are now shown **one** description inside a `textarea` for modification:

Screenshot 2026-04-03 at 11 07 27 AM

includes/Abilities/Meta_Description/SEO_Integration.php
```diff
@@ -0,0 +1,128 @@
...
+	/**
+	 * Detects the currently active SEO plugin.
+	 * Cache is flushed when a plugin is deactivated.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string|null The slug of the active SEO plugin, or null if none detected.
+	 */
+	public static function detect_active_plugin(): ?string {
+		$active_plugin = wp_cache_get( 'wpai_active_seo_plugin', 'wpai' );
```
dkotter: Thoughts about using transients here instead? For a site that doesn't have a cache layer in place, I don't think this will do anything. Whereas with transients, those end up in the cache layer if it exists and if not, falls back to using the database

src/experiments/meta-description/components/MetaDescriptionModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+		<Modal
+			title={ __( 'Meta Description', 'ai' ) }
+			onRequestClose={ onClose }
+			className="ai-meta-description-modal"
+			size="medium"
+		>
+			<div className="ai-meta-description-modal__content">
+				{ /* Generation controls */ }
+				<div className="ai-meta-description-modal__generate">
+					<Button
```
dkotter: Thoughts about moving the location of this button? For me, I'd expect this to show below the textarea, not above. Maybe even inline with the other buttons we show?

tests/Integration/Includes/Abilities/Meta_Description/SEO_IntegrationTest.php
dkotter, line 1: Would be great to get some E2E tests for this as well, once we have everything else settled

includes/Abilities/Meta_Description/SEO_Integration.php
TylerB24890: Definitely a good flag. I'll make that update.

src/experiments/meta-description/components/MetaDescriptionModal.tsx
TylerB24890: I initially added it there because it's also the "Generating..." state indicator. When the modal first loads the textarea is empty, and this tells the user that something is happening. I agree though it does feel a little weird after its loaded being at the top.

I think we can probably get the same point across if it's more inline or just below the textarea, as long as it's pretty obvious. I'll play around with it and see what I come up with.

TylerB24890: @dkotter Thoughts?

**Below Textarea:**
Image

**Inline w/ Actions:**
Image
_Note: Disregard the gray bg on the cancel button. That's where my cursor was when I snapped the screenshot_

dkotter: I'd recommend the second, inline with the other buttons, as that matches the UI we use elsewhere, like image generation and refinement

TylerB24890: Awesome -- that's how I left it so this should be complete. 

If this is looking/working good for you I can put together those E2E tests next. Just let me know if we should update anything else beforehand!

dkotter: This looks good to me, think it's good to move forward on E2E tests.

The only other feedback I'd have is whether we should output this meta description in the scenario where someone doesn't have an SEO plugin installed. Right now you can generate and save it but it isn't anywhere in that situation. Not sure what user expectation is here though I wouldn't be surprised if they'd just assume it will be output

tests/Integration/Includes/Abilities/Meta_Description/SEO_IntegrationTest.php
TylerB24890: Added some basic E2E tests -- unless we restructure how this works completely I think they should be stable through any remaining updates.

dkotter (dismissed):

dkotter:
This all looks good to me, nice work here @TylerB24890! The only last comment/question I had, which I left in a threaded comment but pulling out here for easier tracking, is whether we want to actually render a meta description tag if no SEO plugin is installed?

Right now you can generate the meta description but if you aren't using an SEO plugin (which if we're honest is probably a small percentage of sites) this field is never used anywhere.

Curious for your thoughts here @jeffpaul

TylerB24890:
> This all looks good to me, nice work here @TylerB24890! The only last comment/question I had, which I left in a threaded comment but pulling out here for easier tracking, is whether we want to actually render a meta description tag if no SEO plugin is installed?
> 
> Right now you can generate the meta description but if you aren't using an SEO plugin (which if we're honest is probably a small percentage of sites) this field is never used anywhere.
> 
> Curious for your thoughts here @jeffpaul

That's a good flag. We don't really have a good way to determine if its already being rendered or not unless a known plugin is active (besides parsing the rendered page, which I wouldn't consider a _good_ way)

If we decide to limit this to an SEO plugin being active would we hide the experiment activation all together or just the UI from the block editor?

dkotter:
> If we decide to limit this to an SEO plugin being active would we hide the experiment activation all together or just the UI from the block editor?

I wouldn't personally. I think this is still a good feature to have, even though I'd expect most SEO plugins today to have their own AI integrations. So I think it's good that we change where the data is stored depending on if an SEO plugin is active but I wouldn't disable everything if a plugin is active. Just wondering if we should output the meta tag if no SEO plugin is active but leave everything else as-is.

jeffpaul:
Let's output the meta tag if no SEO plugin is detected, worst case there are dupe tags on that page?

TylerB24890:
> Let's output the meta tag if no SEO plugin is detected, worst case there are dupe tags on that page?

@jeffpaul @dkotter  I've added some logic to hook into `wp_head` & output the meta description if the experiment is enabled and no SEO plugin is detected. Also added unit tests & updated experiment documentation accordingly.

TylerB24890:
@dkotter It appears the inclusion of the `is_supported_*` check here is causing the E2E tests to fail. The `is_supported_for_text_generation()` check requires the AI provider to be fully "connected" at the registration level, which fails in the E2E test environment where the mock plugin only intercepts outgoing HTTP requests.

This is also happening in the recently merged _Content Classification_ experiment E2E tests. I'm pretty positive this did work prior to me merging the most recent `develop` branch though, so I'm not really sure why they're failing now. Any thoughts on how we should approach it?

`generate_text()` returns a `WP_Error` when no provider is available as well -- should we remove the experiment `is_supported_*` check and fallback to the `generate_text` error message?

dkotter (approved):

PR #330, opened 2026-03-20
Add bulk "Generate Alt Text" action to Media Library

### LINKED ISSUE #239: Add generated alt text in Edit Media and bulk generate option in Media Library

Alt text is critical for accessibility and SEO, but it is often missing due to time and effort.  We added alt text generation in #44, but users need an easy, in-context way to generate and review alt text so let's extend what we've done to the Media Library's Edit Media screen and the bulk edit option.

Edit Media integration:

- Add “Generate alt text” button near the alt text field
- Provide suggested alt text (could consider this with a quick apply and edit flow instead of directly adding to the field)

Bulk option:

- Add a Media Library bulk action: “Generate alt text”
- Process selected items and write suggestions into the alt text field (or store as pending suggestions if we want a review step).

zeus2611:
## What?
Closes #239

Adds a **"Generate Alt Text"** bulk action to the WordPress Media Library (list view). When one or more images are selected and the action is applied, alt text is generated sequentially for each image using the existing `ai/alt-text-generation` ability and saved directly to each attachment via the REST API. A live progress notice is shown while processing and replaced with a summary on completion.

## Why?

Alt text generation for individual images already existed in the media modal and attachment edit screen, but there was no way to process multiple images at once. This extends the feature to support bulk workflows, addressing the remaining item from issue #239.

## How?

**PHP (`includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php`):**
- Registers `"Generate Alt Text"` in the Media Library bulk actions dropdown via the `bulk_actions-upload` filter.
- Handles the action via `handle_bulk_actions-upload`: filters selected IDs to image attachments only, then appends `ai_bulk_alt_text=1` and `ai_attachment_ids=<ids>` to the redirect URL.
- Detects the redirect params on page load and enqueues/localizes a new `alt-text-generation-bulk` script with the attachment IDs and a `wp_rest` nonce.

**JavaScript (`src/experiments/alt-text-generation/bulk.ts`):**
- On `DOMContentLoaded`, reads localized data and injects a dismissible WP admin notice.
- Loops through attachment IDs sequentially, calling the existing `runAbility('ai/alt-text-generation', { attachment_id })` utility for each.
- Saves each generated alt text via `apiFetch` to `/wp/v2/media/:id`.
- Updates progress count in the notice after each item; per-item errors are counted but never abort the batch.
- Replaces the notice with a completion summary on finish.

**Build (`webpack.config.js`):**
- Adds `experiments/alt-text-generation-bulk` entry point.

**Tests (`tests/Integration/Includes/Experiments/Alt_Text_Generation/Alt_Text_GenerationTest.php`):**
- 5 new integration tests covering: bulk action registration (enabled/disabled), redirect URL query args for image attachments, non-image filtering, and unrelated action passthrough.

### Use of AI Tools

Implementation was planned with assistance from Claude Code (Anthropic). All code was reviewed, tested, and validated.

## Testing Instructions

1. Activate the plugin locally and ensure the **Alt Text Generation** experiment is enabled.
2. Upload at least 2–3 images to the Media Library.
3. Switch to **List View** in the Media Library (`upload.php`).
4. Select 2 or more images using the checkboxes.
5. Open the **Bulk actions** dropdown — verify **"Generate Alt Text"** appears.
6. Select it and click **Apply**.
7. Confirm the page redirects back to `upload.php` and a progress notice appears: `"Generating alt text: 0 / N…"`.
8. Watch the counter increment as each image is processed.
9. On completion, verify the notice shows a summary: `"Alt text generated for all N images."`.
10. Open each processed image's Edit Media screen and confirm the **Alt Text** field is populated.

1. Tab to the Media Library list view and use `Space` to check image rows.
2. Tab to the **Bulk actions** dropdown and use arrow keys to select **"Generate Alt Text"**.
3. Tab to the **Apply** button and press `Enter`.
4. Verify the progress notice is injected and focusable; the dismiss button (`×`) is reachable by keyboard.

## Screenshots or screencast

https://github.com/user-attachments/assets/d3d388bf-be31-4fa1-8805-04a6b9504814

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-330-6a00850f1afe8dfefb2cad490266598cf5a696f3.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
Looks like there must be a character limit being enforced somewhere as this one (of two sample images used for testing) had an ellipses (`...`) added to the end of the Alt Text field:

Screenshot 2026-03-23 at 3 17 06 PM

I would say we either should render the entire alt text provided or enforce a character limit in the alt text generation prompt?

zeus2611:
Thanks for catching this @jeffpaul — the truncation comes from a hard limit in the Ability class ([Alt_Text_Generation.php:207-209](https://github.com/WordPress/ai/blob/develop/includes/Abilities/Image/Alt_Text_Generation.php#L207-L209)).

This affects all alt text generation (individual + bulk). The `...` gets stored literally in the alt text field, which means screen readers will read it aloud as "dot dot dot" — cutting off meaning and hurting the accessibility we're trying to improve.

The system instruction already soft-limits output to 125 characters ("when possible"). Per [WCAG image alt text guidelines](https://www.w3.org/WAI/tutorials/images/tips/), alt text should be "succinct", not artificially truncated. A full 150-character description serves users better than one cut at 122 characters with `...` appended.

Should we remove the hard truncation and rely on the prompt guidance?

dkotter:
> Thanks for catching this @jeffpaul — the truncation comes from a hard limit in the Ability class ([Alt_Text_Generation.php:207-209](https://github.com/WordPress/ai/blob/develop/includes/Abilities/Image/Alt_Text_Generation.php#L207-L209)).
> 
> This affects all alt text generation (individual + bulk). The `...` gets stored literally in the alt text field, which means screen readers will read it aloud as "dot dot dot" — cutting off meaning and hurting the accessibility we're trying to improve.
> 
> The system instruction already soft-limits output to 125 characters ("when possible"). Per [WCAG image alt text guidelines](https://www.w3.org/WAI/tutorials/images/tips/), alt text should be "succinct", not artificially truncated. A full 150-character description serves users better than one cut at 122 characters with `...` appended.
> 
> Should we remove the hard truncation and rely on the prompt guidance?

Yeah, this limit has been in place since the original PR (see [here](https://github.com/WordPress/ai/pull/156/changes#diff-8f195bc1cf561e9f8f5a286547e480487e8b9ed9e603a30f0fad31e6ebe8716fR207-R210)). I'm not sure on the "why" here though I'd suggest we just remove that limit. I think it's fine to suggest a limit in our system instructions (though LLM's are notorious for not following length requirements) but I don't think we should truncate ourselves

zeus2611:
Removed the hard 125-character truncation in `includes/Abilities/Image/Alt_Text_Generation.php` per @dkotter's feedback. The system instruction's soft limit ("Keep it under 125 characters when possible") remains as guidance for the AI. No tests were affected — no existing tests asserted truncation behavior.

src/experiments/alt-text-generation/bulk.ts
```diff
@@ -0,0 +1,153 @@
...
+			total
+		);
+	}
+
+	const successCount = processed - errors;
+
+	if ( errors === 0 ) {
+		statusParagraph.textContent = sprintf(
+			// translators: %d: number of images
+			__( 'Alt text generated for all %d images.', 'ai' ),
```
dkotter: If only a single image is selected, this message isn't accurate:

`Alt text generated for all 1 images`

We should look to use `_n` here to pass in a singular and plural string

includes/Abilities/Image/Alt_Text_Generation.php
```diff
@@ -24,15 +24,6 @@
...
 class Alt_Text_Generation extends Abstract_Ability {
 
-	/**
-	 * The maximum character length for generated alt text.
-	 *
-	 * @since 0.3.0
-	 *
-	 * @var int
-	 */
-	protected const MAX_ALT_TEXT_LENGTH = 125;
```
dkotter, line 34: This is also referenced in our documentation. We should remove that section: https://github.com/WordPress/ai/blob/6c2865e72aab79713c7970b3bfb86d542de7b310/docs/experiments/alt-text-generation.md#customizing-maximum-alt-text-length

src/experiments/alt-text-generation/bulk.ts
```diff
@@ -0,0 +1,153 @@
...
+	if ( ! data || ! data.attachmentIds || data.attachmentIds.length === 0 ) {
+		return;
+	}
+
+	const { attachmentIds, nonce } = data;
+	const total = attachmentIds.length;
+	let processed = 0;
+	let errors = 0;
+
+	apiFetch.use( apiFetch.createNonceMiddleware( nonce ) );
```
dkotter: Is this needed? I believe `apiFetch` already will add a nonce so I think this is adding a second nonce. Won't cause problems but not sure it's needed

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -197,6 +200,84 @@ public function render_attachment_meta_box( \WP_Post $post ): void {
...
+	 * @param string        $doaction     The bulk action being performed.
+	 * @param list<int>     $post_ids     The list of post IDs to process.
+	 * @return string The redirect URL, possibly with bulk alt text query args appended.
+	 */
+	public function handle_bulk_action( string $redirect_url, string $doaction, array $post_ids ): string {
+		if ( 'ai_generate_alt_text' !== $doaction ) {
+			return $redirect_url;
+		}
+
+		$image_ids = array_values( array_filter( $post_ids, 'wp_attachment_is_image' ) );
```
dkotter, line 236: We could add a user check here as well. The alt text Ability does already check if  the user has edit permissions on the attachment before generating alt text so we don't need that check here but could be nice to add a similar user check just to be safe

src/experiments/alt-text-generation/bulk.ts
```diff
@@ -0,0 +1,153 @@
...
+				throw new Error( __( 'Empty response received.', 'ai' ) );
+			}
+
+			await apiFetch( {
+				path: `/wp/v2/media/${ id }`,
+				method: 'POST',
+				data: { alt_text: altText },
+			} );
+		} catch {
+			errors++;
```
dkotter: Minor but might be nice to surface which images failed so a user can maybe debug those further. Right now they just get told failures happened but no information on what the failure was or which ones failed

dkotter (changes requested):
Tested and overall works good, left a handful of comments. Also would be great to get E2E tests here to cover these changes

src/experiments/alt-text-generation/bulk.ts
zeus2611: Replaced the errors counter with a failedIds array. The error message now shows which attachment IDs failed — e.g. "Alt text generated for 3 of 5 images. Failed attachment IDs: 123, 456."

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
zeus2611: Added `current_user_can( 'upload_files' )` check alongside the `$doaction` guard in `handle_bulk_action()`. The ability's own permission callback still runs per-attachment, but this provides an early exit at the bulk action level.

src/experiments/alt-text-generation/bulk.ts
zeus2611: Agreed — removed `apiFetch.use(createNonceMiddleware(nonce))` from the JS and the `nonce` field from `localize_script` on the PHP side. `apiFetch` handles nonce injection automatically.

includes/Abilities/Image/Alt_Text_Generation.php
zeus2611: Removed the "Customizing Maximum Alt Text Length" section and updated three other references to the 125-char truncation throughout `docs/experiments/alt-text-generation.md`.

src/experiments/alt-text-generation/bulk.ts
zeus2611: Updated — now using `_n()` for proper singular/plural handling:
- Singular: "Alt text generated for %d image."
- Plural: "Alt text generated for all %d images."

jeffpaul (dismissed):

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -197,6 +200,83 @@ public function render_attachment_meta_box( \WP_Post $post ): void {
...
+	 *
+	 * @param array<string, string> $actions The existing bulk actions.
+	 * @return array<string, string> The bulk actions with the generate alt text option added.
+	 */
+	public function register_bulk_action( array $actions ): array {
+		if ( ! $this->is_enabled() ) {
+			return $actions;
+		}
+
+		$actions['ai_generate_alt_text'] = __( 'Generate Alt Text', 'ai' );
```
dkotter: To be safe, should probably use the `wpai` prefix for this (both here and wherever it is used)
```suggestion
		$actions['wpai_generate_alt_text'] = __( 'Generate Alt Text', 'ai' );
```

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -197,6 +200,83 @@ public function render_attachment_meta_box( \WP_Post $post ): void {
...
+		$image_ids = array_values( array_filter( $post_ids, 'wp_attachment_is_image' ) );
+
+		if ( empty( $image_ids ) ) {
+			return $redirect_url;
+		}
+
+		return add_query_arg(
+			array(
+				'ai_bulk_alt_text'  => 1,
+				'ai_attachment_ids' => implode( ',', array_map( 'absint', $image_ids ) ),
```
dkotter: Similar with these:


```suggestion
				'wpai_bulk_alt_text'  => 1,
				'wpai_attachment_ids' => implode( ',', array_map( 'absint', $image_ids ) ),
```

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -197,6 +200,83 @@ public function render_attachment_meta_box( \WP_Post $post ): void {
...
+			return $redirect_url;
+		}
+
+		$image_ids = array_values( array_filter( $post_ids, 'wp_attachment_is_image' ) );
+
+		if ( empty( $image_ids ) ) {
+			return $redirect_url;
+		}
+
+		return add_query_arg(
```
dkotter, line 242: Because we use query args to decide when to run generation, if someone refreshes the page or uses the browser back/forward buttons and ends back on the page, alt text generation will trigger again. After generation we should strip out the query args to prevent this

dkotter (changes requested):
Left a few more comments and would still want some E2E tests here before we merge this

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
zeus2611: Renamed to `wpai_generate_alt_text` throughout — action key, guard check, and all test references updated.

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
zeus2611: Updated to `wpai_bulk_alt_text` and `wpai_attachment_ids` in `handle_bulk_action()`, `maybe_enqueue_bulk_script()`, and all tests.

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
zeus2611: Added `window.history.replaceState()` in `bulk.ts` to strip `wpai_bulk_alt_text` and `wpai_attachment_ids` from the URL after processing completes. Uses `replaceState` (not `pushState`) so no new history entry is created — refresh and back/forward won't re-trigger generation.

zeus2611:
Added E2E tests in `tests/e2e/specs/experiments/alt-text-generation-bulk.spec.js` covering: bulk action visibility in list view, generation flow with progress notice, URL cleanup after completion, and hidden action when experiment is disabled.

dkotter (dismissed):

dkotter:
Looks like some new failing E2E tests here (outside of the known failures we have that are being fixed elsewhere). I think the issue here is we toggle over to the list media view and then we don't toggle back to the grid view, so those tests are failing. I can take a look once we have the existing tests fixed up in a separate PR as it's hard to parse through that right now

zeus2611:
@dkotter All CI tests have passed successfully. I believe this is ready to be merged.

dkotter (approved):

PR #302, opened 2026-03-11
feat: Image Provenance experiment and CDN worker templates (C2PA image signing)

erik-sv:
## Changes since last review

- **Rebased onto PR #294** (which is itself rebased onto latest `develop` 0.6.0)
- **`Image_Provenance` ported to `Abstract_Feature` API** — matches the 0.6.0 feature model introduced in #294
- **121 combined tests passing locally** (102 from #294 + 19 from this PR)
- **CDN workers unchanged** — platform-specific JS/Rust/Node code has no PHP dependency and required no porting

---

## Summary

Adds an **Image Provenance** experiment that signs image attachments with C2PA manifests on upload and injects `C2PA-Manifest-URL` response headers for singular posts with featured images. Also adds CDN edge worker templates for Cloudflare, Lambda@Edge, and Fastly so CDN-served images can propagate the provenance header downstream.

> **Depends on #294** — reuses `C2PA_Manifest_Builder`, `Local_Signer`, and `Signing_Interface` from the Content Provenance namespace. Merge #294 first; the diff against `develop` will then show only the Image Provenance additions.

## What this adds

- **`Image_Provenance` experiment** — hooks `add_attachment` to sign JPEG/PNG/WebP/GIF uploads; stores C2PA manifest JSON and canonical URL in post meta
- **`C2PA-Manifest-URL` header injection** — on singular pages with a signed featured image, injects the header pointing to the manifest REST endpoint
- **REST endpoints** — lookup by URL and retrieve manifest by attachment ID
- **CDN worker templates** — Cloudflare Worker (with KV caching), AWS Lambda@Edge, and Fastly Compute@Edge; all call the WordPress REST lookup endpoint
- **10 integration tests** covering upload signing, MIME filtering, REST lookup, manifest retrieval, header injection, and non-singular guard

## Upload and signing flow

```mermaid
flowchart TD
    A["Media upload (JPEG, PNG, WebP, GIF)"] --> B["add_attachment hook"]
    B --> C{"Supported MIME type?"}
    C -->|No| D["Skip - no meta written"]
    C -->|Yes| E["wp_get_attachment_url()"]
    E --> F["Strip query params to get canonical URL"]
    F --> G["C2PA_Manifest_Builder build c2pa.created"]
    G --> H["Local_Signer RSA-2048 via OpenSSL"]
    H -->|Error| I["Store _c2pa_image_status = error"]
    H -->|Success| J["Write post meta"]
    J --> J1["_c2pa_image_manifest"]
[...]

PR #331, opened 2026-03-20
[Feature]: Issue 300 | Content Resizing

### LINKED ISSUE #300: New Experiment: Content Resizing
## Feature Overview & Value

**Summary:** This feature allows authors to select existing text blocks and use AI to modify their length or density without changing the core meaning with an **Inline Transformation** flow. The experience should feel like a "Smart Edit" where the text morphs in place, allowing for quick iteration.
**Why it matters:** Editors often struggle with pacing. A section might be too verbose (needs condensing) or a key point might feel underdeveloped (needs expanding). This tool acts as an instantaneous editor.

## Use Cases

**The "TL;DR":** A user highlights a long, complex technical paragraph and chooses **Condense** to make it punchier for mobile readers.
**Fleshing out ideas:** A user writes a single sentence premise and chooses **Expand** to have the AI extrapolate it into a full paragraph.
**Fitting Layouts:** A designer needs a paragraph to be exactly 3 lines long to match a column layout; they use this tool to adjust the length accordingly.

### 1. Entry Point (The Block Toolbar)

* **Trigger:** The user selects a standard **Paragraph Block** (or other blocks that have text exposed like Heading and Cover Blocks).
* **The Icon:** A native toolbar button (Sparkle or other AI icon) appears alongside standard controls like Bold/Italic/Link.
* **The Menu:** Clicking the icon opens a **Dropdown Menu** (not a modal).
  * _Options:_ "Shorten", "Expand", "Rephrase".

### 2. The Transformation Process

* **Selection:** The user clicks an option (e.g., "Shorten").
* **Processing:** The block enters a "Loading/Thinking" state (pulsing opacity or skeleton loader).
* **Output:** The new text is returned.

###3. The Review State (The Diff)

**Visuals:** The original text is replaced by the new text, but it is highlighted (e.g., a subtle purple background) to indicate it is a "Suggestion."
**Floating Controls:** A small action bar floats above the block:
* **Accept:** Commits the change and removes the highlight.
* **Retry:** Re-runs the specific command.
* **Undo:** Reverts immediately to the original text.

## Technical Requirements

**Supported Providers:**
* Must support OpenAI, Anthropic, and Google.

**Input Data:** The prompt must include the current block content as the source material.
**Block Compatibility:**
* Primary support: Paragraph Blocks.
* Secondary support (Desirable): List Blocks, Heading Blocks, Cover Blocks.

**Authentication:**
If the user attempts to resize text but has no provider connected, trigger the **Connectors Component** (_if available_), else display an error message and direct folks to `Settings > Connectors`.

## UI Mechanics & Component Definitions

###1. The "Transformation Dropdown"

**Placement:** Anchored to the Toolbar button.
**Items:** List of actions with icons (e.g., Arrows pinching in for "Shorten", Arrows pushing out for "Expand").
**Tone Slider (Future/Optional):** Consider adding a "Tone" sub-menu (Professional, Casual) since the backend logic is identical to resizing (_see #186 for more on this_).

### 2. Error States

**Too Short:** If a user tries to "Condense" a 5-word sentence, return a polite notice ("Text is too short to shorten further").
**Auth Failure:** If the API key fails, display a non-intrusive "toast" error rather than breaking the editor UI.

## Credits

_**Props to @linawiezkowiak, @lwoodmansee, and @rachaelcortellessa for the product definition, and @dkotter for technical review.**_

TylerB24890:
## What?

Closes #300 

This feature allows authors to select existing text blocks and use AI to modify their length or density without changing the core meaning with an Inline Transformation flow.

## Why?

Editors often struggle with pacing. A section might be too verbose (needs condensing) or a key point might feel underdeveloped (needs expanding). This tool acts as an instantaneous editor.

## How?

This PR applies a new experiment "Content Resizing" which, when enabled, adds a toolbar to the `core/paragraph` block with various options to have AI resize the content. There are three options available; Rephrase, Shorten, Expand. 

When the block has AI generated content applied, the toolbar button applies an active state providing a visual cue that the block is using AI content. Additionally, an "Undo" action is added to the toolbar dropdown, allowing editors to undo the AI generated content and revert to the original content.

### Use of AI Tools

* **AI Assistance:** Yes
* **Used for:**
    * Experiment scaffolding
    * Inline documentation
    * Unit test generation

## Testing Instructions

1. Enable the "Content Resizing" experiment from the AI settings page.
2. Navigate to a block editor and select a paragraph block.
3. Use the AI icon in the toolbar to resize the content.
    * Test all three resizing actions.
4. When AI content is applied, confirm the toolbar icon shows an active state indicating AI suggested content is being used.
5. When AI content is applied, confirm the "Undo" action exists and functions in the block toolbar.
    * Clicking "Undo AI changes" will revert the block content to its original state.

## Screenshots or screencast 

**Suggested Content:**
ai-resize_suggestion

**Undo Action**
ai-resize_undo

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-331-e44d5cc63bfb4b614ecd70f7bc2ab70d31cab6f4.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


TylerB24890:
@jeffpaul I took a stab at this and have a functional experiment here, however I took a slightly different approach from the described UI in the issue. I opt'd to use a modal UI for accepting the resized content, similar to the [SEO Meta Descriptions UI](https://github.com/WordPress/ai/pull/318).

One thing I think could improve the experience would be to add a direct comparison UI, perhaps render the original content in the modal so editors can see both side by side? Open to your thoughts here.

Also might need some improvement on the `system-instruction.php`. It works as is, but sometimes the model doesn't retain what I would consider critical links, despite the instruction stating to do so.

jeffpaul:
> One thing I think could improve the experience would be to add a direct comparison UI, perhaps render the original content in the modal so editors can see both side by side? Open to your thoughts here.

Yeah, that could be nice, perhaps noting the +/- on characters/words as well?

> Also might need some improvement on the system-instruction.php. It works as is, but sometimes the model doesn't retain what I would consider critical links, despite the instruction stating to do so.

@sidsector9 I know you've done work on this elsewhere, any advice on how to best retain that content/context?

jeffpaul:
1. If we're able to update from the sparkle icon to the plugin icon in the block toolbar, that might be slightly better:

Screenshot 2026-03-23 at 2 36 49 PM

https://github.com/WordPress/ai/blob/develop/.wordpress-org/icon.svg

2. Let's update `Retry` to `Regenerate` to match other similar buttons in the plugin

3. We can probably get rid of the `Discard` button in the modal as the `X` in the upper right effectively does the same

TylerB24890:
@jeffpaul 
> If we're able to update from the sparkle icon to the plugin icon in the block toolbar, that might be slightly better

I created a variation of the plugin icon for the toolbar:
Screenshot 2026-03-24 at 1 02 58 PM

Let me know if you were envisioning something different.

> Let's update Retry to Regenerate to match other similar buttons in the plugin

Done.

> We can probably get rid of the Discard button in the modal as the X in the upper right effectively does the same

Done.

**What's left:**
* Need to add the original text to the modal still (including +/- word count)
* E2E tests

jeffpaul:
1. Let's tweak the hover text here to "Resize Content" (noting that we may add more items into this toolbar button in the future and can worry about iterating on naming then)

Screenshot 2026-03-27 at 11 42 15 AM

2. I'm not sure that we need the "Has AI Content" state here, can probably just stick with only "Resize Content" for now as someone's likely clicking this button whether the content is human-generated-only or has utilized AI for generation

Screenshot 2026-03-27 at 11 43 32 AM

TylerB24890:
@jeffpaul 
> I'm not sure that we need the "Has AI Content" state here, can probably just stick with only "Resize Content" for now as someone's likely clicking this button whether the content is human-generated-only or has utilized AI for generation

Do we want to leave the icon color change or revert it back to its original state completely? I think some indicator that content was AI generated would be helpful, but I defer to you there.

jeffpaul:
> Do we want to leave the icon color change or revert it back to its original state completely? I think some indicator that content was AI generated would be helpful, but I defer to you there.

I'm fine with the icon color staying as-is.  The only UI bits that show that something is AI-generated is things like a button saying "regenerate" or perhaps a block name.  In general I do like the idea of representing what was AI-generated/assisted where feasible, especially with the coming Content Provenance PRs (#294, #302) so I'm open to something here.  The thing I don't like about "Has AI Content" is that it doesn't make it clear that you could still click that button to resize content again.

TylerB24890:
> > Do we want to leave the icon color change or revert it back to its original state completely? I think some indicator that content was AI generated would be helpful, but I defer to you there.
> 
> I'm fine with the icon color staying as-is. The only UI bits that show that something is AI-generated is things like a button saying "regenerate" or perhaps a block name. In general I do like the idea of representing what was AI-generated/assisted where feasible, especially with the coming Content Provenance PRs (#294, #302) so I'm open to something here. The thing I don't like about "Has AI Content" is that it doesn't make it clear that you could still click that button to resize content again.

That makes sense. I've removed the "Has AI Content" label but left the color change. There is still the option to revert the AI content as well when a paragraph is AI _resized_

jeffpaul:
Wondering if we need the "Undo AI changes" option when there's already the "Undo" button in the toolbar that will roll back the same change?

Screenshot 2026-04-16 at 1 14 28 PM

includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,249 @@
...
+	 *
+	 * @since x.x.x
+	 */
+	protected function input_schema(): array {
+		return array(
+			'type'       => 'object',
+			'properties' => array(
+				'content' => array(
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
```
dkotter: Since the Abilities API doesn't actually call `sanitize_callback` we can remove this

includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,249 @@
...
+		/**
+		 * Filters the prompt for the content resizing.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param string $prompt The prompt to use for the content resizing.
+		 * @param string $action The resizing action to perform.
+		 * @return string The filtered prompt.
+		 */
+		$prompt = (string) apply_filters( 'wpai_content_resizing_prompt', $prompt, $args['action'] );
```
dkotter: I would probably move this filter into the `structure_prompt` method, just filtering the prompt prior to returning it

includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,249 @@
...
+		return wp_kses_post( $result );
+	}
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since x.x.x
+	 */
+	protected function permission_callback( $args ) {
+		if ( ! current_user_can( 'edit_posts' ) ) {
```
dkotter: Since we're resizing content that is within an individual post, I'm wondering if we should pass the post ID as an input schema so we can validate edit permission at the post level here?

includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,249 @@
...
+	 */
+	private function structure_prompt( $content, $action = self::ACTION_DEFAULT ) {
+		$prompt_parts = array();
+
+		// Determine the action-specific instruction.
+		$action_desc = 'Rephrase the content using different wording and sentence structure while preserving the exact same meaning, tone, and level of detail. The output should be approximately the same length as the input.';
+		if ( 'shorten' === $action ) {
+			$action_desc = 'Condense the following text to roughly half its current length. Preserve the core meaning, key facts, and tone. Remove redundancy and filler. Do not add new information.';
+		} elseif ( 'expand' === $action ) {
+			$action_desc = 'Expand the following text to roughly 1.5 to 2 times its current length. Add supporting detail, elaboration, or examples that are consistent with the original meaning and tone. Do not introduce contradictory information.';
```
dkotter: It seems like all of this is more system instructions instead of the actual user message. Our `get_system_instruction` method does allow you to pass in data that is then accessible in our system instruction files, I'd suggest we pass the `$action` into that and add this content into our instructions

includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,249 @@
...
+		// Default arguments.
+		$args = wp_parse_args(
+			$input,
+			array(
+				'content' => null,
+				'action'  => self::ACTION_DEFAULT,
+			),
+		);
+
+		$content = normalize_content( $args['content'] ?? '' );
```
dkotter: So this does strip HTML, meaning if someone wants to resize content that has things like bold, italics, links, etc those will all be removed

includes/Abilities/Meta_Description/Meta_Description.php
```diff
@@ -301,14 +301,9 @@ private function get_prompt_builder( string $prompt ) {
...
-		// Return a more specific error if there isn't a model that supports text generation.
-		if ( ! $builder->is_supported_for_text_generation() ) {
-			return new WP_Error(
-				'unsupported_model',
-				esc_html__( 'Meta description generation failed. Please ensure you have a connected provider that supports text generation.', 'ai' )
-			);
-		}
-
-		return $builder;
+		return $this->ensure_text_generation_supported(
```
dkotter, line 313: Looks like we're fixing a bug here unrelated to this PR. I'd suggest pulling this out into a different PR so we can get this merged in

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
```diff
@@ -0,0 +1,343 @@
...
+		fill="currentColor"
+	>
+		<path d="M17.65 6.35A7.96 7.96 0 0012 4c-4.42 0-7.99 3.58-7.99 8s3.57 8 7.99 8c3.73 0 6.84-2.55 7.73-6h-2.08A5.99 5.99 0 0112 18c-3.31 0-6-2.69-6-6s2.69-6 6-6c1.66 0 3.14.69 4.22 1.78L13 11h7V4l-2.35 2.35z" />
+	</svg>
+);
+
+/**
+ * Undo icon: arrow curving back to the left.
+ */
+const ICON_UNDO = (
```
dkotter: Lots of icons inline here, may be nice to move those to their own file and just import here

dkotter (changes requested):
Handful of comments but in testing, works well for me!

In addition, looks like we're still tracking these:

- [ ] Need to add the original text to the modal still (including +/- word count)
- [ ] E2E tests

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
TylerB24890: Yeah I was actually looking for a little input on if these were the right icons to use -- wasn't sure if there was a standard library we should be loading from or if custom icons like this was fine. 

If these look good I can definitely move them to their own files.

dkotter: There is an `Icon` component that WordPress ships with that we could look to use, though not sure it covers all of these or not: https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library. We can also use Dashicons as well. We've done both across this plugin

jeffpaul: Ideally we stray away from Dashicons as a retired component iirc

TylerB24890: I moved the icons to an `icons.tsx` file and instead of using my own AI icon I'm importing the existing one from `./routes/ai-icon.tsx` for consistency.

TylerB24890:
> Wondering if we need the "Undo AI changes" option when there's already the "Undo" button in the toolbar that will roll back the same change?
> 
> Screenshot 2026-04-16 at 1 14 28 PM

@jeffpaul I've removed the icon and `originalContent` attribute that handled the "Undo AI Changes" functionality. The one downfall I see with this is the "Undo" action doesn't retain across page loads. If someone generates AI content for a paragraph, saves and closes the editor, they can no longer "undo" the AI content as the editor has lost its original state. My original implementation allowed them to "Undo" AI changes for as long as the block retains the attribute.

Let me know if you want to keep it with the core "Undo" action or if I should add that logic back here.

FYI I am still storing an attribute indicating there are AI changes on the block so we can color the icon indicator, but the attribute is now much smaller with just a boolean value.

includes/Abilities/Meta_Description/Meta_Description.php
TylerB24890: @dkotter PR here: https://github.com/WordPress/ai/pull/465

TylerB24890:
@dkotter All feedback has been resolved. I opened a new PR for the `MetaDescription` experiment fixes: https://github.com/WordPress/ai/pull/465, added E2E tests, and added the original content to the modal window with a `+/-` word count indicator. 

Please take a look and let me know if you run into any issues!

includes/Experiments/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,94 @@
...
+		return 'content-resizing';
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function load_metadata(): array {
+		return array(
+			'label'       => __( 'Content Resizing', 'ai' ),
+			'description' => __( 'Shorten, expand, or rephrase selected block content using AI.', 'ai' ),
```
dkotter:
```suggestion
			'description' => __( 'Shorten, expand, or rephrase selected block content. Requires an AI connector that includes support for text generation models.', 'ai' ),
```

src/experiments/content-resizing/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const { aiContentResizingData } = window as any;
+
+// Register the aiResized attribute on paragraph blocks so we can visually flag
+// blocks whose content was generated via the content-resizing ability.
+addFilter(
+	'blocks.registerBlockType',
+	'ai/content-resizing-attribute',
+	( settings, name ) => {
+		if ( name !== 'core/paragraph' ) {
```
dkotter: Do we need the is enabled check here? Minor thing as this file should never be loaded if the experiment isn't enabled but for consistency, may want that here as well

includes/Abilities/Content_Resizing/system-instruction.php
```diff
@@ -0,0 +1,38 @@
...
+return <<<INSTRUCTION
+You are an editorial assistant that transforms text content (denoted by `<content>` tags) while preserving meaning and intent. The content may contain inline HTML tags (such as strong, em, a, code).
+
+Goal: {$action_desc}
+
+Requirements:
+- Return only the transformed text, nothing else
+- Do not include any preamble, explanation, or commentary
+- Preserve all inline HTML links present in the original content.
+- Return content in the same format as it was provided.
```
dkotter: Not sure we need both of these as they essentially are asking the same thing. I'd suggest maybe collapsing into one to save a few tokens:


```suggestion
- Return content in the same format as it was provided. For example, preserve any inline HTML like links.
```

includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,232 @@
...
+				'post_id' => array(
+					'type'        => 'integer',
+					'description' => esc_html__( 'The ID of the post to resize content for.', 'ai' ),
+				),
+				'content' => array(
+					'type'        => 'string',
+					'description' => esc_html__( 'The block content to resize.', 'ai' ),
+				),
+				'action'  => array(
+					'type'        => 'enum',
```
dkotter: I believe this should be string
```suggestion
					'type'        => 'string',
```

includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -0,0 +1,232 @@
...
+			return new WP_Error(
+				'content_not_provided',
+				esc_html__( 'Content is required to resize.', 'ai' )
+			);
+		}
+
+		// "shorten" action requires a minimum word count.
+		if (
+			'shorten' === $args['action'] &&
+			str_word_count( wp_strip_all_tags( $content ) ) < self::SHORTEN_MIN_WORDS
```
dkotter: So if I remember correctly (ran into this elsewhere) `str_word_count` doesn't work on non-Latin characters so we may want to replace this with something else

dkotter (changes requested):
Left a few new comments but in testing, this works really well and looks good to me!

includes/Abilities/Content_Resizing/Content_Resizing.php
TylerB24890: I'm passing an `enum` property though, wouldn't this tell it it needs to be one of the defined `enum` property values?

includes/Abilities/Content_Resizing/Content_Resizing.php
TylerB24890: You're right -- I didn't realize that.

I implemented a new `count_words()` helper function that mimics the way the `@wordpress/wordcount` package handles it.

includes/Abilities/Content_Resizing/Content_Resizing.php
dkotter: My understanding (and I could be wrong here, would need to look up the REST schema to validate), is that `enum` is not a valid `type` argument. Instead you use `enum` as it's own property to determine which values are allowed. So in this case the value type we want is a string and then that string should be one of a few values, which is set in the `enum` property. If that doesn't seem correct though, let me know as we have a few other places in our current code that does this

dkotter: Assuming docs are correct, looks like `enum` isn't a valid `type`: https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#primitive-types

dkotter: Was curious to track this to the code-level and seems to back that up: https://github.com/WordPress/wordpress-develop/blob/a95dd0e0351a9fb2ef124ea1b8be63bf6887b946/src/wp-includes/rest-api.php#L2207

TylerB24890: Wow now I'm curious where and when I decided `enum` was a valid `type` property here 😅 

Thanks for the context and reference links! I've updated accordingly.

dkotter (dismissed):

dkotter (approved):

PR #337, opened 2026-03-22
Allow has_ai_credentials() filter fallback for non-API-key connectors

### LINKED ISSUE #336: Support local AI providers without API keys in credential checks
### Description
When only a local/self-hosted AI provider is installed (no API key required), the AI settings page incorrectly reports that no valid AI connector is configured.

Expected behavior:
- Local providers that do not require API-key auth should be able to work with the AI plugin.

Current behavior:
- The AI settings page shows: "The AI plugin requires a valid AI Connector to function properly..." even when a local provider plugin is installed and configured.

1. Install WordPress 7.0 + AI plugin (v0.6.0).
2. Install and activate a local provider plugin, for example:
   - https://github.com/Fueled/ai-provider-for-ollama/
3. Do not configure any cloud API key connector.
4. Open `wp-admin/options-general.php?page=ai`.
5. Observe the error notice that says no valid AI connector is configured.

Image

Image

Error text shown in settings page:

> The AI plugin requires a valid AI Connector to function properly. Verify you have one or more AI Connectors configured here.

Context:
- Local-hosted providers such as Ollama/llama.cpp-style integrations may not need API credentials.

### Environment info
- WordPress version: 7.0
- AI plugin version: 0.6.0
- Local provider example: ai-provider-for-ollama
- Theme: not relevant to reproduce (tested on admin settings page)
- Device/OS: macOS desktop
- Browser: Chrome (also reproducible in other browsers)

- [x] Yes

- [x] Yes (exception: local provider plugin required to reproduce)

- [x] Not sure

### Additional note
A filter-based extension point in `has_ai_credentials()` would allow third-party local providers to declare credential availability for non-API-key connectors while preserving default behavior for API-key connectors.

raftaar1191:
## What?
Closes https://github.com/WordPress/ai/issues/336

Adds a filter-based fallback to `has_ai_credentials()` so third-party/local providers that do not use API keys can still mark credentials as configured.

## Why?
The current configured-state check only returns `true` when an `ai_provider` connector uses `authentication.method = api_key` and the related setting is non-empty. Local/self-hosted connectors (for example connectors using `authentication.method = none`) can be fully usable but still fail this configured check, which triggers the settings notice and blocks expected behavior.

## How?
1. Cache registered connectors in `$connectors = wp_get_connectors()`.
2. Keep existing `api_key` logic unchanged.
3. If no API-key credential is found, call:
   `apply_filters( 'wpai_has_ai_credentials', false, $connectors )`
4. Update the filter docs to reflect the second argument is the full connectors array.

### Use of AI Tools
Drafting and implementation assistance was done with GitHub Copilot (GPT-5.3-Codex). All changes were reviewed and validated before submission.

## Testing Instructions
1. Check out this PR branch.
2. Ensure no AI provider API key options are set.
3. Register an `ai_provider` connector that does not use API-key auth (for example auth method `none`).
4. Add a filter callback to `wpai_has_ai_credentials` returning `true` when that connector is present.
5. Verify `\WordPress\AI\has_ai_credentials()` returns `true`.
6. Verify AI settings no longer show the "requires a valid AI Connector" configured-state notice in this scenario.
7. Remove the filter and verify behavior returns to previous default (`false` without API-key credentials).

1. Open the AI settings page using keyboard navigation only.
2. Verify focus order and controls are unchanged.
3. Confirm the configured-state notice visibility changes only based on credential/filter state, with no keyboard regressions.

## Screenshots or screencast
N/A (no UI/layout changes; behavior-only change).

|Before|After|
|-|-|
|Configured check could not be overridden with connector context for non-API-key providers.|Configured check can be overridden via `wpai_has_ai_credentials` with access to registered connectors.|

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-337-4c5d711c42cff2aac36924b2dcab08d8ea69cbe2.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -331,7 +331,7 @@ function has_ai_credentials(): bool {
 	 * Allows third-party plugins to declare credential availability for
 	 * connectors that do not rely on API key settings.
 	 *
-	 * @since 0.6.1
+	 * @since x.x.x
 	 *
 	 * @param bool  $has_credentials Whether AI credentials are available.
 	 * @param array $connectors      The registered connectors.
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -308,6 +308,8 @@ function has_ai_credentials(): bool {
 
 	$connectors = wp_get_connectors();
 
+	$has_credentials = false;
+
 	foreach ( $connectors as $connector_data ) {
 		if ( 'ai_provider' !== $connector_data['type'] ) {
 			continue;
@@ -322,7 +324,8 @@ function has_ai_credentials(): bool {
 			continue;
 		}
 
-		return true;
+		$has_credentials = true;
+		break;
 	}
 
 	/**
@@ -336,7 +339,7 @@ function has_ai_credentials(): bool {
 	 * @param bool  $has_credentials Whether AI credentials are available.
 	 * @param array $connectors      The registered connectors.
 	 */
-	return (bool) apply_filters( 'wpai_has_ai_credentials', false, $connectors );
+	return (bool) apply_filters( 'wpai_has_ai_credentials', $has_credentials, $connectors );
 }
 
 /**
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -306,7 +306,9 @@ function has_ai_credentials(): bool {
 		return false;
 	}
 
-	foreach ( wp_get_connectors() as $connector_data ) {
+	$connectors = wp_get_connectors();
+
+	foreach ( $connectors as $connector_data ) {
 		if ( 'ai_provider' !== $connector_data['type'] ) {
 			continue;
 		}
@@ -323,7 +325,18 @@ function has_ai_credentials(): bool {
 		return true;
 	}
 
-	return false;
+	/**
+	 * Filters whether AI credentials are available.
+	 *
+	 * Allows third-party plugins to declare credential availability for
+	 * connectors that do not rely on API key settings.
+	 *
+	 * @since 0.6.1
+	 *
+	 * @param bool  $has_credentials Whether AI credentials are available.
+	 * @param array $connectors      The registered connectors.
+	 */
+	return (bool) apply_filters( 'wpai_has_ai_credentials', false, $connectors );
 }
 
 /**


includes/helpers.php
```diff
@@ -323,7 +325,18 @@ function has_ai_credentials(): bool {
...
 	}
 
-	return false;
+	/**
+	 * Filters whether AI credentials are available.
+	 *
+	 * Allows third-party plugins to declare credential availability for
+	 * connectors that do not rely on API key settings.
+	 *
+	 * @since 0.6.1
```
dkotter:
```suggestion
	 * @since x.x.x
```

includes/helpers.php
```diff
@@ -323,7 +325,18 @@ function has_ai_credentials(): bool {
...
+	 *
+	 * @param bool  $has_credentials Whether AI credentials are available.
+	 * @param array $connectors      The registered connectors.
+	 */
+	return (bool) apply_filters( 'wpai_has_ai_credentials', false, $connectors );
```
dkotter: Wondering if this filter should apply in all scenarios. Right now, this only fires if we don't have any AI credentials. Feels like we might want to always filter the return value, both in situations where AI credentials are found and not found

dkotter (changes requested):

includes/helpers.php
raftaar1191: @dkotter Thanks for the suggestion. Updated the @since annotation to use a placeholder version (x.x.x) as recommended.

includes/helpers.php
raftaar1191: @dkotter Good point — I agree this should be consistent across all scenarios.
I’ve updated the logic so the wpai_has_ai_credentials filter is now applied unconditionally, allowing it to modify the return value whether credentials are present or not.

This should provide better flexibility for third-party integrations. Let me know if you’d like any adjustments.

dkotter (approved):

PR #153, opened 2025-12-19
Improve AI Experiments page UI with header, accordion, and entry points

Jameswlepage:
CleanShot 2025-12-19 at 11 50 42@2x

## Summary

This PR enhances the AI Experiments settings page with improved UI and extensibility:

- **Full-width admin header** with AI icon, page title, and quick-access buttons for Docs and Contribute
- **Refined experiment list layout** using a cleaner grid with border separators instead of card-per-item
- **Entry points system** allowing experiments to display contextual links (Dashboard, Try, etc.) next to their toggle
- **Settings accordion drawer** for experiments with custom settings, using accessible toggle button with slide animation
- **Shared admin styles** in new `_common.scss` for reuse across AI admin pages (MCP, Request Logs, etc.)

## Screenshots

The experiments page now has:
1. A header matching the WordPress privacy-settings style with icon and action buttons
2. White page background for better visual consistency
3. Experiments displayed in a cleaner list format with entry point links inline
4. Collapsible settings drawers with smooth animations

## Technical Changes

- `Abstract_Experiment`: Added `get_entry_points()` and `has_settings()` methods for UI extensibility
- `Settings_Page.php`: New header structure, grid layout, entry points display, and accordion drawer with inline JS
- `helpers.php`: Added `get_ai_icon_svg()` and `get_ai_icon_data_uri()` helpers for icon rendering
- `_common.scss`: New shared styles for admin headers and white page backgrounds
- `settings/index.scss`: Updated styles for grid layout, entry point links, settings toggle, and drawer animation

## Dependency Note

The following feature PRs will need to be rebased onto this PR (or updated after merge) as they also modify the experiments page:

- #148 - Extended Providers
- #149 - AI Request Logging
- #151 - Type Ahead
- #152 - MCP

These PRs currently include similar header styling changes that would become redundant after this PR merges. They should be updated to:
1. Remove any duplicated header/common styles now in `_common.scss`
2. Use the new `get_entry_points()` method to add their entry point links
3. Use the new `has_settings()` method if they have custom settings

## Test Plan

- [ ] Visit Settings > AI Experiments and verify the new header appears with icon, title, Docs and Contribute buttons
- [ ] Verify page has white background
- [ ] Enable experiments and verify entry point links appear for experiments that define them
- [ ] If an experiment has settings, verify the gear icon appears and clicking it toggles the settings drawer
- [ ] Test accordion animation is smooth (0.2s slide-down)
- [ ] Verify toggle states are properly indicated (blue when expanded)
[...]

PR #149, opened 2025-12-19
Add AI Request Logging experiment with observability dashboard

Jameswlepage:
ai-logs

## What?

Adds a comprehensive AI Request Logging experiment that provides observability for all AI operations:
- Real-time logging of every AI request (provider, model, tokens, duration, cost estimate)
- React-powered admin dashboard under **Settings → AI Request Logs**
- REST API endpoints for programmatic access
- Configurable retention period and automatic cleanup

## Why?

AI operations are opaque—users have no visibility into what requests are being made, how much they cost, or whether they're succeeding. This experiment provides the observability surface needed for debugging, cost tracking, and usage analysis.

## How?

Architecture uses the decorator pattern to wrap the SDK's HTTP transporter:

- `Logging_Integration` wraps the transporter on `wp_loaded`/`admin_init` using the public `setHttpTransporter()` API
- `Logging_Http_Transporter` decorates requests, capturing timing and delegating to `Log_Data_Extractor`
- `Log_Data_Extractor` parses request/response payloads with 4 filter hooks for extensibility:
  - `ai_request_log_providers` - customize provider detection
  - `ai_request_log_context` - filter context data
  - `ai_request_log_tokens` - custom token extraction
  - `ai_request_log_kind` - request type detection
- `AI_Request_Log_Manager` coordinates schema, repository, and cost calculation
- `AI_Request_Cost_Calculator` estimates costs based on provider pricing tables

Supports 14 providers out of the box: OpenAI, Anthropic, Google, Fal, Cloudflare, Groq, Grok, HuggingFace, DeepSeek, Ollama, OpenRouter, Azure, Cohere, and Mistral.

## Testing Instructions

1. Enable the AI Request Logging experiment in **Settings → AI → Experiments**
2. Navigate to **Settings → AI Request Logs** and toggle "Logging enabled"
3. Trigger an AI-powered feature (e.g., Title Generation, Excerpt Generation)
4. Verify the log table populates with request details (provider, model, tokens, duration)
5. Test filtering by provider, status, and date range
6. Verify "Purge logs" clears all entries
7. Disable the experiment and confirm no new logs appear

[...]

PR #340, opened 2026-03-25
Add modern wp-build DataForm route for AI settings page

### LINKED ISSUE #103: Modernize admin UI to use React and WP components

The plugin currently uses a legacy PHP-based admin settings experience.  In alignment with the recommendations from [core ticket #64308](https://core.trac.wordpress.org/ticket/64308), let's look to update the plugin’s admin interfaces (provider selection, experiment toggles, and other admin-side functionality) to use the modern WP design system and React-based components.

- [ ] Replace the current PHP-rendered settings UI with a React implementation that leverages the `@wordpress/components` package.
- [ ] Follow the patterns and styling defined in the official Figma Design Library and ensure consistency with the Site Editor and other canonical plugins.
- [ ] Refactor admin screens for: Provider selection and configuration
- [ ] Refactor admin screens for: Experiment toggles and descriptions
- [ ] Refactor admin screens for: Any future experiment settings or diagnostics interfaces
- [ ] Update plugin documentation to note that all new and updated UIs should be built using React and the WordPress Components design system.
- [ ] Reference prior examples such as [Preferred Languages PR #546](https://github.com/swissspidy/preferred-languages/pull/546) as implementation guidance.

jorgefilipecosta:
## Summary

- Adds a new wp-build route (`routes/ai-home/`) that renders the AI settings page using `@wordpress/dataviews` DataForm and `@wordpress/admin-ui` Page components
- Exposes AI settings via the REST API (`show_in_rest`) so the frontend can read/save them
- Replicates the existing settings page layout using DataForm card layouts with collapsible sections
- Adds a page header with AI flask icon, subtitle, and Docs/Contribute action buttons (matching the design from #153)

## Key changes

- **`routes/ai-home/stage.tsx`** — Full DataForm-based settings page with:
  - General Settings card (global AI toggle)
  - Editor Experiments card (6 experiment toggles)
  - Admin Experiments card (1 experiment toggle)
  - Disabled checkbox state when global AI is off
  - REST API integration for loading/saving settings
  - Page header with AI icon and action buttons
- **`routes/ai-home/style.scss`** — Centered layout (max-width 680px) matching the Connectors page design
- **`includes/Settings/Settings_Registration.php`** — Added `show_in_rest => true` to all settings
- **`package.json`** — Added `@wordpress/admin-ui` and `@wordpress/dataviews` dependencies

## Note on `tools/ensure-boot-asset.cjs`

The `wp-build` generated page template (`page-wp-admin.php`) hardcodes a reference to `build/modules/boot/index.min.asset.php` for dependency metadata. In Gutenberg core this file is generated from `packages/boot/`, but plugins don't have a local boot package — they rely on WordPress core's `@wordpress/boot` script module.

The `tools/ensure-boot-asset.cjs` script creates a small PHP proxy that defers to core's boot asset at runtime. This is necessary until `@wordpress/build` adds plugin-aware boot asset resolution (i.e., falling back to core's `wp-includes/js/dist/script-modules/boot/index.min.asset.php` when no local boot package exists).

## Test plan

- [ ] Run `npm run build:routes` then visit Settings > AI (New)
- [ ] Verify header shows AI icon, title, subtitle, Docs and Contribute buttons
- [ ] Toggle experiments on/off and click Save — changes should persist on reload
- [ ] Disable the global AI toggle — experiment checkboxes should become disabled
- [ ] Enable the global AI toggle — experiment checkboxes should become editable again
- [ ] Verify settings sync with the old page at Settings > AI

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-340-f5a6a4eaa9d3979e3b4254ed089cdd25b7529aec.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

## Screenshot
Screenshot 2026-03-25 at 15 09 31


gziolo:
@dkotter or @justlevine, can you verify that the proposed approach to the build step is compatible as the intermediate step? It would allow converge on one approach in other PRs, namely I was thinking about this work from @Jameswlepage:
- https://github.com/WordPress/ai/pull/149

jeffpaul:
Ideally we're able to show Features at the top of the page, here's some low-fi mockups of what this could look like: https://github.com/WordPress/ai/issues/234#issuecomment-4090992020

includes/bootstrap.php
```diff
@@ -137,7 +138,7 @@ static function () {
 function plugin_action_links( array $links ): array {
 	$settings_link = sprintf(
 		'<a href="%1$s">%2$s</a>',
-		admin_url( 'options-general.php?page=ai' ),
+		admin_url( 'options-general.php?page=ai-wp-admin' ),
```
dkotter, line 142: I'm assuming the `-wp-admin` part is required? Or can we leave this as just `ai`?

includes/bootstrap.php
```diff
@@ -152,6 +153,144 @@ function plugin_action_links( array $links ): array {
 	return $links;
 }
 
+/**
+ * Gets feature group metadata for the settings UI.
+ *
+ * @since 0.6.0
```
dkotter: We try and leave `@since` statements as `x.x.x` and we then replace those when we do the actual release, as we don't always know what the version number will be when merging a PR (and in this case, we've already release 0.6.0)

includes/bootstrap.php
```diff
@@ -209,10 +353,53 @@ function initialize_features(): void {
 		$settings_registration = new Settings_Registration( $registry );
 		$settings_registration->init();
 
-		// Initialize admin settings page.
+		// Register admin settings page menu item.
 		if ( is_admin() ) {
-			$settings_page = new Settings_Page( $registry );
```
dkotter, line 214: Any reason to not just keep the `Settings_Page` class and put all the `admin_menu` code in there instead of in this file?

routes/ai-home/ai-icon.tsx
dkotter, line 1: Assuming we don't need the `routes` directory in the final release, we'll want to ensure that's excluded in our `.distignore` file

tools/ensure-boot-asset.cjs
dkotter, line 1: Can't quite tell if we need to ship this file but if not, we'll want to ensure that's excluded in our `.distignore` file

dkotter (changes requested):
Left a few minor comments but overall this looks okay to me. Not sure I love the actual UI but that's probably a separate conversation.

One thing we are missing here is previously we have a single button to enable/disable things. We've now reverted back to a checkbox that requires you to also hit the save button. Any chance we can bring that single button click back, as that was a requested enhancement we added a few releases back.

dkotter:
> @dkotter or @justlevine, can you verify that the proposed approach to the build step is compatible as the intermediate step? It would allow converge on one approach in other PRs, namely I was thinking about this work from @Jameswlepage:

Changes to the build script look fine to me, assuming this question is around the changes in `package.json`? Seems to properly support building our current scripts and also support the new Dataviews UI. Would be great if we could use a single approach for both but no concern from me on how this currently works.

gziolo:
I tested this PR and here are my findings:

- **`npm run build` and `npm run build:routes`** — both work correctly.
- **Browser testing** — everything works as expected.
- **`npm run start`** — the Settings > AI menu disappears. The issue is that `wp-scripts start` and `wp-build --watch` run concurrently and share the same `build/` folder. One wipes the other's output, so the generated PHP files needed for the admin menu registration get lost.

**Proposed fix:** Use separate build folders for each tool so they don't interfere with each other during development.

includes/bootstrap.php
jorgefilipecosta: Yah the wp-admin part is required it is wp-build works.

includes/bootstrap.php
jorgefilipecosta: Done — replaced all `@since 0.6.0` with `@since x.x.x` in bootstrap.php (5 occurrences).

includes/bootstrap.php
jorgefilipecosta: Done — kept the `Settings_Page` class with a static `init( Registry $registry )` method that encapsulates the admin menu registration, script module data filter, and the build-assets-missing fallback. Bootstrap now calls `Settings_Page::init( $registry )`.

routes/ai-home/ai-icon.tsx
jorgefilipecosta: Done — added `/routes` to `.distignore`.

tools/ensure-boot-asset.cjs
jorgefilipecosta: Done — added `/tools` to `.distignore`. The file is a build-time helper only and should not ship in the plugin zip.

jorgefilipecosta:
Addressed the `npm run start` conflict. The root cause is webpack's `output.clean.keep` regex only preserving `fonts/` and `images/`, which wipes all wp-build output on every rebuild.

Regarding separate build folders — both `wp-build` (`BUILD_DIR` hard-coded in `build.mjs`) and `Asset_Loader.php` (hard-codes `build/` paths) would need upstream changes, so that's not feasible right now.

Instead, extended the `clean.keep` regex in `webpack.config.js` to also preserve wp-build directories (`modules`, `pages`, `routes`, `scripts`, `styles`). This lets both tools coexist in `build/` during watch mode — webpack still cleans its own stale output normally.

jorgefilipecosta:
Hi @gziolo, @dkotter thank you a lot for the reviews, I think all the feedback was applied.

includes/bootstrap.php
justlevine: Long-term or temporary due to current usage assumptions in the library?

justlevine:
> This is necessary until @wordpress/build adds plugin-aware boot asset resolution

>  would need upstream changes, so that's not feasible right now.

> Yah the wp-admin part is required it is wp-build works.

@jorgefilipecosta @gziolo how _future-compat_ is this current config? 

It's one thing to be dogfooding a library to provide early feedback, but based on the lack of basic "plugin" support and these coexistence workarounds y'all had to find, how confident do you feel that our config won't require undue maintenance while the lib has a chance to stabilize?

If it _is_ unstable, I ultimately defer to @dkotter and @jeffpaul as to whether they think it's a good idea to assume the potential tech debt, but I do want to remind folks that we're not just about to become a canonical plugin, but we're about to become _the first canonical plugin that WordPress directly promotes in Core_, and we don't want a breaking `0.3.0` holding up a patch or security release while we're still getting our feet wet.

I'm not sure what parts of this new build process are required for the latest dataviews lib, nor what about the new build process delayed WP7.0/how it was eventually resolved and whether it's relevant to us here, but assuming knowing both help decide whether this is a conversation worth having, or premature worry.

---

(Changes themselves look fine to me 🚀)

gziolo:
@justlevine, all fair feedback. I wasn’t aware of all the limitations of wp-build surfaced in this PR. I’m less worried about that as Gutenberg and WP core also use it, so in theory it should improve substantially over time.

@jorgefilipecosta, wp-scripts is flexible enough to output to a different folder. If I recall correctly in the past Gutenberg was using two different build folders to avoid race conditions between scripts and script modules. The solution you proposed works today but it really brittle and something unexpected can pop up in the future as soon wp-build will need its cleanup. It all really depends on how long two build systems will coexist. Therefore, I’m too worried about it, but it’s important to keep it in mind.

dkotter (dismissed):
Overall this looks good to me, noting we do lose the single click, toggle things on that we have right now (I think I flagged this in Slack or on GitHub somewhere). That does feel like a regression, as I know have to click a checkbox and then click the save button instead of just clicking a single button. If it's possible to update that, I think that would be great but not something that needs to be part of this PR.

Edit: also looks like a failing E2E test that would be great to clean up

jorgefilipecosta:
> @justlevine, all fair feedback. I wasn’t aware of all the limitations of wp-build surfaced in this PR. I’m less worried about that as Gutenberg and WP core also use it, so in theory it should improve substantially over time.
> 
> @jorgefilipecosta, wp-scripts is flexible enough to output to a different folder. If I recall correctly in the past Gutenberg was using two different build folders to avoid race conditions between scripts and script modules. The solution you proposed works today but it really brittle and something unexpected can pop up in the future as soon wp-build will need its cleanup. It all really depends on how long two build systems will coexist. Therefore, I’m too worried about it, but it’s important to keep it in mind.

Good point @gziolo, I updated to use two distinct folders as suggested.

jorgefilipecosta:
Given that I did the small two folders change I will need another approval.

jorgefilipecosta:
> Overall this looks good to me, noting we do lose the single click, toggle things on that we have right now (I think I flagged this in Slack or on GitHub somewhere). That does feel like a regression, as I know have to click a checkbox and then click the save button instead of just clicking a single button. If it's possible to update that, I think that would be great but not something that needs to be part of this PR.
> 
> Edit: also looks like a failing E2E test that would be great to clean up

Hi @dkotter 
CI is green. I will happily iterate on the UI as follow up to get auto saving and something closer to the mockup @jeffpaul shared, I guess as immediate next steps we should replace checkbox with toggle control and remove the save to have auto save.

gziolo (approved):
Tested locally — both build folders work correctly in production and development modes. The two-folder approach cleanly avoids the watch-mode race condition. Approved, with agreed follow-ups to be tackled separately (toggle controls + auto-save, UI closer to mockups).

package.json
```diff
@@ -39,21 +51,26 @@
     "@types/wordpress__block-editor": "^15.0.0",
     "@types/wordpress__blocks": "^12.5.18",
     "@types/wordpress__edit-post": "^8.4.2",
+    "concurrently": "^9.2.1",
+    "@wordpress/build": "^0.10.0",
     "@wordpress/e2e-test-utils-playwright": "^1.37.0",
     "@wordpress/env": "^10.37.0",
+    "@wordpress/route": "^0.2.0",
```
justlevine, line 58: @jorgefilipecosta was there a particular reason this version was chosen when the latest version available is 0.9.0? 

There's [no changelog in the repo](https://github.com/WordPress/gutenberg/tree/trunk/packages/route)...

PR #341, opened 2026-03-25
[Fix]: Issue 264 | Anthropic with Title Gen

### LINKED ISSUE #264: Error using Anthropic with Title Gen
### Description

As flagged by @gziolo in https://github.com/WordPress/ai/pull/228#issuecomment-3943514795, there's an error when generating titles with Anthropic (while generating excerpts and summaries work as expected):

Screenshot 2026-02-23 at 10 07 30

Also per https://github.com/WordPress/ai/pull/228#issuecomment-3983159946 the issue exists in the current released version of the plugin and is not an issue in the #228 PR branch (or just in the `develop` branch).

1. Add an Anthropic API key
2. enable title gen
3. attempt to generate a title, see error as noted above

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

TylerB24890:
## What?

Closes #264 

This PR removes the multiple candidates from the Title Generation experiment and makes it return only one. All E2E and unit tests updated accordingly.

## Why?

As described in https://github.com/WordPress/ai/issues/264 -- Anthropic and OpenAI no longer support multiple candidate responses. This PR ensures the title generation logic continues to work even if only one of those models is currently connected.

## How?

Remove all references to candidates within the Title Generation experiment. With the removal of multiple candidates per response, the modal UI has been removed and titles are just inserted into the post directly. Editors can then regenerate the title as needed using the existing toolbar button.

### Use of AI Tools

* **AI Assistance:** Yes
* **Used for:**
    * Outlining everywhere candidates were referenced and used
    * Update E2E and Unit Tests accordingly

## Testing Instructions

1. Ensure the "Title Generation" experiment is enabled.
2. Ensure only an Anthropic connector is active.
3. Navigate to a post and click the "Regenerate" button.
4. Confirm a title is generated and inserted into the post title field.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-341-5290d7c5064dc81fd5344d07a21d76fd77eb68e8.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Title_Generation/Title_Generation.php
+++ b/includes/Abilities/Title_Generation/Title_Generation.php
@@ -24,15 +24,6 @@ use function WordPress\AI\normalize_content;
 class Title_Generation extends Abstract_Ability {
 
 	/**
-	 * The default number of candidates to generate.
-	 *
-	 * @since 0.1.0
-	 *
-	 * @var int
-	 */
-	protected const CANDIDATES_DEFAULT = 3;
-
-	/**
 	 * {@inheritDoc}
 	 *
 	 * @since 0.1.0
@@ -41,23 +32,15 @@ class Title_Generation extends Abstract_Ability {
 		return array(
 			'type'       => 'object',
 			'properties' => array(
-				'content'    => array(
+				'content' => array(
 					'type'              => 'string',
 					'sanitize_callback' => 'sanitize_text_field',
 					'description'       => esc_html__( 'Content to generate title suggestions for.', 'ai' ),
 				),
-				'context'    => array(
+				'context' => array(
 					'type'              => 'string',
 					'sanitize_callback' => 'sanitize_text_field',
 					'description'       => esc_html__( 'Additional context to use when generating title suggestions. This can either be a string of additional context or can be a post ID that will then be used to get context from that post (if it exists). If no content is provided but a valid post ID is used here, the content from that post will be used.', 'ai' ),
-				),
-				'candidates' => array(
-					'type'              => 'integer',
-					'minimum'           => 1,
-					'maximum'           => 10,
-					'default'           => self::CANDIDATES_DEFAULT,
-					'sanitize_callback' => 'absint',
-					'description'       => esc_html__( 'Number of titles to generate', 'ai' ),
 				),
 			),
 		);
@@ -93,9 +76,8 @@ class Title_Generation extends Abstract_Ability {
 		$args = wp_parse_args(
 			$input,
 			array(
-				'content'    => null,
-				'context'    => null,
-				'candidates' => self::CANDIDATES_DEFAULT,
+				'content' => null,
+				'context' => null,
 			),
 		);
 
@@ -134,7 +116,7 @@ class Title_Generation extends Abstract_Ability {
 		}
 
 		// Generate the titles.
-		$result = $this->generate_titles( $content, $context, $args['candidates'] );
+		$result = $this->generate_titles( $content, $context );
 
 		// If we have an error, return it.
 		if ( is_wp_error( $result ) ) {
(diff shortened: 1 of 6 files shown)


includes/Abilities/Title_Generation/Title_Generation.php
```diff
@@ -229,10 +211,9 @@ protected function meta(): array {
 	 *
 	 * @param string $content The content to generate title suggestions for.
 	 * @param string|array<string, string> $context Additional context to use.
-	 * @param int $candidates The number of titles to generate.
-	 * @return array<string>|\WP_Error The generated titles, or a WP_Error if there was an error.
+	 * @return array<string>|\WP_Error The generated title wrapped in an array, or a WP_Error if there was an error.
 	 */
-	protected function generate_titles( string $content, $context, int $candidates = 1 ) {
+	protected function generate_titles( string $content, $context ) {
```
dkotter: What are your thoughts on taking this further? Right now this method is still named `generate_titles` even though it only generates a single title. Also, we return an array that contains a single title.

I'd suggest at least updating this method and we could think about changing our output schema so we only return a single item instead of an array that contains a single item (which also would simplify our JS logic). That is technically a breaking change though I'm not concerned with that, since this is still marked as an experiment. But open to opinions here

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -189,72 +89,39 @@ export default function TitleToolbar(): JSX.Element | null {
...
 
 		try {
-			const generatedTitles = await generateTitles(
+			const generatedTitle = await generateTitle(
 				postId as number,
 				content
 			);
-			setTitles( generatedTitles );
-			openModal();
+			editPost( { title: generatedTitle } );
```
dkotter, line 93: I know we are purposely changing the approach in this PR to automatically set the title instead of showing that in a modal. Wondering if that's worth a discussion first? I'm fine with it but noting it is a change to how it currently works, where we give the user a chance to review, edit and accept the title before using it

TylerB24890: I had that thought too. I guess if they didn't like the title they could just regenerate it or refresh the page to revert it, but it does feel a little out-of-place since almost everything else gives you the opportunity to review before accepting.

Should we implement a popover UI that shows the suggested title with an _Apply_ button or something? That should be pretty simple to implement.

includes/Abilities/Title_Generation/Title_Generation.php
TylerB24890: I've applied all the necessary changes to reflect a single title being returned. Updated all function names, tests and removed the array from the response.

TylerB24890: Not sure why the coverage check is failing here. I've added several new unit tests to try and offset it, but can't seem to get the patch coverage to pass 🤔

dkotter: Yeah it's finicky. As long as the changes in a PR have tests I don't pay much attention to that coverage check so you're good here

src/experiments/title-generation/components/TitleToolbar.tsx
```diff
@@ -25,129 +17,42 @@ import { store as noticesStore } from '@wordpress/notices';
...
+	) {
+		return response.title;
+	}
 
-			return [];
-		} )
-		.catch( ( error ) => {
-			throw new Error( `Error generating titles: ${ error.message }` );
-		} );
+	throw new Error( __( 'No title suggestion was generated.', 'ai' ) );
```
dkotter, line 55: This isn't working now, I get the following when an error happens:

Error notice when title generation fails

In addition, we'll want to ensure we still render the error message that comes from the server as this will provide more details to the user when something is wrong (for example, if they don't have a Provider configured that supports text generation)

src/experiments/title-generation/components/TitleToolbar.tsx
dkotter: Honestly I think this fine for now. If we get feedback wanting to see the result before it's applied we can think about this again but for now, this works for me

dkotter (changes requested):

src/experiments/title-generation/components/TitleToolbar.tsx
TylerB24890: Error message display should be resolved. Please let me know if you're still experiencing issues.

dkotter (approved):

PR #355, opened 2026-03-30
feat: add category filtering and tagging for abilities in Ability Table

### LINKED ISSUE #344: Add category drop down in the ability explorer page to filter out the abilities based on their categories.
The Abilities Explorer only allows filtering by Provider (Core, Plugin, Theme), which describes where an ability comes from. There is no way to filter by what an ability does, making it difficult to find abilities in large lists when you know the functional area you're interested in (e.g. image handling, content editing) but not the source.

-------------------

Add a Category filter dropdown that groups abilities by functional area (Image, Editorial, Content, System, etc.). Categories are auto-derived from ability slugs via keyword matching, with WordPress filter hooks (ai_ability_category_map, ai_ability_tags) so third-party plugins can extend or override them. The filter behaves identically to the Provider dropdown server-side, persisted in the URL, submitted via the same Filter button. 

Image

HILAYTRIVEDI:
## What?

Closes #344

Adds a Category filter to the Abilities Explorer, allowing users to filter abilities by functional area (e.g., Image, Editorial, Content).

## Why?

The Abilities Explorer currently only allows filtering by Provider (Core, Plugin, Theme), which makes it difficult to find abilities based on what they do, especially in large lists. This PR addresses the need for functional filtering by introducing a Category filter, making it easier to discover abilities by their purpose (e.g., image handling, content editing).
References: [Original issue] (summarize: lack of functional filtering in Abilities Explorer).

## How?

This PR adds a Category filter dropdown to the Abilities Explorer UI. Categories are auto-derived from ability slugs using keyword matching, and can be extended or overridden by third-party plugins via the ai_ability_category_map and ai_ability_tags filter hooks. The filter works identically to the Provider dropdown: it is server-side, persists in the URL, and is submitted via the same Filter button.

## Testing Instructions

1. Go to the Abilities Explorer in the admin interface.
2. Observe the new Category filter dropdown next to the Provider filter.
3. Select a category (e.g., Image, Editorial, Content) and click the Filter button.
4. Confirm that the list of abilities updates to show only those matching the selected category.
5. Change the Provider filter and verify that both filters work together as expected.
6. Check that the selected filters persist in the URL and after page reload.

## Testing Instructions for Keyboard

1. Tab to the Category filter dropdown.
2. Use arrow keys to navigate and select a category.
3. Tab to the Filter button and press Enter.
4. Verify that the filtered results are accessible and that focus order is logical.
5. Tab through the Provider filter and ensure all controls are reachable and operable via keyboard.

## Screenshots or screencast 
https://www.awesomescreenshot.com/video/50968973?key=5acd67e6e7dc53b4aa9d85fe075a9709

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-355-0126bf0e80a02d0e8e21ce8925e596bfb6ce1205.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/Abilities_Explorer/Ability_Handler.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Handler.php
@@ -132,7 +132,6 @@ class Ability_Handler {
 		$ability_name = end( $parts );
 		$namespace    = $parts[0] ?? '';
 
-
 		/**
 		 * Filters the keyword-to-category map used when auto-tagging abilities.
 		 *
@@ -141,7 +140,7 @@ class Ability_Handler {
 		 * UI; each value is an array of slug substrings that trigger that category.
 		 *
 		 * Example:
-		 *   add_filter( 'ai_ability_category_map', function( $map ) {
+		 *   add_filter( 'wpai_ability_category_map', function( $map ) {
 		 *       $map['WooCommerce'] = [ 'product', 'order', 'cart' ];
 		 *       $map['SEO']         = [ 'meta', 'seo', 'sitemap' ];
 		 *       return $map;
@@ -152,16 +151,16 @@ class Ability_Handler {
 		 * @param array<string, array<string>> $map Category label => keyword slugs.
 		 */
 		$tag_keywords = apply_filters(
-			'ai_ability_category_map',
-			[
-				'Image'     => [ 'image', 'alt-text' ],
-				'Editorial' => [ 'summariz', 'excerpt', 'title', 'review' ],
-				'Content'   => [ 'post', 'terms', 'taxonomy', 'page' ],
-				'System'    => [ 'environment', 'diagnostic', 'debug' ],
-			]
+			'wpai_ability_category_map',
+			array(
+				'Image'     => array( 'image', 'alt-text' ),
+				'Editorial' => array( 'summariz', 'excerpt', 'title', 'review' ),
+				'Content'   => array( 'post', 'terms', 'taxonomy', 'page' ),
+				'System'    => array( 'environment', 'diagnostic', 'debug' ),
+			)
 		);
 
-		$tags = [];
+		$tags = array();
 		foreach ( $tag_keywords as $tag => $keywords ) {
 			foreach ( $keywords as $keyword ) {
 				if ( false !== strpos( $ability_name, $keyword ) ) {
@@ -172,14 +171,14 @@ class Ability_Handler {
 		}
 
 		// Core abilities always get System tag.
-		if ( in_array( $namespace, [ 'wordpress', 'wp', 'core' ], true ) ) {
+		if ( in_array( $namespace, array( 'wordpress', 'wp', 'core' ), true ) ) {
 			if ( ! in_array( 'System', $tags, true ) ) {
 				$tags[] = 'System';
 			}
 		}
 
 		$unique = array_unique( $tags );
-		$unique = ! empty( $unique ) ? $unique : [ 'Other' ];
+		$unique = ! empty( $unique ) ? $unique : array( 'Other' );
 
 		/**
 		 * Filters the final resolved category tags for a specific ability.
@@ -188,7 +187,7 @@ class Ability_Handler {
 		 * specific ability slug, bypassing keyword matching entirely.
 		 *
 		 * Example:
-		 *   add_filter( 'ai_ability_tags', function( $tags, $slug ) {
+		 *   add_filter( 'wpai_ability_tags', function( $tags, $slug ) {
 		 *       if ( 'my-plugin/generate-meta-description' === $slug ) {
 		 *           return array( 'SEO' );
 		 *       }
@@ -200,7 +199,7 @@ class Ability_Handler {
 		 * @param array<string> $tags Resolved category tags for this ability.
 		 * @param string        $slug The full ability slug, e.g. 'my-plugin/do-thing'.
 		 */
-		return (array) apply_filters( 'ai_ability_tags', $unique, $slug );
+		return (array) apply_filters( 'wpai_ability_tags', $unique, $slug );
 	}
 
 	/**
@@ -212,7 +211,7 @@ class Ability_Handler {
 	 */
 	public static function get_all_tags(): array {
 		$abilities = self::get_all_abilities();
-		$tags      = [];
+		$tags      = array();
 
 		foreach ( $abilities as $ability ) {
 			$tags = array_merge( $tags, $ability['tags'] );
--- a/includes/Experiments/Abilities_Explorer/Ability_Handler.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Handler.php
@@ -105,7 +105,7 @@ class Ability_Handler {
 			'name'          => $ability->get_label(),
 			'description'   => $ability->get_description(),
 			'provider'      => self::detect_provider( $name, $meta ),
-			'tags'          => self::get_ability_tags( $ability ),
+			'categories'    => self::get_ability_categories( $ability ),
 			'input_schema'  => $ability->get_input_schema(),
 			'output_schema' => $ability->get_output_schema(),
 			'raw_data'      => array(
@@ -120,14 +120,14 @@ class Ability_Handler {
 	}
 
 	/**
-	 * Get tags for an ability.
+	 * Get categories for an ability.
 	 *
 	 * @since x.x.x
 	 *
 	 * @param \WP_Ability $ability Ability object.
-	 * @return array<string> Tags for the ability.
+	 * @return array<string> Categories for the ability.
 	 */
-	public static function get_ability_tags( \WP_Ability $ability ): array {
+	public static function get_ability_categories( \WP_Ability $ability ): array {
 		$slug          = $ability->get_name();
 		$category_slug = $ability->get_category();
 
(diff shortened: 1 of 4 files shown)


includes/Experiments/Abilities_Explorer/Ability_Handler.php
```diff
@@ -118,6 +119,110 @@ private static function format_single_ability( \WP_Ability $ability ): array {
 		);
 	}
 
+	/**
+	 * Get tags for an ability based on its slug.
+	 *
+	 * @since 0.6.0
```
dkotter: All `@since` statements need updated to `x.x.x`, as we'll replace that with the actual version number when prepping a release

src/experiments/abilities-explorer/index.js
```diff
@@ -541,17 +541,19 @@ import './index.scss';
...
-			AiAbilityExplorer.init();
-		} );
-	} else {
-		AiAbilityExplorer.init();
-	}
+	initWhenReady( function () {
+		if ( aiAbilityExplorer?.enabled ) {
+			AiAbilityExplorer.initTestRunner();
+			AiAbilityExplorer.initCopyButtons();
+			AiAbilityExplorer.initValidation();
```
dkotter: Any benefit to calling these individual init methods instead of just continuing to call the main `init` method?

includes/Experiments/Abilities_Explorer/Ability_Handler.php
```diff
@@ -118,6 +119,110 @@ private static function format_single_ability( \WP_Ability $ability ): array {
...
+
+		$tags = array();
+		foreach ( $tag_keywords as $tag => $keywords ) {
+			foreach ( $keywords as $keyword ) {
+				if ( false !== strpos( $ability_name, $keyword ) ) {
+					$tags[] = $tag;
+					break;
+				}
+			}
+		}
```
dkotter: Not sure this is the best approach. We're basically trying to map Ability names to some pre-defined keywords, correct? That seems really fragile and not likely to work well with any Abilities beyond the ones we register ourselves.

Abilities are already required to have a category assigned, could we not use that as the filtering here instead?

dkotter (changes requested):

includes/Experiments/Abilities_Explorer/Ability_Handler.php
HILAYTRIVEDI: @dkotter , Thank you I will update it.

src/experiments/abilities-explorer/index.js
HILAYTRIVEDI: @dkotter , yes you are correct . I added that so that when ever the window is loaded then it will execute but looking at the code more in depth `init()` is perfectly fine.

includes/Experiments/Abilities_Explorer/Ability_Handler.php
HILAYTRIVEDI: @dkotter Good call -  `WP_Ability::get_category()` exists and is exactly what we should be using here. Replaced the keyword map approach with a direct call to get_category(), falling back to 'Other' if no category is set. Also kept the wpai_ability_tags filter so third parties can still override per-slug if needed. The wpai_ability_category_map filter is removed since it's no longer needed.

Thank you

includes/Experiments/Abilities_Explorer/Ability_Handler.php
```diff
@@ -105,6 +105,7 @@ private static function format_single_ability( \WP_Ability $ability ): array {
 			'name'          => $ability->get_label(),
 			'description'   => $ability->get_description(),
 			'provider'      => self::detect_provider( $name, $meta ),
+			'tags'          => self::get_ability_tags( $ability ),
```
dkotter: Since we're grabbing Ability categories now, I'd suggest we update all uses of `tag` to be `category` or `categories` across all the files in this PR

includes/Experiments/Abilities_Explorer/Ability_Table.php
```diff
@@ -29,6 +29,15 @@
  */
 class Ability_Table extends \WP_List_Table {
 
+	/**
+	 * Full list of abilities before pagination, used to derive filter options.
+	 *
+	 * @since 0.6.0
```
dkotter: These need updated as well in this file

dkotter (changes requested):

dkotter (commented):
Would also be great to add some E2E tests to cover the changes here, verifying the category dropdown renders and works as expected

includes/Experiments/Abilities_Explorer/Ability_Table.php
HILAYTRIVEDI: Done , I updated it . 👍

includes/Experiments/Abilities_Explorer/Ability_Handler.php
HILAYTRIVEDI: Done ,  The code is now uses category or categories instead if tags.

HILAYTRIVEDI:
> Would also be great to add some E2E tests to cover the changes here, verifying the category dropdown renders and works as expected

@dkotter I added the E2E test for the category dropdown.

includes/Experiments/Abilities_Explorer/Ability_Handler.php
```diff
@@ -118,6 +119,71 @@ private static function format_single_ability( \WP_Ability $ability ): array {
...
+	 * @since x.x.x
+	 *
+	 * @param \WP_Ability $ability Ability object.
+	 * @return array<string> Categories for the ability.
+	 */
+	public static function get_ability_categories( \WP_Ability $ability ): array {
+		$slug          = $ability->get_name();
+		$category_slug = $ability->get_category();
+
+		$category_label = 'Other';
```
dkotter:
```suggestion
		$category_label = esc_html__( 'Other', 'ai' );
```

includes/Experiments/Abilities_Explorer/Ability_Table.php
```diff
@@ -141,6 +162,42 @@ static function ( $a, $b ) use ( $orderby, $order ) {
 		$this->items = array_slice( $abilities, ( $current_page - 1 ) * $per_page, $per_page );
 	}
 
+	/**
+	 * Get sorted unique categories derived from the already-fetched ability list.
+	 *
+	 * @since 0.6.0
```
dkotter:
```suggestion
	 * @since x.x.x
```

includes/Experiments/Abilities_Explorer/Ability_Table.php
```diff
@@ -141,6 +162,42 @@ static function ( $a, $b ) use ( $orderby, $order ) {
...
+
+		return $categories;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * Adds data-categories attribute so the JS category filter can work client-side.
+	 *
+	 * @since 0.6.0
```
dkotter:
```suggestion
	 * @since x.x.x
```

includes/Experiments/Abilities_Explorer/Ability_Table.php
```diff
@@ -141,6 +162,42 @@ static function ( $a, $b ) use ( $orderby, $order ) {
...
+		$categories = array_unique( $categories );
+		sort( $categories );
+
+		return $categories;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * Adds data-categories attribute so the JS category filter can work client-side.
```
dkotter: Looking at this closer, do we actually need this method? It claims it's here to support client-side filtering but doesn't look like we have client-side filtering in this PR. I think we can just remove this

includes/Experiments/Abilities_Explorer/Ability_Handler.php
```diff
@@ -118,6 +119,71 @@ private static function format_single_ability( \WP_Ability $ability ): array {
...
+	}
+
+	/**
+	 * Get all unique categories across all abilities.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array<string> Sorted list of all unique categories.
+	 */
+	public static function get_all_categories(): array {
```
dkotter: I'm not seeing this method used anywhere. We do have a similar `get_unique_categories` method we add, wondering if we should remove that and use this method instead? Or vice-versa?

includes/Experiments/Abilities_Explorer/Ability_Handler.php
```diff
@@ -118,6 +119,71 @@ private static function format_single_ability( \WP_Ability $ability ): array {
...
 
+	/**
+	 * Get categories for an ability.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param \WP_Ability $ability Ability object.
+	 * @return array<string> Categories for the ability.
+	 */
+	public static function get_ability_categories( \WP_Ability $ability ): array {
```
dkotter: Now that we return just the standard Ability category, might make more sense to rename this method to `get_ability_category` and change the return here from an `array` to just a `string`. Seems slightly over complex to return an array with a single item when we could just return a string unless I'm missing something here

dkotter (changes requested):
Sorry for all the back and forth here, this is looking really good and I think we're close, just a few final things I've noticed.

HILAYTRIVEDI:
> Sorry for all the back and forth here, this is looking really good and I think we're close, just a few final things I've noticed.

@dkotter no worries at all . I am new to the AI contribution in core WordPress. So I am actually grateful for your time you are giving in review and feedback. I really appreciate it . As moving forward in other PRs I will keep this feedbacks in mind before review. 🙇

dkotter (approved):

PR #356, opened 2026-03-31
Only add the Generate Alt Text button to selected image blocks

### LINKED ISSUE #16: Duplicate "Regenerate Alt Text" button appears in Block Editor sidebar when navigating between blocks
While testing the Block Editor in WordPress 7.0 RC1, a UI issue was observed where the "Regenerate Alt Text" button appears multiple times in the right sidebar.

This occurs when navigating between different blocks, particularly when switching in and out of Image blocks. The button is duplicated instead of being rendered once per block selection.

### Steps to Reproduce

- Open the Block Editor for a post or page.
- Insert or select an Image block.
- Navigate between different blocks (e.g., Paragraph → Image → Paragraph → Image).
- Observe the right sidebar (Block settings panel).
- Expected Behavior
- Only one "Regenerate Alt Text" button should be displayed for the selected Image block.

### Actual Behavior

- Multiple "Regenerate Alt Text" buttons are displayed in the sidebar.

### Additional Notes

- This may be caused by improper component unmounting or repeated rendering when switching between blocks.
- The issue persists during the same editing session once triggered.

### Environment

- WordPress: 7.0 RC1
- Editor: Block Editor (Gutenberg)
- Browser: Microsoft Edge & Google Chrome
- OS: Windows 11

### Attachments

- Screenshot showing duplicate buttons
- Screencast video demonstrating the issue

You may find screenshot & screencast on this link: https://drive.google.com/drive/folders/1w2mjSUQQtKW82riiCaruckMsnz8lZ4eX

dkotter:
## What?

Closes https://github.com/WordPress/ai-provider-for-openai/issues/16

Ensure we add the alt text generation controls only to selected image blocks

## Why?

As reported in the above issue, if you have an image block within a pattern and you toggle back and forth between the image and another block in the pattern, the alt text generation button ends up displaying weird (that issue mentions it duplicating but I wasn't able to reproduce that).

To fix this, we just need to ensure we only show our controls on **selected** image blocks.

## How?

Instead of checking if the current block is an image, we also add a check to ensure that block is currently selected

### Use of AI Tools

None

## Testing Instructions

1. Pull this branch down and run `npm i && npm run build`
2. Turn on the Alt Text Generation Experiment
3. Add an image into a post and with it selected, ensure you see a button in the sidebar to generate alt text
4. Insert a pattern
5. Ensure there's an image block within the pattern
6. Select that image block and ensure you see the same button in the sidebar
7. Using the Content controls in the pattern sidebar, click into another block within the pattern
8. Ensure you no longer see the alt text controls
9. Click back into the image block and ensure it shows again

## Screenshots

|Before|After|
|-|-|
| Alt text controls showing in the pattern editor when they shouldn't | Alt text controls not showing in the pattern editor |

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-356-04555a045d3962143ffbbd933a0af71d46561017.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/alt-text-generation/index.tsx
+++ b/src/experiments/alt-text-generation/index.tsx
@@ -35,24 +35,22 @@ const { aiAltTextGenerationData } = window as AltTextGenerationData;
  */
 const withAltTextGeneration = createHigherOrderComponent( ( BlockEdit ) => {
 	return ( props: BlockEditProps ) => {
-		// Only add controls to the selected image block.
-		if ( props.name !== 'core/image' || ! props.isSelected ) {
-			return <BlockEdit { ...props } />;
-		}
-
-		// Don't render if disabled.
-		if ( ! aiAltTextGenerationData?.enabled ) {
-			return <BlockEdit { ...props } />;
-		}
+		// Only show controls if we have a selected image block and the experiment is enabled.
+		const showControls =
+			props.name === 'core/image' &&
+			props.isSelected &&
+			aiAltTextGenerationData?.enabled;
 
 		return (
 			<>
 				<BlockEdit { ...props } />
-				<AltTextControls
-					clientId={ props.clientId }
-					attributes={ props.attributes }
-					setAttributes={ props.setAttributes }
-				/>
+				{ showControls && (
+					<AltTextControls
+						clientId={ props.clientId }
+						attributes={ props.attributes }
+						setAttributes={ props.setAttributes }
+					/>
+				) }
 			</>
 		);
 	};
--- a/src/experiments/alt-text-generation/index.tsx
+++ b/src/experiments/alt-text-generation/index.tsx
@@ -17,6 +17,7 @@ import type { ImageBlockAttributes } from './types';
 interface BlockEditProps {
 	clientId: string;
 	name: string;
+	isSelected: boolean;
 	attributes: ImageBlockAttributes;
 	setAttributes: ( attributes: Partial< ImageBlockAttributes > ) => void;
 }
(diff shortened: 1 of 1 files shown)


gziolo (dismissed):
The proposed change makes perfect sense.

gziolo (commented):
Minor suggestion: from a React standpoint, it might be better to always render `<BlockEdit>` in a stable position (inside the fragment) and only conditionally render `<AltTextControls>`. Right now, toggling `isSelected` and changing other conditions causes `<BlockEdit>` to shift between being a standalone element and a fragment child, which can trigger unnecessary unmount/remount cycles.

Something like:

```jsx
const showControls =
    props.name === 'core/image' &&
    props.isSelected &&
    aiAltTextGenerationData?.enabled;

return (
    <>
        <BlockEdit { ...props } />
        { showControls && (
            <AltTextControls
                clientId={ props.clientId }
                attributes={ props.attributes }
                setAttributes={ props.setAttributes }
            />
        ) }
    </>
);
```

I haven't located a concrete bug caused by this here, but based on past experience, keeping the component tree structure stable avoids subtle re-mounting and therefore order issues. Not a blocker — just a refinement to consider as a follow-up.

dkotter:
> Minor suggestion: from a React standpoint, it might be better to always render `<BlockEdit>` in a stable position (inside the fragment) and only conditionally render `<AltTextControls>`. Right now, toggling `isSelected` and changing other conditions causes `<BlockEdit>` to shift between being a standalone element and a fragment child, which can trigger unnecessary unmount/remount cycles.

@gziolo That's a good idea, thanks for the feedback. Updated that here: 76baf5790ceaf1e6846e25fa090939054c491241

gziolo (approved):

PR #359, opened 2026-04-01
Feature/content guidelines

### LINKED ISSUE #322: Integrate Content Guidelines into generation and refinement workflows

AI experiments and features (e.g., title/summary generation, image generation, Review Notes, and Refine from Notes) currently lack awareness of site-specific editorial rules.  This can lead to outputs that are inconsistent with a site’s tone, policies, or standards.  So let's ensure content generation and refinement features respect site-specific [Content Guidelines](https://make.wordpress.org/ai/2026/02/03/content-guidelines-a-gutenberg-experiment/) by passing them as context to the AI Client.

### Requirements / Dependencies

Requires Gutenberg v22.7+ (or future WordPress core support, e.g. 7.1+).

Depends on:

* Content Guidelines REST API
* Content Guidelines custom post type

### Proposed solution

* Detect and retrieve Content Guidelines via the Gutenberg experimental REST API and custom post type.
* Pass relevant guidelines as context to the AI Client for:
  * Title, summary, and excerpt generation
  * Image generation prompts
  * Suggest Refinements / Generate Review Notes
  * Refine from Notes (content updates)
  * Whatever additional, future Features or Experiments would be best informed by Content Guidelines details
* Scope guidelines appropriately (site-wide vs contextual relevance if available).
* Ensure prompts instruct the model to follow guidelines without hallucinating or overfitting.

saarnilauri:
## What?

Closes #322

Adds Content Guidelines integration to the AI plugin, allowing all AI abilities to respect site-wide editorial standards defined via Gutenberg's `wp_content_guideline` custom post type (introduced in Gutenberg 22.7+).

## Why?

When sites define content guidelines (tone, style, image standards, etc.), AI-generated content should follow them. Without this integration, AI abilities operate without awareness of the site's editorial standards, producing output that may not align with the site's voice and style.

## How?

**New `Content_Guidelines` service** (`includes/Services/Content_Guidelines.php`):
- Singleton that fetches and caches guidelines from the `wp_content_guideline` CPT
- Supports four categories: `site`, `copy`, `images`, `additional`, plus per-block guidelines
- Formats guidelines as XML-tagged strings (`<content-guidelines>`) for prompt injection
- Truncates per-category content to 2000 chars (filterable via `wpai_max_guideline_length`)
- Can be disabled via the `wpai_use_content_guidelines` filter
- Gracefully returns empty/null when the CPT isn't registered

**Abstract ability opt-in pattern** (`includes/Abstracts/Abstract_Ability.php`):
- New `guideline_categories()` method — subclasses override to declare which categories they use
- New `get_content_guidelines_for_prompt()` helper fetches and formats guidelines for the declared categories
- `get_system_instruction()` auto-appends a guidelines-awareness paragraph when categories are declared

**Per-ability wiring** — each ability declares relevant categories and injects guidelines into its prompt:

| Ability | Categories |
|---|---|
| Title Generation | `site`, `copy` |
| Excerpt Generation | `site`, `copy` |
| Summarization | `site`, `copy` |
| Alt Text Generation | `site`, `images` |
| Image Generation | `site`, `images` |
| Image Prompt Generation | `site`, `images` |
| Review Notes | `site`, `copy`, `additional` |

Review Notes also gains a new `guidelines` review type that flags content violating the editorial standards.

**Helper functions** in `includes/helpers.php`:
- `get_content_guidelines()` — retrieve guidelines (optionally filtered by category)
- `format_content_guidelines_for_prompt()` — format guidelines for prompt injection

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code (Opus 4.6 and Sonnet 4.6) 
Used for: Initial planning and code implementation; finalization of implementation and tests were reviewed and edited by me.

## Testing Instructions

1. Start the local environment: `npm run test:e2e:env:start`
2. Run the full test suite:
   - `npm run test:php` — runs unit/integration tests including `Content_Guidelines_Test` and `Abstract_Ability_Guidelines_Test`
   - `npm run test:e2e` — runs e2e tests including `content-guidelines.spec.js`
3. **Manual testing (without Gutenberg 22.7+):** Open any post, use title generation, excerpt generation, or any other AI ability — verify they work normally without errors (graceful degradation when the `wp_content_guideline` CPT doesn't exist).
4. **Manual testing (with Gutenberg 22.7+):** If the Content Guidelines feature is available, create a content guideline post with site/copy/images categories, then verify AI-generated content reflects those guidelines in the prompt (inspect via debug logging or network requests).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-359-8a6137adce2f1a84254c0411510dbbc617240af1.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


saarnilauri:
@jeffpaul I was not able to move the card on the kanban board. Now the work is ready for review.

includes/Abilities/Excerpt_Generation/Excerpt_Generation.php
```diff
@@ -23,6 +23,15 @@
  */
 class Excerpt_Generation extends Abstract_Ability {
 
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since 0.7.0
```
dkotter: This can be set to `x.x.x` (here and anywhere else it's used), as we'll update this to the proper version number when we do a release

includes/Abilities/Excerpt_Generation/Excerpt_Generation.php
```diff
@@ -224,6 +233,12 @@ static function ( $key, $value ) {
 			$content .= "\n\n<additional-context>" . $context . '</additional-context>';
 		}
 
+		// Inject content guidelines if available.
+		$guidelines = $this->get_content_guidelines_for_prompt();
+		if ( $guidelines ) {
+			$content .= "\n\n" . $guidelines;
+		}
```
dkotter: So right now we're injecting the guidelines into the prompt. I'm wondering if it would be better to inject these into the system instructions? Curious if you tested both approaches and found this worked better? I guess I would expect the LLM to better follow guidelines that are part of it's instructions instead of the prompt but I haven't tested that myself

includes/Abstracts/Abstract_Ability.php
```diff
@@ -101,9 +101,47 @@ abstract protected function permission_callback( $input );
...
+	 *
+	 * @param string|null $block_name Optional block name for block-specific guidelines.
+	 * @return string Formatted guidelines XML string, or empty string.
+	 */
+	protected function get_content_guidelines_for_prompt( ?string $block_name = null ): string {
+		$categories = $this->guideline_categories();
+		if ( empty( $categories ) ) {
+			return '';
+		}
+		return \WordPress\AI\format_content_guidelines_for_prompt( $categories, $block_name );
```
dkotter: So far we've followed the approach of using `use` statements instead of referencing the fully qualified name. So could add `use function WordPress\AI\format_content_guidelines_for_prompt` to the top of this file and then change this to just 
```suggestion
		return format_content_guidelines_for_prompt( $categories, $block_name );
```

includes/Abstracts/Abstract_Ability.php
```diff
@@ -101,9 +101,47 @@ abstract protected function permission_callback( $input );
...
+	 * Uses guideline_categories() to determine which categories to include.
+	 * Returns empty string when guidelines are unavailable or no categories declared.
+	 *
+	 * @since 0.7.0
+	 *
+	 * @param string|null $block_name Optional block name for block-specific guidelines.
+	 * @return string Formatted guidelines XML string, or empty string.
+	 */
+	protected function get_content_guidelines_for_prompt( ?string $block_name = null ): string {
+		$categories = $this->guideline_categories();
```
dkotter: I know this won't cause problems but if we know we have a defined list of supported categories, wondering if we'd want to verify the categories here match those? Basically dropping any categories that aren't in our supported list?

includes/Services/Content_Guidelines.php
```diff
@@ -0,0 +1,334 @@
...
+	);
+
+	/**
+	 * Default maximum character length per guideline category.
+	 *
+	 * @since 0.7.0
+	 *
+	 * @var int
+	 */
+	private const DEFAULT_MAX_GUIDELINE_LENGTH = 2000;
```
dkotter: Is this just to keep prompt bloat down? Or is there a limit to the guidelines on the Gutenberg side? If it's the former would be interesting to see what impacts this has in real scenarios. Are content guidelines often more than this? Does sending more than this cause problems? Just want to ensure if we put a limit here that we have a good reason to back that up

includes/Services/Guidelines.php
```diff
@@ -0,0 +1,334 @@
...
+			}
+
+			$tag_name = self::CATEGORY_TAG_NAMES[ $category ] ?? $category;
+			$content  = $guidelines[ $category ];
+
+			if ( mb_strlen( $content, 'UTF-8' ) > $max_length ) {
+				$content = mb_substr( $content, 0, $max_length, 'UTF-8' );
+			}
+
+			$parts[] = '<' . $tag_name . '>' . $content . '</' . $tag_name . '>';
```
dkotter, line 236: I've not tried setting guidelines yet so not sure if it supports things like HTML. Do we want any sort of sanitization/escaping around `$content` here?

tests/e2e/specs/experiments/content-guidelines.spec.js
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Internal dependencies
+ */
+const {
+	enableExperiment,
+	enableExperiments,
+} = require( '../../utils/helpers' );
+
+test.describe( 'Content Guidelines Integration', () => {
```
dkotter: So unless I'm missing something, not sure these tests are needed? Seems they just test things that we're already testing elsewhere. We could look to add the Gutenberg plugin into our test environment and we could then add tests that enable that plugin, add content guidelines and then test one of the Experiments to ensure it works

dkotter (changes requested):

includes/Abilities/Excerpt_Generation/Excerpt_Generation.php
saarnilauri: Done.

includes/Abstracts/Abstract_Ability.php
saarnilauri: Cleaned up.

includes/Abstracts/Abstract_Ability.php
saarnilauri: Refactored to drop unsupported categories.

includes/Abilities/Excerpt_Generation/Excerpt_Generation.php
saarnilauri: Good point I refactored to inject them in to the system instructions. I think in most cases the system prompts are given a higher priority by the LLMs.

includes/Services/Content_Guidelines.php
saarnilauri: From the brief research I've done, concise and atomic instructions have been shown to yield better results. The other reason for keeping them short is to set a limit on token usage.

As far as I understand, the guidelines on the Gutenberg side don't have any such limits—at least, nothing in the UI indicates otherwise.

includes/Services/Guidelines.php
saarnilauri: The guidelines are already sanitized in Gutenberg. However, I added escaping as an extra safety measure, since a guideline could contain the `<` character, which might confuse the LLM—given that XML tags are inserted into the system instructions to create clearly defined sections for the LLM to recognize.

tests/e2e/specs/experiments/content-guidelines.spec.js
saarnilauri: was planning to add more valuable tests, but since the test environment didn't have the Gutenberg plugin available, I simplified them and ended up not thinking them through enough. Good catch—I've pruned these.

includes/Services/Content_Guidelines.php
gziolo: The same rules as with SKILLS.md should apply even when the guideline CPT can store excessively long text.

saarnilauri: @gziolo are you refering to this when you say `SKILLS.md` https://github.com/WordPress/agent-skills/blob/trunk/CONTRIBUTING.md#keep-it-small ?

gziolo: Yes, exactly — the same [Keep it small](https://github.com/WordPress/agent-skills/blob/trunk/CONTRIBUTING.md#keep-it-small) principle from agent-skills applies here. Content guidelines are conceptually equivalent to `SKILL.md` instruction bodies — both get injected directly into the LLM context window, so the same sizing constraints apply.

That said, 2000 chars feels too tight. The ecosystem benchmarks for instruction content:

- **Anthropic** recommends <5,000 tokens (~20,000 chars) for `SKILL.md` instruction bodies — [skills docs](https://docs.anthropic.com/en/docs/claude-code/skills), [engineering blog](https://www.anthropic.com/engineering/equipping-agents-for-the-real-world-with-agent-skills)
- **OpenAI's** [function calling guide](https://developers.openai.com/cookbook/examples/o-series/o3o4-mini_prompting_guide) shows effective instructions with examples already run 500–800 chars *per function*
- **Google ADK** calls the tool docstring ["the single most important factor"](https://google.github.io/adk-docs/tools-custom/function-tools/) for correct tool use

Since content guidelines are self-contained (no `references/` offloading), I'd suggest **~5,000 chars** (~1,250 tokens) — enough for procedural steps with examples while keeping token usage in check. The real variable is how many guidelines get injected simultaneously.

gziolo:
Based on the post from @aagam-shah: [Guidelines Lands in Gutenberg 22.7](https://make.wordpress.org/ai/2026/03/23/guidelines-lands-in-gutenberg-22-7/). The final name is going to be Guidelines, and the plan is to drop the "content" prefix from the naming – [confirmed here](https://make.wordpress.org/ai/2026/03/23/guidelines-lands-in-gutenberg-22-7/#comment-125). It would be good to align with that direction in this PR and use the final name wherever possible.

Overall, this PR looks very solid and nicely tackles integration of general-purpose guidelines into existing functionality 👏🏻

saarnilauri:
I will update the naming

includes/Abilities/Image/Generate_Image.php
```diff
@@ -236,6 +245,12 @@ protected function generate_image( string $prompt, ?string $reference_image = nu
 	 * @return \WP_AI_Client_Prompt_Builder|\WP_Error The prompt builder, or a WP_Error on failure.
 	 */
 	private function get_prompt_builder( string $prompt, ?string $reference_image = null ) {
+		// Inject content guidelines if available.
+		$guidelines = $this->get_content_guidelines_for_prompt();
+		if ( $guidelines ) {
+			$prompt .= "\n\n" . $guidelines;
+		}
```
dkotter: We currently don't have system instructions set for image generation so I understand why we're adding these guidelines to the user prompt. But I think it would make sense to use system instructions so we match all other Abilities. We wouldn't need to use our `get_system_instruction` helper method, we could just build those guidelines and pass them in, like:

```php
if ( $guidelines ) {
	$instruction .= 'The following content guidelines represent the site's editorial standards. Apply them where relevant. Do not fabricate content to satisfy guidelines. If guidelines conflict with the input, prioritize accuracy.';
	$instruction .= "\n\n" . $guidelines;
	$prompt_builder->using_system_instruction( $instruction );
}
```

includes/Services/Content_Guidelines.php
dkotter: I agree, I think we can bump this up. 5000 characters sounds like a good place to start with

includes/Abstracts/Abstract_Ability.php
```diff
@@ -101,19 +102,75 @@ abstract protected function permission_callback( $input );
...
+	 * @param string|null            $filename   Optional. Explicit filename to load. If not provided,
+	 *                                           attempts to load `system-instruction.php` or `prompt.php`.
+	 * @param array<string, mixed>   $data       Optional. Data to expose to the system instruction file.
+	 *                                           This data will be extracted as variables available in the file scope.
+	 * @param string|null            $block_name Optional. Block name for block-specific guidelines.
 	 * @return string The system instruction for the feature.
 	 */
-	public function get_system_instruction( ?string $filename = null, array $data = array() ): string {
-		return $this->load_system_instruction_from_file( $filename, $data );
+	public function get_system_instruction( ?string $filename = null, array $data = array(), ?string $block_name = null ): string {
```
dkotter: This is just a thought, open to opinions on this. But since we already allow you to pass in an array of data in the `$data` argument, instead of adding another argument here, we could support the block name in something like `$data['block_name']` and then maybe unset it after we use it. My main concern here is we're adding another argument for one single use case

dkotter (changes requested):

includes/Abstracts/Abstract_Ability.php
saarnilauri: I think your concern is valid. I refactored this to use the `$data` parameter.

includes/Abilities/Image/Generate_Image.php
saarnilauri: Refactored to use system instructions.

saarnilauri:
> Based on the post from @aagam-shah: [Guidelines Lands in Gutenberg 22.7](https://make.wordpress.org/ai/2026/03/23/guidelines-lands-in-gutenberg-22-7/). The final name is going to be Guidelines, and the plan is to drop the "content" prefix from the naming – [confirmed here](https://make.wordpress.org/ai/2026/03/23/guidelines-lands-in-gutenberg-22-7/#comment-125). It would be good to align with that direction in this PR and use the final name wherever possible.
> 
> Overall, this PR looks very solid and nicely tackles integration of general-purpose guidelines into existing functionality 👏🏻

The naming is now updated.

includes/Services/Content_Guidelines.php
saarnilauri: I bumbed it to 5000

jeffpaul:
@saarnilauri once we get the 0.7.0 release out, will likely need to resolve merge conflicts here to get this ready for merge in the 0.8.0 release cycle

dkotter (changes requested):
Overall code looks good here but a few things of note:

1. Currently doesn't work for me. Doing some quick debugging, it appears the Content Guidelines post type stores in a draft state so our query doesn't work. Not sure if there's something I need to do to get those published or if this is expected
2. We've added a few new Abilities since this was opened and we should look to add guidelines support to those: Content Classification and Meta Descriptions

saarnilauri:
I will look into the new abilities and add the guideline injections.

gziolo:
Heads up, the process of renaming CPT to `wp_guideline` is in progress:
- https://github.com/WordPress/gutenberg/pull/77147

I also noticed that @iamchughmayank started some work around taxonomies:
- https://github.com/WordPress/gutenberg/pull/77156

While the current code works, we need to watch the Gutenberg release process and adjust support as needed.

jeffpaul:
@saarnilauri let us know when this is ready again for review/testing, thanks!

saarnilauri:
> * [Guidelines CPT: Changes slug from wp_content_guidelines to wp_guidelines gutenberg#77147](https://github.com/WordPress/gutenberg/pull/77147)
> 

I have updated the post type to match the new slug but keep the legacy slug for < 23.0 

> * [Registers wp_guideline_type taxonomy gutenberg#77156](https://github.com/WordPress/gutenberg/pull/77156)
> 
> While the current code works, we need to watch the Gutenberg release process and adjust support as needed.

The support for the guideline taxonomy is now baked in.

> 1. Currently doesn't work for me. Doing some quick debugging, it appears the Content Guidelines post type stores in a draft state so our query doesn't work. Not sure if there's something I need to do to get those published or if this is expected

Gutenberg's REST controller saves new guideline posts as 'draft' by default. Our query only matched 'publish', silently returning no guidelines for the common case. Now accept both statuses (matching Gutenberg's own `get_guidelines_post()` behavior) and order by date so the most recent guideline wins when multiple posts exist.

> 2. We've added a few new Abilities since this was opened and we should look to add guidelines support to those: Content Classification and Meta Descriptions

These new abilities have now guidelines injected to them.

So, everything is ready for your review.

The not yet implemented PR https://github.com/WordPress/gutenberg/issues/77230 in the Gutenberg, requires a larger refactoring and we need to keep tracking it's progress. It might be better to create a new issue related to it.

includes/Services/Guidelines.php
```diff
@@ -0,0 +1,394 @@
...
+	 *
+	 * @since x.x.x
+	 */
+	private function __construct() {}
+
+	/**
+	 * Resolves the registered post type slug at runtime.
+	 *
+	 * Checks for the canonical slug first, falls back to the legacy slug
+	 * for backward compatibility with Gutenberg < 23.0.
```
dkotter: I'd almost lean towards just making Gutenberg 23.0 as our minimum to avoid this backwards compat code, though the main issue is that version isn't actually released yet. It looks like 23.0 is scheduled for this week, any concern with holding this PR until then and then removing this backward compat code? cc/ @jeffpaul

dkotter (commented):
Tested and works well for me now. Left one comment about maybe waiting until Gutenberg 23.0 comes out before merging this, so we can just make that the minimum version we support. But otherwise this looks good to go, thanks for all the work here!

includes/Services/Guidelines.php
saarnilauri: Seems that 23.0 prerelease is out. I have not found clear release date for it. 

@jeffpaul you can make the call.

jeffpaul: I'd also prefer to just set the min to 23.0 and merging this once that version is released.  I would say we could still merge this and then hold the AI plugin release, but that could end up being a blocker so let's just get this PR to a `done` state and merge once Gb releases.

jeffpaul (dismissed):
Tested copy and images guidelines and title generation and featured image generation resulted as expected.

includes/Services/Guidelines.php
gziolo: GB should be out tomorrow with all the changes necessary.

justlevine:
> maybe waiting until Gutenberg 23.0 comes out before merging this, so we can just make that the minimum version we support. But otherwise this looks good to go

Just passing by and saw this. Are we picking up a new dependency in the form of the Gutenberg plugin or is there some availability checking that I'd see if I actually pulled this PR down to review more deeply?

dkotter:
> > maybe waiting until Gutenberg 23.0 comes out before merging this, so we can just make that the minimum version we support. But otherwise this looks good to go
> 
> Just passing by and saw this. Are we picking up a new dependency in the form of the Gutenberg plugin or is there some availability checking that I'd see if I actually pulled this PR down to review more deeply?

It is the latter. We are _not_ including Gutenberg as a dependency within the plugin. This is a progressive enhancement that only runs if the right version of Gutenberg is installed and if the Guidelines experiment is turned on.

includes/Services/Guidelines.php
saarnilauri: @jeffpaul I will clean up the backwards compat code tomorrow.

saarnilauri: Backwards compat code removed.

saarnilauri:
Ready for the review

dkotter (approved):
Looking good now. Looks like the meta keys they use were also updated in the latest version of Gutenberg so I had to push a fix for those (going from `_content_guidelines` to just `_guidelines`. Also had some flaky E2E tests not related to this PR I fixed up.

Thanks for all the work here @saarnilauri!

PR #361, opened 2026-04-01
Update our preferred models

dkotter:
## What?

Updates our preferred models to more recent ones for the three default providers

## Why?

We maintain a list of our preferred models for text generation, image generation and vision handling. This allows us to be more confident that when someone uses one of our features with the three default providers, things will work (even if new models are released).

This does mean we need to maintain that list and periodically update it, which is what this PR does. Some providers (like OpenAI) have lots of models to choose from and while they often provide guidance on which models to choose, there is still some room for individual opinions here. I made my best judgement on the models to use but open to other opinions here.

## How?

Updates our preferred model list with the most recent suggested models

### Use of AI Tools

None

## Testing Instructions

We're updating the text generation and vision models in this PR so need to test experiments that use those, ideally with the three default AI provider plugins

1. Pull these changes down
2. Activate one of the three default AI provider plugins
3. Enable Title Generation, alt text generation and image generation
4. Test all three and ensure they work
5. Run tests again with the other default AI providers

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-361-c1c4dab15661feeb41f36490a71b18a18799d779.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -143,7 +143,11 @@ function get_preferred_models_for_text_generation(): array {
 	$preferred_models = array(
 		array(
 			'anthropic',
-			'claude-haiku-4-5',
+			'claude-sonnet-4-6',
+		),
+		array(
+			'google',
+			'gemini-3-flash-preview',
 		),
 		array(
 			'google',
@@ -151,11 +155,11 @@ function get_preferred_models_for_text_generation(): array {
 		),
 		array(
 			'openai',
-			'gpt-4o-mini',
+			'gpt-5.4-mini',
 		),
 		array(
 			'openai',
-			'gpt-4.1',
+			'gpt-4.1-mini',
 		),
 	);
 
@@ -237,15 +241,7 @@ function get_preferred_image_models(): array {
 		),
 		array(
 			'openai',
-			'gpt-image-1',
-		),
-		array(
-			'openai',
 			'gpt-image-1-mini',
-		),
-		array(
-			'openai',
-			'dall-e-3',
 		),
 	);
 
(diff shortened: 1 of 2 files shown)


jeffpaul (approved):

drzraf:
Providers must likely be retrieved from https://models.dev/  (or at least given the option to retrieve the list from there). It will avoid all the NIH of handcrafting a list that get outdated every week. That's what most agents do nowadays.

dkotter:
> Providers must likely be retrieved from https://models.dev/ (or at least given the option to retrieve the list from there). It will avoid all the NIH of handcrafting a list that get outdated every week. That's what most agents do nowadays.

I think **must** is strong language here. I don't love the idea of us having to rely on some unknown third party site that we have to continually ping to get this information.

The important thing to remember here is these are the models we set as **preferred**. A user can easily choose the model they want on an individual feature by feature basis (and the models we allow them to choose from are pulled directly from each individual provider, like OpenAI, so that will be the most up to date).

Any model we set as **preferred** will be a model we have tested to ensure it works properly with our features. If this list is dynamic, there's no way for us to verify that is accurate.

PR #332, opened 2026-03-20
Replace the error message the AI Client returns with a more specific one if image requests fail

dkotter:
## What?

Ensure we show a more understandable error message if image generation / refinement / editing fails.

## Why?

At the moment, image generation is supported by the Google and OpenAI provider plugins but image editing / refinement is only supported by Google. If you only have OpenAI connected, you can generate an image but if you then try and refine that image, you'll receive an error message.

Also, if you have no providers connected that support image generation and you turn on the image generation experiment, the UI is still available to use (which this is how all of our experiments currently work). If you try and generate an image you will get an error message.

These messages come from the AI Client and we just capture and display them. But the error messages aren't super specific and don't make it completely clear what the actual problem is, so this PR replaces those errors with more specific ones.

## How?

If an error happens when we make an image generation request, replace the error message the AI Client returns from:

`No models found that support image_generation for this prompt.`

to

`Image generation failed. Please ensure you have a connected provider that supports image generation.`

If an error happens when we make an image edit request, replace the error message the AI Client returns from:

`No models found that support image_generation for this prompt.`

to

`Image refinement failed. Please ensure you have a connected provider that supports image refinement, not just image generation.`

### Use of AI Tools

None

## Testing Instructions

1. Check out this PR
2. Install the OpenAI Provider plugin
3. Activate that and configure it properly
4. Turn on image generation and generate an image
5. Click the Refine button after generating and enter a refinement prompt
6. Ensure you get the more specific error message now
7. Disable the OpenAI Provider
8. Try and generate an image
9. Ensure you get a specific error message

## Screenshots

Error message that shows when a provider doesn't support image generation

Error message that shows when a provider doesn't support image editing

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-332-c39db2c7360f30675f63344b74a6da59d9ba0b8a.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #362, opened 2026-04-01
Add `is_supported_*` checks for better error handling

dkotter:
## What?

Use the AI Clients built-in `is_supported_*` checks to show better errors when a connected provider isn't supported. Follow up to #332

## Why?

Right now if you try and use a provider for something it doesn't support, this is caught and an error message is shown. But this error message isn't super helpful.

It's also best practice to utilize the AI Client's built-in `is_supported_*` checks which also allows us to output more specific error messages.

## How?

Updates all of our AI Abilities to use the proper `is_supported_*` method and return an error if the prompt isn't supported.

Adds a new `get_prompt_builder` method to all of those Abilities to handle this, making it easier to abstract out the building of the prompt builder.

Fixes a bug where errors weren't showing when they occurred during generation of review notes. Also removed an extra "Error" text when an error happened during title generation.

### Use of AI Tools

None

## Testing Instructions

1. Pull down this PR
2. Turn on all Experiments that utilize AI
3. Don't connect any AI Provider
4. Test each Experiment and verify you get an error message with a specific error
5. Connect an AI Provider and test again ensuring everything works

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-362-dd2f9379a327416d6e6a35f51aa4a5563986ea7a.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Abilities/Excerpt_Generation/Excerpt_Generation.php
```diff
@@ -224,11 +224,35 @@ static function ( $key, $value ) {
...
-			->generate_text();
+			->using_model_preference( ...get_preferred_models_for_text_generation() );
+
+		// Return a more specific error if there isn't a model that supports text generation.
+		if ( ! $prompt_builder->is_supported_for_text_generation() ) {
+			return new WP_Error( 'unsupported_model', esc_html__( 'Excerpt generation failed. Please ensure you have a connected provider that supports text generation.', 'ai' ) );
+		}
+
+		return $prompt_builder;
 	}
```
TylerB24890, line 255: Would it make sense to add a method to the `Abstract_Ability` for this? With some defaults or even an `abstract` method requiring extending classes to implement their own

dkotter: Open to thoughts here and I had considered this. The problem is not all Abilities use AI / make requests so this isn't something we will want to make required. And each Ability is slightly different in terms of the final prompt builder so I struggled figuring out a great way to abstract this out that made sense.

That said, in looking at this again, the actual `is_supported_*` checks are the same across all so I have now pushed a new commit that abstracts those out as new methods on the abstract class, so that reduces code duplication a bit: 45085c6afc9824471aaf031b46e914e618bb8a54

jeffpaul (dismissed):

includes/Abilities/Excerpt_Generation/Excerpt_Generation.php
TylerB24890: I think that's a good solution -- text is by far the most common thing to be generated, so having a shared method for it makes sense.

TylerB24890 (approved):
All looks good to me. 
~~Could we apply the same `ensure_text_generation_supported()` approach to the [`Content_Classification` ability](https://github.com/WordPress/ai/blob/develop/includes/Abilities/Content_Classification/Content_Classification.php#L385-L408)? Happy to open a separate PR for that if you'd like, or apply it to this one.~~

**Edit:** Disregard the `Content_Classification` note. I'll apply it in #240 after this is merged in.

dkotter:
> All looks good to me. ~Could we apply the same `ensure_text_generation_supported()` approach to the [`Content_Classification` ability](https://github.com/WordPress/ai/blob/develop/includes/Abilities/Content_Classification/Content_Classification.php#L385-L408)? Happy to open a separate PR for that if you'd like, or apply it to this one.~
> 
> **Edit:** Disregard the `Content_Classification` note. I'll apply it in #240 after this is merged in.

Sorry, didn't see your edit until I had already pushed up change so this is done now in dd2f9379a327416d6e6a35f51aa4a5563986ea7a

PR #364, opened 2026-04-02
Utilize the new client-side Abilities API

### LINKED ISSUE #346: Executing summarization ability with @wordpress/abilities in WordPress 7.0 doesn't work due to invalid schema
### Description

I was testing the changes mentioned in https://make.wordpress.org/core/2026/03/24/client-side-abilities-api-in-wordpress-7-0/ with the summarization ability but couldn't get it executed.

Example snippet:

```js
/**
 * Generates an AI summary for our review text using the WordPress Abilities API.
 *
 * The Abilities API is a Script Module (`@wordpress/abilities`), so it must be
 * loaded via dynamic `import()` from classic scripts. Server-registered abilities
 * (like `ai/summarization`) are available when `@wordpress/core-abilities` is
 * enqueued, which WP 7.0 does on all admin pages.
 *
 * Falls back to `null` when the Abilities API or the ai/summarization ability
 * is not available.
 *
 * @param text The review text to summarize.
 * @return The generated summary, or null on failure.
 */
export async function generateSummary(
	text: string
): Promise< string | null > {
	try {
		const { executeAbility } = await import(
			/* webpackIgnore: true */ '@wordpress/abilities'
		);

		const result = await executeAbility( 'ai/summarization', {
			content: text,
			length: 'medium',
		} );
		return typeof result === 'string' && result.length > 0
			? result
			: null;
	} catch ( error ) {
		console.error( 'Error generating AI summary:', error );
		return null;
	}
}
```

The first error:

```
Schema compilation error: Error: schema is invalid: data/properties/length/type must be equal to one of the allowed values, data/properties/length/type must be array, data/properties/length/type must match a schema in anyOf
Error generating AI summary: Error: Ability "ai/summarization" has invalid input. Reason: Invalid schema provided for validation.
```

After fixing the type in I got another one:

```
Schema compilation error: Error: strict mode: unknown keyword: "sanitize_callback"
Error generating AI summary: Error: Ability "ai/summarization" has invalid input. Reason: Invalid schema provided for validation.
```

This one I'm not sure how to fix.

_No response_

### Environment info

- WordPress 7.0 RC
- AI 0.6.0

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

dkotter:
## What?

Closes #346

Ensure we properly use the new client-side Abilities API and ensure we fix all errors with existing Abilities

## Why?

We currently have a helper method that attempted to use the client-side Abilities API, falling back to making direct API requests if it wasn't available. This was based on the older client-side API and so needed updating with the new version that is shipping with WordPress 7.0 (see [post](https://make.wordpress.org/core/2026/03/24/client-side-abilities-api-in-wordpress-7-0/)).

In doing this, ran into a number of issues that prevented things from working properly. These issues were silently ignored when using direct API requests but the validation the client-side API uses is more strict and thus surfaced these errors. All of those have been fixed up here.

## How?

- Update our `runAbility` helper function to use the new client-side Abilities API
- Remove the use of `sanitize_callback` from all of our Ability schemas as this isn't supported by the Abilities API, just the REST API
- Ensure we have proper sanitization ourselves due to the removal of the above
- Ensure the `@wordpress/core-abilities` script module is loaded anytime an Ability needs it
- Fix an issue where block content was sometimes being returned as an object instead of a string, causing errors with the client-side Abilities API
- Fix some new eslint errors that surfaced after updating dependencies

### Use of AI Tools

Used Cursor running GPT-5.3 Codex to figure out how to properly load the `@wordpress/core-abilities` script module based on our current setup. Also used it to investigate what schema options are supported and verify `sanitize_callback` was not. All other work done by me

## Testing Instructions

2. For each Experiment that utilizes AI (everything besides the Abilities Explorer), turn those on one-by-one
3. Test the Experiment you turned on, ensuring you have the browser console open. The Experiment should function the same as it always has and you shouldn't see any errors or messages in the console. Importantly, shouldn't see `[AI] Client ability execution unavailable. Falling back to REST.`
4. Continuing testing each Experiment ensuring they all still work as expected

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-364-e99ea977818770e88d4c0becfb594c2a9933b9d8.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (dismissed):

jeffpaul:
Holding on merge here as its unclear from https://github.com/WordPress/ai/issues/346#issuecomment-4183282577 whether we should proceed with this PR or await a change in Gutenberg/core that we may then need to adjust for in the AI plugin.

dkotter:
Making a note for myself that this PR needs updated to account for the newly merged Abilities, namely Content Classification and Meta Descriptions

Edit: Now updated in b3846a1b2f59ecb2aa64f3543cfffdee648a552d and e99ea977818770e88d4c0becfb594c2a9933b9d8. If additional Abilities get merged prior to this PR being merged, will update again

gziolo:
> Remove the use of sanitize_callback from all of our Ability schemas as this isn't supported by the Abilities API, just the REST API

It would be helpful to have it handled automatically when `sanitize_callback` is provided in the schema.

dkotter:
> > Remove the use of sanitize_callback from all of our Ability schemas as this isn't supported by the Abilities API, just the REST API
> 
> It would be helpful to have it handled automatically when `sanitize_callback` is provided in the schema.

I'm assuming this is something that would need to be fixed upstream in Core, does that sound correct?

gziolo:
> I'm assuming this is something that would need to be fixed upstream in Core, does that sound correct?

Technically speaking, it should go upstream.

However, we could test everything in practice here for the input, too. All abilities extend `Abstract_Ability` which extends `WP_Ability`. There is `normalize_input` method that we could override:

https://github.com/WordPress/wordpress-develop/blob/8270db8c06e89784c63350dff374c4cd237cf924/src/wp-includes/abilities-api/class-wp-ability.php#L444-L455

This would integrate with both regular `execute()` calls ([called here](https://github.com/WordPress/wordpress-develop/blob/8270db8c06e89784c63350dff374c4cd237cf924/src/wp-includes/abilities-api/class-wp-ability.php#L613)), and explicit `nonmalize_input()` calls (REST API).

For the output, we could tap in [here](https://github.com/WordPress/wordpress-develop/blob/8270db8c06e89784c63350dff374c4cd237cf924/src/wp-includes/abilities-api/class-wp-ability.php#L647) and override `do_execute()` method:

https://github.com/WordPress/wordpress-develop/blob/8270db8c06e89784c63350dff374c4cd237cf924/src/wp-includes/abilities-api/class-wp-ability.php#L561-L571

It would definitely make it easier to add the logic to WordPress core after battle testing with actual usage, since we have a good set of use cases here.

dkotter:
> However, we could test everything in practice here for the input, too. All abilities extend `Abstract_Ability` which extends `WP_Ability`. There is `normalize_input` method that we could override:
> 
> https://github.com/WordPress/wordpress-develop/blob/8270db8c06e89784c63350dff374c4cd237cf924/src/wp-includes/abilities-api/class-wp-ability.php#L444-L455
> 
> This would integrate with both regular `execute()` calls ([called here](https://github.com/WordPress/wordpress-develop/blob/8270db8c06e89784c63350dff374c4cd237cf924/src/wp-includes/abilities-api/class-wp-ability.php#L613)), and explicit `nonmalize_input()` calls (REST API).

Thanks for the feedback @gziolo, this all makes sense to me. I've opened #481 to make this change. There's a few other things in this particular PR that are probably worth keeping but I'll likely pull those changes out to their own PR as well and then close this one out since we're taking a different approach now

dkotter:
I now also have #482 open that pulls the other relevant code out of this PR. As such, going to close this out now in favor of those two new PRs

PR #377, opened 2026-04-08
Update experiments description to include AI provider model supports

### LINKED ISSUE #335: Display AI model type requirements for each experiment

Users enabling experiments don't know which AI model capabilities are required to use each feature effectively. Currently, experiment titles and descriptions don't clarify:

- Whether an experiment requires a **text model** (e.g., content generation, summaries)
- Whether an experiment requires an **image model** (e.g., alt text generation, image analysis)

This causes confusion when users enable experiments expecting them to work, only to find the AI provider they've configured doesn't support the required model type. It also prevents users from making informed decisions about which experiments to enable based on their AI provider's capabilities.

Display the required AI model type(s) for each experiment in the UI. Add a visual indicator (badge, label, or icon) next to each experiment title that shows which model type(s) it requires:

**Example format:**
```
☐ Excerpt Generation [Text Model]
  Generates excerpt suggestions from content

☐ Alt Text Generation [Image Model]
  Generates descriptive alt text for images using AI vision models.

☐ Image Generation [Image Model]
  Generate featured images and inline images using AI

☐ Review Notes [Text Model]
  Reviews post content block-by-block and adds Notes with suggestions for Accessibility, Readability, Grammar, and SEO.

☐ Content Summarization [Text Model]
  Summarizes long-form content into digestible overviews

☐ Title Generation [Text Model]
  Generates title suggestions from content
```

**Implementation details:**
- Add a `models_required` property to each experiment definition
- Display model type badges next to experiment titles (similar to WordPress plugin tags/labels)
- Use semantic colors:
  - Text Model: one color
  - Image Model: different color
- Include documentation so users understand what model types their AI provider must support

This helps users:
- Choose experiments that match their AI provider's capabilities
- Avoid enabling features that won't work with their current setup
- Understand why certain experiments may fail or produce lower quality results

Image

theskinnyghost:
## What?

Closes #335

Updates the description of all existing experiments to include a note about what feature the AI provider should support (e.g. text generation).

## Testing Instructions
1. Open the Settings > AI admin page
2. Verify all experiments in the "Editor Experiments" section include a description and the description outlines which feature the AI provider should support.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-377-00227f750ed0f038427180b973347cfb9c7a6932.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
Perhaps the alt text one should be "vision-based image analysis." instead of just "vision." to be a bit more descriptive?  Alternatively we adjust these to be "Requires and AI provider that includes a model-type model."?

dkotter:
> Perhaps the alt text one should be "vision-based image analysis." instead of just "vision." to be a bit more descriptive? Alternatively we adjust these to be "Requires and AI provider that includes a model-type model."?

I don't have any strong preference on wording here but as far as I'm aware, "vision" in the model sense will always refer to image analysis, not sure there's a "vision" model that doesn't support that. That said, I wouldn't expect the average WordPress user to understand that so being more specific isn't a bad thing. I'd probably suggest the latter there, maybe also changing `provider` to `connector`:

`Requires an AI connector that includes support for vision models.`

`Requires an AI connector that includes supports for text generation models.`

`Requires an AI connector that includes support for image generation models.`

theskinnyghost:
Thanks for your feedback @dkotter @jeffpaul. I've updated the descriptions as per your suggestions. 
Let me know what you think.

dkotter (approved):

PR #376, opened 2026-04-07
Use toggle UI, auto-save, and inline DataForm settings for experiments

jorgefilipecosta:
## Summary
- Replace checkbox controls with toggle switches on the AI settings page
- Auto-save settings on each toggle change instead of requiring a manual "Save Changes" button click
- Show contextual snackbar notices (e.g. "AI enabled.", "Content Classification disabled.")
- Allow experiments to declare custom settings fields via `get_settings_fields()` on `Abstract_Feature`, rendered as inline DataForms below the experiment toggle when enabled
- Migrate Content Classification's custom settings (taxonomy strategy, max suggestions) from server-rendered HTML to DataForm fields with `show_in_rest` support
- Custom experiment settings use a separate Save button that only appears when changes are pending

## Test plan
- [ ] Visit the AI settings page and verify toggles render instead of checkboxes
- [ ] Toggle a setting and verify it auto-saves with a contextual snackbar message
- [ ] Disable the global AI toggle and verify feature toggles become disabled
- [ ] Enable Content Classification and verify the strategy and max suggestions fields appear inline below the toggle
- [ ] Disable Content Classification and verify the sub-settings disappear
- [ ] Change the taxonomy strategy, verify the Save button appears, click it, and confirm the snackbar and value persistence on reload
- [ ] Verify `/wp/v2/settings/` REST API returns `wpai_feature_content-classification_field_strategy` and `wpai_feature_content-classification_field_max_suggestions`
- [ ] Run E2E tests: `npm run test:e2e -- --grep "Plugin settings"` and `npm run test:e2e -- --grep "Content Classification"`

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-376-e0bddc272f4101c04a2f3b49366f5c84034bd8be.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #379, opened 2026-04-08
Allow bulk enabling/disabling experiments in groups

### LINKED ISSUE #334: Improve Experiments UI for scalability with categorization and bulk management
Users must manually check/uncheck each individual experiment checkbox. When there are multiple experiments in a section, it's tedious and time-consuming to toggle all of them at once. This is especially problematic for:

- **Testing workflows** – Quickly enabling all experiments to test new features
- **Auditing/cleanup** – Disabling all experiments before performance testing or site audits
- **Future scalability** – As more experiments are added, managing them individually becomes increasingly time-consuming

This feature would provide a quick way to enable/disable all experiments in a category without clicking each checkbox individually.

Image

Add a "select all" checkbox at the top of each experiments section header:

1. **Editor Experiments** section – add a master checkbox in the header to select/deselect all editor experiments at once
2. **Admin Experiments** section – add a master checkbox in the header to select/deselect all admin experiments at once

The checkbox should:
- Toggle all checkboxes in that section when clicked
- Show three-state behavior (all checked, all unchecked, mixed state) following WordPress conventions
- Be positioned prominently in the section header, similar to the pattern used in:
  - WordPress Posts list table

theskinnyghost:
## What?

Closes #334 

Add controls to bulk enable / disable all features in an experiment group.

Branched off #376 which contains the work of swapping checkboxes with toggles. The issue mentioned using a checkbox but since we've moved away and using toggles now I thought a `ToggleControl` would be more appropriate.

## Testing Instructions
- Visit the AI settings page and verify a new section is shown on the group headings and should contain a toggle to bulk manage all features in that group
- The toggle should allow enabling / disabling all features at once. If the group has a mixed state (i.e. some features enabled, some features disabled) then the toggle should act as "Enable All".
- Settings should be auto-saved with a contextual snackbar message showing
- If the global AI toggle is disable, the bulk manage toggle should become disabled
- Run E2E tests: `npm run test:e2e -- --grep "Plugin settings"`

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-379-c3cdb8211a00790c7c3e0f1870a75c7bd8797c4b.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/routes/ai-home/stage.tsx
+++ b/routes/ai-home/stage.tsx
@@ -41,8 +41,15 @@ interface PageData {
 	features: FeatureData[];
 }
 
+interface SelectAllToggleProps extends DataFormControlProps< AISettings > {
+	featureSettingNames: string[];
+	groupLabel: string;
+	globalEnabled: boolean;
+}
+
 const FEATURE_SETTING_PATTERN = /^wpai_feature_(.+)_enabled$/;
 const GLOBAL_FIELD_ID = 'wpai_features_enabled';
+const SELECT_ALL_FIELD_PREFIX = 'wpai_select_all_';
 
 function isRecord( value: unknown ): value is Record< string, unknown > {
 	return typeof value === 'object' && value !== null;
@@ -113,6 +120,30 @@ function getDefaultLabel( key: string ): string {
 
 function getSectionId( groupId: string ): string {
 	return `feature-group-${ groupId.replace( /[^a-zA-Z0-9_-]/g, '-' ) }`;
+}
+
+function getSelectAllFieldId( groupId: string ): string {
+	return `${ SELECT_ALL_FIELD_PREFIX }${ groupId.replace(
+		/[^a-zA-Z0-9_-]/g,
+		'_'
+	) }`;
+}
+
+function createSelectAllEditComponent(
+	featureSettingNames: string[],
+	groupLabel: string,
+	globalEnabled: boolean
+): React.ComponentType< DataFormControlProps< AISettings > > {
+	return function SelectAllEdit( props ) {
+		return (
+			<SelectAllToggle
+				{ ...props }
+				featureSettingNames={ featureSettingNames }
+				groupLabel={ groupLabel }
+				globalEnabled={ globalEnabled }
+			/>
+		);
+	};
 }
 
 function buildFallbackFeatureGroups(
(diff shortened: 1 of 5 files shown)


jeffpaul:
Flagging as `blocked` to determine if we want to wait for #376 to get merged in or if we're ok merging this PR down into that one (defer to @dkotter for preference on approach here)

jeffpaul:
I'm not certain the placement of the toggle and the grey section background color feels quite right, but the bigger concern I have is that the toggle and text label don't quite feel right.

Screenshot 2026-04-08 at 10 40 43 AM
Screenshot 2026-04-08 at 10 40 33 AM

Perhaps something more like how the paragraph block in the post editor handles typography settings where the primary label doesn't change but the description does?

Screenshot 2026-04-08 at 10 42 17 AM

So perhaps here that would be...

after sitting here with this for a bit, I think the best approach might actually be buttons/links for each state, especially since you can be in a state where you've enabled a couple experiments but don't have an easy way to disable all if the toggle hasn't already been flipped to enable all?  

So perhaps `Enable all` / `Disable all` or `Activate all` and `Deactivate all` as link options at the en of the section description text or as buttons at the bottom of each section?

dkotter:
> Flagging as blocked to determine if we want to wait for https://github.com/WordPress/ai/pull/376 to get merged in or if we're ok merging this PR down into that one (defer to @dkotter for preference on approach here)

Ideally #376 gets merged in prior to 0.7.0 going out so I think we just hold on this one until that is merged and this can then be rebased off of `develop`. So a `blocked` state makes sense here.

> especially since you can be in a state where you've enabled a couple experiments but don't have an easy way to disable all if the toggle hasn't already been flipped to enable all?

Yeah, I think this is key and what I would expect. With the current approach if I have a couple experiments enabled I can't bulk disable them all without flipping the toggle twice.

> So perhaps Enable all / Disable all or Activate all and Deactivate all as link options at the en of the section description text or as buttons at the bottom of each section?

I think this will be a nicer, not so "in your face" approach. Was trying to find an example in core to provide as inspiration but didn't find anything in a quick search. But a simple links at the bottom of each section that allow you to quickly enable or disable everything in that section makes sense to me.

dkotter:
@theskinnyghost Looks like this got autoclosed when I merged #376 in and I don't see a way for me to reopen. May need to recreate this PR, unless you can reopen it?

PR #404, opened 2026-04-09
refactor: cleanup `Main` bootstrap class

justlevine:
## What?

Closes 

This PR cleans up some of root-file tech debt. More specifically it:

1. Colocates all constants into the root `ai.php` file
2. Cleans out `inc/bootstrap.php` for a `WordPress/AI/Main` class
3. Moves the requirement checks out of the `Main`  bootstrapper into a `WordPress/AI/Requirements` class.

And then some cleanup resulting from that.
- Moving the new settings meta functions from the global namespace into the settings class where they are used.
- Add an admin notice for missing `build/build.php`
- Remove the unused `Loader::is_initialized()` method, and a `public init()` method.

## Why?

Pre-7.0 cleanup. The `Requirements` class will likely also become the registry for per-specific features.

## How?

See diff.

### Use of AI Tools

Opencode + GLM 5.1 to do a first pass of backfilling unit tests, followed by manual deslopping and backfilling cases it didn't get correctly.

## Testing Instructions

`npm run test:php`

## Screenshots or screencast 

|Before|After|
|-|-|
|||

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-404-fa21ba61b42dfa7e2bdda82e0197bf67ba59e588.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


justlevine:
~I'm not sure why these tests are failing. They're passing locally. In the meantime,~ feedback appreciated on the architecture 🙇

edit: was the now-enforced dependency checks looking for the `build.php`. Added a test const to skip it.

includes/Requirements.php
```diff
@@ -0,0 +1,198 @@
...
+					$asset_file = WPAI_PLUGIN_DIR . 'build/build.php';
+					if ( file_exists( $asset_file ) ) {
+						require_once $asset_file; // phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.UsingVariable
+						return true;
+					}
+
+					return false;
+				},
+				'error_message' => static fn() => esc_html__( 'The plugin assets are not built. This is most likely because you downloaded the plugin from the GitHub repository without building the assets. Please run `nvm use && npm ci && npm run build` to build the assets.', 'ai' ),
+			),
```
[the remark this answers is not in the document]
justlevine: Using a new WPAI_IS_TEST to bypass

tests/Integration/Includes/MainTest.php
```diff
@@ -0,0 +1,111 @@
...
+						public function is_experiment() {
+							return false;
+						}
+
+						public function initialize() {
+							throw new \Exception( 'Initialization failed.' );
+						}
+					}
+				);
+			}
```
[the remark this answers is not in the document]
justlevine: No we're testing that `Main` doesn't error.

includes/Features/Loader.php
```diff
@@ -51,14 +51,23 @@ public function __construct( Registry $registry ) {
 		$this->registry = $registry;
 	}
 
+	/**
+	 * Initializes the Loader by registering and initializing features.
+	 * @since x.y.z
```
dkotter: These need updated to `x.x.x` across all files to ensure they get caught during a release

includes/Features/Loader.php
```diff
@@ -197,15 +206,4 @@ public function initialize_features(): void {
...
 	}
-
-	/**
-	 * Checks if features have been initialized.
-	 *
-	 * @since 0.6.0
-	 *
-	 * @return bool True if initialized, false otherwise.
-	 */
-	public function is_initialized(): bool {
```
dkotter, line 208: I'm not overly concerned here but any risk others may be using this since it's marked as `public`? Not saying we need to deprecate it but may at least want to ensure we flag this in our changelog as a breaking change

includes/Settings/Settings_Page.php
```diff
@@ -91,4 +91,187 @@ static function () {
...
+
+		$metadata = array(
+			'groups'   => $groups,
+			'features' => $features,
+		);
+
+		/**
+		 * Filters settings metadata passed to the settings route client.
+		 *
+		 * @since x.x.x
```
dkotter: Since this filter was introduced in 0.7.0 and isn't being renamed here (just moved) I think we should set this `@since` statement to match


```suggestion
		 * @since 0.7.0
```

includes/Settings/Settings_Page.php
```diff
@@ -91,4 +91,187 @@ static function () {
...
+				'label'       => __( 'Other Features', 'ai' ),
+				'description' => __( 'Additional AI-powered features.', 'ai' ),
+				'order'       => 90,
+			),
+		);
+
+		/**
+		 * Filters feature group metadata used by the settings UI.
+		 *
+		 * @since x.x.x
```
dkotter: Similar comment here. This filter already exists in the last release, we're just moving the location in this PR.


```suggestion
		 * @since 0.7.0
```

dkotter (changes requested):

includes/Features/Loader.php
justlevine: I don't think so. `Loader` is `final` and not a singleton, so there's no valid scenario for reuse/extension.

 If you want we can make the whole class `@internal` to make that clearer for the future.

includes/Settings/Settings_Page.php
justlevine: Yeah, this branch was cut right before 0.7.0. Don't think there's any other `@since`s that snuck through beyond the two you flagged.

dkotter (approved):
Looks good to me and tests well!

PR #434, opened 2026-04-15
Reduce token usage when running Review Notes

dkotter:
## What?

Try to reduce the token usage when running Review Notes

## Why?

It was reported that it can be easy to run into hourly token limits when running Review Notes. While some of this is unavoidable (for instance, if you have a really long post), we do control how much extra context we send with each request and we can reduce that a bit while still maintaining quality results.

## How?

- For each block being reviewed, we send the content of that block and a certain amount of content around that block. Previously that was set to up-to 2000 characters before and after. This PR reduces that to 800
- We had extra instructions around this context and that has been removed (similarly to how we [removed this from alt text generation](https://github.com/WordPress/ai/pull/357#pullrequestreview-4068270533))

In testing on a fairly short article (12 paragraphs, 5 headings), here's the before and after token usage:

| Before | After | Savings |
|----|----| ---- |
| 27,237 prompt tokens | 20,224 prompt tokens | 29.55% reduction |

### Use of AI Tools

None for the code in this PR. Did use Claude Code running Opus 4.6 to build a helper plugin to log token usage to the debug log

## Testing Instructions

2. Enable the Review Notes experiment
3. Create a post
4. Trigger Review Notes from the sidebar
5. Ensure it still works as expected

## Changelog Entry

> Changed - Reduce the context we send when running Review Notes to decrease the amount of tokens used.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-434-7bf85be448996e49b8cbc18648d7848a41bfb702.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (dismissed):
Tested Review Notes with this change and things continue to work as expected

PR #436, opened 2026-04-16
Add WP-CLI command for bulk alt text generation

### LINKED ISSUE #435: Add WP-CLI command for bulk alt text generation

The alt text generation feature currently requires manual interaction — either one image at a time via the editor/media modal, or selecting images in the Media Library bulk action (which processes sequentially in the browser tab). For sites with large existing media libraries (hundreds or thousands of images missing alt text), there's no efficient way to backfill.

Common scenarios where this matters:
- **Migration**: A site migrating to WordPress that has years of images without alt text
- **Accessibility audits**: Remediating an existing site to meet WCAG compliance
- **Ongoing maintenance**: Processing new uploads added by contributors who skip alt text

Add a `wp ai alt-text generate` WP-CLI command — the plugin's first CLI integration. It queries images missing alt text and processes them server-side in batches using the existing `ai/alt-text-generation` ability.

### Command usage

```bash
# Generate alt text for all images missing it
wp ai alt-text generate

# Preview what would be processed
wp ai alt-text generate --dry-run

# Regenerate for specific images (even if alt text exists)
wp ai alt-text generate --ids=42,55,100 --force

# Process in small batches with custom delay
wp ai alt-text generate --batch-size=5 --delay=1000
```

### Flags

| Flag | Type | Default | Description |
|------|------|---------|-------------|
| `--batch-size` | int | 20 | Images per batch |
| `--dry-run` | flag | false | Preview without changes |
| `--force` | flag | false | Regenerate even if alt text exists |
| `--ids` | string | — | Comma-separated attachment IDs |
| `--delay` | int | 500 | Milliseconds between API calls |

### Implementation approach

- **New file**: `includes/CLI/Alt_Text_Command.php` — the command class, under the `WordPress\AI\CLI` namespace (auto-resolved by the existing PSR-4 autoloader)
- **Modified file**: `includes/Main.php` — register the CLI command conditionally when `WP_CLI` is defined
- **Reuses the existing ability**: calls `wp_get_ability('ai/alt-text-generation')->execute()` for each image, same as the REST endpoint
- **Saves via `update_post_meta()`** directly — no REST overhead
- **Memory-safe**: clears object cache between batches to handle large libraries
- **Progress bar**: uses `WP_CLI\Utils\make_progress_bar()` for real-time feedback
- **Summary table**: reports generated, decorative, skipped, and failed counts at the end
- **Handles decorative images**: saves empty alt text so they aren't reprocessed

### Pre-flight checks

The command validates before processing:
1. The `ai/alt-text-generation` ability is registered (experiment must be enabled)
2. AI credentials are configured

This also opens the door for future CLI commands (e.g., `wp ai excerpt generate`, `wp ai title generate`) using the same pattern.

adamsilverstein:
## Summary

- Adds `wp ai alt-text generate` — the plugin's first WP-CLI command
- Queries images missing alt text and processes them server-side in batches using the existing `ai/alt-text-generation` ability
- Supports `--batch-size`, `--dry-run`, `--force`, `--ids`, and `--delay` flags
- Includes progress bar, summary table, memory management for large libraries, and graceful error handling

Closes #435

## Test plan

- [ ] `wp ai alt-text generate --dry-run` lists images missing alt text without making changes
- [ ] `wp ai alt-text generate --batch-size=2` processes images with progress bar, verify alt text saved in DB
- [ ] `wp ai alt-text generate --ids=<id1>,<id2> --force` regenerates for specific images even if they already have alt text
- [ ] Running with alt-text experiment disabled shows a clear error message
- [ ] Running without AI credentials configured shows a clear error message
- [ ] `--delay` flag adds visible pause between API calls
- [ ] Large batch runs don't exhaust memory (object cache cleared between batches)

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-436-d9e7779d5335e28533b44b79c4926247844aa029.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

## AI use
I used Claude code to write this PR with a series of prompts. I tested the command manually and reviewed the code.


includes/CLI/Alt_Text_Command.php
```diff
@@ -0,0 +1,347 @@
...
+		$this->ensure_admin_user();
+
+		$ability = wp_get_ability( 'ai/alt-text-generation' );
+		if ( ! $ability ) {
+			WP_CLI::error( 'The ai/alt-text-generation ability is not registered. Make sure the Alt Text Generation experiment is enabled in Settings > AI.' );
+			return; // WP_CLI::error() exits, but this satisfies static analysis.
+		}
+
+		if ( ! has_valid_ai_credentials() ) {
+			WP_CLI::error( 'No valid AI credentials found. Configure a provider in Settings > AI.' );
```
dkotter:
```suggestion
			WP_CLI::error( 'No valid AI credentials found. Configure a provider in Settings > Connectors.' );
```

includes/CLI/Alt_Text_Command.php
```diff
@@ -0,0 +1,347 @@
...
+					}
+				)
+			);
+		}
+
+		$query_args = array(
+			'post_type'      => 'attachment',
+			'post_mime_type' => 'image',
+			'post_status'    => 'inherit',
+			'posts_per_page' => -1,
```
dkotter: Any concern here on sites with lots of attachments, thinking multiple thousands? Wonder if we should instead implement batch processing, where we only query for a set number at a time and then continue to iterate until we've processed them all?

includes/CLI/Alt_Text_Command.php
```diff
@@ -0,0 +1,347 @@
...
+
+	/**
+	 * Displays the list of images that would be processed in a dry run.
+	 *
+	 * @param int[] $attachment_ids Array of attachment IDs.
+	 */
+	private function display_dry_run( array $attachment_ids ): void {
+		$items = array();
+		foreach ( $attachment_ids as $id ) {
+			$alt     = get_post_meta( $id, '_wp_attachment_image_alt', true );
```
dkotter, line 282: Seems we default to only getting images without alt text set unless the `$force` argument is passed. Should we be using that argument here so the results we show are accurate to what would actually be processed?

includes/CLI/Alt_Text_Command.php
```diff
@@ -0,0 +1,347 @@
...
+		$ids_flag   = Utils\get_flag_value( $assoc_args, 'ids', '' );
+
+		$attachment_ids = $this->get_attachment_ids( $ids_flag, $force );
+
+		if ( empty( $attachment_ids ) ) {
+			WP_CLI::success( 'No images found matching the criteria.' );
+			return;
+		}
+
+		WP_CLI::log( sprintf( 'Found %d image(s) to process.', count( $attachment_ids ) ) );
```
dkotter: I also wonder if we need some sort of "Are you sure?" protection here to prevent someone from accidentally running this to generate alt text on thousands of images (and thus rack up a nice API bill). You could argue that if someone has access to WP-CLI and they decide to run this, that's on them but also think it would be nice to use the `confirm` method to get user input first

dkotter (changes requested):
Left a few comments, open to discussion on those. Also looks like PHPStan is failing so that needs looked at

includes/CLI/Alt_Text_Command.php
adamsilverstein: makes sense, maybe with a flag you can pass to skip that.

includes/CLI/Alt_Text_Command.php
adamsilverstein: a batch capability makes sense, I'll add that!

includes/CLI/Alt_Text_Command.php
adamsilverstein: right, good point!

includes/CLI/Alt_Text_Command.php
adamsilverstein: Done — switched the message to "Settings > Connectors" in 7cc6c47.

includes/CLI/Alt_Text_Command.php
adamsilverstein: Reworked this in 7cc6c47: `get_attachment_ids()` is replaced by `count_matching_attachments()` plus a cursor-based `fetch_attachment_batch()`. `process_images()` now pulls one batch (size `--batch-size`) at a time and walks forward by post ID, so we never hold more than one batch worth of IDs in memory regardless of library size. The existing `--ids` mode still works against the bounded list.

includes/CLI/Alt_Text_Command.php
adamsilverstein: Fixed in 7cc6c47 — `display_dry_run()` now takes the same `$force` flag and uses `fetch_attachment_batch()` so the preview matches what the live run would actually process. Capped the table at the first 100 rows with a "... and N more" line so dry-runs on large libraries stay readable.

includes/CLI/Alt_Text_Command.php
adamsilverstein: Added in 7cc6c47 — there is now a `WP_CLI::confirm()` prompt before any processing that shows the count, and a `--yes` flag to skip it for scripted/CI use.

includes/Main.php
```diff
@@ -94,6 +94,11 @@ public function load(): void {
 		// Defer feature initialization to the 'init' action.
 		add_action( 'init', array( $this, 'initialize_features' ), 15 );
 
+		// Register WP-CLI commands.
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			\WP_CLI::add_command( 'ai alt-text', CLI\Alt_Text_Command::class );
+		}
```
dkotter: Thoughts on moving this into the `Alt_Text_Generation` Experiment class? Having it here gives us a single location to register any WP-CLI commands in the future but if we put this in the `register` method of the `Alt_Text_Generation` class, the command will only ever be added if the experiment is enabled

adamsilverstein: That works for me; this is my first contribution to the plugin so I'm familiar with the architecture here so I defer to your suggestions. Once the feature graduates from Experiment we can make the command generally available.

adamsilverstein: Moved the registration into `Alt_Text_Generation::register()` in d9e7779

dkotter (approved):

adamsilverstein:
🎉

PR #437, opened 2026-04-16
[Feature]: AI Request Logging

### LINKED ISSUE #345: Add usage safeguards to AI Client (limits, visibility, and cost awareness)

As originally requested by @ifahimreza via https://make.wordpress.org/core/2026/03/24/introducing-the-ai-client-in-wordpress-7-0/#comment-48592:

> This looks like a great step toward a unified AI layer in WordPress. One concern, though, is that since plugins can send prompts through the AI Client, it may be helpful to have safeguards for usage. For example, allowing site owners to set **monthly token/request limits** and a simple **WP-Admin usage dashboard** that shows calls by plugin, with estimated costs and alerts. That could help avoid surprise bills and build more trust in the ecosystem.

TylerB24890:
**Original PR:** https://github.com/WordPress/ai/pull/149

Screenshot 2026-05-07 at 11 16 00 AM

## What?

Adds a comprehensive AI Request Logging experiment that provides observability for all AI operations:
- Real-time logging of every AI request (provider, model, tokens, duration)
- React-powered admin dashboard under **Tools → AI Request Logs**
- REST API endpoints for programmatic access
- Configurable retention period and automatic cleanup

Closes #345.

## Why?

AI operations are opaque—users have no visibility into what requests are being made, how much they cost, or whether they're succeeding. This experiment provides the observability surface needed for debugging, cost tracking, and usage analysis.

## How?

Architecture uses the decorator pattern to wrap the SDK's HTTP transporter:

- `Logging_Integration` wraps the transporter on `wp_loaded`/`admin_init` using the public `setHttpTransporter()` API
- `Logging_Http_Transporter` decorates requests, capturing timing and delegating to `Log_Data_Extractor`
- `Log_Data_Extractor` parses request/response payloads with 4 filter hooks for extensibility:
  - `ai_request_log_providers` - customize provider detection
  - `ai_request_log_context` - filter context data
  - `ai_request_log_tokens` - custom token extraction
  - `ai_request_log_kind` - request type detection
- `AI_Request_Log_Manager` coordinates schema, repository, and cost calculation
- ~~`AI_Request_Cost_Calculator` estimates costs based on provider pricing tables~~

## Use of AI Tools
* **AI assistance:** Yes
* **Tool(s):** Claude
* **Model(s):** Opus 4.7
* **Used for:**
    * Thorough review and analysis of original PR codebase & changes
    * Assistance in updating unit tests
    * Assistance in tracking down duplicate logic/unnecessary function wrappers.

## Testing Instructions

1. Enable the AI Request Logging experiment in **Settings → AI**
2. Trigger an AI-powered feature (e.g., Title Generation, Excerpt Generation)
3. Navigate to the logs via **Tools → AI Request Logs**
4. Verify the log table populates with request details (provider, model, tokens, duration)
5. Test filtering by provider, status, and date range
6. Verify "Purge logs" clears all entries
7. Disable the experiment and confirm no new logs appear

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-437-4f9b9f52bf4fa2cc356cad0a388876cfa6c4e4cd.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


build/admin/ai-request-logs.asset.php
dkotter, line 1: Is there a reason these build files are included here? Seems these should be gitignored like all other built assets but let me know if I'm missing something

includes/Admin/Provider_Metadata_Registry.php
```diff
@@ -0,0 +1,297 @@
...
+use function get_option;
+use function get_transient;
+use function is_array;
+use function is_object;
+use function is_string;
+use function md5;
+use function set_transient;
+use function sprintf;
+use function trim;
+use function wp_json_encode;
```
dkotter: I would remove these. We haven't been following this pattern of using `use` statements for core WordPress functions

includes/Admin/Provider_Metadata_Registry.php
```diff
@@ -0,0 +1,297 @@
...
+	 */
+	public static function get_metadata(): array {
+		if ( ! class_exists( AiClient::class ) ) {
+			return array();
+		}
+
+		$registry    = AiClient::defaultRegistry();
+		$providers   = array();
+		$overrides   = self::get_branding_overrides();
+		$credentials = get_option( 'wp_ai_client_provider_credentials', array() );
```
dkotter: This isn't accurate anymore. Credentials aren't stored under a single key but one key per connector. I'm also wondering if these rest of this code is accurate with the latest changes coming to 7.0?

includes/Admin/Provider_Metadata_Registry.php
```diff
@@ -0,0 +1,297 @@
...
+
+			/** @var \WordPress\AiClient\Providers\DTO\ProviderMetadata $metadata */
+			$metadata = $provider_class::metadata();
+			$brand    = $overrides[ $metadata->getId() ] ?? array();
+
+			$providers[ $metadata->getId() ] = array(
+				'id'              => $metadata->getId(),
+				'name'            => $metadata->getName(),
+				'type'            => $metadata->getType()->value,
+				'icon'            => $brand['icon'] ?? $metadata->getId(),
```
dkotter: Providers can include their own logos now which I think we'd want to use here?

includes/Admin/Provider_Metadata_Registry.php
```diff
@@ -0,0 +1,297 @@
...
+				'initials'        => $brand['initials'] ?? self::get_initials( $metadata->getName() ),
+				'color'           => $brand['color'] ?? '#1d2327',
+				'url'             => $brand['url'] ?? '',
+				'tooltip'         => $brand['tooltip'] ?? '',
+				'keepDescription' => ! empty( $brand['keepDescription'] ),
```
dkotter: Providers can also provide a description now

includes/Experiments/AI_Request_Logging/AI_Request_Logging.php
```diff
@@ -0,0 +1,138 @@
...
+use WordPress\AI\Experiments\Experiment_Category;
+use WordPress\AI\Logging\AI_Request_Log_Manager;
+use WordPress\AI\Logging\AI_Request_Log_Page;
+use WordPress\AI\Logging\Logging_Integration;
+use WordPress\AI\Logging\REST\AI_Request_Log_Controller;
+use WordPress\AI\Settings\Settings_Registration;
+use function add_action;
+use function esc_attr;
+use function esc_html_e;
+use function register_setting;
```
dkotter: Remove

includes/Experiments/AI_Request_Logging/AI_Request_Logging.php
```diff
@@ -0,0 +1,138 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Provides AI request logging for observability and cost tracking.
+ *
+ * @since 0.6.0
```
dkotter: Update all of these to `x.x.x`

includes/Experiments/AI_Request_Logging/AI_Request_Logging.php
```diff
@@ -0,0 +1,138 @@
...
+				'default'           => AI_Request_Log_Manager::DEFAULT_RETENTION_DAYS,
+				'sanitize_callback' => array( $this, 'sanitize_retention_days' ),
+			)
+		);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function render_settings_fields(): void {
```
dkotter: We have a new approach of rendering settings so I believe this will need updated to support that

includes/Logging/REST/AI_Request_Log_Controller.php
```diff
@@ -0,0 +1,386 @@
...
+	 * @param \WP_REST_Request $request Request.
+	 * @return \WP_REST_Response|\WP_Error
+	 */
+	public function get_log( WP_REST_Request $request ) {
+		$log_id = $request->get_param( 'id' );
+		$log    = $this->manager->get_log( $log_id );
+
+		if ( ! $log ) {
+			return new WP_Error(
+				'ai_log_not_found',
```
dkotter:
```suggestion
				'wpai_log_not_found',
```

includes/Logging/AI_Request_Cost_Calculator.php
```diff
@@ -0,0 +1,395 @@
...
+	 *
+	 * Prices are sourced from each provider's public pricing page.
+	 * The prefix-matching strategy in estimate() means shorter keys
+	 * act as catch-alls for versioned variants (e.g. "gpt-4o" matches
+	 * "gpt-4o-2024-08-06"). More specific keys must appear before
+	 * shorter prefixes that would otherwise shadow them.
+	 *
+	 * @var array<string, array<string, array{input: float, output: float}>>
+	 */
+	private static array $model_costs = array(
```
dkotter: Not sure I love this, it's just lots of code for us to maintain. Also as new models are added, we'll need to add those here. Typically when this happens providers will lower the cost on older models but when we change that here, I'm assuming that will then reflect for older requests which won't be accurate. So we'd almost need to store a timestamp next to each cost to know when to apply the proper cost to each request. I'm of the opinion that we focus this experiment on tracking requests but not on tracking costs

includes/Logging/AI_Request_Log_Manager.php
```diff
@@ -0,0 +1,418 @@
...
+declare( strict_types=1 );
+
+namespace WordPress\AI\Logging;
+
+use function add_action;
+use function get_option;
+use function max;
+use function update_option;
+use function wp_next_scheduled;
+use function wp_schedule_event;
```
dkotter: Remove

includes/Logging/AI_Request_Log_Page.php
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * Returns the plugin icon asset URL for the screen header.
+	 *
+	 * @since 0.6.0
+	 *
+	 * @return string Asset URL or an empty string.
+	 */
+	private function get_plugin_icon_url(): string {
+		$svg_path = WPAI_PLUGIN_DIR . '.wordpress-org/icon.svg';
```
dkotter: I don't believe this will work as we don't ship the `.wordpress-org` directory with the final release

includes/Logging/AI_Request_Log_Schema.php
```diff
@@ -0,0 +1,295 @@
...
+ * @since x.x.x
+ */
+class AI_Request_Log_Schema {
+	// Schema management necessarily uses direct queries against the dedicated log table.
+	// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.DirectDatabaseQuery.SchemaChange
+
+	/**
+	 * Database table name (without prefix).
+	 */
+	public const TABLE_NAME = 'ai_request_logs';
```
dkotter:
```suggestion
	public const TABLE_NAME = 'wpai_request_logs';
```

includes/Logging/Log_Data_Extractor.php
```diff
@@ -0,0 +1,624 @@
...
+		$this->provider_patterns = $this->get_default_provider_patterns();
+	}
+
+	/**
+	 * Returns the default provider detection patterns.
+	 *
+	 * @return array<string, array<string>> Provider name => URL patterns.
+	 */
+	private function get_default_provider_patterns(): array {
+		$patterns = array(
```
dkotter: I've probably asked this elsewhere in this PR but do we need to maintain a list like this? Would be ideal if this just worked for any connected AI provider and not force us to maintain a list like this

includes/Asset_Loader.php
```diff
@@ -63,6 +63,15 @@ public static function enqueue_script( string $handle, string $file_name, ?array
 			);
 		}
 
+		// Remove dependencies that are not registered in the current WP version
+		// (e.g. wp-ui from bundled @wordpress/dataviews on older installs).
```
dkotter: Is this needed? We only support WP 7.0+ so if someone happens to be testing on an older version, I'd expect other issues to crop up

src/experiments/summarization/functions/useSummaryGeneration.ts
dkotter, line 1: Curious on these changes and the ones in `src/features/image-generation/components/GenerateFeaturedImage.tsx` and `src/utils/editor-dispatch.ts`. They don't seem related to this PR at all

stubs/wordpress-ai-client.php
dkotter, line 1: Is this needed? So far we've been ignoring PHPStan errors for WP 7.0 specific things, knowing those will be resolved once 7.0 is released and upstream stub packages are updated.

phpstan.neon.dist
```diff
@@ -18,6 +18,7 @@ parameters:
 			min: 70400
 			max: 80400
 		bootstrapFiles:
+				- stubs/wordpress-ai-client.php
```
dkotter: See question above

webpack.config.js
```diff
@@ -91,8 +138,21 @@ module.exports = {
 	},
 
 	plugins: [
-		// Include WP's plugin config.
-		...defaultConfig.plugins,
+		...plugins,
+
+		// Copy DataViews CSS into build/ so the plugin can enqueue it
+		// on WP versions where wp-dataviews is not a registered style.
```
dkotter, line 124: If all the code changes in this file are only to support older WP versions, I think we can just remove all of this as we only support 7.0+ anyway

includes/Admin/Provider_Metadata_Registry.php
```diff
@@ -0,0 +1,297 @@
...
+	 * Defines manual branding overrides per provider ID.
+	 *
+	 * @return array<string, array<string, mixed>>
+	 */
+	private static function get_branding_overrides(): array {
+		/* translators: %s: Provider name. */
+		$link_template = esc_html__( 'Create and manage your %s API keys in these account settings.', 'ai' );
+
+		return array(
+			'anthropic'   => array(
```
dkotter: I would love to not have to support a long list of providers like this. I think the information we actually care about can come from the provider metadata and anything else I'm personally fine just leaving generic, instead of having to keep a bunch of overrides like this

src/admin/components/provider-icons.tsx
dkotter, line 1: I'd suggest getting rid of these icons and just relying on the logo from the provider metadata, if it exists

dkotter (changes requested):
Overall this looks decent to me but have a couple bigger questions:

1. I'd love to not have to manually support a bunch of providers. If possible, I'd suggest we remove any hardcoded provider references and rely on what information we can get from the AI Client
2. I'd suggest we drop cost tracking / estimation from this as it gets overly complicated. My opinion is that should be implemented at the provider level

justlevine:
(I know this is draft, but when it's time for the real thing I hope that the AI disclosure section is backfilled in the PR description. Would help significantly with the code-review, even beyond the specific smells @dkotter already flagged)

build/admin/ai-request-logs.asset.php
TylerB24890: Unfortunately I'm not sure on the answer here. I would assume it's safe to ignore, but as I only resolved merge conflicts and fixed the `DataViews` implementation from the [original PR](https://github.com/WordPress/ai/pull/149) I don't have the context behind these initial decisions. I'll review and ignore if safe to do so.

includes/Admin/Provider_Metadata_Registry.php
TylerB24890: I tested locally against 7.0 and didn't run into any issues, but I assume it's likely because I had this `wp_ai_client_provider_credentials` key saved prior.

src/experiments/summarization/functions/useSummaryGeneration.ts
TylerB24890: Yeah not really sure where those came from or the purpose behind the `editor-dispatch` file. Those were carried over from the previous PR. I assume they were a remnant from the AI used for the original code generation, but I can't say for certain. 

I've removed the `editor-dispatch` file and the changes in the unrelated components.

includes/Experiments/AI_Request_Logging/AI_Request_Logging.php
TylerB24890: Updated to use `get_settings_fields()`

webpack.config.js
TylerB24890: It seems WP 7.0 doesn't actually register the `wp-dataviews` **stylesheet** handle. I tried removing this but the UI ended up breaking.

Is there a different handle for the `wp-dataviews` stylesheet I should be targeting? Left this in place for now until we figure this out.

TylerB24890:
@dkotter All PR comments are resolved except the `wp-dataviews` stylesheet additions to the webpack config (left a reply on your original comment).

Also want to flag the failing static analysis test; PHPStan can't find the `WordPress\AiClient\Providers\Http\Contracts\HttpTransporterInterface` object and because it's a PHPStan internal error, not a regular analysis error, the `ignoreErrors` patterns don't apply. Not sure exactly how to resolve that without a stub file. Any thoughts there?

Also added an AI disclosure to the PR comment. I can't speak to the original PR's AI use (where 90% of this code came from), but I documented what I used it for in here.

includes/Logging/AI_Request_Log_Manager.php
```diff
@@ -0,0 +1,382 @@
...
+
+	/**
+	 * Minimum allowed retention in days.
+	 */
+	public const MIN_RETENTION_DAYS = 1;
+
+	/**
+	 * Maximum allowed retention in days.
+	 */
+	public const MAX_RETENTION_DAYS = 365;
```
dkotter: Thoughts on this? Wondering if we want a hard cap on how long data is stored? I can see some sites wanting this to keep logs clean but it also seems like most sites would want logs lasting forever

includes/Logging/AI_Request_Log_Manager.php
```diff
@@ -0,0 +1,382 @@
...
+
+	/**
+	 * Default maximum number of rows.
+	 */
+	public const DEFAULT_MAX_ROWS = 100000;
```
dkotter: Haven't gotten to the code that uses this yet but I'm assuming this limits how many entries are in the database? If so, worth a discussion on if this is needed, combined with my question above. I would assume most sites will want logs that last forever so having what seems like a fairly arbitrary number based on maximum rows doesn't seem ideal

includes/Logging/AI_Request_Log_Manager.php
```diff
@@ -0,0 +1,382 @@
...
+ * a simple public API for logging AI requests.
+ *
+ * @since x.x.x
+ */
+class AI_Request_Log_Manager {
+
+	/**
+	 * Option key for logging enabled.
+	 */
+	public const OPTION_LOGGING_ENABLED = 'wpai_feature_ai-request-logging_logging_enabled';
```
dkotter: Is this different from the normal feature enablement option? Seems like this plus the `is_logging_enabled` and `set_logging_enabled` methods could just check if the feature itself is enabled like all other features do

includes/Logging/AI_Request_Log_Page.php
```diff
@@ -0,0 +1,199 @@
...
+	/**
+	 * Outputs the root DOM node for the React app.
+	 */
+	public function render_page(): void {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			return;
+		}
+		?>
+		<div class="wrap ai-request-logs">
+			<div class="ai-admin-header">
```
dkotter: Minor thing but wondering if we may want the header of this page to match the header of our settings page, at least in terms of the icon and title we output (don't need the documentation links or AI enable toggle). Just for sake of consistency as users go from one page to the next

includes/Logging/AI_Request_Log_Page.php
```diff
@@ -0,0 +1,199 @@
...
+	 * Builds the provider metadata payload sent to the React app.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array<string, array<string, mixed>>
+	 */
+	private function get_provider_metadata(): array {
+		if ( ! function_exists( 'wp_get_connectors' ) ) {
+			return array();
+		}
```
dkotter: Since 7.0 is our minimum and this function exists there, I think we can remove this check

includes/Logging/AI_Request_Log_Manager.php
dkotter: Actually looking at the UI now does seem like we have duplicate enable toggles and retention settings, both on the log page and the normal settings page. I'd suggest we remove the enable toggle from the log page and just rely on the enable toggle within our normal settings page. I'm not strongly opinionated on the retention setting but ideally that doesn't show in both places so we should use one or the other

includes/Logging/Log_Data_Extractor.php
```diff
@@ -0,0 +1,621 @@
...
+	 * Returns the provider detection patterns, derived from the connectors registry.
+	 *
+	 * @return array<string, array<string>> Provider name => URL patterns.
+	 */
+	private function get_default_provider_patterns(): array {
+		// Slugs whose API host doesn't contain the slug itself. Google's API runs at googleapis.com.
+		$overrides = array( 'google' => 'googleapis' );
+		$patterns  = array();
+
+		if ( function_exists( 'wp_get_connectors' ) ) {
```
dkotter: Can remove this check

includes/Logging/Logging_Integration.php
```diff
@@ -0,0 +1,109 @@
...
+			return;
+		}
+
+		try {
+			$registry = AiClient::defaultRegistry();
+
+			// Get the current transporter.
+			$current_transporter = $registry->getHttpTransporter();
+
+				// Don't wrap if already wrapped.
```
dkotter: Extra tab here

dkotter (changes requested):
Doing some quick testing, I'm seeing an issue where a request is being shown as initiated by the provider plugin. As an example, if I use the Title Generation experiment that this plugin adds, I'd expect to see the AI plugin as the source but I'm seeing the OpenAI Provider plugin as the source.

And maybe I'm missing it but at a code-level, I see lots of code around various filter options, like filtering by log type, status, provider, user ID, cursor ID/timestamp (not sure what that is). But in the UI, I'm not seeing any of those as filter options. I see date filtering, token filtering, and general search but nothing else. If none of those other options exist, ideally we clean up our code to remove anything referencing that. If they do exist, I guess I just need direction on how to see those filters.

Edit: also would be great to get some E2E tests here to verify the log page exists and can be accessed when the experiment is turned on. Maybe even trigger a request to ensure it shows there

src/admin/ai-request-logs/components/LogsTable.tsx
```diff
@@ -0,0 +1,691 @@
...
+									href={ metadata.url }
+									target="_blank"
+									rel="noopener noreferrer"
+									className="ai-request-logs__popover-link"
+								>
+									{ __( 'API Key Settings', 'ai' ) }
+								</a>
+							) }
+							<a
+								href="admin.php?page=ai-provider-credentials"
```
dkotter: This URL isn't accurate anymore

TylerB24890:
@dkotter Thanks for the follow up! I'll work to get these items resolved.

Re:
> And maybe I'm missing it but at a code-level, I see lots of code around various filter options, like filtering by log type, status, provider, user ID, cursor ID/timestamp (not sure what that is). But in the UI, I'm not seeing any of those as filter options. I see date filtering, token filtering, and general search but nothing else. If none of those other options exist, ideally we clean up our code to remove anything referencing that. If they do exist, I guess I just need direction on how to see those filters.

Those filters can be found here: (Operation, Provider/Model, Status, Token Range, Type)
Screenshot 2026-05-01 at 1 10 19 PM

I'm not really sure why there's _user_ filter logic or "cursor ID/timestamp" as those aren't options here, but the others are there. I'll have to dive into what that "cursor ID" is referring to, and I'm pretty sure we can remove anything regarding users and filtering (though we are tracking the user who initiated the request).

Re: E2E Tests;
That's definitely on the radar. Didn't want to get those implemented until we had things more finalized/squared away though, in case we needed to change things up at all. If you're feeling good about the general UI, location, etc... of this experiment then I can plan to get some initial tests in place.

includes/Logging/AI_Request_Log_Manager.php
TylerB24890: Yeah that makes sense. Perhaps we make _forever_ the default with a filter option to limit it?

includes/Logging/AI_Request_Log_Manager.php
TylerB24890: Yeah this key does match the existing toggle option key. Makes sense to remove this UI and leave it to the original settings page.

includes/Logging/AI_Request_Log_Manager.php
TylerB24890: Yeah I suppose bloat isn't much of a concern here since things are stored in their own tables. Agreed this logic seems unnecessary. Will work to remove it.

includes/Logging/AI_Request_Log_Page.php
TylerB24890: That makes sense to me.

src/admin/ai-request-logs/components/LogsTable.tsx
TylerB24890: I think we'll localize this URL so we don't have to worry about keeping it updated if things change again in the future.

dkotter:
> Those filters can be found here: (Operation, Provider/Model, Status, Token Range, Type)

Hmm.. weird, the only thing I see in that dropdown is Token Range:

AI Logging filters

Maybe the other filters only show up when needed and I don't have enough variation in logs here to trigger those?

includes/Logging/AI_Request_Log_Manager.php
dkotter: Yeah, if I was building this myself (acknowledging hard to say for sure what the original PR was considering when writing this), I think I'd set this to just log forever and maybe not even provide a way to change that. We do already have the ability to clear logs if someone wants to clear things out.

That said, I'm fine if we want a filter around this just in case

includes/Logging/AI_Request_Log_Page.php
dkotter: Just as a quick follow up, I know our settings page is now using Data Views so if it looks like it's going to be hard to bring that header over without converting this entire new settings page to Data Views, maybe we leave it alone for now.

includes/Logging/AI_Request_Log_Manager.php
TylerB24890: I removed the default retention logic and added a filter to allow users to enable it if they want. The cron-job is only registered if the retention filter is set and returns a number greater than `0`.

includes/Logging/AI_Request_Log_Manager.php
TylerB24890: Removed the settings from the Logs page. Left the "Purge All Logs" button, but the only place to enable/disable logging is from the AI settings page now.

Also removed the Log Retention Days input as that can only be configured via the filter now.

src/admin/ai-request-logs/components/LogsTable.tsx
TylerB24890: I've localized this value from the server-side now so it matches the correct URL. **However,** in testing, it seems the Gutenberg plugin changes this page name again to `options-connectors-wp-admin` vs the core `options-connectors` slug. Is that expected? Right now all links to `options-connectors` would be broken if the Gutenberg plugin is active.

The implementation here matches how other experiments handle it, but with the Gutenberg plugin changing the page slug these links become broken when that is active.

includes/Logging/AI_Request_Log_Page.php
TylerB24890: Wrapped the existing page in the DataViews `<Page>` component. This gives us a consistent header without having to redo this into a full DataView UI. Perhaps we can do that as a follow up PR?

TylerB24890:
@dkotter I've made some updates here per the latest feedback. 

I also did some checking into the filters; I haven't been able to reproduce the issue you saw with a single filter showing, but there is a 1 hour cache on the filter options/values. I'm wondering something happened during that process where only the Tokens values were cached for the filters 🤔 

I did add the ability to filter by user here, so maybe the updates I made there will resolve your filter UI? If not let me know and I'll have to dig into it some more.

Regarding the "cursor/timestamp" logic; that appears to be only for the REST implementation and offers a faster index-based pagination. The React UI uses the typical `page` based offset, but this implementation at the REST API is meant for users who are extending this (e.g. building dashboards or exporters.) It removes the need for the `OFFSET` clause and says "give me what comes after this exact row", which is much faster when dealing with a high-volume of logs. I can see it being useful for the case of building a custom dashboard or something, but it seems unnecessary in this initial implementation. Should we remove that or just leave it as is?

src/admin/ai-request-logs/components/LogsTable.tsx
dkotter: Hmm.. I was not aware of that URL changing, wondering if that's going to be part of WP 7.0? If so, there's other areas we'll need to update cc / @jeffpaul

dkotter:
> I did add the ability to filter by user here, so maybe the updates I made there will resolve your filter UI? If not let me know and I'll have to dig into it some more.

Filters are showing and working for me now so either you fixed it or cache cleared on my end 🤷🏻 

> It removes the need for the OFFSET clause and says "give me what comes after this exact row", which is much faster when dealing with a high-volume of logs. I can see it being useful for the case of building a custom dashboard or something, but it seems unnecessary in this initial implementation. Should we remove that or just leave it as is?

I usually don't love having code in place to support some hypothetical future that may never be needed but since we already have the code in place, I think I'm fine just leaving that for now

includes/Logging/AI_Request_Log_Page.php
dkotter: > This gives us a consistent header without having to redo this into a full DataView UI. Perhaps we can do that as a follow up PR?

Yeah, I think that's fine. Noting though that it now seems like the main settings container is not fully extending to the edge of the container, either on the right or left:

Screenshot 2026-05-04 at 1 37 34 PM

TylerB24890:
@dkotter 
> I usually don't love having code in place to support some hypothetical future that may never be needed but since we already have the code in place, I think I'm fine just leaving that for now

I'm happy to remove it if we want. Given this is just an _experiment_ I don't feel its necessary to have this type of future-proof pagination in here. Just let me know!

In the meantime I'm going to get some E2E tests in place.

dkotter:
This is looking good to me. I think final thing needed here is some discussion around the introduction of a new database table in this PR. I don't have any concerns with that but others may.

If we decide we're fine with that, I think final step is to ensure we are okay with the structure of the table and the data we are storing as it's harder to change that once this goes live. cc / @jeffpaul

includes/Logging/AI_Request_Log_Schema.php
```diff
@@ -0,0 +1,281 @@
...
+	 *
+	 * @since x.x.x
+	 */
+	private function create_table(): void {
+		global $wpdb;
+
+		$table_name      = $this->get_table_name();
+		$charset_collate = $wpdb->get_charset_collate();
+
+		$sql = "CREATE TABLE {$table_name} (
```
dkotter, line 116: I asked for feedback on this new table and schema on our last AI call and one question was raised around the logging of cached tokens. We log input and output tokens here but not cached tokens.

Looking into it, I think this will require changes both in the PHP AI Client and within each AI Provider plugin, as right now the AI Client has a [`TokenUsage` class](https://github.com/WordPress/php-ai-client/blob/trunk/src/Results/DTO/TokenUsage.php) that doesn't consider cached tokens.

And looking at the OpenAI Provider plugin as an example, it [removes all usage](https://github.com/WordPress/ai-provider-for-openai/blob/841ef4a39b96f88b7b2aa6fb8f87c9b2a49edee5/src/Models/OpenAiTextGenerationModel.php#L482) information after sending it to the `TokenUsage` class so you can't even manually pull that information yourself.

TylerB24890: Hmmm, that's a good flag. I didn't really even think about cached tokens as a separate metric here. Are cached tokens returned as a value in that `usage` key? I think supplying them as another parameter to the `GenerativeAiResult` object would be the best path forward, as that already stores `tokenUsage`.

Given the `GenerativeAiResult` object doesn't currently do anything for cached tokens is that something we could add support for in like a phase 2 of this experiment or should we aim to get that squared away now?

dkotter: I didn't dive too far into it but I think upstream changes are needed to fully support this, both in the PHP AI Client and in each individual provider plugin. As such, I wouldn't put that as a blocker here

TylerB24890: Yeah I reviewed a bit and that was my understanding as well 👍

jeffpaul:
Using WordPress 7.0-RC4, OAI provider Version 1.0.3, and the latest ZIP here I'm not seeing any data appearing in the AI Request Logs screen.

jeffpaul:
Tested locally again, still nothing coming through.  Tested on Playground with WordPress 7.0-RC4 running, OAI Version 1.0.3, and things worked as expected, so maybe just something off with my local.

jeffpaul (dismissed):

assets/svg/wpai-icon.svg
jeffpaul, line 1: May end up replacing this with the updated WPORG icon asset from #546

dkotter: Ah, yep, good catch. I can update that now that #546 has been merged

dkotter: Actually it doesn't look like that icon was even being used, it was already defaulting to the same icon we use everywhere else. I've gone ahead and removed that extra icon file

dkotter (approved):

phpstan.neon.dist
```diff
@@ -40,5 +40,7 @@ parameters:
 			analyse:
 				- tests/
 				- vendor/
+				- includes/Logging/Logging_Integration.php
+				- includes/Logging/Logging_Http_Transporter.php
```
justlevine, line 44: I know I'm late here, but I'm not seeing it in the previous comments.
 @TylerB24890 @dkotter Why aren't we analyzing these files?

dkotter: I believe the issue here is these files reference code coming from the AI Client and that code isn't stubbed yet, so PHPStan throws errors. And the errors thrown were internal to PHPStan so they weren't errors we could suppress individually.

I couldn't find the discussion in a quick search but the idea was we could either stub the AI Client ourselves or ignore those files until the upstream WordPress Core stubs are updated. We decided on the latter but open to revisiting that decision

justlevine: Definitely no need to revisit here, we should be able to update to the latest WordPress stubs as soon as wp-cli-stubs [cuts a new release](https://github.com/php-stubs/wp-cli-stubs/pull/24). If it's not out this week though, I'll open a PR to add in a `# Reason why we're ignoring these`

Issue #325, opened 2026-03-19
Integrate media features and experiments with Gutenberg's experimental Media Editor


jeffpaul:
## What problem does this address?

As we continue to add media-related AI features/experiments, we're currently targeting the media experience within the post editor and within the Media Library.  We may want to consider also integrating with the experimental [Gutenberg media editor](https://github.com/WordPress/gutenberg/issues/72734) as a forward-leaning media editing experience within WordPress.

## What is your proposed solution?

If Gutenberg is enabled and the Media Editor experiment is activated, then perhaps we consider interjecting our media-related AI features/experiments into that experience?

<img width="2166" height="126" alt="Image" src="https://github.com/user-attachments/assets/06f1df3c-75b7-49a8-af45-2c82aa528b17" />

TylerB24890:
@jeffpaul I've been playing around with this and the new Media Editor and have something working for the Alt Text field, _however_ there are currently no filters or Slots/Fills in the new Media Editor, which means we have to go through `DataViews` with `registerEntityField()` to get there. This also means we can't hook into the existing field output.

I played around with a version of this where we just added a new "Generate Alt Text" button to the end of the field list, but it felt disconnected not being inline with or right next to the Alt Text input field.

So then I looked into replacing the core `alt_text` input with our own, wrapped in the AI Experiment. This works (see screenshot below) but I couldn't just _replace_ the alt text field because because `registerEntityField` runs asynchronously from `usePostFields()` (which is what the Media Editor uses), and whatever field is registered last wins. Because they run async from each other we don't know which field (core or ours) would be registered first.

To get around this I'm setting a timeout of `1ms` on the `registerEntityField()` method. This seems to be working to override the core field, but I'm not sure how stable it is given the unknown time it might take for `usePostFields()` to complete. In the event that it takes longer than `1ms` to complete, the core field might override our field instead. The other downside is we can't retain the current field ordering, because they are registered independently from each other. Our `alt_text` field is being added to the end of the current Media fields because we can't hook into the current field rendering.

Here is a screenshot of what I have implemented:
<img width="1077" height="504" alt="Image" src="https://github.com/user-attachments/assets/368147e4-a354-4d28-8a54-09a7fd0231cd" />

**All this said, I think we have a couple paths forward:**
1. I can open the PR for what I have for testing and we can proceed with that for now.
[...]

PR #446, opened 2026-04-17
[Feature][PoC]: Issue 325 | Experimental Media Editor | AI Alt Text

TylerB24890:
## What?

See #325 

This PR integrates the **AI Alt Text generation** experiment into the new [experimental Media Editor](https://github.com/WordPress/gutenberg/issues/72734) in Gutenberg.
It is intended as a Proof of Concept/MVP implementation only.

**Related:**
* https://github.com/WordPress/gutenberg/issues/55238
* https://github.com/WordPress/gutenberg/issues/72734
* https://github.com/WordPress/gutenberg/issues/73771

## Why?

Currently all AI experiments only target the existing core media library and modals. This is sufficient for the time-being, but as WordPress continues to evolve its important that the AI features available in the legacy media UI are also available in the upcoming media UI.

## How?

The new Media Editor **does not** currently provide any filters or Slot/Fill components for extending the field output. This means we are not able to hook into the media editor fields and add new components within the existing fieldset. We must register our field separately with `registerEntityField` to the `attachment` post type. By default this appends the registered fields to the bottom of the form, with no way to control the render order. 

To ensure the Alt Text field and the "Generate" button are rendered together, we're overriding the core Alt Text input and rendering it with a `<Button />` to trigger the Alt Text generation. This moves the Alt Text field to the bottom of the UI, but keeps the "Generate" button inline/next-to the Alt Text `<TextareaControl />` component. 

**Additionally,** because `registerPostTypeSchema` runs asynchronously from `usePostFields` we can't accurately determine which `Field` component will be registered first. Like all objects, the last reference will take priority, meaning if the core `Field` is registered _after_ our custom `Field`, it will override our override.

To get around this, we are subscribing to the `postType` state and re-registering the `Field` once the state changes to the `attachment` post type with a `100ms` delay. This ensures our custom `Field` (and `Button`) are registered after the core `Field` and therefore taking priority in the render. Subscribing to the `postType` state ensures this re-register only happens once, as the `postType` is set only once.

### Use of AI Tools

* **AI assistance:** Yes
* **Tool(s):** Claude
* **Model(s):** Opus 4.6
* **Used for:** 
    * Thorough review & explanation of the `media-editor` package.
    * Approach proposal and feasibility assessment.
    * Support in understanding the asynchronous nature of `registerPostTypeSchema` and `usePostFields`
    * Base implementation; final code was reviewed and revised by me.

## Suggested Next Steps
1. Monitor the Media Editor (and Media Modal) experiment in the Gutenberg plugin for the extensibility implementation. Update this PR accordingly.

## Testing Instructions

1. Pull down and build a copy of the latest Gutenberg plugin and activate in a local installation.
4. Pull down the branch and build.
5. Navigate to the **Experiments** settings page in the Gutenberg plugin settings and activate the Media Editor experiment _and_ the Media Editor Modal experiment.
6. Navigate to the **AI** settings and enable the **Alt Text Generation** experiment.
7. Navigate to a post and insert the Image block. Select an image from the Media Library.
8. In the block toolbar, click the **Edit Image** button to open the inline Image Editor.
9. Confirm the Alt Text field in the Image Editor is accompanied by the "Generate Alt Text" button.
10. Click the "Generate Alt Text" button and confirm the suggested text is inserted directly into the Alt Text textarea field.
11. Confirm the button changes to "Regenerate" after a suggestion is in place.
12. Navigate back to the post editor and select the Image block. Click the Crop icon to open the Media Editor Modal experiment.
13. In the modal sidebar, toggle over to the "Details" UI:
Screenshot 2026-05-07 at 10 48 18 AM
14. Confirm the same "Generate Alt Text" input is available. Repeat steps 9-11 for the modal UI.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
|Screenshot 2026-04-17 at 11 47 23 AM|  Screenshot 2026-04-17 at 11 46 17 AM|

_Note: Notice the alt text has been moved to the bottom of the UI with the Generate button. This is a result of not having any existing hooks to the Field rendering._

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.
> Development Update - Development related updates.

Added - New feature.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-446-154b2af10818269ebfd8105933f2975dbfcfca18.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -159,6 +160,29 @@ private function maybe_enqueue_media_script(): void {
...
 
+	/**
+	 * Conditionally enqueues assets for the experimental Gutenberg media editor.
+	 *
+	 * Requires the Gutenberg media editor experiment to be enabled.
+	 *
+	 * @since x.x.x
+	 */
+	private function maybe_enqueue_media_editor_script(): void {
+		if ( ! $this->is_enabled() || ! is_plugin_active( 'gutenberg/gutenberg.php' ) ) {
```
dkotter: Technically don't need the `is_enabled` check here as that should have already run prior to calling the `register` method (which in turn is what calls this method)

src/experiments/alt-text-generation/components/AltTextControls.tsx
```diff
@@ -190,12 +210,7 @@ export function AltTextControls( {
 				{ /* Decorative image notice */ }
 				{ isDecorative && (
 					<div style={ { marginBottom: '12px' } }>
```
dkotter, line 210: Looks like the `DecorativeNotice` component is wrapped in a `div` with a bottom margin applied so I'm assuming this will be wrapped twice. Do we want to get rid of the outer `div` here?

src/experiments/alt-text-generation/media-editor.tsx
```diff
@@ -0,0 +1,71 @@
+/**
+ * Alt text generation integration for Gutenberg's experimental Media Editor.
+ *
+ * Re-registers the core `alt_text` entity field on the `attachment` post type
+ * to render an AI generate/regenerate control next to the standard textarea.
```
dkotter: We've been moving away from explicitly saying `AI`, since that's more or less implied with anything in this plugin:


```suggestion
 * to render a generate/regenerate control next to the standard textarea.
```

src/experiments/alt-text-generation/media-editor.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { MediaEditorAltTextControl } from './components/MediaEditorAltTextControl';
+import type { MediaEditorAttachment } from './types';
+
+interface MediaEditorWindow extends Window {
+	__experimentalMediaEditor?: boolean;
+}
+
+const { __experimentalMediaEditor } = window as MediaEditorWindow;
+
+if ( __experimentalMediaEditor ) {
```
dkotter: Would probably read nicer to have an early exit statement here instead of a larger nested `if` statement if the media editor doesn't exist

src/experiments/alt-text-generation/components/MediaEditorAltTextControl.tsx
```diff
@@ -0,0 +1,131 @@
...
+ * Internal dependencies
+ */
+import { getButtonLabel, DecorativeNotice } from './AltTextControls';
+import { generateAltText } from '../../../utils/generate-alt-text';
+import type { MediaEditorAttachment } from '../types';
+
+/**
+ * MediaEditorAltTextControl component.
+ *
+ * Replaces the core `alt_text` field to add an AI generate/regenerate button.
```
dkotter:
```suggestion
 * Replaces the core `alt_text` field to add a generate/regenerate button.
```

src/experiments/alt-text-generation/components/MediaEditorAltTextControl.tsx
```diff
@@ -0,0 +1,131 @@
...
+		}
+	};
+
+	return (
+		<div className="ai-alt-text-controls ai-alt-text-controls--media-editor">
+			{ /* Core alt text textarea (mirrors the field being replaced). */ }
+			<TextareaControl
+				label={ field.label }
+				hideLabelFromVision={ !! hideLabelFromVision }
+				value={ currentAlt || '' }
```
dkotter, line 99: Not sure if this is a bug from Gutenberg or a bug here in our code but in testing this out, the alt text I see populated in this field is the alt text stored globally with the image, not the alt text set at the post level.

To reproduce:

1. Upload an image directly to the Media Library
2. Set alt text for that image in the Media Library
3. Create a post and insert that image into the content
4. With the image selected, find the alt text field in the sidebar. It should show the alt text you set within the Media Library
5. Change this to something else and save the post
6. Click the Edit Media button in the block toolbar and find the alt text field in the sidebar
7. Note the contents of this field reflects the alt text from the Media Library, not from the image override

dkotter (changes requested):

src/experiments/alt-text-generation/components/AltTextControls.tsx
TylerB24890: I think we actually want to remove it from the `DecorativeNotice` component, because this `marginBottom` applies to the _Apply_ and _Dismiss_ buttons in this UI (where the other UI doesn't have those buttons).

src/experiments/alt-text-generation/media-editor.tsx
TylerB24890: We have to apply that to the `subscribe` and `doRegister()` methods in that case, but it does look better.

src/experiments/alt-text-generation/components/MediaEditorAltTextControl.tsx
TylerB24890: @dkotter I'm not sure, but I think this is expected? The image editor is meant to be an inline version of the Attachment edit page (e.g. `post.php?post=[attachment_id]&action=edit`). When you open the "Edit Image" feature from the block editor you can see the URL change to exactly that, but if you navigate to that URL from the browser directly you get the classic attachment Edit page. When you "Edit Image" from the block editor you are actually editing the global attachment, not just the instance of it inside the post.

If that _is not_ what you would expect, then that would be a bug with the Media Editor itself as it's the same experience without this PR.

dkotter: Ah, okay, yeah I understand now. I do think this is more confusing now as it drops you into the normal post editor experience so it feels like I'm editing that image specifically for that post, not in a global context. But yeah, not something for us to fix here though I'd expect other people to have a similar complaint so potentially something for us to track upstream

TylerB24890:
Updated this experiment to support the Media Editor Modal experiment as well, per the latest comment in #325 ([ref](https://github.com/WordPress/ai/issues/325#issuecomment-4394741109)). Same caveat applies here regarding not being able to replace the `alt_text` field in its same position (we have to re-register the field entirely to override it.)

**Edit:** Also updated the testing instructions in the PR description accordingly.

dkotter (approved):
This looks good to me and tests well! Anything else you're waiting on here @TylerB24890 or is this ready to be taken out of draft and merged?

TylerB24890:
> This looks good to me and tests well! Anything else you're waiting on here @TylerB24890 or is this ready to be taken out of draft and merged?

Good to go from my end!

PR #456, opened 2026-04-22
Add support for the new `gpt-image-2` model

dkotter:
## What?

Adds the new `gpt-image-2` model to our preferred model list

## Why?

OpenAI [recently announced](https://openai.com/index/introducing-chatgpt-images-2-0/) the new `gpt-image-2` model and it supposedly is significantly better than the previous model and it's also cheaper to use.

## How?

- Add `gpt-image-2` to our preferred model list
- Remove `gpt-image-1-mini` from that list

### Use of AI Tools

None

## Testing Instructions

1. Checkout this PR
2. Setup the OpenAI provider
3. Turn on image generation
4. Generate an image and ensure it works

## Changelog Entry

> Added - `gpt-image-2` to our preferred model list.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-456-e0b1ac7e00901ddfb3111204575159eab3ea3ca5.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -237,11 +237,11 @@ function get_preferred_image_models(): array {
 		),
 		array(
 			'openai',
-			'gpt-image-1.5',
+			'gpt-image-2',
 		),
 		array(
 			'openai',
-			'gpt-image-1-mini',
+			'gpt-image-1.5',
 		),
 	);
 
--- a/tests/Integration/Includes/HelpersTest.php
+++ b/tests/Integration/Includes/HelpersTest.php
@@ -527,13 +527,13 @@ class HelpersTest extends WP_UnitTestCase {
 		$this->assertIsArray( $result[4], 'Fifth model should be an array' );
 		$this->assertCount( 2, $result[4], 'Fifth model should have 2 elements' );
 		$this->assertEquals( 'openai', $result[4][0], 'Fifth model provider should be openai' );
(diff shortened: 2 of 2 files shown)


jeffpaul (approved):
Is there a way we can push these sorts of updates without having to ship a plugin release?

dkotter:
> Is there a way we can push these sorts of updates without having to ship a plugin release?

So this has been discussed multiple times previously but we could remove the preferred models list and rely on the AI Client and individual providers to determine what model to use. This _could_ result in the latest models being used without any code changes (in this case, `gpt-image-2`) but there's no guarantee that will be the case (depends on the sorting, if any, a provider brings and also depends on what order the AI Client lands on).

My preference is to maintain this preferred list and it's not hard for us to update these versions and have those go out in a subsequent release, but I do understand the benefit to removing this list and relying on the AI Client to decide.

PR #459, opened 2026-04-22
Add C2PA Monitor experiment

### LINKED ISSUE #421: WordPress should detect C2PA manifests on upload

WordPress extracts image metadata at upload — EXIF, IPTC, and a subset of XMP — through `wp_read_image_metadata()`. It does not detect C2PA Content Credentials.

C2PA manifests are now embedded in images from AI generators (DALL-E 3, Adobe Firefly, Google Gemini, Microsoft Copilot), camera hardware (Google Pixel, Samsung Galaxy S25, Leica, Sony, Nikon), and infrastructure providers (Cloudflare Images). These manifests carry machine-readable provenance: what tool created the image, whether AI was involved, and the image's modification history.

WordPress's image processing pipeline (both GD and Imagick) destroys C2PA manifests during subsize generation. The original uploaded file survives on disk, but the window to read the manifest is at upload — before processing begins.

No WordPress plugin or core feature detects C2PA manifests today.

A read-only experiment that extends WordPress's image metadata awareness to include C2PA. The experiment registers through `Abstract_Feature`, toggleable through the existing AI plugin settings like any other experiment.

At upload, it hooks into the attachment pipeline, reads the original file via `wp_get_original_image_path()`, and captures a structured record in postmeta (`_wpai_monitor_record`). The record includes:

- Traditional metadata (curated EXIF/IPTC/XMP fields WordPress already partially extracts but never surfaces)
- C2PA presence detection (scanning JPEG APP11 markers, PNG `caBX` chunks, WebP RIFF `C2PA` chunks)
- C2PA claim summary (claim generator, digital source type, action history — decoded from the JUMBF manifest store)

The C2PA parsing requires JUMBF box reading and CBOR decoding utilities that don't currently exist in the plugin. These would be introduced as shared infrastructure, available to any experiment that works with C2PA data. See reference implementations in #294.

**Design constraints:**

- Read-only — does not modify images, manifests, or core attachment fields
- Fail-open — any error during capture is caught at the experiment boundary; the upload always succeeds
- No external dependencies — no outbound HTTP, no external services; pure PHP within the plugin's existing requirements
- No upload delay — capture targets < 500ms median on images under 15MB

**Out of scope:**

- Signing or re-signing images
- Cryptographic verification of manifests
- Preserving manifests through WordPress's image processing pipeline
- UI for displaying captured metadata

lnispel:
Read-only feature that detects [C2PA Content Credentials](https://c2pa.org/) in uploaded JPEG/PNG/WebP images at the `add_attachment` hook, captures the raw manifest store to a sidecar file under `wp-content/uploads/ai-c2pa/`, and persists a structured `_wpai_monitor_record` postmeta entry for downstream consumers.

## What?

Closes #421

Adds a new `C2pa_Monitor` experiment that:

- Detects C2PA segments via streaming magic-byte/segment walks: JPEG APP11/JUMBF, PNG `caBX`, WebP RIFF `C2PA`. Hard byte caps throughout.
- Reassembles JPEG manifests fragmented across multiple APP11 markers by tracking Box Instance Numbers (per ISO 19566-5). Continuation segments do not need the `c2pa`/`jumb` token in their first 64 bytes.
- Streams raw manifest bytes to disk under `uploads/ai-c2pa/<attachment_id>.<format>.c2pa`, hashing in flight (SHA-256). Postmeta stores the hash, length, and relative sidecar path - not the bytes themselves.
- Creates the sidecar directory on demand with `.htaccess` (Apache deny) and `index.php` hardening. nginx operators must add a deny rule manually (documented in the experiment README).
- Wraps the entire capture path in a fail-open `try / catch ( Throwable )` boundary: errors land in the record's `errors[]` array and the upload itself is never blocked.
- Adds no external dependencies, no Composer additions, and no outbound HTTP. Pure PHP, compatible with the plugin's PHP 7.4 floor.
- Surfaces a **Content Credentials** status column in the Media Library list view (sortable), a field in the Attachment Details panel / media modal, and a meta box on the Edit Media screen.

## Why?

C2PA Content Credentials are increasingly embedded in images uploaded to WordPress sites (camera firmware, AI image generators, editorial signing tools), but WordPress's image processing pipeline destroys them - APP11 segments don't survive GD/Imagick re-encoding. Capturing the manifest at `add_attachment`, before subsize generation, is the only point in the WordPress lifecycle where the original bytes are still intact.

This PR establishes that capture as read-only infrastructure: the manifest data becomes available to downstream consumers (admin UI, REST endpoints, verification tooling) without each of them having to re-parse containers.

## How?

Implementation lives entirely under `includes/Experiments/C2pa_Monitor/` with one entry point:

- `C2pa_Monitor::capture_for_attachment()` is hooked to `add_attachment` at priority 20, gated by MIME on `image/jpeg`, `image/png`, and `image/webp`.
- `Format_Detector` walks containers and returns a location descriptor (segments + total length) without reading payload bytes.
- `Manifest_Reader` consumes that descriptor, streams the bytes via `fread` into a hash context and an in-memory buffer, and produces an immutable `Raw_Manifest` value object.
- `Sidecar_Writer` persists the bytes, ensuring the directory and hardening files exist exactly once.
- `Record` normalizes the structured payload and persists it as JSON-encoded postmeta at `_wpai_monitor_record`.

The admin UI is entirely read-only and consists of three surfaces wired in `register()`:

- **Media Library column** (`manage_media_columns` / `manage_media_custom_column`): "Content Credentials" column showing ? Credentials, No credentials, or - (not scanned). Sortable via `manage_upload_sortable_columns` + `pre_get_posts`. Hover tooltip via inline CSS on `admin_head-upload.php`.
- **Attachment Details / media modal** (`attachment_fields_to_edit`): same three-state badge with a visible help text paragraph. `show_in_edit => false` suppresses it on the Edit Media screen to avoid duplicating the meta box.
- **Edit Media meta box** (`add_meta_boxes_attachment`): "Content Credentials" side meta box with the badge and a `<p class="description">` help paragraph.

**Reviewing this PR.** The diff is large but cleanly partitioned. The work was developed in five layers and the integration branch's commits are organized that way:

1. Register the experiment (scaffolding, no behavior).
2. Define the record schema (`Raw_Manifest`, `Record`, tests).
3. Detect image formats (`Format_Detector`, fixtures, tests).
4. Read and persist manifests (`Manifest_Reader`, `Sidecar_Writer`, tests).
5. Wire the intake hook (`register()` body, `capture_for_attachment`, end-to-end tests).
6. Admin UI (Media Library column, sortable column, Attachment Details field, Edit Media meta box).

Each layer is independently coherent if you'd rather review incrementally; otherwise, the integrated diff stands on its own.

### Use of AI Tools

AI assistance: Yes
Tool(s): Cursor, Claude
Model(s): Claude Sonnet 4.6
Used for: Architecture and PR-segmentation planning, initial code drafts, test scaffolding. Final implementation, all design decisions, and the byte-level format work were reviewed and edited by me.

## Testing Instructions

**Capture pipeline**

1. Enable the AI plugin globally and toggle on **C2PA Monitor** under experiments.
2. Upload a JPEG with embedded C2PA credentials (e.g., from the [C2PA public test files](https://github.com/c2pa-org/public-testfiles)).
3. Inspect the attachment's postmeta for `_wpai_monitor_record` - it should be JSON with `c2pa.present: true`, a SHA-256 hash, and a `sidecar_path_relative` value.
4. Confirm the sidecar file exists at `wp-content/uploads/ai-c2pa/<attachment_id>.jpeg.c2pa` and that its bytes match `manifest_sha256`.
5. Upload a JPEG without C2PA - the postmeta should record `c2pa.present: false` and no sidecar should be written.
6. Upload a non-image (e.g., a `.txt` file) - no postmeta should be written at all.
7. Disable the experiment and re-upload - no postmeta should be written.

**Admin UI**

8. With the experiment enabled, open **Media > Library** in list view. A **Content Credentials** column should appear, showing ? Credentials for the image from step 2, and No credentials for the image from step 5. Hovering each cell should show a tooltip.
9. Click the **Content Credentials** column header to sort - credentials-present images should appear first.
10. In the Media Library, click an attachment to open the **Attachment Details** panel (modal). A **Content Credentials** field should appear with the status badge and a plain-text explanation beneath it.
11. Navigate to `wp-admin/upload.php?item=<id>` for the same attachment. The same **Content Credentials** field should appear in the details form.
12. Navigate to `wp-admin/post.php?post=<id>&action=edit`. A **Content Credentials** meta box should appear in the sidebar with the badge and help text. The field from step 11 should **not** appear in the main column (it is suppressed with `show_in_edit => false` to prevent duplication).

**Verification**

13. Clicking ? Credentials opens [verify.contentauthenticity.org](https://verify.contentauthenticity.org/) in a new tab. Download the original image from your Media Library and drag it into the verify tool to run a full cryptographic check. (Direct URL pre-fill is not supported because WordPress admin URLs are not reachable by the verify tool's fetcher from outside the session.)

Automated coverage:

- `Format_Detector`: magic bytes, single-segment APP11, multi-segment reassembly, interleaved APP0/APP1/APP2 around C2PA, generic JUMBF (non-C2PA) ignored, truncated input, `JPEG_MAX_SEGMENTS` cap (positive and negative), PNG/caBX, WebP simple + extended (VP8X) + odd-length padding.
- `Manifest_Reader`: byte-exact roundtrip for JPEG/PNG/WebP, multi-segment reassembly, deterministic SHA-256, `MAX_MANIFEST_BYTES` rejection, missing file, empty segments, bad offsets.
- `Sidecar_Writer`: write + roundtrip, hardening files, format sanitization, overwrite, multi-attachment coexistence, custom `.htaccess` preserved across `ensure_dir()`.
- `Record`: roundtrip, defaults on empty input, JSON-not-serialize storage format, null on corrupt JSON, null when absent.
- `C2pa_Monitor` end-to-end: JPEG/PNG/WebP present, JPEG absent, unsupported MIME, fail-open on bogus ID, truncated JPEG, `duration_ms` recorded, `add_attachment` hook actually fires, file-deleted-on-disk produces `errors[0].stage = 'resolve_path'`, column enabled/disabled, column cell states (present/absent/dash/malformed), tooltip present in column, tooltip absent on detail surfaces, `show_in_edit` false, `helps` key populated, meta box registration, meta box renders help paragraph, verify link has no `source=` query param, sort column registration and query modification.

Synthetic fixtures are generated at runtime so no binary blobs land in the repo and there is no third-party fixture licensing question.

## Deferred (out of scope for this PR)

- [#953](https://github.com/WordPress/ai/issues/953) - JUMBF box reader and CBOR decoder; populating `c2pa.decoded` claim summary (claim generator, digital source type, action history).
- [#954](https://github.com/WordPress/ai/issues/954) - Cryptographic verification in-browser via the C2PA JS SDK (`@contentauth/sdk`, targeting 1.4.0) - would allow verification without leaving the admin and without relying on publicly reachable URLs.
- [#955](https://github.com/WordPress/ai/issues/955) - Media Library grid view indicator. The C2PA CR trustmark is limited to products on the C2PA conforming products list and denotes validated credentials, so this ships as a neutral plugin-specific badge using "detected" language, with the CR icon revisited if WordPress joins the list.
- [#956](https://github.com/WordPress/ai/issues/956) - Preserving manifests through WordPress's GD/Imagick subsize pipeline.
- [#957](https://github.com/WordPress/ai/issues/957) - Pre-filling the CAI Verify tool with a publicly reachable attachment URL (`?source=`).

## Screenshots or screencast

Demo video posted in thread: [c2pa_monitor.mp4](https://github.com/user-attachments/assets/43b2a9bc-a64d-4c99-ab85-b7239aa2ef11)

## Changelog Entry

> Added - New Experiment: C2PA Monitor - read-only detection of [C2PA Content Credentials](https://c2pa.org/) in uploaded JPEG/PNG/WebP images. Captures the raw manifest store to a sidecar file under `wp-content/uploads/ai-c2pa/` and stores a structured `_wpai_monitor_record` postmeta entry for downstream consumers. Adds a sortable Content Credentials column to the Media Library and surfaces the credential status on the Attachment Details and Edit Media admin screens. Fail-open and never blocks an upload. JUMBF/CBOR claim decoding and in-browser cryptographic verification deferred to follow-up PRs.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-459-b9a1984a0dea6778ec7a3dff324a5f79bb801c8f.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,299 @@
...
+}
+
+/**
+ * C2PA Monitor experiment class.
+ *
+ * Hooks into add_attachment and captures a structured `_wpai_monitor_record`
+ * for every uploaded image. The capture is read-only, fail-open, and never
+ * blocks the upload pipeline.
+ *
+ * @since 0.7.0
```
dkotter: We'll want to update all `@since` statements in this PR to be `x.x.x` as we'll replace those with the proper version number during the release process

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,299 @@
...
+	 *
+	 * @var int
+	 */
+	public const MAX_SCAN_BYTES = 67108864; // 64 MB.
+
+	/**
+	 * JSON-LD context URL embedded in every stored postmeta record.
+	 *
+	 * Interim URL resolving via raw.githubusercontent.com. Migrate to
+	 * https://w3id.org/openverifiable/v1 once the w3id.org PR is merged
```
dkotter: This mentions a PR, can we add the link to that here so it can easily be tracked?

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,299 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function load_metadata(): array {
+		return array(
+			'label'       => __( 'C2PA Monitor', 'ai' ),
+			'description' => __( 'Detects C2PA Content Credentials in uploaded images and stores the raw manifest plus a structured record in postmeta. Read-only and fail-open; never blocks an upload.', 'ai' ),
+			'category'    => Experiment_Category::ADMIN,
+			'stability'   => 'experimental',
```
dkotter, line 107: We now have a new `capability` property that we'll want to set to `none` here

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,299 @@
...
+			if ( ! self::is_supported_mime( $mime ) ) {
+				$should_persist = false;
+				return;
+			}
+
+			$path = self::get_original_path( $attachment_id );
+			if ( '' === $path || ! is_readable( $path ) ) {
+				$errors[] = array(
+					'stage'   => 'resolve_path',
+					'message' => 'Attachment file is not readable.',
```
dkotter: Haven't seen where we output these error messages yet but if these end up being user facing, we should pass these through translation. Example:
```suggestion
					'message' => esc_html__( 'Attachment file is not readable.', 'ai' ),
```

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,299 @@
...
+	 * Falls back to get_attached_file() when wp_get_original_image_path() does
+	 * not return a usable path (non-image attachments, edited media, etc.).
+	 *
+	 * @since 0.7.0
+	 *
+	 * @param int $attachment_id Attachment ID.
+	 * @return string Absolute filesystem path, or empty string when unresolved.
+	 */
+	private static function get_original_path( int $attachment_id ): string {
+		if ( function_exists( 'wp_get_original_image_path' ) ) {
```
dkotter: Do we need this check? Our minimum WP version is 7.0 and this function appears to have been introduced in 5.3

includes/Experiments/C2pa_Monitor/README.md
dkotter, line 1: Right now documentation for each experiment lives under `docs > experiments` so probably want to move this there

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 All notable changes to this project will be documented in this file, per [the Keep a Changelog standard](http://keepachangelog.com/), and will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased] - TBD
+### Added
+- New Experiment: C2PA Monitor — read-only detection of [C2PA Content Credentials](https://c2pa.org/) in uploaded JPEG/PNG/WebP images. Captures the raw manifest store to a sidecar file under `wp-content/uploads/ai-c2pa/` and stores a structured `_wpai_monitor_record` postmeta entry for downstream consumers. Fail-open and never blocks an upload. JUMBF/CBOR claim decoding deferred to a follow-up PR.
```
dkotter: No need to add changelog entry here, just within the PR description. We'll populate this when we do a release

readme.txt
```diff
@@ -20,6 +20,7 @@ This plugin is built on the [AI Building Blocks for WordPress](https://make.word
 
 * **Abilities Explorer** – Browse and interact with registered AI abilities from a dedicated admin screen.
 * **Alt Text Generation** - Generate descriptive alt text for images to improve accessibility.
+* **C2PA Monitor** – Detects [C2PA Content Credentials](https://c2pa.org/) in uploaded images and stores the raw manifest plus a structured record alongside the attachment. Read-only and fail-open; never blocks an upload.
```
dkotter, line 26: Likely want a similar line in the README.md file

includes/Experiments/C2pa_Monitor/Sidecar_Writer.php
```diff
@@ -0,0 +1,143 @@
...
+		if ( ! file_exists( $index ) ) {
+			file_put_contents( $index, "<?php\n// Silence is golden.\n" );
+		}
+
+		$htaccess = trailingslashit( $basedir ) . '.htaccess';
+		if ( file_exists( $htaccess ) ) {
+			return;
+		}
+
+		$rules = "# C2PA Monitor sidecar files - block direct web access (Apache).\n"
```
dkotter, line 116: Seems this only applies to Apache but won't impact nginx or other providers. How concerned are we with these files being publicly accessible? If they shouldn't be accessed, may be best to store these in a non-public location instead of relying on an htaccess file

dkotter (changes requested):

jeffpaul:
@lnispel are you able to update from code review here?

jeffpaul:
@lnispel checking back again to see If you're able to make updates based on code review here?  I'm keeping this on our roadmap but continuing to punt from releases awaiting updates on your end, so if you're unlikely to get back to this in the near future please let us know so we can see if there's someone else who could pick this up... thanks!

lnispel:
> @lnispel checking back again to see If you're able to make updates based on code review here?  I'm keeping this on our roadmap but continuing to punt from releases awaiting updates on your end, so if you're unlikely to get back to this in the near future please let us know so we can see if there's someone else who could pick this up... thanks!

Thanks for the push, Jeff. I'm still in a bit of a personal transition so my time is spotty right now. I'll continue to refine this to get things more settled but I am open to someone else assisting or focusing on the code contributions specifically if it's holding up the roadmap/someone else is interested.

readme.txt
lnispel: Done - added a matching C2PA Monitor bullet under Current Features in root README.md, linking to docs/experiments/c2pa-monitor.md.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Thanks - good catch. This is still outstanding: the C2PA files currently use @since 0.7.0. Per CONTRIBUTING / RELEASE_INSTRUCTIONS we should switch new code to @since x.x.x so release can fill the real version. I'll follow up with that change shortly.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/Sidecar_Writer.php
lnispel: Good question. We kept sidecars next to uploads (with .htaccess / index.php hardening as defense-in-depth for Apache) rather than moving them out of the public tree, because the sidecar bytes are a verbatim extract of the C2PA manifest already embedded in the original attachment - WordPress serves that original publicly and only strips C2PA from generated subsizes. So the sidecar does not expose anything that isn't already reachable via the original image URL, and co-locating keeps lifecycle parity (backups / migrations / deletion with the attachment).

Documented this under **Access considerations** in docs/experiments/c2pa-monitor.md. Happy to revisit if maintainers prefer a non-public store anyway.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Done - human-readable error / exception messages in C2pa_Monitor and Sidecar_Writer are now wrapped with esc_html__() / esc_html__() + sprintf (text domain i), so they're translation-ready if they surface to users.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Agreed. The CONTEXT_URL docblock currently references "the w3id.org PR" without a link. I don't have a stable public PR URL handy for that migration yet - if you (or Jeff) have the preferred link, I can drop it into the docblock; otherwise I'll add a clearer @todo with the target repo/path once we know it.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/README.md
lnispel: Done - moved the experiment docs to docs/experiments/c2pa-monitor.md (and removed the old includes/Experiments/C2pa_Monitor/README.md), matching the other experiments.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Agreed - removed the unction_exists( 'wp_get_original_image_path' ) guard. Minimum supported WP is 7.0 and that API has been around since 5.3, so the direct call is fine. The test was updated to match.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Done - load_metadata() now sets 'capability' => 'none' (read-only admin experiment), and the integration test asserts get_capability() returns 
one.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

CHANGELOG.md
lnispel: Done - removed the C2PA entry from CHANGELOG.md Unreleased. Changelog will be populated at release; the PR description covers the change for now.

---\n_Reply posted via the Cursor coding assistant on behalf of @lnispel._

lnispel:
## Review follow-up

Replied on all of @dkotter's inline review threads.

**Addressed in recent commits:**
- capability => 'none'
- i18n for user-facing error strings (esc_html__)
- removed obsolete unction_exists( 'wp_get_original_image_path' ) guard
- docs moved to docs/experiments/c2pa-monitor.md
- CHANGELOG Unreleased entry removed
- matching bullet added to root README.md
- sidecar public-access rationale documented under Access considerations
- PHPStan is_array() finding fixed
- branch merged with latest develop

**Still open (called out in the thread replies):**
1. Switch new @since 0.7.0 tags to @since x.x.x
2. Add a concrete link for the w3id.org CONTEXT_URL migration PR (need the preferred URL)

---
_Posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Done - I swept the C2PA source and test/fixture files from @since 0.7.0 to @since x.x.x, matching the repo's release-process convention for new code.

Also, for the separate CONTEXT_URL migration link: there was an earlier w3id PR at https://github.com/perma-id/w3id.org/pull/6007, but it was closed for inactivity. We'll reopen or replace that PR, then I can add the concrete link back into the docblock thread once there's an active URL to point at.

---
_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: https://github.com/perma-id/w3id.org/pull/6376

lnispel:
Demo Video

https://github.com/user-attachments/assets/43b2a9bc-a64d-4c99-ab85-b7239aa2ef11

lnispel:
Screenshot 2026-08-09 211013

jeffpaul:
I grabbed test files from https://github.com/c2pa-org/public-testfiles.  Some testing feedback:

1. Are there any other variations shown in the Content Credentials column besides "✓ Credentials" and "No credentials"?  We may want to consider the column naming and the names of the options displayed to make sure they're more clear to users new to the c2PA concept; nothing immediately different/better comes to mind, but something I'm trying to consider.
2.  When clicking "✓ Credentials" is there a way we can either pre-fill that page with the image URL (or the image itself if the site/image isn't publicly available via URL) or is there a more automated method that the image could be sent off for verification and then updating the status in that column?
3. There's no Content Credentials shown on the Attachment details screen (e.g. https://example.com/wp-admin/upload.php?item=7) or Edit Media screen (e.g. https://example.com/wp-admin/post.php?post=7&action=edit), can we include on those views as well?

jeffpaul:
1. Same question as above.
 
2. Testing via Playground likely means my test images aren't fully available on the internet to access via link so clicking the "✓ Credentials" link gets to something like https://verify.contentauthenticity.org/?source=https%3A%2F%2Fplayground.wordpress.net%2Fscope%3Acurious-chic-river%2Fwp-content%2Fuploads%2F2026%2F08%2Fa-bad-01-scaled.jpg which doesn't load the image into the inspector on the CAI verify site.

3a. Looks like on the "Edit Media" screen that the CC info is separate from the metabox set up for that info and the "No credentials" and "✓ Credentials" text don't include the on-hover helper text.

Screenshot 2026-08-13 at 1 59 27 PM

3b. On the "Attachment details" screen the on-hover text shows but gets overlapped by some element and isn't fully rendered

Screenshot 2026-08-13 at 2 01 27 PM

Screenshot 2026-08-13 at 2 01 35 PM

lnispel:
Thanks @jeffpaul, addressing all three items from your Aug 13 testing pass:

**1. Column states**

There are three: "? Credentials", "No credentials", and "-" (dash - no scan record, e.g. uploaded before the experiment was enabled or a non-image MIME type). The labels are intentionally minimal for now. Once we validate using the C2PA JS SDK and can surface decoded claim details (claim generator, digital source type, signing authority), we'll have the context needed to make the language more specific and can refine from there.

**2. Verify link / ?source= pre-fill**

Good catch - the `?source=` parameter was a mistake and I've reverted it. WordPress attachment URLs are not publicly reachable from the CAI verify tool's servers regardless of whether the site is "live", because verification runs outside the admin session (local, Playground, staging, and any auth-gated site all fail the same way). The link now points at the bare `verify.contentauthenticity.org` URL, and a help text note below each credential badge tells the user to download the original image and drag it into the tool.

The right long-term fix is in-browser verification via `@contentauth/sdk`, which can fetch the attachment via the user's authenticated browser session and run the check inline - no redirect needed. I've noted that as a deferred follow-up in the PR description.

**3a. Edit Media - duplicate display and missing tooltip**

The CC info was appearing in both the main column (via `attachment_fields_to_edit`) *and* the sidebar meta box on `post.php?post=<id>&action=edit`. Fixed by setting `'show_in_edit' => false` on the attachment field, which is the same approach `Alt_Text_Generation` uses in this codebase. The meta box is the only surface on Edit Media now.

The tooltip was absent because `print_column_styles()` only fires on `admin_head-upload.php`. Rather than extending the CSS tooltip to more screens, I replaced it with a visible `<p class="description">` help paragraph on the meta box - more appropriate for that screen's layout anyway.

**3b. Attachment details - tooltip clipping**

The CSS tooltip's `position: absolute; bottom: calc(100% + 6px)` was getting clipped by the media modal's `overflow: hidden` parent. Fixed by using the `'helps'` key on the `attachment_fields_to_edit` field instead, which WordPress renders natively as `<p class="help">` beneath the field - no absolute positioning, nothing to clip.

---

**Corrected w3id.org link** (also in this commit): the `CONTEXT_URL` docblock referenced pull/6007 but the PR that actually merged was #6376 - fixed in both `C2pa_Monitor.php` and `docs/experiments/c2pa-monitor.md`.

**PR description** updated to reflect the current state: `Closes #421`, the three admin UI surfaces are documented in What/How/Testing Instructions/Changelog Entry, "No UI changes" is replaced with a link to the demo video, and the Deferred list now accurately reflects what remains (JUMBF/CBOR decoding, in-browser SDK verification, grid view badge).

---
_Reply posted via the Cursor coding assistant on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+		return 'c2pa-monitor';
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function load_metadata(): array {
+		return array(
+			'label'       => __( 'C2PA Monitor', 'ai' ),
+			'description' => __( 'Detects C2PA Content Credentials in uploaded images and stores the raw manifest plus a structured record in postmeta. Read-only and fail-open; never blocks an upload.', 'ai' ),
```
dkotter: Is this accurate? Seems we store the raw manifest in a sidecar file, not postmeta as this seems to imply
```suggestion
			'description' => __( 'Detects C2PA Content Credentials in uploaded images and stores the raw manifest in a sidecar file plus a structured record in postmeta. Read-only and fail-open; never blocks an upload.', 'ai' ),
```

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	 *
+	 * Only output on the Media Library screen (admin_head-upload.php) and only
+	 * when the experiment is enabled.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return void
+	 */
+	public function print_column_styles(): void {
+		if ( ! $this->is_enabled() ) {
```
dkotter: No need for this check as `register` only fires if the experiment is enabled

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	/**
+	 * Registers the C2PA status column in the Media Library list table.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array<string, string> $columns Existing columns.
+	 * @return array<string, string>
+	 */
+	public function add_media_column( array $columns ): array {
+		if ( ! $this->is_enabled() ) {
```
dkotter: Same comment here, this check already runs prior to calling `register`

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+		$record = json_decode( $raw, true );
+		if ( ! is_array( $record ) || ! isset( $record['c2pa']['present'] ) ) {
+			return '<span aria-label="' . esc_attr__( 'Not scanned', 'ai' ) . '">—</span>';
+		}
+
+		if ( $record['c2pa']['present'] ) {
+			$tooltip = $with_tooltip
+				? ' data-wpai-tooltip="' . esc_attr__( 'Unverified — credentials were detected but have not been validated. Click to open the Content Authenticity Initiative verify tool.', 'ai' ) . '"'
+				: '';
+			return '<a href="' . esc_url( 'https://verify.contentauthenticity.org/' ) . '" target="_blank" rel="noopener noreferrer"'
```
dkotter: Likely no need to run the URL through escaping here as it's hardcoded

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	 *   was enabled, or a non-image MIME type).
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $column_name The column being rendered.
+	 * @param int    $post_id     The attachment post ID.
+	 * @return void
+	 */
+	public function render_media_column( string $column_name, int $post_id ): void {
+		if ( 'wpai_c2pa' !== $column_name || ! $this->is_enabled() ) {
```
dkotter: Same comment here, no need for the `is_enabled` check here as this check already runs prior to calling `register`

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	 * modal's overflow container.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array<string, mixed> $form_fields Existing form fields.
+	 * @param \WP_Post             $post        The attachment post object.
+	 * @return array<string, mixed>
+	 */
+	public function add_attachment_fields( array $form_fields, \WP_Post $post ): array {
+		if ( ! $this->is_enabled() ) {
```
dkotter: Same comment here, this check already runs prior to calling `register`

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	 * Fires on the `add_meta_boxes_attachment` action, which runs when loading
+	 * the classic `post.php?post=<id>&action=edit` screen for an attachment.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param \WP_Post $post The attachment post object.
+	 * @return void
+	 */
+	public function add_attachment_meta_box( \WP_Post $post ): void {
+		if ( ! $this->is_enabled() ) {
```
dkotter: Same comment here, this check already runs prior to calling `register`

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	/**
+	 * Marks the Content Credentials column as sortable.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array<string, string|array<int, string|bool>> $columns Sortable columns map.
+	 * @return array<string, string|array<int, string|bool>>
+	 */
+	public function register_sortable_column( array $columns ): array {
+		if ( ! $this->is_enabled() ) {
```
dkotter: Same comment here, this check already runs prior to calling `register`

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	 * sort; those with no credentials (0) come next; unscanned attachments
+	 * (no sort meta row) appear last via a separate JOIN.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param \WP_Query $query The current query.
+	 * @return void
+	 */
+	public function sort_by_c2pa_column( \WP_Query $query ): void {
+		if ( ! $this->is_enabled() ) {
```
dkotter: Same comment here, this check already runs prior to calling `register`

includes/Experiments/C2pa_Monitor/Manifest_Reader.php
```diff
@@ -0,0 +1,142 @@
+<?php
+/**
+ * Reads raw C2PA manifest store bytes from a located image segment.
+ *
+ * @package WordPress\AI
+ */
+
+declare( strict_types=1 );
+// phpcs:disable WordPress.WP.AlternativeFunctions
+// phpcs:disable WordPressVIPMinimum.Functions.RestrictedFunctions
```
dkotter: I'm assuming there are functions we are using here that will throw errors without these lines. Would be great though to document why here so it's clear to anyone looking at this code why we disabled these sniffs

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	 *
+	 * Stores a single integer: 1 = credentials present, 0 = absent.
+	 * Written alongside POSTMETA_KEY so the Media Library can ORDER BY it.
+	 * Not written when no scan record exists (unsupported MIME / pre-existing upload).
+	 *
+	 * @since x.x.x
+	 *
+	 * @var string
+	 */
+	public const SORT_META_KEY = '_wpai_c2pa_present';
```
dkotter, line 53: We now have an `Uninstall` script that runs when the plugin is deleted. This cleans up some things so we don't leave a bunch of data behind. This PR now adds a handful of things:

1. Meta stored at `_wpai_monitor_record`
2. Meta stored at `_wpai_c2pa_present`
3. Sidecar file

Do you think any of this should be cleaned up when the plugin is deleted? We made the decision to leave most meta behind in case anyone still needs that for something but seems like deleting `_wpai_c2pa_present` may be nice as that's not super useful by it's own. And may be worth discussing deleting the sidecar files

jeffpaul:
1. Ok, I think we'll want to look at iteration on all the copy shown but I prefer to get this experiment into the plugin and then based on community feedback can iterate on the copy in a subsequent release.

2. Please open a new issue to document that long-term fix as something we can follow-up on in the 1.4.0 release.  And it looks like there are 4 bullets in that "Deferred" section in the PR description, so would be great to have issues created for each of them so we can track improvements after getting the base experiment into the plugin.

Separately, for WordPress installs that do have publicly available image links, could those image links be more easily passed to the verification site/tool to avoid the WordPress site owner having to download/upload?

3a. That works, thanks!

3b. The "Content Credentials" label is misaligned and showing a row above the "✓ Credentials" value for that image, let's please ensure they're aligned on the same row

Screenshot 2026-08-18 at 12 06 42 PM

includes/Experiments/C2pa_Monitor/Sidecar_Writer.php
```diff
@@ -0,0 +1,143 @@
+<?php
+/**
+ * Writes raw C2PA manifest bytes to a sidecar file under the uploads dir.
+ *
+ * @package WordPress\AI
+ */
+
+declare( strict_types=1 );
+// phpcs:disable WordPress.WP.AlternativeFunctions
+// phpcs:disable WordPressVIPMinimum.Functions.RestrictedFunctions
```
dkotter: Same comment here, we should document the why behind disabling these

includes/Experiments/C2pa_Monitor/Format_Detector.php
```diff
@@ -0,0 +1,478 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @param resource              $fh                  File handle, positioned at the segment payload.
+	 * @param int                   $payload_length      Length of this segment's payload, in bytes.
+	 * @param int                   $payload_offset      Absolute offset of the segment payload.
+	 * @param array<int, true>      $approved_instances  Set of Box Instance Numbers already classified as C2PA. Updated by reference when a new sequence is approved.
+	 * @return array{0:int,1:int}|null [offset, length] of inner JUMBF bytes, or null.
+	 */
+	private function extract_jpeg_app11_jumbf_slice( $fh, int $payload_length, int $payload_offset, array &$approved_instances ): ?array {
```
dkotter, line 317: I don't know if this is accurate as I'll admit I don't fully understand what it's saying but just flagging that passing things through Claude, it returned the following issue from this method. If this isn't accurate, feel free to ignore:

**APP11 header size is wrong (12 vs 8), real C2PA JPEGs are captured corrupted**

`extract_jpeg_app11_jumbf_slice()` assumes `CI(2) + En(2) + Z(8) = 12` and slices from `$payload_offset + 12`. Per ISO/IEC 19566-5 Annex B, `Z` is 4 bytes, followed by `LBox(4) + TBox(4)` repeated in every segment. Confirmed in the code: it reads a 12-byte header, checks `JP` at 0 and `En` at 2, then slices from +12 — which lands on `TBox` with `LBox` dropped, and every continuation segment splices its repeated LBox/TBox inline. The sidecar isn't a valid JUMBF store and `manifest_sha256` won't match what any verifier computes. The fixtures can't catch it: `Fixtures::write_jpeg_with_c2pa()` uses `pack('J', 1)` (8-byte Z) and omits LBox/TBox, so it encodes the same misreading. Needs a real C2PA asset (c2pa-org/public-testfiles) before merge. PNG/WebP unaffected.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,629 @@
...
+	public function sort_by_c2pa_column( \WP_Query $query ): void {
+		if ( ! $this->is_enabled() ) {
+			return;
+		}
+
+		if ( ! is_admin() || ! $query->is_main_query() || 'wpai_c2pa' !== $query->get( 'orderby' ) ) {
+			return;
+		}
+
+		$query->set( 'meta_key', self::SORT_META_KEY );
```
dkotter: So the method claims that sorting will still show unscanned documents but I don't think that's the case as those won't have this meta key set and I believe they'll be removed from the sort results

dkotter (changes requested):

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Good catch, and you were right. Fixed in b3a3403 — the description now reads:

> Detects C2PA Content Credentials in uploaded images, writes the raw manifest to a sidecar file, and stores a structured record in postmeta. Read-only and fail-open; never blocks an upload.

That matches where the bytes actually land: postmeta holds the SHA-256, the byte length and the relative sidecar path, never the manifest itself.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Removed in b3a3403. You're right that `register()` only runs for enabled experiments, so every one of these guards was dead weight — all of them are gone from the class now.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Removed in b3a3403 along with the rest of the redundant `is_enabled()` guards.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Removed in b3a3403 along with the rest of the redundant `is_enabled()` guards.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Removed in b3a3403 along with the rest of the redundant `is_enabled()` guards.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Removed in b3a3403 along with the rest of the redundant `is_enabled()` guards.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Removed in b3a3403 along with the rest of the redundant `is_enabled()` guards.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Removed in b3a3403 along with the rest of the redundant `is_enabled()` guards.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Agreed — the URL is a hardcoded literal, so `esc_url()` bought nothing. Removed in 1607ed9; the anchor now embeds the literal directly.

includes/Experiments/C2pa_Monitor/Manifest_Reader.php
lnispel: Fair point — `Format_Detector` explained its two disables and these did not. Documented in 1607ed9:

- `WordPress.WP.AlternativeFunctions` — manifests are read with streaming `fopen`/`fread`/`fseek` so the bytes can be hashed in flight. `WP_Filesystem::get_contents()` would pull the whole file into memory, which is exactly what `MAX_MANIFEST_BYTES` exists to avoid.
- `WordPressVIPMinimum.Functions.RestrictedFunctions` — the only paths read here come from `wp_get_original_image_path()` for the attachment being processed.

includes/Experiments/C2pa_Monitor/Sidecar_Writer.php
lnispel: Documented in 1607ed9, same as `Manifest_Reader`:

- `WordPress.WP.AlternativeFunctions` — the sidecar write uses `file_put_contents( ..., LOCK_EX )` because it has to be atomic and byte-exact. `WP_Filesystem` has no locking equivalent and is not guaranteed to be initialised during an upload request.
- `WordPressVIPMinimum.Functions.RestrictedFunctions` — every path written here is under `wp_upload_dir()['basedir'] . '/' . self::SUBDIR`.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Yes, all three should be cleaned up — thanks for catching that this PR was adding data the uninstall routine did not know about. Implemented in 1607ed9:

- `delete_meta()` now calls `delete_post_meta_by_key()` for both `_wpai_monitor_record` and `_wpai_c2pa_present`. Both keys are written by this plugin and are meaningless without it.
- A new `delete_c2pa_sidecars()` removes the `*.c2pa` files plus the `.htaccess` / `index.php` hardening files that `Sidecar_Writer::ensure_dir()` writes, then removes the directory itself — but only once it is empty, so anything an operator dropped in there survives along with the directory.

Both run behind the existing `wpai_remove_data_on_uninstall` filter, so opting out preserves the sidecars too. Nothing is lost that cannot be re-derived: the attachment originals still carry the same manifest, so re-enabling the experiment can rebuild it.

Covered by three additions to `UninstallTest`: removal on the default path, preservation under the opt-out filter, and a case seeding a foreign file in the sidecar directory to prove the directory and that file both survive.

includes/Experiments/C2pa_Monitor/Format_Detector.php
lnispel: That analysis was correct, and it was a real bug — thanks for flagging it rather than skipping it. Fixed in b3a3403.

The prefix is `CI(2) + En(2) + Z(4) = 8` bytes, not 12. The old code read a 12-byte prefix and sliced from `+12`, which landed mid-way into the box header and dropped `LBox`, so the reassembled sidecar was not a valid JUMBF store and `manifest_sha256` would not have matched any verifier.

What the parser does now:

- First segment (`Z === 1`): slice from `payload + 8`, and validate before accepting it — `TBox` at +4 must be `jumb`, the inner `TBox` at +12 must be `jumd`, and the 16 bytes at +16 must be the C2PA type UUID.
- Continuation segments (`Z > 1`): the `LBox`/`TBox` pair is repeated in every segment, so those 8 bytes are skipped and the slice starts at `payload + 16`.

You were also right that the fixtures encoded the same misreading. `Fixtures::write_jpeg_with_c2pa()` used `pack('J', 1)` for an 8-byte `Z` and omitted `LBox`/`TBox` entirely, so it could never have caught this. The generator now emits the correct 4-byte `Z` and a real box header, and the multi-segment fixture repeats the header per segment the way a real encoder does.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: You were right — the first implementation set `meta_key` + `orderby=meta_value_num`, which turns into an INNER JOIN and silently drops every attachment without the key. Fixed in b3a3403.

It now sets a named `meta_query` clause with `relation => OR`, pairing an `EXISTS` clause against `wpai_c2pa_sort` with a `NOT EXISTS` clause, and orders by the named clause. Unscanned attachments are retained and sort last.

There is a regression test for exactly this: it seeds three attachments (credentials present, credentials absent, never scanned) and asserts all three come back from the sorted query rather than just the two carrying meta.

lnispel:
## Review round follow-up

@dkotter — I've replied inline on all 14 threads from your Aug 18 pass. Apologies for the silence: most of these were actually fixed on the 18th, but I never closed the loop on the threads themselves, so from your side the review looked ignored.

**Fixed in b3a3403**, the push right after your review:

- Experiment description now says the manifest goes to a sidecar file, with only a structured record in postmeta.
- All seven redundant `is_enabled()` guards removed — you were right that `register()` only runs when the experiment is enabled.
- APP11 prefix corrected from 12 to 8 bytes, with proper `LBox`/`TBox` handling per ISO/IEC 19566-5 Annex B. That analysis you passed along was correct and it was a real bug: the reassembled sidecar would not have been a valid JUMBF store. You were also right that the fixtures encoded the same misreading, so the generator was fixed too.
- Sortable column no longer drops unscanned attachments. It uses a named `meta_query` clause pairing `EXISTS` with `NOT EXISTS` under `relation => OR`, with a regression test seeding all three states.

**Fixed in 1607ed9**, just pushed:

- Dropped `esc_url()` from the hardcoded verify link.
- Documented why the two PHPCS sniffs are disabled in `Manifest_Reader` and `Sidecar_Writer`, matching what `Format_Detector` already explained.
- Uninstall now clears `_wpai_monitor_record`, `_wpai_c2pa_present`, and the `ai-c2pa` sidecar directory, behind the existing `wpai_remove_data_on_uninstall` filter. Files the plugin did not write are preserved and the directory is only removed once empty. Three new `UninstallTest` cases cover removal, opt-out, and the foreign-file case.

Also cleared three PHPCS failures that were blocking CI (short ternary, early exit, Yoda condition) and merged `develop`, which had fallen 10 commits behind.

---

@jeffpaul — on your Aug 18 items:

1. Agreed on landing the experiment first and iterating on copy from community feedback.
2. Issues are open for every Deferred bullet: #953, #954, #955, #956, #957.

3b. Fixed. The **Content Credentials** label and its value now render on the same row; `print_admin_styles()` ships the alignment rule for the Attachment Details compat field on both `upload.php` and `post.php`.

On your separate question about public image links — yes, technically. The verify tool accepts a `?source=<url>` parameter and fetches the image itself, so on a site whose uploads are publicly reachable that would remove the download/upload step entirely. The reason it is not in this PR is that we cannot tell from inside wp-admin whether a given install qualifies. Playground, local development, staging behind basic auth, and private or members-only sites all produce a link that silently fails, which is a worse experience than no link at all. That is tracked in #957.

The better long-term answer is #954: verifying in the browser with the C2PA JS SDK needs no publicly reachable URL, and behaves identically on a laptop and a production host. It would also let us replace the bare **Credentials** presence badge with an actual validation result, which should make the copy question in item 1 much easier to settle.

Separately, I'm revising #955. I had filed it as a CR icon badge, but the CR pin is a certification mark restricted to products on the C2PA conforming products list, so it is not ours to ship today. I'm rescoping that issue to a neutral, plugin-specific badge, with the trustmark revisited if WordPress joins the conforming products list.

---

_Replied via Cursor on behalf of @lnispel._

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,623 @@
...
+	 * sort; those with no credentials (0) come next; unscanned attachments
+	 * (no sort meta row) appear last. A named clause with `compare => 'EXISTS'`
+	 * uses a LEFT JOIN so rows without the meta key are still returned.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param \WP_Query $query The current query.
+	 * @return void
+	 */
+	public function sort_by_c2pa_column( \WP_Query $query ): void {
```
dkotter: This sorting still doesn't work for me. If I sort by either ASC or DESC, I never see any files that have credentials, they all just disappear

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,623 @@
...
+		if ( ! is_admin() || ! $query->is_main_query() || 'wpai_c2pa' !== $query->get( 'orderby' ) ) {
```
dkotter: Should likely also gate this to the attachment list screen

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,623 @@
...
+	 * and the label-alignment fix for the Attachment Details compat field.
+	 *
+	 * Hooked to admin_head-upload.php (Media Library + upload.php?item=<id>)
+	 * and admin_head-post.php (Edit Media screen).
+	 *
+	 * @since x.x.x
+	 *
+	 * @return void
+	 */
+	public function print_admin_styles(): void {
```
dkotter: So the media modal can load in lots of different places besides just the Media Library or Edit Media screen. This may be fine for now but might need extended in the future

includes/Admin/Uninstall.php
```diff
@@ -221,6 +224,79 @@ private static function delete_meta(): void {
...
+	 * Sidecars hold raw manifest bytes extracted from attachments at upload
+	 * time. The attachment originals still carry the same manifest, so nothing
+	 * is lost here that re-enabling the experiment could not re-derive. Files
+	 * this plugin did not write are left in place, and the directory itself is
+	 * only removed once it is empty.
+	 *
+	 * @since x.x.x
+	 */
+	private static function delete_c2pa_sidecars(): void {
+		$uploads = wp_upload_dir();
```
dkotter: Might want to pass `false` for the [second param here](https://developer.wordpress.org/reference/functions/wp_upload_dir/) so it won't create new directories on what is supposed to be a cleanup routine

includes/Experiments/C2pa_Monitor/Record.php
```diff
@@ -0,0 +1,170 @@
...
+ *     Subject-only JSON Schema (DIF credential-schemas; wrap in a VC at issuance if needed).
+ * @see C2pa_Monitor::CONTEXT_URL for the JSON-LD context embedded in every stored record.
+ *
+ * @since x.x.x
+ */
+class Record {
+	/**
+	 * Required top-level keys for a valid record.
+	 *
+	 * @var string[]
```
dkotter:
```suggestion
	 * @var list<string>
```

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
```diff
@@ -0,0 +1,623 @@
...
+			if ( null === $manifest ) {
+				$errors[] = array(
+					'stage'   => 'read_manifest',
+					'message' => esc_html__( 'The manifest could not be read.', 'ai' ),
+				);
+				return;
+			}
+
+			$writer = new Sidecar_Writer();
+			$rel    = $writer->write( $attachment_id, $manifest );
```
dkotter: If for some reason an image does have valid credentials but this write fails, the catch block fires and we lose the validation of the image. This may be expected but it will end up showing the image has not having valid credentials even though it did

dkotter (changes requested):
Thanks again for all the work here. I've put this through another round of review and testing and have found a few more things. In addition to what I've flagged, there's a couple other things:

1. If I delete an image, the sidecar isn't deleted. I think ideally we hook into `delete_attachment` and clean that up
2. We show the `Content Credentials` section even if the media item isn't an image, say an audio file or video file. Should we only show that section if it's an image?
3. The UI in the modal still isn't great, which I see @jeffpaul flagged as well. Seems the content is vertically aligned in the container instead of top aligned, maybe?

Image

Image

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Fixed in e18a1fc. The root cause was structural: the `meta_query` EXISTS + NOT EXISTS pattern with `relation => OR` emits an unconditional INNER JOIN for the EXISTS side, then reads `ORDER BY wp_postmeta.meta_value` from an arbitrary other meta row (e.g. a serialised `_wp_attachment_metadata` blob) for attachments that have no sort key — those strings sort above `'1'`, so credentialed items get pushed to the back.

Replaced with a `posts_clauses` filter that injects a named `LEFT JOIN ... ON ( post_id = ID AND meta_key = '_wpai_c2pa_present' )`. One joined row per attachment, no `DISTINCT`, and `COALESCE( meta_value + 0, -1 )` gives unscanned items a deterministic -1 so they sort last on DESC and first on ASC.

Also fixed the test that concealed the bug: the factory attachments had no other postmeta, so the ambiguous `ORDER BY` always hit the right value. The test now seeds `_wp_attached_file` and `_wp_attachment_metadata` on every fixture.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Added in e18a1fc. The `posts_clauses` handler now checks `get_current_screen()->base === 'upload'` and returns the clauses unchanged on any other screen.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Agreed — the styles were limited to `admin_head-upload.php` and `admin_head-post.php`, so the block editor, customizer, and any third-party admin page that embeds the media frame would get no CSS. Changed in e18a1fc to `admin_enqueue_scripts` with `wp_register_style( 'wpai-c2pa-monitor', false, array(), false )` + `wp_add_inline_style()`, which fires on every admin page.

includes/Admin/Uninstall.php
lnispel: Fixed in e18a1fc — the call is now `wp_upload_dir( null, false )` so the cleanup path never creates directories.

includes/Experiments/C2pa_Monitor/Record.php
lnispel: Applied in e18a1fc.

includes/Experiments/C2pa_Monitor/C2pa_Monitor.php
lnispel: Fixed in e18a1fc. `$c2pa` is now populated from the manifest facts (including `present: true`) *before* the sidecar write is attempted. If the write throws, the failure goes into `errors[]` with `stage: 'sidecar_write'` and the detection result survives. The outer `RuntimeException` catch (for scanner failures) is now labelled `stage: 'scan'` so the two paths are distinguishable in the stored record.

lnispel:
## Review round 3 follow-up — e18a1fc7

@dkotter — replied inline on all six threads. Summary of the review-body items:

**1. Sidecar not deleted when an image is deleted**

Added a delete_attachment hook that calls a new Sidecar_Writer::delete() method. It globs i-c2pa/<id>.*.c2pa — the literal . after the ID prevents 12.*.c2pa from matching 123.jpeg.c2pa. Postmeta is already removed by WordPress core.

**2. Content Credentials section shown for audio/video**

All three entry points — 
ender_media_column, dd_attachment_fields, and dd_attachment_meta_box — now bail early via wp_attachment_is_image(). Any image MIME type (including GIF, AVIF, etc.) shows the status; audio and video show nothing. The existing factory calls in the test file now create image/jpeg attachments so the assertions continue to pass.

**3. Modal vertical alignment still wrong**

The styles were hooked to dmin_head-upload.php and dmin_head-post.php only, so they were absent for the block editor, customizer, and any third-party page that opens the media frame. Moved to dmin_enqueue_scripts via wp_register_style + wp_add_inline_style, which fires everywhere. The tooltip and compat-field alignment CSS are identical — the alignment issue in the screenshot was caused by the styles simply not being loaded on that screen, not by a wrong rule.

---

_Replied via Cursor on behalf of @lnispel._

lnispel:
## Verified against a live install — 897e1949

I got a local wp-env instance running and reproduced all of this properly rather than reasoning about it, which turned up one more bug and one correction to what I said earlier.

**Reproduced the sorting failure.** Seeded four attachments carrying realistic postmeta (`_wp_attached_file`, `_wp_attachment_metadata`) — one with credentials, one scanned without, one never scanned, one audio file. The old `meta_query` approach produced:

`ddd-audio(none) | ccc-never-scanned(none) | aaa-has-credentials(1) | bbb-no-credentials(0)`

On a **descending** sort the unscanned items come *first* and the credentialed one lands third. That is exactly what you were seeing.

**Correction to my earlier reply.** I said `WP_Query` added `DISTINCT` for the `OR` relation. It did not — no `DISTINCT` appeared. The generated clause was:

`ORDER BY CAST(wp_postmeta.meta_value AS CHAR) DESC`

reading from the unconditioned `wp_postmeta` alias, so for attachments with no sort key it sorted on whatever other meta row it landed on (a path string, a serialised metadata blob), and those sort above `'1'` as `CHAR`. The mechanism I described was right, the `DISTINCT` detail was not.

**After the fix**, same data:

- `DESC` -> `aaa(1) | bbb(0) | ddd(none) | ccc(none)`
- `ASC` -> `ddd(none) | ccc(none) | bbb(0) | aaa(1)`
- 4 of 4 rows returned, no `DISTINCT`

**New bug found while verifying.** Applying the `posts_clauses` filter twice appended the `LEFT JOIN` twice, and MySQL rejects that outright:

`WordPress database error: [Not unique table/alias: 'wpai_c2pa_sort']`

which returns **zero rows** — an empty Media Library, worse than a mis-sort. The filter is registered once so this would not fire in normal operation, but it is one stray `apply_filters` away from a blank screen. Guarded against the duplicate alias in 897e1949 with a regression test.

**Also confirmed on the live install:**

- Image gating — attachment 7 (`audio/mpeg`) renders an empty column; attachment 4 (`image/jpeg`) renders the verify link.
- Sidecar cleanup — deleting attachment 5 removed `5.jpeg.c2pa` and left `55.jpeg.c2pa` untouched, so the ID prefix guard holds.

---

_Posted via Cursor on behalf of @lnispel._

dkotter:
> **3. Modal vertical alignment still wrong**

@lnispel I'm still seeing this problem on the latest commit. Checking the CSS myself, seems the problem is this style: `.compat-field-wpai_c2pa th.label{width:33%}`.

If I change that 33% to 30%, seems to fix the issue for me everywhere I tested. Any particular reason you set that to 33%?

lnispel:
## Modal alignment - fixed in dc4ad17e

@dkotter your 30% is right, and there was no good reason for the 33% - it was a guess, and the comment sitting above it explained the layout wrongly, which is how it got there.

**Why it breaks.** Core does not lay these rows out as table cells. `media-views.css` sets `.compat-item tr { display: block }` and then floats the two halves:

```css
.compat-item .label { min-width: 30%; margin-right: 4%; float: left; }
.compat-item .field { float: right; width: 65%; margin: 1px; }
```

So the budget is `30 + 4 + 65 = 99%`. A 33% label makes it `102%`, the right-floated value can no longer fit beside the label, and it drops onto its own row. That is the misalignment - the value was never vertically off-centre, it was on the next line.

**Measured it** rather than eyeballing, loading core's actual `media-views.css` in headless Chromium with the markup `get_compat_media_markup()` emits, at the 267px sidebar width, and reading the label/value bounding boxes:

| label width | label box | value offset | same line? |
| --- | --- | --- | --- |
| no override | 115px | 18px lower | no |
| `33%` | 88px | 36px lower | no |
| `30%` | 80px | 1px lower | **yes** |

Worth noting the "no override" row: dropping the width entirely does not work either, because the label is a shrink-to-fit float and "Content Credentials" is long enough to take 43% on its own. `min-width: 30%` gives it a floor, not a ceiling, so the width does have to be pinned - just at 30%, matching what core budgeted for.

**Also removed two rules that never did anything:**

```css
.compat-field-wpai_c2pa th.label,
.compat-field-wpai_c2pa td.field { display: table-cell; vertical-align: top }
```

Since core floats both elements, `display` is blockified to `block` per CSS 2.1 §9.7, so `table-cell` and `vertical-align` were dead. The measurement above is byte-identical with and without them. They were the source of the bad "wraps compat fields in a table row" comment, so they are gone and the comment now records the 30/4/65 float budget to stop the number drifting again.

CI is green on the branch, including E2E.

---

_Posted via Cursor on behalf of @lnispel._

lnispel:
Small correction to my last comment, and a heads-up on the red CI.

**The branch had drifted into conflict with `develop`**, which is why no checks ran on the CSS commit at all - GitHub cannot build the merge ref for a conflicted PR, so the `pull_request` workflows were never created. Merged `develop` in (78e8dca1). Three conflicts, all additive:

- `Uninstall.php` - kept both the new `Embedding_Schema` import and the C2PA ones, and both `drop_embeddings_table()` and `delete_c2pa_sidecars()` in the teardown order.
- `README.md` / `readme.txt` - kept the C2PA bullet and took `develop`'s updated Comment Moderation wording.

**The four red jobs are not this PR.** `Test PHP 7.4` and `Test PHP 8.0` fail during `wp-env start`, while the Debian bullseye image installs `sudo`:

```
E: Failed to fetch .../sudo_1.9.5p2-3+deb11u4_amd64.deb  404  Not Found
target wordpress: failed to solve: process "/bin/sh -c apt-get -qy install sudo" did not complete successfully
```

The pinned version was pulled from the security archive, so it depends on which CDN node the runner hits - the log shows different IPs across the three retries. It is hitting everything right now, not just this branch: the run on `develop` itself failed `Test PHP 8.0 WP latest`, and [#feat/ai-workspace](https://github.com/WordPress/ai/actions/workflows/test.yml) lost three of the same six jobs in its latest run.

The other 22 jobs on this branch pass, including E2E, Plugin Check, static analysis and the full PHP 8.1-8.4 matrix. A re-run should clear the four - I do not have permission to trigger one here.

---

_Posted via Cursor on behalf of @lnispel._

lnispel:
> > **3. Modal vertical alignment still wrong**
> 
> @lnispel I'm still seeing this problem on the latest commit. Checking the CSS myself, seems the problem is this style: `.compat-field-wpai_c2pa th.label{width:33%}`.
> 
> If I change that 33% to 30%, seems to fix the issue for me everywhere I tested. Any particular reason you set that to 33%?

Thank you for your patience on this work. I have still been largely going through a personal transition that is preventing me from being more hands on with this code but I appreciate the great reviews. It makes it much easier to address the feedback.

lnispel:
Howdy, just checking in on this again to see how yall are feeling

.github/workflows/test.yml
```diff
@@ -238,6 +238,32 @@ jobs:
       - name: Build assets
         run: npm run build
 
+      - name: Patch wp-env Docker config to redirect bullseye to archive.debian.org
```
dkotter, line 241: This should no longer be needed as this was fixed upstream. I think all the changes in this file can just be reverted

dkotter (changes requested):

.github/workflows/test.yml
lnispel: Good catch, thanks. Reverted all workflow additions (
etry_wait_seconds, the patch-node-modules step) in 8038dcce and merged develop (b9a1984a), which picks up @wordpress/env 11.15.0. All six PHP 7.4/8.0 jobs are now passing with no workarounds.

dkotter (approved):

lnispel:
Thank you @dkotter! 🙏🏼 
How do I go about merging this in?

dkotter:
> Thank you @dkotter! 🙏🏼 How do I go about merging this in?

Nothing needed from you there, hoping to merge this in prior to the next release, just waiting for a final review from @jeffpaul to ensure he's fine with where things have landed

PR #467, opened 2026-04-23
Connector Approval experiment to limit what plugins can access what connectors

### LINKED ISSUE #342: Add permission controls for plugins to use a connected provider
It's currently difficult to control access to connected AI providers. `WP_AI_SUPPORTED` and related filters give some options for programmatic control, but the level of connection provided by centralizing access through the AI client methods isn't matched by the level of control that's enabled. 

(Complements #149 which adds the observability layer)

Exposing controls that allow site admins to allow/revoke or route access to connected AI providers. There are a few layers to this:
### 1. Master off switch
Ability to prevent all access to AI providers. Needed for cases where strict privacy/compliance concerns require hardline prevention of AI functionality.
 - [x] Already enabled via `WP_AI_SUPPORTED`

### 2. Plugin-level permissions
Ability to grant/revoke access to AI providers for specific plugins to use for their own functionality. This allows users to control costs and access on a more granular level than the all-or-nothing approach of 1. Preferably permissions are granted on an opt-in basis so plugins have to be granted explicit access to use connected providers (similar mental model to OAuth scopes).

### 3. Functionality-level permissions
Ability to grant/revoke access to connected providers for specific AI functionality. Grants even more granular control than 2, which can break down along prompt-type-lines (e.g. allow text generation while preventing image gen), or even for particular features (plugins with multiple AI features must request access per feature). This gives users fine-grain control over AI access/spend on the site, especially beneficial for sites with multiple roles that have wp-admin access (e.g. Admin might want to only allow Editors to use specific features, or types of features).

### 4. Provider routing
Ability to route usage to specific providers on plugin/functional basis. Can be used as another cost control, or to improve performance of features that are better suited for one model vs. another. This layers on top of the "on/off" permissions settings to gives users control of "how" their AI features work, not just "if" they work or not.

### LINKED ISSUE #441: Require explicit admin approval for plugin access to Connectors + improve connector secret protection

_This is a concept I think is worth testing out here in the AI plugin and then perhaps migrating to Gutenberg for wider testing ahead of potentially merging this concept into WP core during the 7.1 release cycle to provide better, tighter controls over usage of connections on the Connectors screen._

Currently, once a connector is configured (e.g., Ollama, OpenAI, etc.), any plugin with access to the relevant APIs may be able to use that connector without explicit admin approval.

This creates two related risks:

1. Unauthorized connector usage: A plugin could leverage an existing connector without the an admin intentionally granting permission (e.g., a forms plugin automatically using OpenAI for plugin functionality).

2. Credential exposure risk: If connector credentials are stored in ways accessible to PHP code, a malicious or poorly designed plugin could read API keys or tokens and use them for unintended purposes (including off-site usage).

Together, these gaps reduce admin control and increase security risk so let's look to add a permission layer to the Connectors system that requires an admin to explicitly approve which plugins can use a configured connector.  In parallel, let's evaluate improvements to how connector credentials (API keys, tokens, secrets) are stored and protected to reduce risk of unauthorized access by other plugins.

### Plugin Access Approval

Introduce an approval workflow that requires explicit admin consent before a plugin can use a connector.  Plugins must request access to a specific connector.  An admin UI must show pending connector access requests, allow approve / deny per plugin, and display currently approved plugins per connector.  Connector usage should be blocked unless plugin access is approved and provide clear error messaging when access is denied.

### Connector Secret Protection

Investigate stronger protections for connector credentials.  Strengthen how connector credentials (API keys, tokens, and secrets) are stored and accessed to reduce the risk of unauthorized use by other plugins.  This could include encrypting credentials at rest and limiting exposure to only trusted connector components.

dkotter:
## What?

Closes #441, #342

Adds a new settings screen that allows administrators to determine which plugin and theme can access which AI Connectors

## Why?

WordPress 7.0 is releasing a new Connectors API that allows users to install various Connector plugins and then configure those once.

There is no approval / permission system in place though, meaning once a Connector is set up, any plugin can then use it even if you don't want that. This PR adds in a new experiment to showcase one approach to adding an approval system.

Worth noting this is (on purpose) a granular approval system. An administrator can decide which plugin has access to which connector, it's not an all or nothing approach.

## How?

- Adds a new `Connector Approval` experiment
- When turned on, a new settings page is added (`Settings > AI Connector Approvals`). This settings page will show any pending requests and will also show an approval matrix for each active plugin and theme, allowing you to toggle support for each connector
- Auto detection when a non-approved plugin or theme tries to use an AI Connector. This will then show as a pending request on the settings page and an admin notice will also render
- Hook into `pre_http_request` and determine which plugin/theme is making the request and which provider they are using. If that plugin/theme has not been approved for that provider, stop the request
- Adds necessary tests and documentation

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code, Cursor
Model(s): Opus 4.7, GPT-5.3 Codex
Used for: Used Claude Code for initial review and planning. Tried out a number of approaches before finally landing on the current approach. Testing and refinement done by me. Used Cursor to add tests and documentation.

## Testing Instructions

2. Ensure you have at least one AI Connector in place
3. Turn on the Connector Approval Experiment in `Settings > AI`
4. Go to the Dashboard you should see an admin notice about a plugin requesting access
5. Go to `Settings > AI Connector Approvals` and you should see a pending request for the AI plugin (yes, we even block our own plugin here)
6. Either approve that request or toggle the connector settings for the AI plugin in the Approval matrix below
7. Ensure the admin notice goes away and any AI experiments you have turned on work
8. Remove the approval and test those experiments again and ensure they don't work

## Screenshots or screencast

Connector Approval experiment settings

Admin notice showing that a plugin has been blocked from using a connector

AI Connector Approvals settings page showing pending approvals

AI Connector Approvals settings page showing the Approval matrix

## Changelog Entry

> Added - Connector Approval experiment, allowing administrators the ability to determine which plugins can access which connectors.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-467-b145aa049b37eda1fb16e4871a4bf562799e14d6.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


justlevine:
(I didn't have a chance to pull down and test but wanted to say that this looks super cool and exciting. Does the matrix scale for more than 3 connectors? )

dkotter:
> Does the matrix scale for more than 3 connectors?

Yep, uses the new `wp_get_connectors` function to return all Connectors that are registered as an `ai_provider` type so this matrix will show all AI Connectors.

That said, does bring up two points to consider:

1. Should we worry about a state where someone has a dozen or more Connectors? There's probably a limit where the design starts to fall apart. I'm not overly worried about this right now as it seems like an edge case but something to consider
2. Right now `wp_get_connectors` will always return the three "core" connectors (Anthropic, Google, OpenAI) even if the actual plugins for those don't exist. For instance, in the screenshot in this PR, I have the OpenAI Provider plugin installed but not anything else and you'll see there's toggles for Anthropic and Google. Could argue this is a potential upstream bug that needs looking at, where those default connectors should only be registered if the underlying plugin exists. But we may want to look at handling that within this PR, removing any connector that is registered but isn't actually available due to the plugin not being installed

justlevine:
> 

> That said, does bring up two points to consider:
> 
> 1. Should we worry about a state where someone has a dozen or more Connectors? There's probably a limit where the design starts to fall apart. I'm not overly worried about this right now as it seems like an edge case but something to consider

While I do think it's something we don't need to solve in the first try and can iterate towards, I think it's a lot less of an edge case than it seems. The amount of nondevelopers in the *Claw threads or various harness repos asking for things like complex model fallback chains is surprising, and once folks wake up and realize non-deterministic code is still code, we'll have a lot more than just plugin devs for enterprise testing against multiple models.

> 2. Right now `wp_get_connectors` will always return the three "core" connectors (Anthropic, Google, OpenAI) even if the actual plugins for those don't exist. For instance, in the screenshot in this PR, I have the OpenAI Provider plugin installed but not anything else and you'll see there's toggles for Anthropic and Google. Could argue this is a potential upstream bug that needs looking at, where those default connectors should only be registered if the underlying plugin exists. But we may want to look at handling that within this PR, removing any connector that is registered but isn't actually available due to the plugin not being installed.

AFK so hard for me to check the latest iteration of Connectors, but imo its an upstream bug unless/until `wp_get_connectors( ..., bool $include_recommended = false )` or something is implemented. 

For us, Id consider consider it something that should be addressed as part of this PR. if core doesn't provide a way to filter connectors by `active` - IMO not just `installed` , since definitionally as an `Approval Matrix` anything that isn't available to the system can't be approved _or_ blocked - then we gotta do it ourselves.

itsgajendraSingh:
Hey @dkotter  👋 
I tested this PR on WordPress Playground and wanted to share a couple of suggestions from a UX and behavior perspective.

---

### 1. Sidebar menu not updating immediately after enabling AI

When AI is enabled from Settings → AI, the “AI Connector Approvals” menu item doesn’t appear right away in the sidebar. It only becomes visible after navigating to another tab or changing the view.

**Steps to reproduce:**

- Go to Settings → AI
- Enable AI
- Check the sidebar → “AI Connector Approvals” is not visible
- Navigate to another tab (e.g., Writing/Reading)
- The menu item then appears

**Expected behavior:**
The sidebar should update immediately after enabling AI, without requiring any navigation or refresh.

**Current behavior:**
The sidebar doesn’t seem to react to the state change until a re-render is triggered.

---

### 2. Approval toggles are active even when providers are not connected

Screenshot 2026-05-01 at 2 58 42 PM

In the Approval Matrix, the toggles for providers (Anthropic, Google, Hugging Face, OpenAI) are interactive even when the respective provider is not connected/configured.

**Suggestion:**

- Disable toggles when the provider is not connected
- Optionally add a tooltip like “Provider not connected”

jeffpaul:
As discussed in today's AI contributor call, we'll explore adding in something like https://github.com/ericmann/displace-secrets-manager to help ensure Connector details are kept secret.

dkotter:
> As discussed in today's AI contributor call, we'll explore adding in something like https://github.com/ericmann/displace-secrets-manager to help ensure Connector details are kept secret.

In diving into this more today, this secrets manager is interesting but I don't think fully solves the problem we were discussing. In essence, it provides a standardized way to encrypt and decrypt data in WordPress, allowing us to encrypt API keys before saving and decrypt them when we need to make a request.

I think that's sorely needed in WordPress (and there is already a [Trac ticket](https://core.trac.wordpress.org/ticket/64789) open around this for the Connectors API) but still doesn't solve the issue of preventing someone from accessing an API key a site owner doesn't want them to have.

That said, if we were to add this alongside what's already in this PR, that does get us closer to a "secure" state. But there are still edge cases, like a plugin directly pulling the API key out of the database or just using our built-in encryption/decryption methods that we would need to add. I'm not sure there's a way for us in a plugin to 100% cover all bases.

My suggestion here would be to proceed with this PR (and any feedback given) but then open a separate PR to look at integrating encryption (either with that Secrets Manager or a custom approach). I don't think these two need to be shipped within the same PR and I'm cognizant at the large size of this PR already.

swissspidy:
> Auto detection when a non-approved plugin or theme tries to use an AI Connector. This will then show as a pending request on the settings page and an admin notice will also render

How does this work? Plugins and themes by nature have complete unrestricted access to WordPress, they could just circumvent this.

dkotter:
> How does this work? Plugins and themes by nature have complete unrestricted access to WordPress, they could just circumvent this.

Correct, only so much we can do here from within another plugin. At a high-level, we hook into `pre_http_request` and if the request is for an AI Connector, we try and determine who is making the request and if they have been given permission. If not, we block that request.

But yes, nothing stopping another plugin from hooking in earlier, removing our hook or making a direct curl request and not using the WP HTTP API at all.

This isn't meant to be 100% effective (or I should say I don't think it's possible to be 100% effective), it's meant more as a PoC that can hopefully stir conversation about getting something more permanent in WordPress Core itself.

haktan-suren:
I built a small plugin exploring plugin-level AI Client access control (probably that's why I was tagged here too), so I wanted to share a few implementation tradeoffs from that work https://wordpress.org/plugins/handl-ai-connector-access-control/#is%20attribution%20perfect%3F.

My approach used `wp_ai_client_prevent_prompt`, which is a cleaner interception point because it avoids scanning outbound requests for credentials. The limitation I ran into is that it does not expose enough connector/provider context for granular per-connector approval. This PR’s HTTP-layer approach solves that, but it also depends on stack traces (which is unfortunate!) plus credential/base-URL matching.

Longer term, I think the ideal API would be a higher-level AI Client/Connector hook that passes the resolved connector/provider, caller identity, and request capability before execution. That would avoid both SDK reflection and credential scanning.

A few smaller points:
- Should unknown callers be configurable as allow/block/log? Currently unidentified callers are allowed through.
- It may be useful to add optional audit logging for approved usage, not just pending blocked attempts.
- Since caller identification supports themes, the approval matrix may also need active theme entries so admins can preapprove themes before they generate a pending request.

dkotter:
> My approach used wp_ai_client_prevent_prompt, which is a cleaner interception point because it avoids scanning outbound requests for credentials. The limitation I ran into is that it does not expose enough connector/provider context for granular per-connector approval.

Yeah, my initial exploration used `wp_ai_client_prevent_prompt` but that does not allow the level of granularity I wanted. It's either an all or nothing approach.

> Longer term, I think the ideal API would be a higher-level AI Client/Connector hook that passes the resolved connector/provider, caller identity, and request capability before execution

Yep, I agree and added [those thoughts](https://github.com/WordPress/ai/issues/441#issuecomment-4307862700) on the original issue.

> Should unknown callers be configurable as allow/block/log? Currently unidentified callers are allowed through.

Good question and this is where I'd love some more testing. I didn't run into any unidentified callers in my testing but if that's a common occurrence, likely need to decide if those should be allowed through or not. I'd lean towards allowing them in as part of this exploration work as this is meant as an experiment but open to opinions on that.

> It may be useful to add optional audit logging for approved usage, not just pending blocked attempts.

We have #437 open right now that adds logging. That PR and this PR will work nicely together

> Since caller identification supports themes, the approval matrix may also need active theme entries so admins can preapprove themes before they generate a pending request.

I debated whether I should look at themes but I know some themes add in quite a bit of functionality so likely there will be some that try and make AI requests. That said, I did miss having those show in the approval matrix by default so that's likely a good thing to consider adding.

swissspidy:
> Yeah, my initial exploration used `wp_ai_client_prevent_prompt` but that does not allow the level of granularity I wanted. It's either an all or nothing approach.

We might still have time to get something into 7.0 to make the filter more useful, if anyone wants to open a ticket :)

jeffpaul:
1. Perhaps move the new admin page to under `Tools` and rename to `Connector Approvals` on the page header and in the admin menu (this more closely matches the Experiment name).
2. Can probably hide the alerts on the Connector Approvals page as you'll already see the list of pending requests there and can review / approve / dismiss and the admin notice is mostly duplicative there
3. Could be nice to have a list of the dismissed requests (in case someone mistakenly dismissed, wants to review all dismissed items), but not necessary for this iteration
4. Will be curious how the Approval matrix table scales to sites with a LOT of plugins and themes active, but maybe we leave any special handling there for the next iteration here
5. How is this determining an "AI" connector from a "non-AI" connector (e.g. how does it know Mistral is one to show but Akismet or Jetpack are not?  Could be with the changes in (1) above that making this be for any/all Connector plugins could be helpful and not just AI ones?

dkotter:
> * Perhaps move the new admin page to under `Tools` and rename to `Connector Approvals` on the page header and in the admin menu (this more closely matches the Experiment name).

Done in https://github.com/WordPress/ai/pull/467/commits/1d86430b4c9c5d649fc3ec7898a46cdd7b8f6f1c

> Can probably hide the alerts on the Connector Approvals page as you'll already see the list of pending requests there and can review / approve / dismiss and the admin notice is mostly duplicative there

Good call, removed that from showing on that page in https://github.com/WordPress/ai/pull/467/changes/26cd034268af174ae1e566b4f03e3bb082836b4f

> Could be nice to have a list of the dismissed requests (in case someone mistakenly dismissed, wants to review all dismissed items), but not necessary for this iteration

Yeah, not something we store right now. Dismissing just removes it from the current view but if the plugin tries making a request again, it will show up again (so dismissing doesn't remove it forever)

> Will be curious how the Approval matrix table scales to sites with a LOT of plugins and themes active, but maybe we leave any special handling there for the next iteration here

Definitely open to design ideas here. I'm not as concerned with the list of plugins/themes as this table can just keep expanding down forever (though maybe there's a better approach). More worried if sites have lots of connectors, there's only so many columns we can show before you have to start scrolling (I would assume this is an edge case and most sites will only have one or two AI connectors)

> How is this determining an "AI" connector from a "non-AI" connector (e.g. how does it know Mistral is one to show but Akismet or Jetpack are not? Could be with the changes in (1) above that making this be for any/all Connector plugins could be helpful and not just AI ones?

Right now we use the core `wp_get_connectors` function and then filter down to only showing connectors with the `ai_provider` type, so no other connectors will show here.

If we want this to work for _any_ connector, that's doable though likely better as a follow up as there's probably a few other changes needed.

jeffpaul (approved):

t-hamano:
When the "Connector Approval" experiment setting is enabled, the API key is considered invalid even when the correct key is entered. I didn't understand what was happening until I reached this PR. Is it intended behavior that the API key input itself is put on hold?

dkotter:
> When the "Connector Approval" experiment setting is enabled, the API key is considered invalid even when the correct key is entered. I didn't understand what was happening until I reached this PR. Is it intended behavior that the API key input itself is put on hold?

@t-hamano Do you happen to have any other experiments turned on? I could only reproduce this if I had the Request Logging experiment turned on alongside Connector Approval. Otherwise it allows the API key to be validated and saved just fine. Definitely a bug we'll want to fix but curious if that's what you're running into or if you're seeing this with only having the Connector Approval experiment on?

t-hamano:
> I could only reproduce this if I had the Request Logging experiment turned on alongside Connector Approval.

That's correct. I was able to reproduce the issue in my environment with those settings as well.

image

dkotter:
@t-hamano Thanks for verifying! I have #595 up to fix this and a few other bugs I found while testing. If you have a second to test and verify it fixes your problem, that would be great.

PR #455, opened 2026-04-21
 Compress AI settings page above-the-fold

Adi-ty:
Move the global AI toggle into the page header and replace the large general settings block with an infotip so Features and Experiments appear higher on the page.

## What?

Closes #451

Moves the global “Enable AI” control out of the large General Settings block and into the page header. Replaces the explanatory block with an infotip so the Features and Experiments sections are visible higher on the page.

## Why?
The previous layout pushed the main feature sections too far below the fold because the General Settings block consumed a large amount of vertical space. This change keeps the same behavior while improving scanability and reducing the amount of unused space at the top of the settings page.

## How?
- Renders the global AI toggle in the page header instead of inside the main DataForm settings section.
- Uses an `@wordpress/ui` Popover-based infotip to display the explanatory copy.
- Keeps the same settings key and save flow through core-data and the WP settings API.

### Use of AI Tools

GPT-5.3-Codex was used to assist with code review and PR description drafting.

## Testing Instructions
1. Open the AI settings page in wp-admin.
2. Verify the global AI toggle is shown in the page header.
3. Verify the Features and Experiments sections are visible higher on the page.
4. Hover or focus the infotip icon and confirm the explanatory text appears.
5. Toggle AI on and off and confirm the setting persists after refresh.

## Screenshots or screencast

| Before | After |
|--------|-------|
|  |  |

## Changelog Entry

> Changed - Compress the AI settings page above the fold by moving the global AI toggle into the header and replacing the large general settings block with an infotip.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-455-a40f8cd316d1bafb402c196ad3ba0fd1e38762f9.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #472, opened 2026-04-24
Update settings page to use @wordpress/ui components

simison:
## What?

Updates settings page (`routes/ai-home/*`) to use the latest `@wordpress/ui` components as replacements for any custom CSS and `@wordpress/components` where available. Spinner and ToggleControl are still pulled from the old components, and there's a minimal amount of CSS to match the previous page structure more closely.

No functional changes, but there are some minor visual changes:

- Some spacing and font sizes here and there are slightly different, but we now follow the component system without local overrides, which often come back as bugs as upstream components change; we can't really rely on their CSS selector structures as "public API".
- Spinner loader is positioned more centrally on the page.

## Why?

Makes things more consistent with all the new UIs in WordPress, and moves away from deprecated components. Less custom CSS to maintain.

## Testing Instructions

- Visit the settings page at `/wp-admin/options-general.php?page=ai-wp-admin` and confirm it works and looks good. You can test with and without Gutenberg enabled.
- Test different window widths and different browsers (I've tested Firefox & Chrome), desktop and mobile
- Test toggling different features on/off.

### Before screenshots

image
image
image

### After
image

image

image

## Changelog Entry

Changed - Updated components for the settings page.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-472-846fe986cb43fd096e51a752e9f1a63c828bb356.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jameskoster:
Nice. A couple of design considerations outside of the UI package adoption:

* Dimming the cards when AI is disabled doesn’t feel like a strong pattern—the connection isn’t very clear. It also raises the question of whether the “Enable AI” toggle is necessary at all. As it stands, a user could enable AI and then switch off all related features, which creates a misleading state. That starts to feel like a UX smell. Unless “Enable AI” has a meaningful standalone effect—beyond simply unlocking other options—it might be better to remove it.
* Whats the reason for using a Button to enable/disable Image Generation and Editing rather than a Toggle like all the other options? The inconsistency was a bit unexpected.
* I don't know that the "Enable all" / "Disable all" actions add too much value—shouldn't use be evaluating each experiment individually?
  * They are especially redundant in the "Admin experiments" card since it only contains a single option.
* Do the cards need to be collapsible? Unsure how useful this is.

Totally fine to address these separately, or not at all.

package.json
```diff
@@ -64,7 +64,7 @@
 	},
 	"dependencies": {
 		"@wordpress/a11y": "^4.44.0",
-		"@wordpress/admin-ui": "^1.12.0",
+		"@wordpress/admin-ui": "~1.12.0",
```
justlevine: What's the benefit of locking this to patch?
We bundle our lockfile, GB isn't particularly mindful of semver, and IRL it's the versions bundled with WordPress getting served.

simison: Yep, agreed, it's just my `npm` config doing things on its own, I'll put it back to exact.

I temporarily used `@next` version of packages to check if we had [new](https://github.com/WordPress/gutenberg/pull/76469) `visual` prop yet available for the header icon, but nope.

Edit: done in https://github.com/WordPress/ai/pull/472/commits/42eadd1417ac1664e7b329cac2d6169e1508b2e2

simison:
> Dimming the cards when AI is disabled doesn’t feel like a strong pattern—the connection isn’t very clear. It also raises the question of whether the “Enable AI” toggle is necessary at all. As it stands, a user could enable AI and then switch off all related features, which creates a misleading state. That starts to feel like a UX smell. Unless “Enable AI” has a meaningful standalone effect—beyond simply unlocking other options—it might be better to remove it.

Agreed — it's one aspect I left untouched in this PR and also decided not to pull `Disabled` from `@wordpress/components` to improve UX.

@jameskoster  Agreeing with your other points, too, I think addressing them would tighten the UI. Happy to follow up as I'm trying to keep functional changes out of this PR.

simison:
@jameskoster BTW I noticed Card doesn’t allow “full bleed image as first element”, which I’m sure is a pretty common need.

I used a [one-liner](https://github.com/simison/ai/blob/42eadd1417ac1664e7b329cac2d6169e1508b2e2/routes/ai-home/style.scss#L25-L27) to make it work and couldn’t use `Card.Header` sadly, but I just didn’t want to diverge too much from the existing design in this PR, but happy to implement some good ideas that follow componentry more closely. :-)

The image itself feels a bit restless and could use better cropping, edge-blurring/shadows, or similar effects.

aduth:
> use the latest `@wordpress/ui` components

It's an interesting question whether we should be using _all of them_ as-is. A lot of components are "not ready" purely out of concern for consistency with existing components. Notices being a good example: The design is quite different, and this needs to be reconciled (i.e. `@wordpress/components` needs to be updated) so that they don't look strange placed next to each other. But considering that this is an experimental package, and a relatively self-contained experience, maybe that's okay?

If we wanted to follow Gutenberg, I'd suggest this project should adopt `@wordpress/eslint-plugin` and use [`wordpress/use-recommended-components`](https://github.com/WordPress/gutenberg/blob/HEAD/packages/eslint-plugin/docs/rules/use-recommended-components.md), which is the canonical source of what components should be used. This experience is also enabled through [the new design system MCP server](https://github.com/WordPress/gutenberg/tree/trunk/packages/design-system-mcp), which uses the same source of truth in understanding which components should be recommended for use.

routes/ai-home/stage.tsx
```diff
@@ -888,50 +896,65 @@ function AISettingsPage() {
 	return (
 		<Page
 			title={
-				<>
+				<Stack align="center" gap="xs">
 					<AIIcon />
 					{ __( 'AI', 'ai' ) }
-				</>
+				</Stack>
```
simison: In future we can use the [new](https://github.com/WordPress/gutenberg/pull/76469) `visual` prop instead of Stack.

dkotter:
> It also raises the question of whether the “Enable AI” toggle is necessary at all. As it stands, a user could enable AI and then switch off all related features, which creates a misleading state. That starts to feel like a UX smell. Unless “Enable AI” has a meaningful standalone effect—beyond simply unlocking other options—it might be better to remove it

This is purposely there to allow users an easy, one-click experience to entirely disable everything. Can definitely talk about alternatives here if there's concern with how it's currently done but ideally that should happen in a separate PR as that is a significant deviation from what we currently have. Also worth noting we're exploring moving that option to the header, see #455

> Whats the reason for using a Button to enable/disable Image Generation and Editing rather than a Toggle like all the other options? The inconsistency was a bit unexpected.

Anything in that top section (right now that's just Image Gen) is considered a "stable" feature. Whereas everything below is an "experimental" feature. We want a distinct UI between the two to try and make it more obvious. My assumption is the use of a Button here instead of a Toggle was to make things stand-out a bit more but I'm open to changing that.

> I don't know that the "Enable all" / "Disable all" actions add too much value—shouldn't use be evaluating each experiment individually?
> They are especially redundant in the "Admin experiments" card since it only contains a single option.

This was added specifically due to a few user requests that wanted an easy way to enable/disable everything in a section. Could re-evaluate that though if there's differing thoughts.

Note there's likely to be additional "Admin experiments" in the future which is why those buttons exist there (though they could be removed and added back once that is true).

> Do the cards need to be collapsible? Unsure how useful this is.

To my point above about the differences between "stable" and "experimental" features, originally we were going to have "experimental" features in a collapsed state by default. We haven't done that yet but did use the collapsible UI to support that

routes/ai-home/stage.tsx
```diff
@@ -888,50 +896,66 @@ function AISettingsPage() {
...
 			}
+			ariaLabel={ __( 'AI', 'ai' ) }
```
simison: Otherwise, `aria-label` is empty since we don't have a string as `title`.

Before

Image

After

Image

jeffpaul:
As @dkotter flagged, there are changes pending in #455 that I'd want to have made ahead of merging this here (as that further compresses some rather verbose portions of the settings page)

simison:
@jeffpaul yep no prob, I'll wait for that to merge first.

jameskoster:
> This is purposely there to allow users an easy, one-click experience to entirely disable everything

That’s a fair use case, but a checkbox/switch doesn’t feel like the right control here; it isn’t actually changing a setting by itself, just toggling whether other settings can be toggled. If the goal is simply to quickly disable everything, a dedicated action (button) might be more appropriate, shown only when at least one AI feature is enabled.

This kind of overlaps with the "Disable all" buttons too. A single button would simplify the UI quite a bit.

> Anything in that top section (right now that's just Image Gen) is considered a "stable" feature. Whereas everything below is an "experimental" feature. _We want a distinct UI between the two to try and make it more obvious_.

I worry the current pattern is a bit ambiguous. I'm not sure the average user would grok the distinction. It might be simpler (and more accessible) to clarify the difference in the copy.

routes/ai-home/stage.tsx
```diff
@@ -2,18 +2,13 @@
...
 import { Page } from '@wordpress/admin-ui';
-import {
-	Button,
-	ExternalLink,
-	Notice,
-	Spinner,
-	ToggleControl,
-} from '@wordpress/components';
+import { Spinner, ToggleControl } from '@wordpress/components';
+import { Badge, Button, Card, Link, Notice, Stack, Text } from '@wordpress/ui';
```
mirka: Just a heads up here, at least in Gutenberg itself, we currently gate the usage of `@wordpress/ui` components to those we have explicitly allowlisted through the `@wordpress/use-recommended-components` eslint rule. This is because there are still some design consistency and popover compatibility [issues](https://github.com/WordPress/gutenberg/issues/76135) with legacy components that should be resolved before safely starting widespread use.

It should be fine if you're _only_ using `@wordpress/ui` components, but if you're going to use a mix with `@wordpress/components` and `@wordpress/dataviews`, there may be places where you run into design consistency or popover stacking issues. Your call as to whether you're ok with being vigilant about these potential issues, or if you want to play it safe.

dkotter:
> it isn’t actually changing a setting by itself, just toggling whether other settings can be toggled

Note this isn't 100% accurate as toggling this does save a `wpai_features_enabled` setting which we check in a few places. So yes, if this is toggled off you can't change the other settings but when this is toggled, we do update the value of `wpai_features_enabled`

simison:
Spin off the logo changes to a separate PR:

- https://github.com/WordPress/ai/pull/488

routes/ai-home/stage.tsx
simison: Won't be needed as of https://github.com/WordPress/ai/pull/488

simison:
A couple more spin-off PRs:
- https://github.com/WordPress/ai/pull/490
- https://github.com/WordPress/ai/pull/491

routes/ai-home/style.scss
```diff
@@ -8,127 +8,35 @@
...
-	width: 100%;
-	aspect-ratio: 16 / 9;
-	background: #f0f0f0;
-	display: flex;
-	align-items: center;
-	justify-content: center;
-	overflow: hidden;
-
+	margin-bottom: var(--wp-ui-card-padding);
+	margin-top: calc(-1 * var(--wp-ui-card-padding)); // Negative to bump against top border.
```
simison: Implementing hero images in Cards in the design system here:
- https://github.com/WordPress/gutenberg/pull/77856

simison:
@dkotter @jeffpaul aight, [header toggle](https://github.com/WordPress/ai/pull/455) is now merged, and I moved some of the changes from this PR to their own PRs. 

Rebased, and this one is good for a test/review again.

dkotter (approved):

dkotter (dismissed):

simison:
Thanks, @dkotter. I rebased again; feel free to merge, as I don't have access myself.

jeffpaul:
@mirka any other tweaks you'd suggest here before merging this in?

simison:
FYI, updated Card+image to follow the undocumented approach I just [learned elsewhere](https://github.com/WordPress/gutenberg/pull/77856#pullrequestreview-4222838895); it doesn't require CSS offsets. Might still need to follow up on updating the Card structure as the component improves, but this seems fine for now.

routes/ai-home/stage.tsx
```diff
@@ -586,51 +595,50 @@ function VisualCardToggle( {
...
+						>
+							{ checked
+								? __( 'Disable', 'ai' )
+								: __( 'Enable', 'ai' ) }
+						</Button>
+						{ checked && (
+							<Badge intent="stable">
+								{ __( 'Enabled', 'ai' ) }
+							</Badge>
+						) }
```
mirka: This is a rather strange UI pattern, and I noticed that Jay flagged it as well.

A concrete case against it would be for assistive technologies — it's generally not good to change button labels dynamically, and it's also semantically unclear what the checked state of the button is since the badge is shown separately unconnected to the button. But aside from that, I'd say it's also unconventional enough that sighted users may be confused by the button + badge combo too. I think a normal toggle would be the safer choice here.

simison: As far as design goes +1, tho I simply kept it as designed so far and simply switched to use components. :-)

Might be a good follow-up to offer using toggle?

simison: Moved to another more focused PR:

- https://github.com/WordPress/ai/pull/505

cc @mirka @jameskoster

simison: Merged above, and UX problems with the component usage solved. :-)

simison:
Rebased with https://github.com/WordPress/ai/pull/505 merged, and updated PR description screenshots. Good for a review. Now this is pretty good with all the harder/distracting problems that came up merged in their own PRs. :-)

jeffpaul (approved):
Works well, thanks!!!

simison:
Thanks! @jeffpaul, mind hitting merge for me? :-) I don't have access.

jeffpaul:
Waiting on @dkotter final code review, will let him mash that big green button when he's ready

dkotter (approved):

PR #479, opened 2026-04-27
Hide title generation until content exists

### LINKED ISSUE #391: Regenerate option is being shown even though there’s no content yet.
### Description

ThisThis appears to be a UX issue: the “Regenerate” option is visible even when no content has been added. The button should be hidden or disabled until there is existing content to regenerate.

1. Go to the Post
2. Click add new post
3.  Click on the Title.
4. You will see the regenerate options even though we haven't added yet anything.

Image

### Environment info

- WordPress: 7.0-RC2
- Device: Windows 11
- Browser: Chrome
- Plugin: AI Dev plugin version 0.6.0
- Theme: Twenty Twenty-Five

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [x] Not sure

dhrupo:
## What?
Closes #391

Hides the Title Generation control until the post has body content, and adds an e2e regression test for that empty-state behavior.

## Why?
The current UI shows the title generation toolbar even when there is no content to generate from. That creates a confusing empty-state flow and matches the behavior reported in #391.

## How?
The title toolbar now reads live editor state with `useSelect` and returns `null` when the edited post content is empty. The existing Generate/Regenerate labeling stays the same once content exists. An e2e test was added to verify the toolbar stays hidden for a blank post and appears after content is added.

### Use of AI Tools
AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5
Used for: Drafting the implementation and regression test; final changes were reviewed and edited before submission.

## Testing Instructions
1. Enable the Title Generation experiment.
2. Create a new post with no title and no body content.
3. Focus the title field and confirm the Title Generation control is not shown.
4. Add body content to the post.
5. Focus the title field again and confirm the control appears with the `Generate` label.
6. For a post that already has a title and body content, confirm the control still appears with the `Regenerate` label.

Ran locally:
- `npm run typecheck`
- `npm run lint:js -- src/experiments/title-generation/components/TitleToolbar.tsx tests/e2e/specs/experiments/title-generation.spec.js`

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-479-ab80b8f446e7e3e7e1fa0ab9e58ddce5b705110e.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/title-generation/components/TitleToolbar.tsx
+++ b/src/experiments/title-generation/components/TitleToolbar.tsx
@@ -14,7 +14,7 @@ import {
 	ToolbarGroup,
 	ToolbarButton,
 } from '@wordpress/components';
-import { dispatch, select, useDispatch } from '@wordpress/data';
+import { dispatch, useDispatch, useSelect } from '@wordpress/data';
 import { store as editorStore, PostTypeSupportCheck } from '@wordpress/editor';
 import { useState } from '@wordpress/element';
 import { update } from '@wordpress/icons';
@@ -78,8 +78,15 @@ interface TitleToolbarProps {
 export default function TitleToolbar( {
 	isStandalone = false,
 }: TitleToolbarProps ): React.JSX.Element | null {
-	const postId = select( editorStore ).getCurrentPostId();
-	const title = select( editorStore ).getEditedPostAttribute( 'title' );
+	const { postId, content, title } = useSelect( ( select ) => {
+		const editor = select( editorStore );
+
+		return {
+			postId: editor.getCurrentPostId(),
+			content: editor.getEditedPostContent() || '',
+			title: editor.getEditedPostAttribute( 'title' ) || '',
+		};
+	}, [] );
 
 	const { editPost } = useDispatch( editorStore );
 
@@ -94,6 +101,7 @@ export default function TitleToolbar( {
 		setGeneratedTitle( '' );
 	};
 
+	const hasContent = content.trim().length > 0;
 	const hasTitle = title.trim().length > 0;
 
 	let buttonLabel: string = __( 'Generate', 'ai' );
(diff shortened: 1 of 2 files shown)


dkotter (commented):
Tested and things work great. That said, we also have #492 open that adds the same thing but for the Summarization experiment. The approach taken there was to keep the UI but have it disabled until we hit a certain content length (100 characters by default). Ideally we use the same approach for both tasks so we don't end up with different experiences.

@dhrupo Any thoughts on the approach taken in #492 and if it's worth updating this PR to match? Or vice-versa, if you think that PR should update to match this one?

jeffpaul:
I also prefer the approach in #492, so let's update to showing the button disabled (and perhaps a tooltip or on-hover message noting the minimum characters needed like is done in 492)?

dhrupo:
I think #492’s approach makes sense for summarization because it’s a persistent sidebar surface and can explain why the action is disabled. For title generation here, I preferred hiding because it’s a toolbar/title-focus control and a disabled button there felt less clear.

So I’d lean toward keeping the UX pattern dependent on the surface:
- sidebar UI: visible + disabled + explanation
- toolbar UI: hidden until usable

I do agree we should keep the eligibility logic aligned where possible. We should avoid drifting into inconsistent eligibility rules. If #492 lands with the minimum-length approach, I’m fine revisiting #479 afterward to see whether we want to align the underlying threshold/eligibility logic more closely, even if the presentation differs by surface.

jeffpaul:
Ok, fair points.  Though I would want to find a way for a user to know what's required to have the title gen experiment be available to them (as the disabled button and helper text does for summarization).  Any ideas on how to handle that in a scenario where you're advocating for not showing a disabled title gen button @dhrupo?

dkotter:
Closing this out as this as been resolved now in #581

PR #481, opened 2026-04-27
Ensure any `sanitize_callback`'s in our Abilities input schema are executed

dkotter:
## What?

Related to  #346

Ensure any `sanitize_callback`'s set in our Abilities input schema are run

## Why?

There's conversation on #346 and #364 around the execution of `sanitize_callback`'s in the Abilities API. Right now those are never executed and previously would also cause issues with the client-side Abilities API (which #364 attempted to fix by removing).

Now that we don't need to remove those to make the client-side API work, this PR takes the approach of overriding the parent `normalize_input` method and adding our own handling, based on the input schema, for custom `sanitize_callback` methods. This is set to match closely what the REST API does and is meant to show a proof of concept on how this could be done to attempt to get feedback and then addressed upstream.

## How?

- In our `Abstract_Ability` class, override the `normalize_input` method and call our own custom sanitize method
- Within this method, iterate through the schema and run any `sanitize_callback` method we find. If a `WP_Error` is returned, return that instead
- Override the `validate_input` method as well to short-circuit if a `WP_Error` is returned from our sanitization
   
Note this PR ensures any existing `sanitize_callback` methods are run but doesn't add in any that may be missing. That could be done in this PR but will probably tackle that in a separate PR to keep this clean.

### Use of AI Tools

AI assistance: Yes
Tool(s): Cursor
Model(s): GPT-5.5
Used for: Exploring how the REST API handles sanitization and then building an approach based on that. Manual testing and refinement done by me

## Testing Instructions

Hard to fully test without putting some breakpoints in place to ensure the `sanitize_callback`'s are now executed but main thing to test is ensuring all existing features still work

## Changelog Entry

> Added - Ensure any `sanitize_callback` methods defined in our Abilities input schema are executed properly.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-481-0f7ad18c6e6c636c6baf076085da4e2307efd16e.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


gziolo:
I’ll take a closer look tomorrow. It appears to be correct at first glance.

gziolo:
It looks like it's way more complex than I anticipated, so I spent more time researching how this integrates with the REST API lifecycle.
A few takeaways:

- Core REST only honors `sanitize_callback` at the top level of registered args. `rest_sanitize_value_from_schema()` doesn't recognize the keyword for nested properties (see [#49960](https://core.trac.wordpress.org/ticket/49960)). This PR is going further than REST does, which is reasonable for Abilities, but worth noting in the docblock so readers don't assume parity.
- The ordering is the part I'd want to think about most carefully after further investigation. `WP_Ability::execute()` runs `normalize_input` → `validate_input`, and this PR puts sanitization inside `normalize_input` as discussed before. However, effectively sanitize runs before validate. REST flipped that ordering in [#37192](https://core.trac.wordpress.org/ticket/37192) (4.6) precisely because validators were seeing coerced values instead of raw input. The `'7'` → `7` default test shows the same shape: `absint` runs before `'type' => 'integer'` validation gets a chance. Worth deciding whether that's intentional for Abilities or whether we want a separate sanitize step after `validate_input`.
- Callback signature: REST passes `($value, $request, $param)`, this passes `($value)`. Extending after release might not be so straightforward. I'd suggest `($value, $param_path)` with a path accumulator threaded through the recursion. Single-arg callbacks like `sanitize_text_field` keep working, and callbacks that need to know which property they're sanitizing (shared sanitizers, error messages) get the path.

gziolo:
I have just landed https://github.com/WordPress/wordpress-develop/pull/12025, which documents the current behavior on `trunk` in WordPress core for REST API request params behavior (`sanitize_callback`, `validate_callback`, `arg_options`). It’s part of a larger effort to unify schema handling for the Abilities API between the JavaScript client and the PHP server implementation. For more details, refer to this ticket: https://core.trac.wordpress.org/ticket/64955.

As part of the effort, I’ve been contemplating how REST API-specific handling embedded within schemas aligns with abilities. My inclination leans towards the belief that input and output schemas in abilities should remain unchanged, as they accurately represent the data structure. Consequently, all other aspects can be addressed through a collection of filters that operate at various stages of ability execution. In my opinion, the most significant challenge lies in the fact that REST API only validates and sanitizes parameters in REST API requests, yet it treats the schemas returned for the output of endpoints as information rather than a contract that undergoes validation.

dkotter:
At this point going to close this out. If this is something we want to continue to iterating on, happy to re-open but seems like there's discussion upstream that is more valuable than continuing to keep this PR fresh

PR #482, opened 2026-04-28
Utilize the new client-side Abilities API

### LINKED ISSUE #346: Executing summarization ability with @wordpress/abilities in WordPress 7.0 doesn't work due to invalid schema

dkotter:
## What?

Closes #346 

Ensure we properly use the new client-side Abilities API

## Why?

We currently have a helper method that attempted to use the client-side Abilities API, falling back to making direct API requests if it wasn't available. This was based on the older client-side API and so needed updating with the new version that is shipping with WordPress 7.0 (see [post](https://make.wordpress.org/core/2026/03/24/client-side-abilities-api-in-wordpress-7-0/)).

Note this code was pulled directly from #364, as that PR is no split into this one and #481

## How?

- Update our `runAbility` helper function to use the new client-side Abilities API
- Ensure the `@wordpress/core-abilities` script module is loaded anytime an Ability needs it
- Fix some schema validations issues, particularly ensuring block content is always returned as a string and numbers returned as integers. Otherwise the client-side validation fails and it falls back to making direct API requests (which do work but isn't what we want here)

### Use of AI Tools

AI assistance: Yes
Tool(s): Cursor
Model(s): GPT-5.3 Codex
Used for: Determining how to properly load the `@wordpress/core-abilities` script module based on our current setup. Also used it to debug schema validation errors. Most work was then done by me

## Testing Instructions

2. For each feature that utilizes AI (everything besides the Abilities Explorer), turn those on one-by-one
3. Test the feature you turned on, ensuring you have the browser console open. The feature should function the same as it always has and you shouldn't see any errors or messages in the console. Importantly, shouldn't see `[AI] Client ability execution unavailable. Falling back to REST.`
4. Continuing testing each feature ensuring they all still work as expected

## Changelog Entry

> Fixed - Ensure we properly use the new client-side Abilities API

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-482-cc24d04690da2c26b371fe65996f2141cb59c6ef.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -147,6 +147,7 @@ private function maybe_enqueue_media_script(): void {
 			return;
 		}
 
+		wp_enqueue_script_module( '@wordpress/core-abilities' );
```
gziolo: Out of curiosity, can the build tool detect this script module dependency and automatically include it in the asset file for the entry point that needs it? It would be a more future-proof approach rather than remembering to add it manually.

src/utils/run-ability.ts
```diff
@@ -49,12 +54,12 @@ const logFallbackWarning = () => {
 
 	// eslint-disable-next-line no-console
 	console.warn(
-		'[AI] wp.abilities.executeAbility is unavailable. Falling back to REST.'
+		'[AI] Client ability execution is unavailable. Falling back to REST.'
```
gziolo, line 61: Is this still necessary with WordPress 7.0 set as minimum version?

dkotter: I think we want to keep the fallback as there are still times that's used (for instance, if schema validation fails, the client-side API rejects the request and we fallback to direct API requests which will normally work).

I'm fine getting rid of the `console.warn` if we want, though makes it easier to test

gziolo: That's fine to leave it in that case. I wasn't sure what was the reasoning behind it 👍🏻

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
dkotter: I'll admit my knowledge on this is limited so I could be wrong here but my understanding is that since this is a script module and not a classic script, dependency extraction doesn't work here. I think we'd have to move away from using `wp_enqueue_script` for all of our scripts and convert those all into modules we import via `wp_enqueue_script_module`, along with some other changes. So this seemed like the easiest approach.

But if I'm wrong on this and there's a better approach, definitely interested in learning that.

One thing we could do that I had considered is update our `Asset_Loader::enqueue_script` method to auto include this script module when needed. That would reduce the duplication here and new experiments in the future wouldn't need to have this line added. Would just need an extra argument added to `Asset_Loader::enqueue_script` to determine if this module should be enqueued or not

gziolo: @sirreal, would you mind sharing insights what’s possible with WordPress 7.0 as minimum version in regards to mixing scripts and script modules?

gziolo: This isn’t a blocker. I’m not even sure we can do much about it without larger changes.

sirreal: [Changeset 61587](https://core.trac.wordpress.org/changeset/61587) landed in WordPress to allow Classic Scripts to depend on Script Modules.

Usage looks like this:

```php
wp_register_script_module( '@example/module', /*…*/ );

wp_enqueue_script(
	'example-classic',
	'/classic.js',
	array( 'classic-script-dependency' ), // Classic script dependencies.
	null,
	array(
		// Script module dependencies are declared here.
		'module_dependencies' => array( '@example/module' ),
	)
);
```

There's a related issue about updating build tooling:

https://github.com/WordPress/gutenberg/issues/75196

gziolo: Nice, thank you @sirreal for all the pointers 💯

dkotter: I've added an extra argument to our `enqueue_script` helper method that if set to `true`, will load the `@wordpress/core-abilities` script module: https://github.com/WordPress/ai/pull/482/changes/3e8a6292ffcf32b3018b83cdbfedd077aa80dfc4

Note in testing, if I just set that as a `module_dependencies` it wouldn't work, I still needed to call `wp_enqueue_script_module( '@wordpress/core-abilities' )`

sirreal: I think this has discovered an issue with the script->module dependencies. If an eagerly evaluated script depends on a script module, the `importmap` may be printed after the script, so the script will be unable to resolve the module. Was the error you saw like this?

> TypeError: Failed to resolve module specifier '@wordpress/core-abilities'

You should _not_ need to enqueue the script module in this case. I'm opening a ticket for this.

> [!NOTE]
> This _should_ work as-is, but I'll add workarounds and other follow-up questions.

Is there a compelling reason not to use modules more broadly in this project? Module->module dependencies should work without issue.

For the script that depends on the abilities module, adding `defer` is a workaround that should resolve the issue.

```php
wp_enqueue_script(
  /*…*/
  array(
    'in_footer' => true,   // No need to print a deferred script early.
    'strategy' => 'defer', // Add `defer` attribute
    'module_dependencies' => array( '@wordpress/core-abilities' ),
  ),
);
```

This ensures that the script is evaluated after the importmap is available.

sirreal: [Here's the Trac ticket.](https://core.trac.wordpress.org/ticket/65165)

dkotter: > Was the error you saw like this?
> 
> TypeError: Failed to resolve module specifier '@wordpress/core-abilities'

I get an error around not being able to find the Ability, example `Error: Ability not found: ai/title-generation`. This goes away once I add `wp_enqueue_script_module( '@wordpress/core-abilities' )`

> Is there a compelling reason not to use modules more broadly in this project? Module->module dependencies should work without issue.

Just not the way things have been setup before this. I'm not opposed to moving that direction but that's likely a larger discussion and task.

> For the script that depends on the abilities module, adding defer is a workaround that should resolve the issue.

We do already set `'strategy' => 'defer'`. I added `'in_footer' => true` to test but that doesn't fix things either.

sirreal: This seems to work, although there are some issues with it:

```php
wp_register_script(
  'example',
  false,
  array( 'wp-api-fetch', 'wp-url', 'wp-data', 'wp-i18n', ),
  false,
  array(
    'in_footer' => true,
    'strategy' => 'defer',
    'module_dependencies' => array( '@wordpress/abilities', '@wordpress/core-abilities' ),
  ),
);
wp_enqueue_script( 'example' );
wp_add_inline_script(
  'example',
  <<<JAVASCRIPT
  ( async () => {
    await import( '@wordpress/core-abilities' );
    const m = await import( '@wordpress/abilities' );
    window.m = m;
    console.log({m});
  } )()
  JAVASCRIPT,
);
```

It seems like the abilities modules expect several classic WordPress scripts to be present (see the dependency array in the example script).

Calling the `@wordpress/abilities` module's `getAbilities()` function is an array with 2 entries. That may be because I don't have this plugin registered? Can you provide details about the steps to discover the issue you're seeing?

dkotter: Here's the steps I'm following:

1. Checkout this PR and run `npm i && npm run build`
2. Comment out this line so the script module isn't loaded: https://github.com/WordPress/ai/pull/482/changes#diff-9b683a19a4d64a079b42ade32de8038f235d7b12f619ae623796f7a9eff667cdR69
3. Go to `Settings > AI` and turn on Title Generation (you don't even need an AI Connector setup to test this)
4. Go to a post, open the browser console, click into the title and click the Regenerate button that shows
5. You should see an error in the console: `Error: Ability not found: ai/title-generation`. Right after that, should see a message: `[AI] Client ability execution is unavailable. Falling back to REST.`. If using the client-side API doesn't work, we fallback to making direct REST requests using `apiFetch`
6. Uncomment out the script module line above and try the steps again
7. This time you should see an error around invalid schema due to `sanitize_callback`, which will be addressed when the next beta version of 7.0 is released. So it tried to use the `executeAbility` function but that failed due to schema validation issues and it then falls back to using `apiFetch` again

Looking at your code example, there are some differences to what I've been trying. I've only been including `@wordpress/core-abilities` as a script module and only been importing `@wordpress/abilities`. I tried adjusting that to include both but still get the same error.

I even replaced my code with the following (adjusted slightly from your example) and got the same error (if I comment out `wp_enqueue_script_module( '@wordpress/core-abilities' )`):

```php
wp_register_script(
  'example',
  false,
  array( 'wp-api-fetch', 'wp-url', 'wp-data', 'wp-i18n', ),
  false,
  array(
    'in_footer' => true,
    'strategy' => 'defer',
    'module_dependencies' => array( '@wordpress/abilities', '@wordpress/core-abilities' ),
  ),
);
wp_enqueue_script( 'example' );
wp_add_inline_script(
  'example',
  <<<JAVASCRIPT
  ( async () => {
    await import( '@wordpress/core-abilities' );
    const m = await import( '@wordpress/abilities' );
    window.m = m;
    m.executeAbility('ai/title-generation');
  } )()
  JAVASCRIPT,
);
```

Screenshot 2026-05-05 at 10 18 02 AM

sirreal: Thanks, I tested this out and was able to reproduce.

As far as I can tell, it's a race condition. It has more to do with wp-data than with the module system. I didn't see any initial state in the DOM, so the abilities store has an empty ability list until the request to `/wp-json/wp-abilities/v1/abilities` completes and data is present in the store. Eager evaluation here that expects abilities data to be present is extremely likely to fail because the initial abilities are `[]`.

If I run the following on the browser console, I see the abilities are present (after the page has loaded and data is in the store).

```js
const wpAbilities = await import('@wordpress/abilities');
wpAbilities.getAbilities();
// [ …4 items… ]
```

I believe that manually enqueuing the module in your case changes the race conditions without eliminating them.

An example like this makes the race clear:

```php
wp_register_script(
	'example',
	false,
	array( 'wp-api-fetch', 'wp-url', 'wp-data', 'wp-i18n' ),
	false,
	array(
		'in_footer' => true,
		'strategy' => 'defer',
		'module_dependencies' => array( '@wordpress/abilities' ),
	),
);
wp_enqueue_script( 'example' );
wp_add_inline_script(
	'example',
	<<<JAVASCRIPT
	( async () => {
		const wpAbilities = await import( '@wordpress/abilities' );
		console.log( 'Initial abilities: %o', wpAbilities.getAbilities().map( ab => ab.name ) );
		const wpData = window.wp.data;
		const selectAbilities = wpData.select( wpAbilities.store )
		const {promise, resolve} = Promise.withResolvers();
		const unsubscribe = wpData.subscribe(
			(...args) => {
				if ( wpAbilities.getAbilities().length ) {
					resolve()
				}
			}
		);
		await promise;
		unsubscribe();
		console.log( 'After subscription: %o', wpAbilities.getAbilities().map( ab => ab.name ) );
	} )()
	JAVASCRIPT,
);
```

In my testing I see:

```
Initial abilities: []
After subscription: (4) ['core/get-site-info', 'core/get-environment-info', 'ai/title-generation', 'ai/get-post-details']
```

westonruter: I believe this race condition is resolved by explicitly waiting for the `ready` promise which is newly exported from `@wordpress/core-abilities` as of https://github.com/WordPress/gutenberg/pull/77254.

So, @dkotter, your example in https://github.com/WordPress/ai/pull/482#discussion_r3189977241 should be made reliable by adding `await ready` as in the following:

```php
wp_register_script(
  'example',
  false,
  array( 'wp-api-fetch', 'wp-url', 'wp-data', 'wp-i18n', ),
  false,
  array(
    'in_footer' => true,
    'strategy' => 'defer',
    'module_dependencies' => array( '@wordpress/abilities', '@wordpress/core-abilities' ),
  ),
);
wp_enqueue_script( 'example' );
wp_add_inline_script(
  'example',
  <<<JAVASCRIPT
  ( async () => {
    const { ready } = await import( '@wordpress/core-abilities' );
    await ready; // ⬅️⬅️⬅️⬅️ ✨ NEW ✨✨✨✨✨
    const m = await import( '@wordpress/abilities' );
    window.m = m;
    m.executeAbility('ai/title-generation');
  } )()
  JAVASCRIPT,
);
```

sirreal: Thanks, that `ready` promise seems like exactly what's missing here. In my testing it fixes the race condition.

gziolo: @westonruter’s detective work is truly impressive! 👍🏻

I’m curious to know if there are any suggestions on how to improve this functionality in WordPress core. This way, plugin authors wouldn’t have to deal with such a complex and nuanced configuration outside of the editor. Another question is whether there isn't a preexisting similar issue in the post and site editor.

dkotter: Thanks for the pointer @westonruter! Added the new `ready` promise and verified I was able to remove the `wp_enqueue_script_module` now. This is done in c79587687fa9347d813da2b22fe0abb6e91210e2

package.json
```diff
@@ -64,12 +64,14 @@
 	},
 	"dependencies": {
```
westonruter, line 65: Most of these should actually be `devDependencies` since they're only here for the sake of tooling, right? Aren't the packages from core actually being used by the plugin?

src/utils/run-ability.ts
```diff
@@ -69,7 +78,15 @@ const isAbilityNotFoundError = ( error: unknown ): boolean => {
 			: '';
 
 	return (
-		code === 'ability_not_found' || message.includes( 'Ability not found' )
+		// Ability is not registered in the client store yet.
+		code === 'ability_not_found' ||
+		message.includes( 'Ability not found' ) ||
```
westonruter, line 83: Is this safe to look for? The string is translated in core:

```
wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-run-controller.php:87:				__( 'Ability not found.' ),
wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-run-controller.php:148:				__( 'Ability not found.' ),
wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php:164:				__( 'Ability not found.' ),
```

src/utils/run-ability.ts
```diff
@@ -69,7 +78,15 @@ const isAbilityNotFoundError = ( error: unknown ): boolean => {
...
+		// Ability exists but has no callback in client store (server-only ability
+		// not yet bridged to a callback), so use REST instead.
+		message.includes( 'missing callback' ) ||
```
westonruter, line 86: Similarly here, this looks like a string which could be translated.

src/utils/run-ability.ts
```diff
@@ -26,21 +26,30 @@ type RunAbilityOptions = {
...
+			/* webpackIgnore: true */ '@wordpress/core-abilities'
+		);
+		await ready;
+		return ( await import(
+			/* webpackIgnore: true */ '@wordpress/abilities'
+		) ) as AbilitiesModule;
+	} catch {
+		return null;
+	}
+};
```
westonruter, line 52: This looks good 👍🏻

includes/Asset_Loader.php
```diff
@@ -63,12 +64,24 @@ public static function enqueue_script( string $handle, string $file_name ): void
...
+			'in_footer' => true,
+			'strategy'  => 'defer',
+		);
+
+		if ( $include_core_abilities ) {
+			$args['module_dependencies'] = array(
+				'@wordpress/abilities',
+				'@wordpress/core-abilities',
+			);
+		}
```
westonruter, line 100: Good! Per Core-65165

includes/Asset_Loader.php
```diff
@@ -53,8 +53,9 @@ final class Asset_Loader {
 	 *
 	 * @param string $handle    The handle for the script.
 	 * @param string $file_name The script file name without the .js extension.
+	 * @param bool $include_core_abilities Whether to include the core abilities script.
 	 */
-	public static function enqueue_script( string $handle, string $file_name ): void {
+	public static function enqueue_script( string $handle, string $file_name, bool $include_core_abilities = false ): void {
```
westonruter: Instead of a `bool` how about an args array? This will be much easier to read and it will scale to adding more args in the future.


```suggestion
	 * @param array{ include_core_abilities?: bool } $args Args.
	 */
	public static function enqueue_script( string $handle, string $file_name, array $args = array() ): void {
```

includes/Asset_Loader.php
```diff
@@ -63,12 +64,24 @@ public static function enqueue_script( string $handle, string $file_name ): void
 			return;
 		}
 
+		$args = array(
+			'in_footer' => true,
+			'strategy'  => 'defer',
+		);
+
+		if ( $include_core_abilities ) {
```
westonruter: Per above:

```suggestion
		if ( $args['include_core_abilities'] ?? false ) {
```

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -97,7 +97,7 @@ public function register_abilities(): void {
 	 * @since 0.3.0
 	 */
 	public function enqueue_editor_assets(): void {
-		Asset_Loader::enqueue_script( 'alt_text_generation', 'experiments/alt-text-generation' );
+		Asset_Loader::enqueue_script( 'alt_text_generation', 'experiments/alt-text-generation', true );
```
westonruter:
```suggestion
		Asset_Loader::enqueue_script( 'alt_text_generation', 'experiments/alt-text-generation', array( 'include_core_abilities' => true ) );
```

Per below and replicated in the other instances.

westonruter (changes requested):

package.json
dkotter: Open to discussion on this again, originally these were all `devDependencies`. Moved them to normal `dependencies` based on feedback I was given: https://github.com/WordPress/ai/pull/83#discussion_r2557957896

src/utils/run-ability.ts
dkotter: My understanding is the client-side API does it's own validation before passing to the server to do additional validation. That client-side validation can fail and it sometimes returns an error message with an error code (which we look for that code) and sometimes just returns an error message (which is why we have this code here).

See https://github.com/WordPress/gutenberg/blob/c6740febed9d6150bceb86916f47bcedd08e6fbb/packages/abilities/src/api.ts#L173

Would be great if this could be updated to always have a code associated with the error message but this section is meant to get around that.

That said, not actually sure if this check is needed for us, as we're only ever using this when we have an actual Ability with a valid callback so likely this message check can be removed

package.json
westonruter: I don't feel strongly. I don't think there's a right or wrong answer for that.

includes/Asset_Loader.php
dkotter: Updated in https://github.com/WordPress/ai/pull/482/commits/17993408dd43aa13101b529361fc88159d784431

includes/Asset_Loader.php
dkotter: Updated in https://github.com/WordPress/ai/pull/482/commits/17993408dd43aa13101b529361fc88159d784431

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
dkotter: Updated in https://github.com/WordPress/ai/pull/482/commits/17993408dd43aa13101b529361fc88159d784431

includes/Asset_Loader.php
```diff
@@ -86,12 +87,24 @@ public static function enqueue_script( string $handle, string $file_name ): void
 			return;
 		}
 
+		$args = array(
+			'in_footer' => true,
+			'strategy'  => 'defer',
```
westonruter, line 92: Is adding `defer` always safe for all the enqueued scripts? Will this cause any issues? (I'm happy for it to be the default otherwise.)

src/utils/run-ability.ts
westonruter: I see. So this is an client-side `Error` that is being thrown with an untranslated error message.

includes/Asset_Loader.php
dkotter: That's been our default so far (that's not new in this PR, the `in_footer` part is new). So far haven't seen any issues. I could see this being something we add to the new `$extra_args` parameter to turn this off if it ever does cause an issue

src/utils/run-ability.ts
dkotter: Yeah, as far as I can tell. My experience so far is the client-side validation is more strict then the server-side validation. So what I'm attempting here is trying to determine if something failed due to client-side validation and if so, re-attempt it with a direct REST request (which usually succeeds).

includes/Asset_Loader.php
westonruter: Oh yes, I see `defer` is the default below.

westonruter (dismissed):
This looks good to me. I don't have anything else to add.

PR #486, opened 2026-04-29
Add developer settings mode with the ability to set desired provider and model per feature

### LINKED ISSUE #33: Add advanced configuration tools for power users
Description

Provide configuration options and tools that give advanced users more control over AI behavior and provider usage.  These should be opt-in and surfaced in a way that does not overwhelm non-technical users.

Goals

- [ ] Extend settings UI to include an “Advanced” tab or expandable section.
- [x] Add configuration controls for: Model selection or override (where supported by the provider).
- [ ] Add configuration controls for: Adjustable model parameters such as temperature, top_p, and max tokens.
- [ ] Add configuration controls for: Logging or debug mode toggle.
- [ ] Integrate these options into the AI Playground where relevant to reduce duplication.
- [ ] Document how these advanced settings affect provider behavior and plugin performance.

### Acceptance Criteria

- Advanced configuration UI appears only when experimental features are enabled.
- Settings persist and are accessible to feature modules.
- Non-technical users are not exposed to unnecessary complexity.
- Documentation updated to explain each option and its purpose.

### LINKED ISSUE #262: Provider-Level Model Bucketing for Model Selection
Model selection is currently hardcoded in helpers.php via three static priority lists `(get_preferred_models_for_text_generation(), get_preferred_image_models(), get_preferred_vision_models())`. While this was an intentional design decision to prevent surprise model upgrades from breaking things or inflating costs, it creates a few pain points:

- Users with multiple API keys configured have no visibility into which provider is actually handling their tasks.
- There is no UI-level way to express a preference — customization is only possible via PHP filters, which is inaccessible to most users.
- Hardcoding specific model names is brittle given how fast the AI landscape changes, and it forces the team to make opinionated model-tier recommendations that may quickly become outdated.

Introduce provider-level bucketing as a middle ground between full model selection and the current fully hardcoded approach:

- Let users select a preferred provider (Anthropic, Google, or OpenAI) for text and image generation tasks, rather than picking a specific model.
- The internal model priority order within each provider remains intact — preserving the stability guarantees that were the motivation for the original design.
- Default provider values are pre-set based on the current priority logic, so behavior is unchanged for users who don't configure anything.
- This gives users meaningful visibility and control without exposing them to the complexity of individual model names.
- 

As a longer-term direction, a capability-tier bucketing approach (e.g. fast/cheap vs. high-reasoning) could complement this — borrowing from patterns used by tools like oh-my-opencode — with an escape hatch for users who disagree with a bucket's default assignment.

This work also has natural overlap with the Playground feature (#32), which will require dynamic model input regardless, and should be designed in alignment with whatever bucketing approach we land on here.

dkotter:
## What?

Add a new "developer mode" option on the settings page that when toggled on, shows a new settings section on each feature that allows you to choose the desired provider and model.

## Why?

While the average user may not care about the actual model being used for a feature (and likely won't know the difference between the dozens of model options) there are certainly users and situations where setting the desired provider (and optionally model) is desired and beneficial.

Since we don't want to overwhelm users with settings they don't care about, we're hiding these new options behind a "developer mode" that can be toggled on and off.

## How?

- Add a new dropdown menu on our settings page that when clicked, shows a Developer Tools menu with the ability to turn on the selection of provider and model 
- When in this mode, output a new developer settings section for each feature. This section only shows if a feature is toggled on
- Within this new settings section, add a dropdown that allows you to choose one of your connected providers
- When a provider is selected, a new dropdown shows that allows you to choose a model. These models are filtered to the selected provider and filtered to the feature. For example, in the Image Generation feature, you should only see image generation models. This data is delivered by a new REST endpoint as well as a new capability argument added to our feature classes
- Add new helper functions and methods to set the provider (and model, if set) when a request is made. This overrides the defaults we set

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Sonnet 4.6
Used for: Initial review and plan; executed on that plan and wrote tests; final review and tweaking done by me

## Testing Instructions

2. Go to the settings page: `Settings > AI`
3. Ensure you see a new three dot icon in the top toolbar
4. Click on this. Ensure a dropdown is shown with a Model Selection option. Click this and ensure it gets an active state
5. Toggle on an experiment or feature
6. Ensure a dropdown shows to select a provider
7. Select a provider
8. Ensure a new dropdown shows to select a model
9. Select a model from the list
10. Test the feature/experiment out and ensure it still works. We don't currently have built-in logging in order to easily see what provider and model was used but I do have a dev tool that logs this to the debug log that I'm happy to share if desired

## Screenshots or screencast

| Developer mode toggle | Developer mode menu | Provider and Model fields |
| ------ | ----- | ----- |
|    AI plugin settings screen showing the new developer mode icon    |   AI plugin settings screen showing the new developer menu open    |   AI plugin settings screen showing the new provider and model fields    |

## Changelog Entry

> Added - New developer mode that can be toggled on the settings page.
> Added - When developer mode is toggled on, show new settings allowing you to set the desired provider and model for a feature.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-486-87bb0a56413f243acafd00e367a267fb04dcd680.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


itsgajendraSingh:
Hi @dkotter,

I tried testing this PR using the “Try it in Playground” button, but it doesn’t seem to be working.

When Playground opens, it shows a “Blueprint execution failed” error and isn’t able to download/install the plugin from the CI artifact URL.

Screenshot 2026-05-02 at 7 23 05 PM

Screenshot 2026-05-02 at 7 23 27 PM

raftaar1191:
@dkotter PlayGround button is not working 


image

image

dkotter:
@itsgajendraSingh @raftaar1191 The Playground link should work now.

Note the GitHub Action that adds that Playground link is a bit flakey and doesn't seem to always work when a new commit is pushed (seems to fairly often update the referenced build even if that build doesn't exist). I'd recommend for best results to just pull the PR down and test locally.

jeffpaul (changes requested):
Its not immediately clear when clicking the new gear icon and its color/state changing is doing (even with the "Developer mode" on-hover text).  So perhaps we switch to the vertical ellipses like Gutenberg and add one item there to select/check/enable for "Model selection" with description of "Select the specific provider and model per feature"?

Image

dkotter:
> Its not immediately clear when clicking the new gear icon and its color/state changing is doing (even with the "Developer mode" on-hover text). So perhaps we switch to the vertical ellipses like Gutenberg and add one item there to select/check/enable for "Model selection" with description of "Select the specific provider and model per feature"?

This has been updated now (also updated screenshots in the PR description). I did add this menu item in a `DEVELOPER TOOLS` group but can remove that group if we don't want it or can rename it if there's a better phrase

jeffpaul (commented):
The model dropdown appears to be missing its down carret and then when a model is selected the X looks misaligned as well:

Image

dkotter:
> The model dropdown appears to be missing its down carret and then when a model is selected the X looks misaligned as well:

Misalignment was my fault, overaggressive CSS that was impacting that which I've fixed now.

For the missing caret, this is actually expected here. This page is using `DataForm` fields and in diving into those, when using a select input, if there are more than 10 items in that select, it automatically converts to an `adaptiveSelect` component that has built-in search functionality.

So what you're seeing here is you can click into the input and start typing and it will show you options that match. But that means it's slightly different than a normal select that just shows you dropdown items (and that will render that caret).

That said, for now I've forced these inputs to always use the normal `select` component instead of `adaptiveSelect` so now will always be a normal select input (with the caret) regardless of how many options it has. Can revert that back though if we think the search input is nicer (though I had the same issue, wasn't clear to me it actually is a search input)

jeffpaul (commented):
Instead of having all of this behind the loading animation, can the `Provider` label be immediate and the loading animation is the placeholder just for that dropdown?

Image

dkotter:
> Instead of having all of this behind the loading animation, can the `Provider` label be immediate and the loading animation is the placeholder just for that dropdown?

Updated now to show the label with the spinner icon below instead of hiding everything with that spinner

jeffpaul (dismissed):
Awesome, happy to see this merge into 0.9.0 today if you're good with that

phil-sola:
@dkotter @jeffpaul I'd beg to differ that the average user may not care about the model being used. This setting being hidden behind the 3 dots made this way too hard to find without knowing about it yesterday.

For alt text generation it used Claude Sonnet 5 out of the box and cost me nearly $2 for 200 images. Once realising what was happening and finding the setting hidden away and switching the model for alt text generation to Gemini 2.5 flash lite, I was able to do 4,100+ images for $0.25 - Quite the difference.

This could have cost me substantially more, by trusting the plugin to use an efficient model for the task.

jeffpaul:
@phil-sola the target audience is non-technical site owners, exposing provider/model settings for every feature (among other feature-specific settings) is more context than they need.  The problem you're describing should likely be controlled by a different means.

phil-sola:
@jeffpaul Completely understand and perhaps you’re right however I wanted to raise this, as the model selection certainly needs reviewing to prevent this.

Most users, if not manually selecting the model, will expect the plugin to do the right thing by them in terms of quality / cost.

AI is useful but if your method with unexpectedly high bills due to hidden model usage it’s not a great UX.

Just surfacing for a wider discussion as I think it’s useful.

jeffpaul:
@phil-sola if you wouldn't mind summarizing that into a new issue in the repo so I can get that in front of some other folks for review/input before identifying the best way to try and guard around that?

PR #489, opened 2026-04-30
Switch to more specific hooks to render the Generate Image buttons to avoid conflicts

### LINKED ISSUE #487: Generate Image action only toggles accordion and does not trigger generation
### Description

In the Image block, the "Generate Image" option appears inside an accordion. However, clicking on "Generate Image" only toggles (expands/collapses) the accordion and does not trigger any image generation action.

This creates confusion, as the label suggests an actionable feature but behaves like a UI toggle instead.

## Step-by-step reproduction instructions

1. Go to WordPress editor (Gutenberg)
2. Add an Image block
3. Add Caption
4. Locate the "Generate Image" option
5. Click on "Generate Image"
## Expected behaviour

Clicking on "Generate Image" should:

* Either trigger the image generation flow (AI prompt / modal / generation action), OR
* Clearly indicate that it is only a collapsible section (with proper labeling)

## Actual behaviour

Clicking on "Generate Image":

* Only expands/collapses the accordion
* Does not initiate any image generation process

Image

https://github.com/user-attachments/assets/67c6f0a7-d469-4198-b737-f639cd0dc5c0

## WordPress information

* WordPress version: 7.0
* Are all plugins except Gutenberg deactivated? Yes
* Required AI plugin
* Are you using a default theme? Yes

## Device information

* Device: Desktop
* Operating system: Windows / macOS
* WordPress app version: N/A

## Additional context

This might be a UX issue where the label "Generate Image" is misleading if it is intended to only expand a section. It could be improved by:

* Renaming it to something like "Image Generation Settings", OR
* Triggering the actual generation flow on click.

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

dkotter:
## What?

Closes #487

Switch to using `editor.MediaPlaceholder` and `editor.MediaReplaceFlow` as our hooks instead of using `editor.MediaUpload`

## Why?

As reported in #487, there's some scenarios where the Generate Image button is rendering (like above the caption field when that field is selected) where we don't want it.

Seems that when using the `editor.MediaUpload` hook to modify things, these sorts of "inline" components within an image block still show as being an image block even though they are nested within an image block.

I originally had fixed this by checking for a specific prop that only existed on the main image block but that seemed pretty fragile long term. Switch instead to using more specific hooks to add our buttons, both the inline placeholder button and the menu button.

## How?

- Switch away from using `editor.MediaUpload` as our only hook and instead use `editor.MediaPlaceholder` to add our button to the media placeholder and `editor.MediaReplaceFlow` to add our button to the replace menu
- Add a new helper function to ensure we only target the blocks we want (blocks that support images)
- Add new helper functions to ensure our button in the media placeholder stays in the same position (after the Upload button) instead of the default of either at the beginning or at the end

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code, Cursor
Model(s): Sonnet 4.6, GPT-5.5
Used for: Debugging and suggesting solutions. Review and testing of each suggestion done by me.

## Testing Instructions

2. Turn on the Image Gen Feature
3. Create a post and add an image block
4. Select an existing image for this block
5. Add a caption to this image in the block
6. Click into the caption and ensure you don't see an extra `Generate Image` button
7. Add an empty image block and ensure you do still see a `Generate Image` button and that works
8. For an image block that has an image already selected, click into the block and click Replace in the block toolbar. Ensure Generate Image shows in the dropdown and still works
9. Can also install the latest version of the Gutenberg plugin and turn on the `Block fields: Show dataform driven inspector fields on blocks that support them` experiment. This will then show the caption field in the image block sidebar. Ensure no extra Generate Image button shows there

## Changelog Entry

> Fixed - Ensure the Generate Image button doesn't render in contexts that aren't valid.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-489-1bc5fcf26d38d04fa348cbfc1f6cec044238b5f7.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (approved):
Testing via instructions and this works as expected, thanks!

PR #492, opened 2026-05-01
Enhance Summarization feature: add minimum content length filter and disable button for short content

### LINKED ISSUE #390: Generate Review Notes and Generate Summary options present even with no post content
### Description

When I trying to create post I notied without any content showing AI options like "Generate Summary" even AI option description is clearly mentioned "This will create a block that is a summary of the content of this post."

This option should show or enable only if post have some content.

1. Go to admin area
2. Create a new post
3 see in right sidebar where you can see "Generate Summary" option

Image

### Environment info

Wordpress Version : development version (7.0-RC2)
Device: Macbook
Browser: safari, chrome, edge and firefox as well.

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

coderGtm:
## What?

Closes #390 

Disabled the Summarize button in editor for shorter content and adds a filter for minimum content length required to enable the Summarize button (default is 100 characters).

## Why?

Summarize button stays visible even if there is no content or very less content on the page. This is practically useless and can potentially rack up AI costs for the user unknowingly. Disabling it for shorter content length ensures that it is only available when it makes sense.

## How?

1. PHP: `includes/Experiments/Summarization/Summarization.php`
- Added a `wpai_summarization_min_content_length` filter with a default value of 100
- The filtered value is passed to the frontend via `aiSummarizationData.minContentLength`

2. JS Hook — `src/experiments/summarization/functions/useSummaryGeneration.ts`
- Reads `minContentLength` from the localized data (falls back to 100)
- Strips HTML tags from the post content and checks if it's shorter than the minimum
- Exposes `isContentTooShort` and `minContentLength` in the returned object

3. JS Component — `src/experiments/summarization/components/SummarizationPlugin.tsx`
- The "Generate Summary" button is now disabled when `isContentTooShort` is `true`
- When disabled due to short content, the description text changes to: "The summarize button will be enabled when the post content has at least X characters."

4. JS Component — `src/experiments/summarization/components/SummarizationBlockControls.tsx`
The toolbar button is also disabled when content is too short

## Testing Instructions

1. Enable the AI Summarization feature from settings after connecting a connector.
2. Create/Edit a post/page.
3. Remove all the content and check if the Summarization button is disabled.
4. Add content equal to or above 100 characters and see if the button enables.
5. Generate the summary and again make the total content less than 100 characters. See if the toolbar button for regenerating summary gets disabled.
6. Try the above steps again after changing the value using the filter. You can add this to your theme's `function.php` for testing purposes:
  ```php
  add_filter( 'wpai_summarization_min_content_length', function () {
	  return 100; // Change to your desired minimum character count
  } );
```

## Screenshots or screencast

https://github.com/user-attachments/assets/1512c94a-cde5-47ef-9919-c764497a1bdb

## Changelog Entry

> Changed - AI Summarization experiment functionality.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-492-a81bf79c733bccf41ce7b7f906cbb0d1e8b3e355.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### CODE REVIEW

**Comment by coderGtm** [comment]
Okay I think I need to fix the test. Is it okay if I do it once we finalize the implementation?

File: includes/Experiments/Summarization/Summarization.php [hunk 1]
```diff
@@ -101,11 +101,21 @@ public function enqueue_assets( string $hook_suffix ): void {
 
 		Asset_Loader::enqueue_script( 'summarization', 'experiments/summarization' );
 		Asset_Loader::enqueue_style( 'summarization', 'experiments/summarization' );
+		/**
+		 * Filters the minimum content length required to enable summarization.
+		 *
+		 * @since 0.9.0
```
dkotter:
```suggestion
		 * @since x.x.x
```

src/experiments/summarization/components/SummarizationPlugin.tsx
```diff
@@ -31,15 +36,31 @@ export default function SummarizationPlugin() {
...
+
+	const isDisabled = isSummarizing || isContentTooShort;
+
+	let buttonDescription: string;
+
+	if ( isContentTooShort ) {
+		buttonDescription = sprintf(
+			/* translators: %d: minimum number of characters required */
+			__(
+				'The summarization experiment will be enabled when the post content has at least %d characters.',
```
dkotter:
```suggestion
				'Summarization will be available when the post content has at least %d characters.',
```

src/experiments/summarization/functions/useSummaryGeneration.ts
```diff
@@ -110,10 +112,20 @@ export function useSummaryGeneration() {
 		}
 	};
 
+	// Minimum content length required for summarization.
+	const minContentLength: number =
+		aiSummarizationData?.minContentLength ?? 100;
+	const strippedContent = content
+		? content.replace( /<[^>]+>/g, '' ).trim()
+		: '';
+	const isContentTooShort = strippedContent.length < minContentLength;
```
dkotter: WordPress does have a [`@wordpress/wordcount` package](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-wordcount/) that can count words or characters. I'd suggest we use that here.

dkotter (changes requested):
Left a few comments but overall this looks good and tests well.

I left a similar comment on #479 but both that PR and this PR solve the same problem though for different experiments (title generation vs content summarization). Ideally we take the same approach for both to avoid confusion.

I personally like the approach here, where we have a minimum length we look for and we disable the UI instead of removing it. But open to discussion on that.

One other thing we should look to add here that #479 has is adding an E2E test to verify a situation where the minimum content length isn't met, ensuring the button is disabled then.

jeffpaul:
I also prefer the approach here of showing the button disabled with the note of minimal characters needed.  Where reasonable, would be nice to allow that 100 character minimum to be filterable but not something we should expose as an AI setting.

coderGtm:
Hi all

Is there anything else left here to work on? Just following up to make sure we are on the same page.

dkotter (approved):

PR #495, opened 2026-05-01
Move alt text generation button next to alt text field

### LINKED ISSUE #425: Update placement of Alt Text generation buttons

Let's move the Alt Text generation button a bit closer contextually to where the Alt Text field exists in various locations.

On the `Edit Media` page, let's move it to be just under the `Alternative Text` text area and above the description text/link while also removing the standalone Alt Text metabox.

Image

On the `Attachment details` view, let's move it to be just under the `Alternative Text` text area and above the description text/link while also removing the Alt Text section/button further below.

Image

iamsohilvahora:
- Remove separate alt text section/button from the bottom of the
  Attachment Details view
- Use attachment_fields_to_edit to inject the Generate/Regenerate
  button immediately after the Alternative Text textarea in both
  locations
- Update JS init selector from removed metabox ID to the button itself

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-495-7177c6dfb0196a8dfcf06b15e5215ba154c4234f.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


src/experiments/alt-text-generation/media.ts
```diff
@@ -115,6 +113,146 @@ class AltTextMediaControls {
 		} );
 	}
 
+	/**
+	 * Gets the next unbound generate button.
+	 *
+	 * @since 0.3.0
```
dkotter, line 119: Any reason these `@since` statements are set to 0.3.0? If these are all newly added methods, this should be set to `x.x.x` and that will get updated when a release happens

package-lock.json
dkotter, line 1: Curious on why this file needs changed as part of this PR? Since there isn't any updates to `package.json` here I wouldn't expect any changes to the lock file.

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -300,8 +254,8 @@ public function add_button_to_media_modal( array $fields, ?\WP_Post $post ): arr
 		$fields['ai_alt_text'] = array(
 			'label'        => __( 'Alt Text', 'ai' ),
 			'input'        => 'html',
-			'show_in_edit' => false,
-			'html'         => '<div class="ai-alt-text-media-actions"><button id="ai-alt-text-generate-button" class="button button-secondary" type="button" data-attachment-id="' . absint( $post->ID ) . '">' . esc_html( $button_text ) . '</button><span class="spinner" aria-hidden="true" style="margin-left: 8px;"></span><p class="description" aria-live="polite" style="margin-top: 6px; font-size: 12px;"></p></div>',
+			'show_in_edit' => true,
```
dkotter, line 257: I'm seeing a stranded `Alt Text` field now on the single media edit screen:

Image

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -300,8 +254,8 @@ public function add_button_to_media_modal( array $fields, ?\WP_Post $post ): arr
...
+			'html'         => '<div class="ai-alt-text-media-actions"><button id="ai-alt-text-generate-button" class="button button-secondary ai-alt-text-generate-button" type="button" data-attachment-id="' . absint( $post->ID ) . '">' . esc_html( $button_text ) . '</button><span class="spinner" aria-hidden="true" style="margin-left: 8px;"></span><p class="description" aria-live="polite" style="margin-top: 6px; font-size: 12px;"></p></div>',
```
dkotter, line 258: The UI in the media modal view isn't great:

Image

dkotter: Also, the loading states both in the modal and on the single edit screen need some style tweaks:

Image

Image

dkotter (changes requested):

jeffpaul:
@iamsohilvahora did you check #425 to ensure you're catching all the places to update?  Also, it seemed from @CacheMeOwside in that issue that core didn't have hooks available to do this reasonably so I'm curious if the approach here is a bit too hacky until those hooks are available or if you've found an alternate approach?

CacheMeOwside:
Thanks for working on this @iamsohilvahora.

Also, thanks for the ping @jeffpaul! Core still doesn't expose a hook here (although I do see a PR raised for the [Trac ticket](https://core.trac.wordpress.org/ticket/65086)), so any approach today is a workaround.                                                                                                                                 

In this PR, PHP renders the button inside its own "Alt Text" row, then JS moves it next to the desired "Alternative text" textarea and tries to delete the empty row that's left behind. If that cleanup misfires, we end up with a stranded label, likely what @dkotter spotted on the edit screen.

Since Core is the proper long-term fix, it might be worth weighing whether to iterate on this approach or hold #425 until the hook is available in Core. I'm happy to help either way.

jeffpaul:
Yeah, I'd prefer to work off that core hook when its available and not take a hacky approach to try and force it before then.

jeffpaul:
Closing this while we await a hook in core to properly adjust our alt text integration, hopefully something for that will land in WP 7.1

PR #497, opened 2026-05-01
For image gen, move guidelines from being system instructions to being part of the prompt

### LINKED ISSUE #496: Image generation with OpenAI doesn't work if Guidelines are used
### Description

In #359 we added support for the new Guidelines experiment in Gutenberg. This allows you set various guidelines (site, copy, image, block) that others can then use as reference. In our case, we take those guidelines and attach them to our system instructions for various features.

In testing something else, ran into an issue where image generation would return the error `Image generation failed. Please ensure you have a connected provider that supports image generation`. In tracking down what was going on, found out that OpenAI doesn't support system instructions for their image models and so if you have either site or image guidelines set, those get added as system instructions and then the AI Client can't find any models that support image generation, leading to the mentioned error.

It seems some of the Google models also don't support instructions (any of the imagen ones) but the gemini models do. Likely other image models from other providers will also have various degrees of support.

I think there's a few options we can look to take here:

1. Remove the use of guidelines entirely from image generation
2. Instead of setting these as system instructions, add them as part of the prompt. This preserves the guidelines and should work for any image generation model. This probably makes more sense anyway when dealing with image generation (which was my mistake as I recommended this approach to match what we do elsewhere)
3. Keep these as system instructions for models that support it and remove when the model doesn't. This is tricky to do as we don't know the model until the AI Client tries to make the request

1. Install Gutenberg and enable the Guidelines experiment
2. Add a Site guideline and Images guideline
3. Setup OpenAI as your AI Connector
4. Turn on image generation
5. Try and generate an image. You should see an error

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

dkotter:
## What?

Closes #496

For image generation, add matching guidelines as part of the prompt instead of system instructions

## Why?

Turns out some image generation models (like OpenAI) don't support system instructions. This was wrong advice by me on the PR that implemented guideline support (I must have been testing with Google only).

This means that trying to generate images using OpenAI when you have guidelines in place won't work if those are set as system instructions.

## How?

Instead of setting the guidelines as a system prompt, append those to the prompt

### Use of AI Tools

None

## Testing Instructions

1. Pull this PR down
2. Ensure you have Image Generation turned on
3. Configure OpenAI as your AI Connector
4. Install the Gutenberg plugin and turn on the guidelines experiment
5. Add Site and Images guidelines
6. Generate an image and ensure it works

## Changelog Entry

> Changed - For image generation, set guidelines as part of the prompt instead of system instructions.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-497-8e782c556239a910064e4a419afa55b2fdddc977.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Image/Generate_Image.php
+++ b/includes/Abilities/Image/Generate_Image.php
@@ -248,18 +248,18 @@ class Generate_Image extends Abstract_Ability {
 		$request_options = new RequestOptions();
 		$request_options->setTimeout( 90 );
 
+		// Inject guidelines into the prompt. Unlike the other features, we don't
+		// use system instructions here because most image gen models don't support them.
+		$guidelines = $this->get_guidelines_for_prompt();
+		if ( $guidelines ) {
+			$prompt .= "\n\n" . 'The following guidelines represent the site&#039;s editorial standards. Apply them where relevant. If guidelines conflict with the above input, prioritize accuracy.';
+			$prompt .= "\n\n" . $guidelines;
+		}
+
 		$prompt_builder = wp_ai_client_prompt( $prompt )
 			->using_request_options( $request_options )
 			->as_output_file_type( FileTypeEnum::inline() )
 			->using_model_preference( ...get_preferred_image_models() );
-
-		// Inject guidelines as a system instruction to match other abilities.
-		$guidelines = $this->get_guidelines_for_prompt();
-		if ( $guidelines ) {
-			$instruction  = 'The following guidelines represent the site&#039;s editorial standards. Apply them where relevant. Do not fabricate content to satisfy guidelines. If guidelines conflict with the input, prioritize accuracy.';
-			$instruction .= "\n\n" . $guidelines;
-			$prompt_builder->using_system_instruction( $instruction );
-		}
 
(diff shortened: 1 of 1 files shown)


gziolo (approved):
Code changes look good 👍🏻

dkotter:
This was merged into `develop` (see https://github.com/WordPress/ai/commit/9518a7ccb4a7d945c0943dc261bcde6d79134e48) but GitHub isn't reflecting that status here. Going to manually close this out and just assume it's a GitHub glitch

PR #501, opened 2026-05-03
docs: add and refine experiment documentation

### LINKED ISSUE #221: Add missing experiment documentation pages
Several features in the AI Experiments plugin have dedicated markdown documentation under `/docs/experiments`, but a few key areas are missing equivalent docs.  To keep documentation consistent and discoverable, add markdown pages for the following features:

- [ ] Title Generation experiment
- [ ] Experiment Framework (registry, base class, lifecycle)
- [ ] Multi-Provider Support (provider detection, selection, and fallback behavior)

Ideally those docs should:

- Create markdown files under `/docs/experiments/` following the structure and tone of existing experiment docs.
- Clearly describe what each feature does, where (if) it appears in the UI, and how it is intended to be used.
- Include basic implementation notes for developers, including extension points where relevant.
- Cross-link these new docs from any relevant index or README files so they are easy to find (eg see https://github.com/WordPress/ai/pull/219#pullrequestreview-3778219997).

henryperkins:
## Summary

- Adds `docs/experiments/content-resizing.md` and `docs/experiments/title-generation.md` covering end-user behavior, developer details, REST usage, permissions, extension points, and testing notes.
- Refines related experiment documentation and resource links so they point at the current docs structure.
- Removes the redundant `docs/INTEGRATION_GUIDE.md`; the reusable Editorial Guidelines content now lives in the Developer Guide and the Title Generation doc links there.

Docs-only — no source files changed.

## Test plan

- [x] Confirmed the removed integration guide has no remaining internal references.
- [x] Verified the Title Generation Guidelines link points to `docs/DEVELOPER_GUIDE.md#editorial-guidelines`.
- [x] Ran `git diff --check` for whitespace/conflict-marker issues.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-501-0d9e184bb7bda48eb1761c2a16fab3a9b4715cd1.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (commented):
@dkotter the experiment docs here seem ok to me, curious for your take on the integration guide (mostly whether that content is already / should be moved elsewhere)?

dkotter:
> @dkotter the experiment docs here seem ok to me, curious for your take on the integration guide (mostly whether that content is already / should be moved elsewhere)?

I've tweaked the experiment docs a bit to remove unnecessary sections but those look good, thanks for catching those @henryperkins!

For the Integration Guide, I'm not sure the full point of this. I would say that most of the information here is already captured in other docs. It's also overly verbose which makes it hard for a human to consume and understand. It also makes lots of references to specific files and functions which will change over time and lead to this type of documentation from getting out of date.

My recommendation here is to remove that integration guide and we can proceed with merging this PR with just the new experiment docs, but open to other thoughts.

henryperkins:
@dkotter thanks for the feedback! The integration guide, you nailed it. I created it to assist LLMs with integrating plugins that otherwise create their own surfaces instead of utilizing this plugin (either at all or fully). If this is a redundant doc with that being said, I can certainly remove it. I can say it has helped me modernize plugins with AI much more than simply asking it to do so (without the doc). I do see the friction with it naming specific things that will change over time (some likely soon)

jeffpaul:
@henryperkins if there's content that's relevant in that integration guide, perhaps they get moved to other existing docs that could otherwise include that info?

docs/experiments/review-notes.md
jeffpaul, line 1: This will probably cause some merge conflicts with #528, but we can tackle those as that PR gets through final reviews+merge

jeffpaul (approved):

PR #503, opened 2026-05-04
Fix dependabot alerts

dkotter:
## What?

Manually address some Dependabot alerts that can't be fixed automatically

## Why?

Dependabot will open PRs to automatically fix alerts where possible. In some cases, it can't cleanly open a PR. These alerts will show under the Security > Dependabot section but you have to manually fix those. This PR addresses the issues that are easily fixable.

## How?

- Override `serialize-javascript` to install version >=7.0.3
- Override `postcss` to install version >=8.5.10
- Override `minimatch`, specifically when required by `markdownlint-cli`, to 3.1.4

### Use of AI Tools

AI assistance: Yes
Tool(s): Cursor
Model(s): GPT-5.5
Used for: Analyze reports and suggest fixes. Manual testing done by me

## Testing Instructions

Locally remove `node_modules` and ensure you can run `npm install` and that you can run commands like `npm run build` and `npm run lint:js`.

## Changelog Entry

> Security - Bump `serialize-javascript` from 6.0.2 to 7.0.5
> Security - Bump `postcss` from 8.5.10 to 8.5.14
> Security - Bump `minimatch` from 3.0.8 to 3.1.4

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-503-075e1de614a4d923faafcf2f9450f004ea641581.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


package.json
```diff
@@ -92,9 +92,14 @@
...
 		"@wordpress/route": "^0.10.0",
 		"@wordpress/theme": "^0.11.0",
+		"markdownlint-cli": {
+			"minimatch": "3.1.4"
+		},
+		"postcss": ">=8.5.10",
 		"prettier": "npm:wp-prettier@^3",
 		"react": "^18.3.1",
 		"react-dom": "^18.3.1",
+		"serialize-javascript": ">=7.0.3",
```
dkotter, line 102: Seems like these overrides is the best (maybe only?) way to solve these issues but open to other suggestions if there's something I'm missing

justlevine: I'm not by my desk to confirm but the other way to solve them is to dismiss the dependabot warning for being invalid. Minimatch for example - not sure why you're targeting markdowncli and not @wordpress/scripts, but either way my (admittedly faulty) memory tells me that there's no attack vector for us because its only used against our committed or user provided configs.

dkotter: > the other way to solve them is to dismiss the dependabot warning for being invalid

I didn't verify each of these individually, just seemed like a good idea to resolve.

> Minimatch for example - not sure why you're targeting markdowncli and not @wordpress/scripts

So there's multiple versions of `minimatch` being included here from various dependencies. `@wordpress/scripts` doesn't include it directly but some of it's dependencies do include it. `markdownlint-cli` is ultimately the downstream package that has version constraints such that the vulnerable version is included (< 3.1.3). I initially just had an override for `minimatch` directly but it was causing issues with other things and needed the direct scoping to prevent that.

jeffpaul (dismissed):
Functionality still continues to work as expected, thanks!

PR #512, opened 2026-05-06
fix: refine image generation loading state and hide button when image set

### LINKED ISSUE #270: Refine image generation loading state and fix layout shift in Generate Image modal

Follow up on UI feedback from @dkotter in https://github.com/WordPress/ai/pull/258#pullrequestreview-3898186001 to improve the loading experience and resolve layout shift issues in the Generate Image modal.

Two UI issues were identified during review:

1. Loading state needs refinement

The current loading state when generating images could be improved for clarity and consistency.  We should evaluate the best UX pattern both for the standalone Generate Image modal and inline generation within the editor.

2. Layout shift after generation

After an image is generated, selecting actions like Use Image or Generate Variation causes a small layout shift where the container width changes and the preview image shrinks slightly.

zeus2611:
## What?

Closes #270

Refines the image generation loading state and fixes layout shift in the Generate Image modal. Also hides the "Generate featured image" button once a featured image is already set.

## Why?

The loading state had several UX problems reported in #270:
- No visual feedback on first generation (blank space before image appears)
- Layout shift when switching from generating → preview states
- The "Generate featured image" button remained visible after an image  was already set, confusing users
- Button showed no feedback during generation

Additionally, `GenerateImageStandalone` and `GenerateImageInlineModal` had ~400 lines of duplicated state, logic, and JSX with no shared abstraction.

## How?

**Featured image button (`GenerateFeaturedImage.tsx`)**
- Button label switches to "Generating…" and `isBusy` activates during   generation
- Component returns `null` when `featured_media` is set and not currently generating, hiding the button entirely

**Loading state refinement**
- First generation: animated skeleton placeholder (aspect-ratio 1:1, pulsing gray) fills the image area, eliminating blank space
- Subsequent generations: previous image stays visible with a dim-pulse animation while the new image generates, preventing layout shift
- Spinner + progress label centered below the image/skeleton

**Shared component extraction**
- New `useImageGeneration` hook centralises all shared state (`state`, `prompt`, `refinePrompt`, `progress`, `error`) and the `generate()` function used identically in both components
- New `shared/index.tsx` exports four reusable components: `GeneratingState`, `PromptForm`, `ImageHistoryNav`, `RefinePromptForm`
- Both `GenerateImageStandalone` and `GenerateImageInlineModal` now consume the hook and shared components; component files reduced by ~200 lines each
- CSS unified under `.ai-image-generation__*` BEM namespace, removing duplicated rules from both component-specific blocks

**E2E tests**
- Updated selectors in `image-generation.spec.js` and `image-editing.spec.js` to match the renamed CSS classes

### Use of AI Tools

AI assistance: Implementation was planned with assistance from Claude Code (Anthropic). All code was reviewed, tested, and validated.

## Testing Instructions

**Featured image button**
1. Create/open a post without a featured image
2. Open the Document sidebar → confirm "Generate featured image" button is visible
3. Click it — confirm button shows "Generating…" with spinner and is disabled during generation
4. After generation completes — confirm button disappears (featured image is now set)
5. Remove the featured image — confirm button reappears

**Inline modal (block editor)**
1. Insert an Image block → open "Generate Image" modal
2. Submit a prompt — confirm skeleton placeholder appears (first generation, no prior image)
3. After image appears, click "Generate Another Image" — confirm previous image dims with pulse animation while new one generates (no layout shift)
4. Click "Refine Image" → submit refinement — confirm the same dim+skeleton behavior during refine generation
5. Confirm comparison view (before/after) renders correctly after refinement

**Standalone (Media Library)**
1. Navigate to Media → Generate Image
2. Repeat steps 2–5 from Inline modal above
3. Confirm "Save to Media Library" and success notice still work

**Regression**
- Confirm history navigation (prev/next arrows, version counter) works
- Confirm error messages surface correctly on API failure

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| image | Screenshot 2026-05-06 at 4 54 46 PM |

## Changelog Entry

Fixed - Refined image generation loading state: skeleton placeholder on first generation, dim-pulse animation on subsequent generations, no layout shift. Featured image button now hides when image is already set and shows "Generating…" during generation.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-512-b61471f16fabf4b89b30cc45ab4d96824cd6b318.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (dismissed):
Solid improvements, nice work!

src/features/image-generation/hooks/useImageGeneration.ts
```diff
@@ -0,0 +1,126 @@
...
+	const { activeEntry, historyIndex, addToHistory } = imageHistory;
+
+	async function generate(
+		activePrompt: string,
+		referenceImage?: string,
+		refHistoryIndex?: number
+	): Promise< void > {
+		setError( null );
+		setState( 'generating' );
+		setProgress( __( 'Generating image…', 'ai' ) );
```
dkotter: Should we update this to just `Generating…`? With changes in this PR, we now have a mix of both. Some places just say `Generating`, some say `Generating image`. Ideally we settle on one and I think we've gone with just `Generating…` for all other features

dkotter: Note in doing a search, I'm seeing `Generating image` in four different places so we'll want to update all of those

src/features/image-generation/components/GenerateFeaturedImage.tsx
```diff
@@ -21,53 +21,52 @@ import { uploadImage } from '../functions/upload-image';
...
 
 		try {
-			const generatedImageData = await generateImage( content, {
-				onProgress: setProgressMessage,
-			} );
-			const importedImage = await uploadImage( generatedImageData, {
-				onProgress: setProgressMessage,
-			} );
+			const generatedImageData = await generateImage( content );
+			const importedImage = await uploadImage( generatedImageData );
```
dkotter: Do we want to intentionally remove the progress message here? That was added by request to make it clear where we were at in the process, between generating the image prompt, generating the image, generating alt text or importing the image. We lose that now

dkotter (changes requested):
Overall this is looking better, left a few comments inline and have a few other suggestions below.

Just my opinion but not sure centering the loading state looks the best, particularly on the stand-alone page. I'd suggest left align:

Image

This is not a new issue but as long as we're fixing other UI/UX things, may be nice to address these as well. Right now when I generate an image within the media modal, I don't see the actual loading message unless I scroll down:

Image

I think ideally we should scroll to the bottom to ensure you can see that message. Maybe even make the skeleton placeholder smaller?

And then similarly after an image is generated and I'm taking action, like importing. I don't see the loading indicator unless I scroll down:

Image

I think a combination of ensuring we always scroll to the bottom and maybe making things a bit smaller would help here.

A couple other not new things is after saving a generated image, we show a success message above the image (I think a typical admin notice). You don't see that unless you scroll up. I think ideally we should show the success message inline with the other buttons so it's clear something happened.

We also don't disable the save button after saving so someone could save the image multiple times. Maybe that's fine but seems like a nice enhancement to disable that once a successful save happens.

src/features/image-generation/hooks/useImageGeneration.ts
zeus2611: Updated `useImageGeneration.ts` and `MediaLibraryImageEditor.tsx`. The sub-step strings in generate-image.ts (Generating image prompt, Generating image) are intentionally specific — they appear sequentially to tell the user which step they're on, so those stay descriptive.

src/features/image-generation/components/GenerateFeaturedImage.tsx
zeus2611: Reverted to the previous progress message to maintain a clear and consistent view of the current state in the process.

zeus2611:
- Left-aligned the standalone page to match the admin UI convention. For the modal, I kept it centered since the fixed-width container feels more balanced that way, but happy to left-align that too if you'd prefer consistency. Screenshot 2026-05-09 at 6 12 58 PM


- Reduced max-width from 480px to 320px so it fits in the visible modal viewport without scrolling. Screenshot 2026-05-09 at 6 11 06 PM


- Moved the success message below the action buttons, so it's visible without scrolling after a save. And, the "Save to Media Library" button is disabled for the current version once saved. Navigating to another version re-enables it. Resets on Cancel. Screenshot 2026-05-09 at 6 14 06 PM

dkotter (approved):

Issue #509, opened 2026-05-05
Add sorting and filtering plus additional labeling to Comment Moderation


jeffpaul:
## What problem does this address?

Following on from #155, let's look to extend the new Sentiment and Toxicity data to:

* add sorting to the Sentiment and Toxicity columns
* add filtering for the Sentiment and Toxicity data

Let's also update the Activity dashboard widget to render the Sentiment and Toxicity "pills" in the Recent Comments section.

## What is your proposed solution?

Proposed tasks noted above.

PR #516, opened 2026-05-06
Add a new public method to the `Comment_Analysis` class that is used to analyze a comment

### LINKED ISSUE #515: Fatal permission error during subscriber auto-moderation (Comment Moderation experiment)
### Description

Hey team! Following up on the newly merged Comment Moderation experiment ( #155 ). It looks awesome, but I found a specific edge case during local testing regarding user permissions.

**The Issue:**
If a logged-in Subscriber (or any user without the moderate_comments capability) leaves a comment on the front-end, the site throws a WP_Ability::execute permission fatal error.

**The Cause:** `moderate_comment()` hooks into `wp_insert_comment`, which fires under the current user's context. Because `Comment_Analysis` strictly requires the `moderate_comments` capability, the AI framework blocks the execution for non-admins and crashes the comment submission.

**Thoughts on a fix:**
a. Presently, I am temporarily using `wp_set_current_user( $first_found_admin )` inside `moderate_comment()` right before the ability executes and restoring the original user immediately after.

Via the following code:
```php
		$original_user = get_current_user_id();
		
		// Dynamically grab the first Administrator ID on the site
		$admins = get_users( array( 'role' => 'administrator', 'number' => 1, 'fields' => 'ID' ) );
		if ( ! empty( $admins ) ) {
			wp_set_current_user( $admins[0] );
		}
		$analysis = $ability->execute( array( 'comment_id' => $comment_id ) );

		// Restore original user context immediately.
		wp_set_current_user( $original_user );
```

b. For a more permanent architectural fix, we could add a method like `execute_as_system()` to the `Abstract_Ability` base class. This would allow us to safely bypass capability checks for internal system processes (like those triggered automatically by core hooks) without exposing that bypass to the REST API.

I was originally diving into this because I'm looking to pick up issue #509 (Add sorting and filtering plus additional labelling). I'd be happy to submit a PR to patch this permission bug first, based on whatever approach you prefer, so the moderation flow is totally stable before I start adding the filters to it!

1. Install and activate the WordPress AI plugin.
2. Configure any API (Google Gemini in my case) connector with a valid API key.
3. Create a new user with Subscriber role.
4. Log in as the user and comment on any existing post.
5. Observe the error notice on the screen.

Image

### Environment info

- WordPress version: WordPress 7.1-alpha-62299
- AI plugin version: Version 0.8.0
- Active Theme: Twenty Twenty-Five
- Connector used: Google Gemini
- Browser: Opera GX (Chromium)
- Device/OS: macOS

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

dkotter:
## What?

Closes #515 

Add a new `analyze_comment_by_id` method to the `Comment_Analysis` class and use that whenever we want to analyze a comment.

## Why?

If we directly use the `Comment_Analysis` Ability to analyze a comment, it will require the `permission_callback` to pass which requires a user to be logged in and have the `moderate_comments` capability. While this is great and we want that in place, this means we can't use this class to analyze a newly added comment from a non-logged in user or a non-admin/editor user.

## How?

- Adds a new `analyze_comment_by_id` method to the `Comment_Analysis` class
- When the `Comment_Analysis` Ability is called, the execution callback will now use this method (after passing the permission check)
- When we analyze a newly added comment, instead of directly calling the Ability (which will trigger the permission check) instead call this new `analyze_comment_by_id` method. Since we hook this to `wp_insert_comment`, WordPress has already run it's own checks so no need for us to have a permission check here

### Use of AI Tools

AI assistance: Yes
Tool(s): Cursor
Model(s): GPT-5.5
Used for: Analyzing the problem and suggesting some approaches. I considered the approaches and settled on what I felt was right and AI executed on that plan. Final review and testing by me

## Testing Instructions

1. Pull down this PR
2. Turn on the Comment Moderation experiment
3. Ensure you have an AI Connector in place
4. Ensure comments are allowed on your site
5. As a logged-in user, go to the front-end and leave a comment. Ensure this comment is saved
6. Go to the `Comments` screen in the admin and ensure you see this comment and you see Sentiment and Toxicity values
7. Log out
8. Go to a post and leave a comment
9. Ensure that comment goes through
10. Log back in and find that comment on the `Comments` screen and ensure you see this comment and you see Sentiment and Toxicity values
11. From that screen, manually analyze one or more comments and ensure that still works

## Changelog Entry

> Fixed - Ensure comment moderation works properly for comments left by non-logged in users and users without the `comment_moderation` capability

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-516-b62cf7f0b9816f9af48d9fdc7edb7af8147716da.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Comment_Moderation/Comment_Analysis.php
+++ b/includes/Abilities/Comment_Moderation/Comment_Analysis.php
@@ -79,7 +79,22 @@ class Comment_Analysis extends Abstract_Ability {
 	 * @return array{comment_id: int, toxicity_score: float, sentiment: string}|\WP_Error The result of the ability execution.
 	 */
 	protected function execute_callback( $input ) {
-		$comment_id = absint( $input['comment_id'] ?? 0 );
+		return $this->analyze_comment_by_id( absint( $input['comment_id'] ?? 0 ) );
+	}
+
+	/**
+	 * Analyzes a comment by ID from trusted internal plugin flows.
+	 *
+	 * This method intentionally does not perform user capability checks. User-invoked
+	 * ability execution must go through execute(), which runs permission_callback().
+	 *
+	 * @since x.x.x
+	 *
+	 * @param int $comment_id Comment ID.
+	 * @return array{comment_id: int, toxicity_score: float, sentiment: string}|\WP_Error The result of the analysis.
+	 */
+	public function analyze_comment_by_id( int $comment_id ) {
+		$comment_id = absint( $comment_id );
 
 		if ( ! $comment_id ) {
 			return new WP_Error(
@@ -201,6 +216,24 @@ class Comment_Analysis extends Abstract_Ability {
 	 * @return array{toxicity_score: float, sentiment: string}|\WP_Error The analysis result.
 	 */
 	private function analyze_comment( string $content, string $author ) {
+		/**
+		 * Filters the comment analysis result before calling the AI provider.
+		 *
+		 * Returning an array short-circuits the provider call. This is primarily useful for tests
+		 * and integrations that provide their own comment analysis implementation.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param array{toxicity_score: float, sentiment: string}|null $result  Precomputed analysis result.
+		 * @param string                                              $content Comment content.
+		 * @param string                                              $author  Comment author name.
+		 */
+		$pre_result = apply_filters( 'wpai_comment_analysis_result', null, $content, $author );
+
+		if ( is_array( $pre_result ) ) {
+			return $this->sanitize_analysis_result( $pre_result );
+		}
+
 		$prompt = sprintf(
 			"Comment by %s:\n\"\"\"%s\"\"\"",
 			$author,
@@ -229,20 +262,7 @@ class Comment_Analysis extends Abstract_Ability {
 			);
 		}
 
-		// Validate and sanitize the response.
-		$toxicity_score = isset( $parsed['toxicity_score'] )
-			? max( 0, min( 1, (float) $parsed['toxicity_score'] ) )
-			: 0;
-
-		$valid_sentiments = array( 'positive', 'negative', 'neutral' );
-		$sentiment        = isset( $parsed['sentiment'] ) && in_array( $parsed['sentiment'], $valid_sentiments, true )
-			? $parsed['sentiment']
-			: 'neutral';
-
-		return array(
-			'toxicity_score' => $toxicity_score,
-			'sentiment'      => $sentiment,
-		);
+		return $this->sanitize_analysis_result( $parsed );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


Intenzi:
@dkotter I have tested this PR locally and it works perfectly! Posting a comment no longer errors out for Subscribers and the comment is correctly flagged and dropped into the moderation queue. I was able to verify that anonymous comments are also functional with this.

I also really like this approach, keeping it scoped to the specific class instead of altering `Abstract_Ability` makes total sense for right now. And adding the filter for the unit tests is a great touch. Thanks for the quick fix!

**One quick thought that crossed my mind while testing this:** Since this allows the background AI to run on front-end submissions, should we be worried about anonymous comment spam causing API token exhaustion (essentially a Denial of Wallet)? I don't want to distract from this PR since it perfectly fixes the immediate bug.

At a minimum, if a standard spam filter (like Akismet) intercepts the comment first and flags it, such that `'spam' === $comment->comment_approved`, the AI shouldn't run. We might also want to consider a toggle for whether AI runs on guest comments at all.

Would you like me to open a new PR to patch that early exit check as it is a simple fix, or should I open an Issue to track this as a broader discussion about mitigating anonymous comment spam? 

Either way, I'm going to start working on #509 in the morning! 🌞

dkotter:
Thanks for the review @Intenzi!

> Since this allows the background AI to run on front-end submissions, should we be worried about anonymous comment spam causing API token exhaustion (essentially a Denial of Wallet)?

It's a good question and open to thoughts on ways to mitigate this. It's essentially the reason we have the permission check in the first place, to limit the scope of who can initiate an AI request. But in wanting to support non-logged in comments (which I imagine are a good percentage of comments) not sure there's a great way to allow those to be scanned but not allow non-legitimate comments (other than maybe some sort of built-in rate limiting).

> At a minimum, if a standard spam filter (like Akismet) intercepts the comment first and flags it, such that 'spam' === $comment->comment_approved, the AI shouldn't run

Yeah, I like this thought, if a comment has already been marked as spam or marked for moderation, likely no reason for us to analyze that comment.

> We might also want to consider a toggle for whether AI runs on guest comments at all.

We've so far taken the approach of limiting the amount of settings we have but this is likely a decent middle ground approach for the first problem. Maybe have this default to being on but a site can toggle it off if they don't want to analyze guest comments (or they start getting bombarded by spam and it's using too many tokens).

> Would you like me to open a new PR

Always open to review PRs so I think that's a good next step here for any of the points you've raised if you have interest

jeffpaul (approved):
Works well in testing, thanks!

dkotter:
I'm going to go ahead and merge this as it at least makes this comment moderation experiment work as expected. I think we have some good feedback we can follow up on though for the next release.

PR #518, opened 2026-05-07
Add sorting and filtering plus additional labelling to Comment Moderation

### LINKED ISSUE #509: Add sorting and filtering plus additional labeling to Comment Moderation

Intenzi:
## What?

Closes #509 

- Added Sorting, Filtering in Comments screen for Sentiment, Toxicity.
- Updated Activity Widget -> Recent Comments section to include Sentiment, Toxicity pills for each comment.
- Added tests to validate the same.

## Why?

Presently there is no UX to sort and filter in Comments screen for the new Comment Moderation Experiment 's Sentiment and Toxicity Analysis. This acts as a UX enhancement.
By displaying the pills in the Admin dashboard as well for the Activity widget's Recent Comments section, we maintain consistency in our User Experience.

## How?

*   **Sorting and Filtering Logic**:
    *   Adds `wpai_sentiment` and `wpai_toxicity` to the sortable columns via the `manage_edit-comments_sortable_columns` filter.
    *   Hooks into `restrict_manage_comments` to inject custom dropdowns for Sentiment and Toxicity filtering.
    *   Updates the `pre_get_comments` logic (via `handle_sorting_and_filtering`) to intercept list table queries and adds relevant meta query for filtering.

*   **Dashboard Integration & Scoping of Pills**:
    *   Hooks into `get_comment_excerpt` to inject AI sentiment and toxicity "pills" into the Dashboard's Activity widget.
    *   Uses `get_current_screen()` to scope UI injection to the dashboard only.
    *   Introduces the `wpai_comment_moderation_show_dashboard_pills` filter, allowing third-party developers to opt-out of dashboard pill rendering if needed.
    *   Optimises `add_inline_styles` hooked to `admin_head-edit-comments.php` and `admin_head-index.php` for conditional rendering to ensures that the styling is only rendered on the exact required screens.

*   **Automated Testing**:
    *   Adds new integration tests in `Comment_ModerationTest.php` that validate the `meta_query` logic for sentiment and medium toxicity ranges.
    *   Includes visibility tests for the dashboard pills to ensure they only appear on the correct screens and respect the opt-out filter.

### Use of AI Tools

AI assistance: Yes
Tool(s): Antigravity
Model(s): Gemini 3.1 Pro, Gemini 3 Flash
Used for: Gemini built an initial solution, test cases were written via the same. I have manually reviewed, refactored the code and decided on the approach to be taken. 

## Testing Instructions
1. Pull down this PR
2. Turn on the Comment Moderation experiment
3. Ensure you have an AI Connector in place
4. Ensure comments are allowed on your site
5. As a logged-in user (Admin/Editor), go to the front-end and leave a comment. Ensure this comment is saved.
6. Go to the Comments screen in the admin and ensure you see this comment and you see Sentiment and Toxicity values
7. Make multiple comments/atleast one of each sentiment and toxicity.
8. Check Admin Dashboard -> Activity Widget to observe the Recent Comments showing relevant pills.
9. Check Comments Screen for sorting toggles present on Sentiment and Toxicity columns.
10. Click on them to test the same.
11. Check the Comments Screen for Filter Dropdowns for Sentiment and Toxicity.
12. Select any and press the Filter button to observe the filtered out comments.

You may also run the added tests to validate the functioning.

## Screenshots or screencast

| Before | After |
| --- | --- |
| image | image |

| Before | After |
| --- | --- |
| image | image |

Screenshot showcasing filtering by Sentiment:
image

## Changelog Entry

> Added - Sorting, Filtering in Comments screen by Toxicity, Sentiment with labelling in Admin Dashboard for Comments to include relevant pills of Toxicity and Sentiment.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-518-87523929378c2bbf75fb788840048f9d703094a7.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (dismissed):
Tests well, thanks!

_Will wait on @dkotter code review before getting this merged in._

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -593,13 +804,18 @@ public function enqueue_assets( $hook_suffix ): void {
 	 * @since x.x.x
 	 */
 	public function add_inline_styles(): void {
+		$current_hook = current_action();
+
+		$is_comments_screen  = 'admin_head-edit-comments.php' === $current_hook;
+		$is_dashboard_screen = 'admin_head-index.php' === $current_hook;
 		?>
 		<style>
+			<?php if ( $is_comments_screen ) : ?>
```
dkotter: While this works, is a bit messy combing PHP and CSS. I'm thinking a better approach here might be to just scope the CSS based on a unique class for the page we're on. For instance, looks like the dashboard widget has a class `dashboard-comment-wrap` that we could likely use. And the comments page has `edit-comments-php`.

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -593,13 +804,18 @@ public function enqueue_assets( $hook_suffix ): void {
...
 			.column-ai_sentiment,
 			.column-ai_toxicity {
```
dkotter: Not a new issue here but seems these classes never got updated so this style is never applied. Should be `.column-wpai_sentiment` and `.column-wpai_toxicity`

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -252,6 +263,206 @@ public function add_columns( $columns ): array {
...
+	public function add_filter_dropdowns(): void {
+		$current_sentiment = isset( $_GET['wpai_sentiment'] ) ? sanitize_text_field( wp_unslash( $_GET['wpai_sentiment'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$current_toxicity  = isset( $_GET['wpai_toxicity'] ) ? sanitize_text_field( wp_unslash( $_GET['wpai_toxicity'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
+		// Sentiment Dropdown.
+		$sentiments = array(
+			'positive' => __( 'Positive', 'ai' ),
+			'neutral'  => __( 'Neutral', 'ai' ),
+			'negative' => __( 'Negative', 'ai' ),
+		);
```
dkotter: Just leaving as a comment to prompt conversation but right now we reference these sentiment values in multiple places. Might be worth moving these to public class variable so we have a single place we manage these

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -252,6 +263,206 @@ public function add_columns( $columns ): array {
...
+			<?php endforeach; ?>
+		</select>
+
+		<?php
+		// Toxicity Dropdown.
+		$toxicities = array(
+			'low'    => __( 'Low Toxicity (<40%)', 'ai' ),
+			'medium' => __( 'Medium Toxicity (40%-69%)', 'ai' ),
+			'high'   => __( 'High Toxicity (>=70%)', 'ai' ),
+		);
```
dkotter: This is another one that may make sense to make a class variable for easier long-term maintenance

dkotter (changes requested):
This looks really good, nice work here. Left a few comments but also seeing a couple issues when testing.

If I have a comment without a sentiment or toxicity score and I filter by Positive sentiment or Low toxicity, that comment shows. Seems like this should be excluded?

Image

If I end up combining filters to where no results are found, I end up on a screen where all the filters disappear:

Image

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -252,6 +263,206 @@ public function add_columns( $columns ): array {
...
+		if ( ! is_admin() || ! function_exists( 'get_current_screen' ) ) {
+			return;
+		}
+
+		$screen = get_current_screen();
+		if ( ! $screen || 'edit-comments' !== $screen->id ) {
+			return;
+		}
+
+		$meta_query = $query->meta_query->queries;
```
dkotter: Any reason to use this over `$query->query_vars['meta_query']`?

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -252,6 +263,206 @@ public function add_columns( $columns ): array {
...
+
+		return $comment_excerpt . $pills;
+	}
+
+	/**
+	 * Adds filter dropdowns for sentiment and toxicity.
+	 *
+	 * @since x.x.x
+	 */
+	public function add_filter_dropdowns(): void {
```
dkotter, line 438: This is hooked to `restrict_manage_comments`, which I'm not sure where all that fires. Any need for a screen guard here like you have in other methods?

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -252,6 +263,206 @@ public function add_columns( $columns ): array {
 
 		return $new_columns;
 	}
+	/**
```
dkotter, line 375: Need an extra line return above this

dkotter (commented):
We have good unit tests here but would be great to update our E2E tests to have them verify the filters and sorts work

includes/Experiments/Comment_Moderation/Comment_Moderation.php
Intenzi: I agree with you, its utilisation in multiple files is visibly being hardcoded. I can centralize this as a refactor commit.

includes/Experiments/Comment_Moderation/Comment_Moderation.php
Intenzi: No specific reason, I will patch this to use the standard approach. Thanks for catching that!

includes/Experiments/Comment_Moderation/Comment_Moderation.php
Intenzi: Yes! I missed out on the screen guard over here. It is necessary as the hook is tied to `WP_Comments_List_Table` which can be utilised by plugins.  Thankyou for catching that!

Intenzi:
Thankyou for the PR review, testing! @dkotter 

> If I have a comment without a sentiment or toxicity score and I filter by Positive sentiment or Low toxicity, that comment shows. Seems like this should be excluded?

I am unable to reproduce this issue. Can you check if the comment has metadata present in the database?

> If I end up combining filters to where no results are found, I end up on a screen where all the filters disappear:

I did notice this. I've implemented the filters via `restrict_manage_comments` and explored making the filters persistent when results are empty, but since `WP_List_Table` bails out on empty states, I opted to stick with the native WordPress behavior to keep the UI consistent with Core's UX. I could not find a clean alternative to this approach and would like to know how we should proceed for this.

dkotter:
> I am unable to reproduce this issue. Can you check if the comment has metadata present in the database?

Ah, yep, seems that was it. Which does bring up a question on why those values weren't showing in the columns but not related to the filtering.

> but since WP_List_Table bails out on empty states, I opted to stick with the native WordPress behavior to keep the UI consistent with Core's UX

Hmm.. I guess if this is core behavior (to have _all_ filters hide if nothing matches) then I'm fine with sticking with it. Seems odd that's the approach though.

Intenzi:
@dkotter I have implemented the feedback changes and added e2e test for filtering and sorting of comments. PR is ready for re-review 😄 

> Ah, yep, seems that was it. Which does bring up a question on why those values weren't showing in the columns but not related to the filtering.

I have noticed it occurs when a comment fails to be analyzed and then it ends up still containing the metadata (as an empty string value) instead of having no metadata.

dkotter (approved):

PR #523, opened 2026-05-07
Gracefully handle missing AI provider configuration

### LINKED ISSUE #182: Graceful degradation when no AI provider is configured
Ensure all AI Experiments fail gracefully when no AI provider or credentials are configured. Users should receive clear, actionable guidance rather than errors, broken UI, or silent failures.

This applies to both editor-side interactions and admin settings.

Acceptance criteria:

* No experiment triggers fatal errors or broken UI without a provider
* Clear messaging explains why an action is unavailable
* Users are guided toward configuring a provider where appropriate
* Behavior is consistent across all experiments

Takshil-Kunadia:
## What?

Closes https://github.com/WordPress/ai/issues/182

Adds graceful degradation across all AI experiments and features when no AI provider/connector is configured. Users now receive clear, actionable guidance instead of failed API calls or silent failures.

## Why?

When no AI provider is configured, clicking any AI action button (Generate Title, Generate Alt Text, etc.) triggers a request which generated a Error Notice. Comment Moderation also silently fails on new comment. There is no guidance pointing users toward configuring a provider. This PR ensures users understand why an action is unavailable and know exactly where to go to fix it.

## How?

Implemented a centralized provider availability architecture shared across all AI experiments.

- Added helper fns which expose provider state globally by injecting provider availability in window object, avoiding per-experiment duplication.
- Added shared frontend guards that prevents requests when no provider is configured and surfaces a consistent notice with a “Manage Connectors” action.
- Added equivalent PHP-side guards for Comment Moderation flows, including early returns, redirect-based notices, and prevention of queued processing without credentials.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 4.6
Used for: Implementation across multiple files following architectural decisions made collaboratively. All code was reviewed, iterated on, and validated by me.

### Prerequisites
Ensure no AI connector is configured (remove API keys from Settings > Connectors). Enable all experiments in Settings.

### Block Editor (Posts > Add New)

1. Click Generate in the title toolbar > snackbar notice appears with "Manage Connectors" link
2. Open Excerpt panel > click Generate > same notice
3. Click the Summarize button in Post Status > same notice
4. Open Categories panel > click Suggest > same notice
5. Open Meta Description panel > click Generate > same notice
6. Click Generate Review Notes > same notice
7. Click Refine from Notes (requires existing notes) > same notice
8. Add an image block > select it > click Generate Alt Text in inspector > same notice
9. Add an image block > click Generate Image in placeholder > same notice

### Media Library
10. Click an image > attachment details > click Generate button > inline red error text
11. Select multiple images > Bulk Actions > Generate Alt Text > Apply > admin notice at top

### Comments
12. Select comments > Bulk Actions > Analyze Sentiment and Toxicity > Apply > server-rendered error notice with connectors link
13. Hover a comment > click Analyze Sentiment and Toxicity > same notice

### With provider configured
14. Configure any AI connector with a valid API key
15. Repeat steps 1-14 > all features should work normally, no degradation notices

## Screenshots or screencast

Experiments:-
image

Comments Mod:-
image

Alt text from Media Library
image

## Changelog Entry

> Changed - Graceful degradation when no AI provider is configured. All experiments now show actionable guidance directing users to configure an AI Connector, instead of failing silently or showing generic errors.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-523-289b02eddcf9500b02ae603586a5ab61ee91e9cb.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Main.php
```diff
@@ -187,6 +190,20 @@ public function register_ability_category(): void {
...
 	}
 
+	/**
+	 * Outputs provider data as a global JavaScript object.
+	 *
+	 * Every experiment can subscribe to Provider data.
+	 *
+	 * @since x.x.x
+	 */
+	public function print_provider_data(): void {
```
dkotter: Not sure I love the approach of blindly outputting javascript data like this. We do already have an `Asset_Loader` class that each feature uses to load any required JS/CSS. I'm wondering if we can utilize that to load this data, that way it's only ever loaded when needed

includes/Experiments/Comment_Moderation/Comment_Moderation.php
```diff
@@ -532,6 +544,33 @@ public function add_inline_action( $actions, $comment ): array {
...
+	 * Shows an admin notice if the inline action is attempted without a provider.
+	 *
+	 * @since x.x.x
+	 */
+	private function show_missing_provider_notice(): void {
+		$connectors_url = get_provider_availability_data()['connectorsUrl'];
+		$notice_message = sprintf(
+			/* translators: %s: Link to connectors settings page. */
+			__( 'This feature requires a valid AI Connector to function properly. Please set up a provider to use this feature in %s.', 'ai' ),
+			'<a href="' . esc_url( $connectors_url ) . '" target="_blank">' . esc_html__( 'Settings → Connectors', 'ai' ) . '</a>'
```
dkotter: I would probably not have this open a new window
```suggestion
			'<a href="' . esc_url( $connectors_url ) . '">' . esc_html__( 'Settings → Connectors', 'ai' ) . '</a>'
```

src/utils/provider-status.ts
```diff
@@ -0,0 +1,77 @@
...
+
+	( dispatch( noticesStore ) as any ).createInfoNotice(
+		__(
+			'This feature requires an AI Connector to function properly.',
+			'ai'
+		),
+		{
+			id: noticeId,
+			isDismissible: true,
+			type: 'snackbar',
```
dkotter: Not sure we'd want a snackbar notice here as opposed to a normal admin notice; otherwise we now have some errors that output as admin notices and some as snackbars. Seems ideal to standardize that

src/utils/provider-status.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Checks whether an AI provider is available and creates a notice to guide them if not.
+ *
+ * @param noticeId Unique ID for the notice to prevent duplicates.
+ * @return True if a provider is available, false otherwise.
+ */
+export function ensureProvider( noticeId: string ): boolean {
+	const providerStatus = getProviderData();
+
+	if ( providerStatus.hasProvider ) {
```
dkotter, line 41: So this check only looks for a configured provider, it does not check if the provider will work for the request. That may be fine (the request will fail at a later stage and output an error message) but could argue that if we want to add this check, maybe we want to expand the check to ensure we have a provider configured for the specific request

dkotter (changes requested):

includes/Main.php
Takshil-Kunadia: Updated. Moved to `Asset_Loader::add_global_data()` generic method that stores data and loads it via `wp_add_inline_script` onto plugin script enqueued.

src/utils/provider-status.ts
Takshil-Kunadia: Removed `type: 'snackbar'` and `explicitDismiss`. Now renders as a standard admin notice, consistent with other error notices in the plugin. My idea was that its not an error per se, its a step user missed in setting up the plugin. Snackbar would be a softer nudge in the right direction compared to the red highlights of error.

src/utils/provider-status.ts
Takshil-Kunadia: A possible approach is we could dynamically iterate `CapabilityEnum::cases()` and call `$prompt_builder->is_supported()` for each, producing a capabilities map like `{ text_generation: true, image_generation: false, ... }`. The frontend `ensureProvider()` would accept an optional capability parameter which when provided, it checks the map and shows a specific message.

Happy to tackle this in this PR or as a follow-up :)

includes/Experiments/Comment_Moderation/Comment_Moderation.php
Takshil-Kunadia: Fixed. removed target="_blank".

Takshil-Kunadia:
Thanks for the review @dkotter 🙇 . Addressed the feedbacks, please have a look whenever you get the chance. I'll add tests as well.

dkotter (dismissed):
Tested and looks good to me, thanks for the work here! As mentioned, would be ideal to get E2E tests in place but once that's done this can be merged in

src/utils/provider-status.ts
dkotter: I think we leave this as-is for now and can look at this in a follow-up if we think it's worth having

Takshil-Kunadia:
@dkotter updated the tests and added tests covering the changes.

tests/e2e-request-mocking/e2e-request-mocking.php
```diff
@@ -12,9 +12,64 @@
...
 // Mock the HTTP requests and provide known responses.
 add_filter( 'pre_http_request', 'ai_e2e_test_request_mocking', 10, 3 );
 
+/**
+ * Registers REST endpoints for seeding and clearing dummy AI provider credentials.
+ *
+ * POST /ai-e2e/v1/credentials/seed  — sets a dummy provider API key.
+ * POST /ai-e2e/v1/credentials/clear — removes it.
+ */
+function ai_e2e_register_credentials_endpoint() {
```
dkotter, line 27: Can you explain the need for these endpoints? We already have helper functions that clear or set credentials using the UI so just wondering what benefit these provide over using that existing functionality?

tests/e2e-request-mocking/e2e-request-mocking.php
```diff
@@ -12,9 +12,64 @@
...
+	);
+}
+
+/**
+ * Seeds a dummy provider key so has_ai_credentials() returns true.
+ *
+ * @return WP_REST_Response
+ */
+function ai_e2e_seed_credentials() {
+	update_option( 'connectors_ai_openai_api_key', 'sk-fake-e2e-test-key' );
```
dkotter: Currently we use `valid-api-key` as our valid test key and `invalid-api-key` as our invalid test key

tests/e2e-request-mocking/e2e-request-mocking.php
Takshil-Kunadia: The REST endpoints exist because Playwright's `afterAll` and `globalSetup` hooks only receive `requestUtils`.

I tried using the existing `clearConnectors` in `afterAll`, but Playwright throws:

> "context" and "page" fixtures are not supported in "afterAll" since they are created on a per-test basis.

Specifically we need REST for two cases:

1. `global-setup.ts` - seeds credentials before any browser launches (only `requestUtils` available, no `page`)
2. `afterAll` hooks in `settings.spec.js` and `dashboard.spec.js` — these specs clear connectors during their run and need to restore credentials for subsequent specs, but `afterAll` only receives `requestUtils`

tests/e2e-request-mocking/e2e-request-mocking.php
Takshil-Kunadia: Ahh Thanks for that. Updated the key to `valid-api-key` to match the existing test convention.

tests/e2e-request-mocking/e2e-request-mocking.php
Takshil-Kunadia: Before this PR, tests didn't need credentials since there was no frontend provider check. Now that `ensureProvider()` blocks AI actions when no provider is configured, every experiment test needs a valid API key. The global-setup seeds it upfront, and specs that clear connectors (settings, dashboard) re-seed in their `afterAll` via these REST endpoints.

dkotter (approved):

PR #535, opened 2026-05-12
Add an `.npmrc` config file

dkotter:
## What?

Adds in an `.npmrc` config file and updates our minimum node and npm versions.

## Why?

Main reason here is to set a minimum release age required to install an npm package to help avoid the recent plethora of supply chain attacks. This can be done with the `min-release-age` config option in `.npmrc`. This option is only supported on newer versions of npm so this PR also bumps our minimum node version to 24 and npm version to 11.

Note this is set to match what was done in https://github.com/WordPress/gutenberg/pull/78191/

## How?

- Add an `.npmrc` config file with a few config options set
- Remove `package-lock.json` and `node_modules` and re-run `npm install` to rebuild the lock file

### Use of AI Tools

None

## Testing Instructions

Ensure all tests pass on this PR. If desired, manual testing can be done of all existing features

## Changelog Entry

> Developer - Add an `.npmrc` config file.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-535-3eac71291e0197fb2a181de64066937ef31f2ed4.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


.nvmrc
```diff
@@ -1 +1 @@
-22
+24
```
justlevine: I'm worried about doing this before https://github.com/WordPress/gutenberg/issues/72973 . In general those packages don't have the greatest release hygiene and now that we're using [experimental (and fragile) packages like `@wordpress/ui`](https://github.com/WordPress/gutenberg/tree/trunk/packages/ui#ui:~:text=This%20package%20is%20still%20experimental.%20%E2%80%9CExperimental%E2%80%9D%20means%20this%20is%20an%20early%20implementation%20subject%20to%20drastic%20and%20breaking%20changes.) , I'd recommend increased care.

Related:
- https://github.com/WordPress/gutenberg/pull/74310#issuecomment-3824046485

justlevine (commented):
the `.npmrc` config LGTM, but I think we should just bump NPM and keep `node` as is.

.nvmrc
dkotter: Fair point though two thoughts:

1. GitHub is switching to Node24 as the default June 2nd so this gets us ahead of that
2. Is a bit trickier having a newer node version but an older npm version though that's manageable. For instance, with this set to 22, if I run `nvm use` locally it will automatically pull in npm v10. With this set to 24, it will automatically pull in npm v11 (which is the npm version we want). So anyone developing locally will potentially have to manually install the right npm version

justlevine: > GitHub is switching to Node24 as the default June 2nd so this gets us ahead of that

[My understanding](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-versioned-actions) is this is only affects the runner (and actions themselves), but anything that uses our .nvmrc will still work as intended. (If it didn't we'd be seeing deprecation notices, but I thought we took care of them all in a previous sweep).

If I'm incorrect, then I assume it also will affect the gutenberg repo too though if they still cant hit the deadline, we can also update our workflows with `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` 

> So anyone developing locally will potentially have to manually install the right npm version

As someone who calls `nvm use` multiple times an hour this one gives me bigger pause. 🤔

@dkotter - waddya think about us reverting npm to `>=10` too, and only committing the new `.npmrc` (new version props included) in this PR?

It would still allow contributors who are using v11 to benefit from `min-release-age`, although a better defence for the problem for _all_ versions is to use `npm ci` unless you're intentionally trying to update the lockfile. And we get all the other .npmrc stuff immediately. It also looks like [dependabot doesn't use .nvmrc anyway](https://github.com/dependabot/dependabot-core/issues/4808), so that should benefit from the `min-release-age` on transient deps immediately too.

Leaving it at `>=10` would mean that contributors _intentionally changing the lockfile_ still need to remember to `npm i -g npm` or `nvm install ---latest-npm` or whatever the notice says, but that's a much smaller surface area for us to police...🤔

.npmrc
```diff
@@ -0,0 +1,5 @@
+engine-strict = true
+lockfile-version = 3
+min-release-age = 1
+prefer-dedupe = true
+save-exact = true
```
aduth: If this is following Gutenberg, it's worth noting that this is being proposed for removal in https://github.com/WordPress/gutenberg/pull/78196 . That being said, the logic for using `save-exact` might be different between Gutenberg and this project. I could see an argument for using `save-exact` here even if it doesn't make sense for Gutenberg. But I figured I'd raise in case this is copied because it's what's in Gutenberg.

dkotter: This was mostly copy/pasted from Gutenberg so that's good to know. Likely worth us removing as well then

dkotter: Removed in https://github.com/WordPress/ai/pull/535/commits/8ea40d46128e1ed17060d9bc2863182722e96fea

package-lock.json
```diff
@@ -165,22 +165,22 @@
 			}
 		},
 		"node_modules/@babel/core": {
-			"version": "7.29.0",
-			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.29.0.tgz",
-			"integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
+			"version": "7.25.7",
+			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.25.7.tgz",
+			"integrity": "sha512-yJ474Zv3cwiSOO9nXJuqzvwEeM+chDuQ8GJirw+pZ91sCGCyOZ3dJkVE09fTV0VEVzXyLWhh3G/AolYTPX7Mow==",
```
justlevine, line 170: Maybe run an `npm audit fix` too, to be safe? (if you did already disregard, just wasn't mentioned in the PR desc)

.nvmrc
dkotter: Reverted back to the existing node and npm versions. I then set my node version to 22.22.2 and npm version to 11.14.1 and re-built the lock file and pushed to this PR, so the only changes now are the new `.npmrc` file and lock file changes

package-lock.json
dkotter: So I just fixed most of these in #503. The only remaining issues are `showdown`, which unfortunately hasn't had a release since April 2022 so can't override that. And then `ajv` which has a fix but there's multiple dependencies that all install different versions so the amount of overrides needed to fix is pretty large (so I didn't tackle that in #503)

dkotter: All that to say, `npm audit fix` doesn't fix anything. `npm audit fix --force` will fix the `ajv` issues but it does it by downgrading packages we depend on

justlevine: Not concerned about fixing existing alerts here, just want to prevent a flood of dependabot PRs for any transitive deps that got downgraded. non-security updates will get grouped, so if you ran `npm audit fix` after your `rm -rf... npm install` then we're good 👍

justlevine (approved):

PR #537, opened 2026-05-12
Fix AI credentials detection

### LINKED ISSUE #536: has_ai_credentials returns false if key is configured via constant
### Description

Instead of setting the API_KEY for the provider using the GUI and a database option, one can also set the `env` value or a constant in `wp-config.php`:
```
// wp-config.php
define( 'MITTWALD_API_KEY', 'sk-5S...' );
```

This works well, the connectors screen shows this information and displays the provider as connected, all AI requests succeed, the models can fetched, and so on.

In _Settings › AI_, where you can activate the experiments, the experiments can be activated and with the new [Developer mode model controls](https://make.wordpress.org/ai/2026/05/11/whats-new-in-ai-0-9-0/#:~:text=Developer%20Mode%20model%20controls) form 0.9.0 the models can be selected.

The problem is, that the following error is still displayed on the top of the page:

> The AI plugin requires a valid AI Connector to function properly. Verify you have one or more AI Connectors configured.
Image

1. set provider API Key using content or env:
   ```
   // wp-config.php
   define( 'PROVIDER_API_KEY', 'sk-5S...' );
   ```
1. Validate that the key is configured correctly by checking _Settings › Connectors_
1. Go to _Settings › AI_
1. Error shown despite key is configured correctly

_No response_

### Environment info

- WordPress 7.0-RC3
- AI-Plugin Version 0.9.0
- DDEV and mittwald hosting project

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

LukasFritzeDev:
## What?

Closes https://github.com/WordPress/ai/issues/536

- add shared API key detection that checks environment variables, PHP constants, and stored settings
- use the shared detection for both the credentials helper and dashboard widget connection state

## Why?

if the provider API key is set up using constant in `wp-config.php` or an ENV (instead of providing the key via the connectors settings page)

* _Settings › AI_ shows an error saying “The AI plugin requires a valid AI Connector to function properly.”
* the Dashboard widget shows ❌ instead of ✅ for the connector

## How?

* Add helper function `has_api_key_configured` that checks environment variable, PHP constant and database option.
* Use the helper in `has_ai_credentials` and `AI_Status_Widget` to get the correct state information 

### Use of AI Tools

AI assistance: Yes
Tool(s): IDE Code completion
Model(s): Devstral-Small-2-24B-Instruct-2512
Used for: IDE Code completion

## Testing Instructions

* Run e2e tests
* manual tests
  * Setup provider API key as described in issue https://github.com/WordPress/ai/issues/536
  * Go to _Dashboard_: Connector should show connected state
  * Go to _Settings › AI_: said error must not be shown

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| 591160483-e9897f93-659f-4934-843e-409e63390d42 |  No error     |
| Bildschirmfoto 2026-05-12 um 20 23 00 | Bildschirmfoto 2026-05-12 um 20 42 29 |

## Changelog Entry

Fixed connector status in settings page an dashboard widget if key is configured using env or constant

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-537-728f98cf38421469df7c8945e0cc81017c557318.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


LukasFritzeDev:
I’m not sure how to test this in the e2e tests. Sure, we could add a test spec for this, but this would require to set the config value in .wp-env.test.json and this would break other tests. Should set up an additional env with the value and run a seperate test script for this? Do you have any other ideas?

justlevine:
> I’m not sure how to test this in the e2e tests.... Do you have any other ideas?

I don't think we need explicit e2e testing for this, since the outcome - behavior when there are/arent creds - hasn't changed. What's changing here is the internal handling for where credentials can be sourced from, so PHPUnit (Integration) testing should be enough IMO.

gziolo:
Why don't you check directly on the AI provider instance whether it's configured? The connectors for AI providers gets registered automatically based on the info stored in AI provider registry. In fact, the names for the env variable and PHP constant are standardized there.

dkotter:
> I’m not sure how to test this in the e2e tests. Sure, we could add a test spec for this, but this would require to set the config value in .wp-env.test.json and this would break other tests. Should set up an additional env with the value and run a seperate test script for this? Do you have any other ideas?

I agree with @justlevine here that we don't really need an E2E test for this. That said, we do load a test plugin in our test environment and so you could use that to help with this. As an example of something I've done before, you could add something like this in the `e2e-request-mocking.php` file:

```php
// Set the OpenAI API key as a constant when override param is set.
add_action(
	'init',
	static function () {
		if (
			isset( $_GET['ai_api_key_override'] ) &&
			! defined( 'OPENAI_API_KEY' )
		) {
			define( 'OPENAI_API_KEY', 'valid-api-key' );
		}
	},
	1
);
```

And then in your E2E test, do the following:

1. Remove credentials from OpenAI
2. Go to the settings page and ensure the admin notice shows
3. Refresh the settings page with the query param `ai_api_key_override` added
4. Ensure the admin notice no longer shows

Could then replicate this for the dashboard widget

dkotter:
> Why don't you check directly on the AI provider instance whether it's configured? The connectors for AI providers gets registered automatically based on the info stored in AI provider registry. In fact, the names for the env variable and PHP constant are standardized there.

@gziolo Not sure I completely follow? Is this around the question about E2E tests? Or a suggestion on how the code in this PR could be better? Need to do a full review but in looking quickly, does look like it pulls the env and constant name from the Connector Registry.

As far as I know though, that registry doesn't tell us if a Connector is actually configured. Let me know if that's not correct though.

EDIT: Reading again, I think what you're referring to is the AI Client Registry, not the Connectors Registry. In that case I do think we can use that instead though it's not super different from the approach taken here

includes/helpers.php
```diff
@@ -368,6 +368,44 @@ function format_guidelines_for_prompt( array $categories, ?string $block_name =
...
+ * Checks in order: environment variable, PHP constant, database.
+ * Environment variable and constant are only checked when their
+ * respective names are provided.
+ *
+ * @since 1.0.0
+ *
+ * @param array{setting_name?: string, env_var_name?: string, constant_name?: string} $auth Authentication configuration.
+ * @return bool True if an API key is set by one of the configuration options, false otherwise.
+ */
+function has_api_key_configured( array $auth ): bool {
```
dkotter: Based on feedback from @gziolo and looking at things closer, a different approach we can take here is using the AI Client Registry. ~~I'm not super opinionated either way as Core itself uses both approaches but it is slightly less code~~ This is likely the better approach as I think about it more, see https://github.com/WordPress/wordpress-develop/blob/88612f5798c7ec0d2d94044114a2f1e8fe111897/src/wp-includes/connectors.php#L692-L697

```php
$registry        = AiClient::defaultRegistry();
$connectors      = get_ai_connectors();
$has_credentials = false;

foreach ( $connectors as $connector_id => $connector_data ) {
    if (
		! $registry->hasProvider( $connector_id ) ||
		! $registry->isProviderConfigured( $connector_id )
	) {
		continue;
	}

    $has_credentials = true;
    break;
}
```

gziolo:
@dkotter, I'm talking about production code and the fact that AI providers need to be configured before they get discovered by the connectors registry. Even when the AI provider uses db setting, it's the connector that sets everything on the AI provider instance. See:

https://github.com/WordPress/wordpress-develop/blob/7938204d20159fa863e6d34af352bb7557524354/src/wp-includes/connectors.php#L626

includes/helpers.php
justlevine: Only looked briefly, but `::isProviderConfigured()` also seems more future-compat and likely to stay robust enough for whatever Connector changes come in WP7.1.

dkotter:
> the fact that AI providers need to be configured before they get discovered by connectors registry. Even, when the AI provider uses db setting, it's the connector that sets everything on the AI provider insrance.

I _think_ we're saying the same things, apologize for the back and forth (see my [comment here](https://github.com/WordPress/ai/pull/537#discussion_r3230008511) on recommended changes that I think matches what you're saying).

If we just use `wp_get_connectors`, it never tells us what connectors are actually configured. But yes, if we use `AiClient::defaultRegistry`, that should only give us AI providers that have been configured. I'd argue this is more confusing than it needs to be and maybe having a `is_configured` key as part of the data `wp_get_connectors` returns would be nice but regardless, we can work around that here with a better approach than we currently have

LukasFritzeDev:
As suggested, I’ve refactored the function `has_api_key_configured` so that it uses `$registry→sProviderConfigured` instead of custom logic. 

When it comes to testing: 
> so PHPUnit (Integration) testing should be enough IMO.
> _https://github.com/WordPress/ai/pull/537#issuecomment-4434284589_

I’ve not yet added any tests, because I’m still not sure how to add those.

* The helper function `has_ai_credentials` seems not to be tested directly. Should we add tests for this?

* The Settings_Page evaluates the property that is generated in the filter function for `script_module_data_' . self::PAGE_SLUG` in JS. I did not find any tests for the filter function. https://github.com/WordPress/ai/blob/develop/includes/Settings/Settings_Page.php#L82-L82

* `AI_Status_Widget->render` uses the helper. https://github.com/WordPress/ai/blob/develop/includes/Admin/Dashboard/AI_Status_Widget.php#L64-L64 There are tests for this that check the status icons if the correction option is set. Tests could be added that check for the checkbox if the provider key is set via db option, constant or ENV. Would this be a good strategy?

If one of these strategies should be implemented, we would need to mock the `AiClient::defaultRegistry`. I’m not sure how to do this the best way. Any advice?

dkotter:
@LukasFritzeDev Thanks for this PR! I've added some basic unit tests and things look good to me!

dkotter (approved):

PR #540, opened 2026-05-13
Docs: Update `CONTRIBUTING.md` to Include `node and npm` version requirement detail

### LINKED ISSUE #539: Update `CONTRIBUTING.md` to Include `nvm use`

* When a new developer or contributor wants to work on the plugin, they follow the setup instructions provided in the `CONTRIBUTING.md` file.
* At the step `Install dependencies and build assets:`, the documentation currently instructs users to run the following command:
  ```bash
  composer install && npm i && npm run build
  ```
* This setup relies on the default Node.js version installed on the contributor’s machine, rather than the version officially supported by the repository.
* As a result, contributors using an unsupported Node.js version may encounter issues during `npm install` or while building assets.

* Update the setup instructions to include the `nvm use` command before installing dependencies and building assets.
* Since the repository already includes an `.nvmrc` file, running `nvm use` will automatically switch contributors to the correct Node.js version required by the project.
* This helps ensure consistency across development environments and reduces the likelihood of installation or build failures caused by incompatible Node.js versions.

```bash
  composer install && nvm use && npm i && npm run build
  ```

### PR DESCRIPTION
**Pull request by hbhalodia**
## What?

Closes https://github.com/WordPress/ai/issues/539

## Why?

- PR updates the `CONTRIBUTING.md` file to include the `nvm use`, so developers/contributors can use the specified version of node from the repo.

## How?

- Add `nvm use` command to `Install and build assets` step.

### Use of AI Tools

- None

## Testing Instructions

- When installing the plugin first time, it should use the node version that is added in the repository.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
|        |       |

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

Updated `CONTRIBUTING.md` to include `nvm use` command.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-540-c94a4a0748323398801a3a72865e3dc78ff93d1a.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### CODE REVIEW

File: CONTRIBUTING.md [hunk 1]
```diff
@@ -21,7 +21,7 @@ cd ai
 2. **Install dependencies and build assets:**
 
 ```bash
-composer install && npm i && npm run build
+composer install && nvm use && npm i && npm run build
```
dkotter: Minor thing but I'll note there are other options to set the node and npm version outside of `nvm`. While I use `nvm` myself I realize not everyone does so not sure if we want to make that a requirement here.

I'd maybe suggest instead of adding `nvm use` to the step here, we just add a call out that those looking to contribute need to ensure they have the right node and npm version and maybe suggest `nvm` as a solution for that.

hbhalodia: Thanks @dkotter for the review,

>  I'll note there are other options to set the node and npm version outside of nvm

Yes Agree on this, while I mostly use `nvm use` as well.

> I'd maybe suggest instead of adding nvm use to the step here, we just add a call out that those looking to contribute need to ensure they have the right node and npm version and maybe suggest nvm as a solution for that.

I could update the PR as per suggestion, to include that which refer `.nvmrc` for correct node version, and if using nvm to switch node versions, use `nvm use` before `npm install` step.

dkotter: Worth noting it's not just node version, it's npm version as well. These are documented in our [package.json file](https://github.com/WordPress/ai/blob/27ac41fff2f75c3dcd64c252aef11908de5a0635/package.json#L106-L107) so probably a good idea to update instructions to say something along the lines of `verify your node and npm version match the minimum requirements in package.json`

hbhalodia: Hi @dkotter, Have updated the messaging. Can you please review if anything needs to be updated/changed in the wordings, I would be happy to update that.

Thanks,

.nvmrc
```diff
@@ -1 +1 @@
-22
+22.21.1
```
hbhalodia: @dkotter, Have updated this to exact node version used in package.json. Reason being, when I initially cloned the repo, I was having node version `22.14.0`, So nvm use pointed to that version. On build got error, because min required version is `22.21.1`. Hence it should be good to have nvmrc file as the exact version.

Thanks,

dkotter: I guess my concern here is ideally people are using the latest compatible version of node. Right now the latest on the v22 branch is 22.22.3 but this will now force people to always use an older version. I think ideal is keeping this as 22 and if someone has an older version of v22 on their machine they'll need to update that for things to work.

hbhalodia: Makes sense, have not though of that. Will revert the change.

dkotter (approved):

PR #546, opened 2026-05-13
update wporg assets

### LINKED ISSUE #513: Update banner to byte beaker style
Currently the beaker style isn't uniform across the plugin. This is both in core and ai. This will need other tickets, but for now I am going to start here.

Update where possible to use the following treatment:

Image

Props to [shaunandrews](https://github.com/shaunandrews) for iteration and @jeffpaul for feedback

Image

jeffpaul:
## What?

Closes #513.

Follow up to https://github.com/WordPress/gutenberg/pull/78243#issuecomment-4443413141.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
|        |       |

## Changelog Entry

> Changed - Updated plugin banner and icons.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-546-d59a7400f45f949011199bf099fbdec702d30f9b.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter (commented):
This looks good to me but noting there are other places that use the icon that this PR doesn't update, at least:

1. Icon shown in the header of the AI settings page
2. Icon used in the block toolbar for the Content Resizing feature

dkotter:
I quickly tested locally by just copying the SVG markup into our `ai-icon.tsx` file but this new icon doesn't work great at the smaller size we show in the admin. Ends up looking blurry/pixelated, which I realize the icon does that on purpose but doesn't work great at small scale:

New AI icon on settings page

New AI icon in block toolbar

Now maybe this is a mistake on my end and there's a "better" way to pull that SVG markup into our icon file that would resolve this. But wondering if we need a solid line version of the icon to use in the admin?

jeffpaul:
@karmatosed perhaps an alt version that appears better that these lower resolutions/sizes?

karmatosed:
I tried to make it work smaller but it doesn't so lets just use the original one smaller simplified.


smallsize

jeffpaul:
@dkotter are you able to try that ^ image in the settings page and block toolbar?

dkotter:
Does look better but the old icon was 24x24 and this one is 132x184. Would be great if it was the same size or at least a 1:1 ratio, otherwise it doesn't match other icons, in particular in the block toolbar:

New AI icon in block toolbar

karmatosed:
I couldn't get exact square so maybe...
ratio

I actually think the border might be better to be a true 1:1
Frame 9

dkotter:
So this looks better though the border does make it stand out more as compared to other block toolbar icons.

The main issue that I need help on (someone that's better with SVG and CSS) is that this SVG has inline `fill` attributes and that overrides the styling that is automatically added to the block toolbar. So if you hover over the icon there, it should get a blue fill color but that's not working. Same with blocks that have previously been resized using AI, the toolbar icon should default to blue but that's no longer working. I've tried a few CSS tweaks but haven't found one that full resolves it.

Screenshot 2026-05-18 at 2 58 11 PM

Screenshot 2026-05-18 at 2 58 21 PM

dkotter:
> The main issue that I need help on (someone that's better with SVG and CSS) is that this SVG has inline fill attributes and that overrides the styling that is automatically added to the block toolbar

Tried a few other approaches and finally found one that seems to more or less fix the problem so just pushed that up. This is looking good to me now.

jeffpaul (commented):
I'd prefer there wasn't the dark square/rectangular outline around the image in the settings header and block toolbar, but if that's the best we can do in time for v1 then let's go ahead and ship this and come back to try and remove that in a follow-up release.

dkotter:
> I'd prefer there wasn't the dark square/rectangular outline around the image in the settings header and block toolbar, but if that's the best we can do in time for v1 then let's go ahead and ship this and come back to try and remove that in a follow-up release.

Messing with the SVG markup a bit, I was able to remove that border so it looks like this now:

Screenshot 2026-05-19 at 11 53 49 AM

Screenshot 2026-05-19 at 11 53 57 AM

Screenshot 2026-05-19 at 11 53 55 AM

jeffpaul:
Awesome, ship it!

dkotter (approved):

PR #552, opened 2026-05-14
Fix: Bug: Empty space shown for Model field when saved provider no longer exists

### LINKED ISSUE #551: Bug: Empty space shown for Model field when saved provider no longer exists
## Description

In the Developer Settings panel, when a provider and model have been previously selected and saved, removing that provider leaves an empty space where the Model field used to be - even though the Model field itself is not rendered.

### Expected Behavior

The Model field and its container should be fully hidden when the saved provider no longer exists. No empty space should appear between the Provider dropdown and the `Reset to default` link.

### Actual Behavior

The Model field's container is still rendered (taking up vertical space) because `isVisible` only checks `!! data.provider`, which is `true` for any non-empty saved provider ID - even if that provider no longer exists in the available providers list.

### Fix

Update the `isVisible` callback for the `model` field to also verify the saved provider ID exists in the current `providers` array:

```diff
- isVisible: ( data: DeveloperSelection ) => !! data.provider,
+ isVisible: ( data: DeveloperSelection ) =>
+     !! data.provider &&
+     !! providers.find( ( p ) => p.id === data.provider ),
```

## Step-by-step reproduction instructions

1. Open Developer Settings for any feature.
2. Select a provider from the Provider dropdown.
3. Select a model from the Model dropdown.
4. Remove the selected provider (e.g., deactivate/delete the provider integration).
5. Return to Developer Settings for the same feature.

## Screenshots, screen recording, code snippet

https://github.com/user-attachments/assets/644cfe0b-cb8f-4a73-93bb-6f16b9afbb20

Image

## Environment info

- WordPress - 7.0-RC3-62350
- AI plugin - Develop branch

## Please confirm that you have searched existing issues in the repo.

- [x] Yes

## Please confirm that you have tested with all plugins deactivated except the AI plugin.

- [x] Yes

## Please confirm which theme type you used for testing.

- [ ] Block
- [ ] Classic
- [ ] Not sure

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/551

## Why?

- PR removes the extra empty space, when provider and model are selected, and then provider is removed, it shows blank space in model selection area in settings.

## How?

- PR checks for current providers fetched and what is stored in settings. If current providers does not have that data, the field would not render.

### Use of AI Tools

AI assistance: Yes
Tool(s): GitHub Copilot, Claude
Model(s): Claude Opus 4.6
Used for: Used for issue creation and PR description, plus review the approach.

## Testing Instructions

1. Open Developer Settings for any feature.
2. Select a provider from the Provider dropdown.
3. Select a model from the Model dropdown.
4. Remove the selected provider (e.g., deactivate/delete the provider integration).
5. Return to Developer Settings for the same feature.
6. You would see there is no extra spacing.

## Screenshots or screencast

| Before | After |
| ------ | ------ |
| Image | Screenshot 2026-05-14 at 12 43 19 PM |

## Changelog Entry

> Fixed - Empty space shown for Model field when saved provider no longer exists in developer settings.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-552-8b3d41a2b2c9fd70b3581213c2ede39521acb1ec.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/routes/ai-home/components/DeveloperSettings.tsx
+++ b/routes/ai-home/components/DeveloperSettings.tsx
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { Button, Spinner } from '@wordpress/components';
+import { Button, Notice, Spinner } from '@wordpress/components';
 import { DataForm } from '@wordpress/dataviews';
 import type { Field, Form } from '@wordpress/dataviews';
 import { useCallback, useMemo, useRef } from '@wordpress/element';
@@ -105,6 +105,9 @@ export function DeveloperSettings( {
 	);
 
 	const hasSavedSelection = settings.provider !== '' || settings.model !== '';
+	const hasStaleProvider =
+		!! settings.provider &&
+		! providers.find( ( p ) => p.id === settings.provider );
 
 	if ( capability === 'none' ) {
 		return (
@@ -134,6 +137,18 @@ export function DeveloperSettings( {
 			) }
 			{ ! isLoading && ! fetchError && (
 				<>
+					{ hasStaleProvider && (
+						<Notice
+							className="ai-developer-mode-fields__notice"
+							status="warning"
+							isDismissible={ false }
+						>
+							{ __(
+								'The previously selected provider is no longer available. This feature will not function as expected until a valid provider is selected or the selection is reset to default.',
+								'ai'
+							) }
+						</Notice>
+					) }
 					<div ref={ formWrapperRef }>
 						<DataForm< DeveloperSelection >
 							data={ settings }
--- a/routes/ai-home/components/DeveloperSettings.tsx
+++ b/routes/ai-home/components/DeveloperSettings.tsx
@@ -1,11 +1,12 @@
 /**
  * WordPress dependencies
  */
-import { Button, Notice, Spinner } from '@wordpress/components';
+import { Button, Spinner } from '@wordpress/components';
 import { DataForm } from '@wordpress/dataviews';
 import type { Field, Form } from '@wordpress/dataviews';
 import { useCallback, useMemo, useRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
+import { Notice } from '@wordpress/ui';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


dkotter (commented):
Thanks for the fixing this, tested and works as expected.

The one thing worth discussing is right now this properly hides the Model dropdown in the situation where someone removes a Connector. But when we make the AI request, we still try and use that Connector even though it doesn't exist.

While you could argue that's expected, the fact that we show the Provider dropdown as being set to `Default` is confusing here.

I wonder if we should output an error message above that dropdown letting the user know their previous chosen provider no longer exists and they need to update that our the feature won't work.

Or maybe add a new option to the dropdown in this scenario that has the previously set provider name and some helper text, like `OpenAI - note this provider no longer exists and should be reset`

jeffpaul:
> I wonder if we should output an error message above that dropdown letting the user know their previous chosen provider no longer exists and they need to update that our the feature won't work.

I'd prefer this path

hbhalodia:
Thanks for the review @dkotter @jeffpaul, I have updated the PR to add the `Warning Notice` above the dropdown. Could you please review the PR?

Below is how it would be,
Screenshot 2026-05-15 at 2 54 45 PM

Let me know if copy text needs to be updated? or `warning` needs to be changed to `error`?

**Update:** I just noticed that, we have some place used `Notice` from `@wordpress/ui`. I am updating the PR to use the same here, so we can have less tech debt while we migrate/move from components to ui on the developerSettings page.

Ref - https://github.com/WordPress/ai/blob/7888332e6c649b2f2ad2dce6ce8161ce7540d030/routes/ai-home/stage.tsx#L1028-L1049

Screenshot 2026-05-15 at 3 14 51 PM

Thanks,

dkotter (approved):
Looks good, thanks for the work here!

PR #560, opened 2026-05-15
Add encryption for AI Connector API keys

dkotter:
## What?

See comment: https://github.com/WordPress/ai/pull/467#issuecomment-4390742183

Ensures any AI Connector API key is passed through encryption when saved and then decrypted when used.

## Why?

WordPress 7.0 introduces a new Connectors API and a big part of that is supporting AI Connectors that authenticate via API keys. These keys are currently stored as plain-text in the database, meaning those can "easily" be leaked. Encrypting those keys increases the level of security here.

This is meant as a PoC to demonstrate how this could be done. Ultimately it would be great to get similar changes in WordPress Core.

## How?

- Adds a new Key Encryption experiment users can opt-in to
- When turned on, will automatically encrypt all AI Connector API keys using the [Displace Secrets Manager](https://github.com/ericmann/displace-secrets-manager) package. The code for this was copied into a new `includes/Vendor` directory and a namespace was added to avoid collisions with sites that are directly using that plugin. Note we don't copy all the code over but just the pieces we need. File names have been changed slightly to work with our autoloading and a custom namespace added to each file but otherwise all code was directly copied
- If the experiment is turned off or if the global AI toggle is turned off or if the plugin is disabled, all keys are automatically decrypted so things still work
- When things are turned back on or the plugin is reactivated, things are encrypted again

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 4.7, Opus 4.8
Used for: Iterated on an initial plan and then executed on that plan. Final iteration and testing done by me. Was also used to evaluate the Displace Secrets Manager plugin and figure out the best way to bring that into our repo.

## Testing Instructions

1. Pull down this PR and run `npm i && npm run build`
2. Setup at least one AI Connector with valid credentials
3. Turn on the Key Encryption experiment
4. You shouldn't notice anything different. All experiments should still work as expected
5. Inspect your database and search for the AI Connector API key in the options table. For example, if using OpenAI, the API key is stored under the option name `connectors_ai_openai_api_key`. This value of this option should be blank
6. Search within the options table for the encrypted API key. For OpenAI, this is stored under `_secret_ai/openai_api_key`. There should be an encrypted value
7. Try various combinations of disabling the experiment, globally disabling AI features, disabling the AI plugin, etc. Ensure the keys are encrypted/decrypted properly and you're not left in a state where you can't access the keys anymore

## Changelog Entry

> Added - New experiment, Key Encryption, that will encrypt any AI Connector API keys.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-560-c1115d73270fc003bff0281727830bd0879e6b88.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter:
Point for discussion:

As mentioned in the PR description, this PR relies on the [Displace Secrets Manager plugin](https://github.com/ericmann/displace-secrets-manager) to do the actual encryption and decryption.

I have not yet included that plugin with this PR so you have to manually install that to test things. While I think that's fine for testing, ideally we find a better way if this is something we want to actually release.

A couple options to consider:

1. Include the plugin via composer
2. Copy the plugin into our repo directly
3. If the plugin makes it to WordPress.org, we can include it from there once the experiment is turned on (basically auto-install and activate it)
4. Don't rely on that plugin at all and build our own encyrption/decryption functionality

There's pros and cons to all of those approaches. I'd lean towards option 2 for now even though that does mean bloating our repo a bit. I think option 3 is likely the best approach but does mean we end up installing and activating a plugin without the user fully understanding that's happening.

jeffpaul:
@ericmann - I'd be curious for your thoughts on the above, as this could be a good testing ground for this approach as one that could make its way towards WP core with lessons learned from usage within the AI plugin

justlevine:
As I wasn't on the call 2 weeks ago, and the AI-summaries leave much to be desired... Is there a specific reason not to ship and hook in our own Encryption class? What are the cons here?

Putting aside any misgivings I have about bringing in reliance on a non-established 3pp (nothing stopping them from shipping their own integration with the AI plugin), the solution itself seems like overkill for what we need.

dkotter:
> As I wasn't on the call 2 weeks ago, and the AI-summaries leave much to be desired... Is there a specific reason not to ship and hook in our own Encryption class? What are the cons here?

The plugin used here was being floated as a solution that could make it into Core. So like many other things we have here, this was an attempt to try that out and prove it works to hopefully help move that conversation around.

That said, see point 4 in my [comment](https://github.com/WordPress/ai/pull/560#issuecomment-4461423239). I'm happy to build out a custom encryption class if we think that's a better approach.

jeffpaul:
Also happy to build or use another approach, especially if there's something that's potentially a MUCH better connection for core.  But lacking that, getting something out there as part of an Experiment will get us some real-world feedback to continue iterating towards something that perhaps in 7.1/7.2 makes it to core.

gziolo:
I didn’t find any reference to https://core.trac.wordpress.org/ticket/64789 in this PR. For the record, I initiated a related discussion on WordPress Trac about how to store two-way encrypted values in WordPress, addressing the need for connectors.

jeffpaul:
@dkotter for the purposes of getting this experiment into the plugin for further testing, let's go with option 2 (copy plugin into our repo) so we can get this merged/released and iterate later as we get feedback / discussion evolves

dkotter:
> @dkotter for the purposes of getting this experiment into the plugin for further testing, let's go with option 2 (copy plugin into our repo) so we can get this merged/released and iterate later as we get feedback / discussion evolves

Okay, this was a bit more complicated (since it was set up as a normal plugin and can't just copy a plugin into another plugin and have things directly work) but this is done now. I didn't copy over everything (there's some admin and WP-CLI things we don't directly need yet) and I added a custom namespace to the files I did bring over to prevent conflicts with sites that may be using that plugin directly.

But this should be working now without any need for extra dependencies.

jeffpaul:
Could be an oddity of my testing, but trying to save API keys is now giving me an error using Playground to test this PR:

> It was not possible to connect to the provider using this key.

dkotter:
> Could be an oddity of my testing, but trying to save API keys is now giving me an error using Playground to test this PR:
> 
> > It was not possible to connect to the provider using this key.

Nice catch, was able to consistently reproduce this myself on a fresh install (or an install that hadn't set up a Connector previously):

1. Install the AI plugin from this branch (or use the Playground link)
2. Go the `Settings > AI`
3. Toggle on `Enable AI`
4. Toggle on Key Encryption
5. Go to `Settings > Connectors`
6. Install/activate one of the connectors there
7. Add your API key and hit save
8. You'll see an error message
9. If you disable Key Encryption and try again, it will now save. If you delete the key and try saving again, it will now save. Just fails the first time

I've pushed changes now that should fix the above but let me know if you're still running into issues.

jeffpaul (commented):
Tested via Playground with OAI credentials and experiments continued to work as expected (did not check for API key in DB though because I'm not even sure how/where to via Playground).  Will test separately locally.

jeffpaul (approved):
My local shows that the keys are properly encrypted and features and experiments work as expected.

johnbillion:
This approach suffers from the exact problem that has been discussed on https://core.trac.wordpress.org/ticket/64789, the default encryption relies on a secret stored in a constant in wp-config.php that is not stable and can cause data loss for end users.

> Tying it to a key in this file risks data loss if it gets rotated or a user migrates/exports/copies their data to another environment that uses different keys.

I like the idea of an extensible encryption API that allows users/hosts to connect it to a KMS, but I honestly cannot see any way for it to safely use the wp-config.php constants as a default provider for encryption keys.

jeffpaul:
> I like the idea of an extensible encryption API that allows users/hosts to connect it to a KMS, but I honestly cannot see any way for it to safely use the wp-config.php constants as a default provider for encryption keys.

@johnbillion any suggestion on an alternate path to consider as an iteration on this experiment?

dkotter:
Worth noting the preferred/recommended approach is for a site to set their own custom `WP_SECRETS_KEY` constant. While that doesn't completely solve this point:

> the default encryption relies on a secret stored in a constant in wp-config.php that is not stable and can cause data loss for end users

It's less likely that a custom constant will change compared to things like `LOGGED_IN_KEY` or `LOGGED_IN_SALT`. And it also supports a `WP_SECRETS_KEY_PREVIOUS` constant so if someone wants to manually rotate the key, they can without breaking anything (noting all of that is more technically focused and I wouldn't expect a non-technical site owner to understand how to set this up).

One thing we don't do here is enforce that custom constant. If it's there we use it but if not, we fallback to core constants that are more likely to run into rotation problems. A potential follow-up here would be to enforce the use of `WP_SECRETS_KEY` and not support any sort of fallback.

johnbillion:
A dedicated constant for a key (such as the `WP_SECRETS_KEY`/`WP_SECRETS_KEY_PREVIOUS` used by this implementation) does partly mitigate the issue, but it doesn't remove the fundameltal risk of data loss when moving data between different hosts or environments and the master encryption key changes as a result.

At some point we'll need to weigh up the advantages of encrypting sensitive info by default versus the risk of it being lost if the master encryption key is lost or unexpectedly changes. (The `WP_SECRETS_KEY_PREVIOUS` solution fixes intentional rotation).

At the moment my gut is telling me that a two-way encryption API must be backed a stable key that a site owner can't get caught out by when they move their database and the key suddenly differs or isn't available. A KMS can solve that problem, but notably if it's a host-specific KMS that doesn't go with the site owner when they move to a new web host then they're still in the same situation.

Maybe this is a UX problem more than a technical one. A site owner needs to be loudly made aware if something happens that causes their encrypted data can no longer be decrypted, so they can act upon it. That's fine for something like an API key that can usually be recovered by other means or regenerated, but if other features start encrypting data that cannot be recovered then I think the recovery workflow UX becomes the blocker.

johnbillion:
Couple of links:

- https://laravel.com/docs/12.x/encryption
- https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html

PR #577, opened 2026-05-19
Fix: Content Classification: Make character count locale-aware

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/571

## Why?

- PR updates the content clasification experiment for suggestion to use dynamic wordCount based on user's locale.
- It uses `words, characters_excluding_spaces, or characters_including_spaces` for wordCount.

## How?

- Updates the `useContentClassification` to use `wordCountType` as suggested here and most of places, https://github.com/WordPress/gutenberg/blob/1ecc3d2fe9c819af35ef312563805bc16a103036/packages/block-library/src/post-time-to-read/edit.js#L51-L61

### Use of AI Tools

- None

## Testing Instructions

- Change the site language to Japanese.
- Open a post.
- Copy the following Japanese text multiple times to create content: `こんにちは。`
- Confirm that the Suggest Categories button should enable now.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| Screenshot 2026-05-19 at 2 33 21 PM | Screenshot 2026-05-19 at 2 33 21 PM |

## Changelog Entry

> Added - Determine wordCountType for content classification experiment to use words or characters based on users locale.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-577-e0c4bb41034c64677d275db6634afc5b98a7b2dc.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #545, opened 2026-05-13
Feature: Disable Review notes until minimum content threshold is reached

coderGtm:
## What?

Closes #390 

Adds the same minimum-content gating used by Content Summarization to Review Notes, requiring at least 100 characters before Review Notes can run, making that threshold filterable, and showing descriptive disabled-state text in the editor UI.

## Why?

Review Notes could previously be triggered on very short or empty post content, which does not provide enough content for meaningful AI review suggestions.

Content Summarization already avoids this by requiring a minimum amount of content before enabling the feature. This change aligns Review Notes with that existing behavior so users get a more consistent UX across AI editor features, and so Review Notes only becomes available when there is enough content to review.

## How?

This PR updates the Review Notes experiment to follow the same enablement pattern as Content Summarization.

Implementation details:
- Adds a new filter, `wpai_review_notes_min_content_length`, with a default value of `100`
- Localizes the minimum content length to the editor script as `minContentLength`
- Adds shared Review Notes availability logic in the editor based on character count
- Disables the main `Generate Review Notes` button when post content is below the threshold
- Shows descriptive helper text explaining when Review Notes becomes available
- Disables the single-block `Generate Review Note` menu action when the post content is below the threshold
- Prevents short-content posts from triggering Review Notes requests
- Adds integration test coverage for the default and filtered minimum content length
- Adds e2e coverage for the disabled/enabled Review Notes states and the single-block disabled state
- Updates the empty-post e2e scenario to reflect the new minimum-length gating behavior

### Use of AI Tools

AI assistance: Yes  
Tool(s): GitHub Copilot  
Model(s): GPT-5-class model via Copilot  
Used for: Implementation guidance, test updates, and PR drafting; final code and description were reviewed and edited by me.

## Testing Instructions

1. Enable Experiments globally in the AI plugin settings.
2. Enable the `Review Notes` experiment.
3. Open a new post in the editor with content shorter than 100 characters.
4. Open the post settings sidebar.
5. Confirm the `Generate Review Notes` button is visible but disabled.
6. Confirm the helper text says Review Notes will be available at 100 characters.
7. Add enough content so the post reaches at least 100 characters.
8. Confirm the `Generate Review Notes` button becomes enabled.
9. Select a reviewable block and open the block options menu.
10. Confirm `Generate Review Note` is disabled when the post is below the threshold.
11. Confirm `Generate Review Note` becomes available once the post content meets the threshold.
[...]

PR #581, opened 2026-05-19
Fix: locale-aware (words/characters) minimum content length across AI content experiments

### LINKED ISSUE #390: Generate Review Notes and Generate Summary options present even with no post content

### LINKED ISSUE #391: Regenerate option is being shown even though there’s no content yet.

### LINKED ISSUE #571: Content Classification:  Make character count locale-aware
### Description

Content classification requires more than 150 **words** to be enabled. However, CJK languages like Japanese and Chinese do not use spaces between words. Consequently, this feature cannot be utilized even with sufficient content.

We should adopt a locale-aware approach rather than hard-coding the words. So, as the word count type, we should use the translated text.

https://github.com/WordPress/gutenberg/blob/1ecc3d2fe9c819af35ef312563805bc16a103036/packages/block-library/src/post-time-to-read/edit.js#L51-L61

https://github.com/WordPress/ai/blob/396fcc33bf39a77ea817ffc03f52e87a9c16b4b1/src/experiments/content-classification/components/SuggestionPanel.tsx#L77

This text should also change depending on whether it is counted by words or characters.

> Add more content to enable AI suggestions (approximately 150 words).
> Add more content to enable AI suggestions (approximately 150 characters).

- Change the site language to Japanese.
- Open a post.
- Copy the following Japanese text multiple times to create content: `こんにちは。`
- Confirm that the Suggest Categories button remains disabled.

Image

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

### LINKED ISSUE #578: Content Resizing: The "shorten" action does not detect the character length of Japanese text
### Description

Similar to #571

The "shorten" action requires that a paragraph contain more than 5 words. However, CJK languages like Japanese and Chinese do not use spaces between words. Consequently, this feature cannot be utilized even with sufficient content.

We should adopt a locale-aware approach rather than hard-coding the words. So, as the word count type, we should use the translated text.

https://github.com/WordPress/gutenberg/blob/1ecc3d2fe9c819af35ef312563805bc16a103036/packages/block-library/src/post-time-to-read/edit.js#L51-L61

https://github.com/WordPress/ai/blob/396fcc33bf39a77ea817ffc03f52e87a9c16b4b1/src/experiments/content-classification/components/SuggestionPanel.tsx#L77

1. Change the site language to Japanese.
2. Open a post.
3. Copy the following text and paste it multiple times inside a paragraph block: `あああああああああああ`
4. Execute the Shorten action
5. Confirm that an error notice is displayed.

https://github.com/user-attachments/assets/c9b80e9b-4347-42dc-b7c5-c70044f7e6ff

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/578, https://github.com/WordPress/ai/issues/391, https://github.com/WordPress/ai/issues/390

## Why?

- PR uses `wordCountType` for counting words or characters based on the users locale, standardize across all the experiments.

## How?

- PR uses the same technique that Gutenberg uses for wordCount. To check for user's locale and update the countType based on `_x()` language pack.
- Adds the filter to change based on the experiments for the min length criteria.
- See detailed comment here - https://github.com/WordPress/ai/pull/581#issuecomment-4658950572

### Use of AI Tools

- Yes, Claude Code, Opus 4.8,4.7.
- Used for implementation phase with standardize across the experiments.

## Testing Instructions

1. Change the site language to Japanese.
2. Open a post.
3. Copy the following text and paste it multiple times inside a paragraph block: `あああああああああああ`
4. Execute the Shorten action
5. Confirm that no error notice is displayed and word shorten is working as expected.

## Screenshots or screencast

https://github.com/user-attachments/assets/02da5ec9-9eb8-489b-b7b2-e98aef12f5f9

## Changelog Entry

> Fixed - Add wordCountType to check for user's locale and update to count character or words and standardize the min content length accross the experiments.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-581-58d3f86e4fa025094db7f818b77c984f23230767.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


src/experiments/content-resizing/components/ContentResizingToolbar.tsx
hbhalodia: Hi Team, Do we need to update the copy here to use `characters`? based on wordCountType and also on similar places?

IMO we should update that, waiting for feedback before applying the change.

Thanks,

dkotter: So a couple thoughts here.

We have a few places that use the `@wordpress/wordcount` package and now two separate places that use the `wordCountType` to determine if we should look at words or characters. Wondering if we extract all that out into a shared helper so we're not duplicating code?

In addition, we're inconsistent right now on if we look at words or characters (not just based on locale but just in general). For instance, we default to looking at words here and for Content Classification but I believe Content Summarization defaults to looking at characters.

I'm thinking it might be nice to standardize this across the plugin to avoid these inconsistencies. This may also simplify the code a bit if we decide to standardize on characters instead of words, avoiding the need to change things depending on locale.

Any thoughts on these?

dkotter: Also noting we have #545 open that adds a minimum content length to Editorial Notes and it uses characters.

jeffpaul: In practice and testing, I've seen AI reviews and feedback much more useful when there's a LOT more content available for it to utilize.  However, I don't know that setting something like 500 words is realistic for all use cases for the AI plugin so perhaps something much lower like 100 words or 100 characters (noting neither are super specific so much as a round number).  To some extent here, I don't care too much about the minimum length so much as we set something and use it across the plugin and allow that to be filterable.  So let's just YOLO into _something_ and can iterate if feedback shows we need to raise or lower that minimum threshold.

t-hamano: As one approach, how about considering both words and characters? In other words, the label displays both words and characters.

```javascript
_n(
	'+%1$d word (+%2$d characters)',
	'+%1$d words (+%2$d characters)',
	magnitude,
	'ai'
),
magnitude,
charMagnitude
```

Furthermore, the threshold check might be able to consider both of those.

```javascript
if ( action === 'shorten' ) {
	const wordCount = count( blockContent, 'words', {} );
	const charCount = count( blockContent, 'characters_excluding_spaces', {} );
	if ( wordCount < SHORTEN_MIN_WORDS && charCount < SHORTEN_MIN_CHARS ) {
		noticesDispatch.createErrorNotice();
		return;
	}
}
```

hbhalodia: > We have a few places that use the @wordpress/wordcount package and now two separate places that use the wordCountType to determine if we should look at words or characters. Wondering if we extract all that out into a shared helper so we're not duplicating code?

Sure I would make a helper function, that checks for type.

> This may also simplify the code a bit if we decide to standardize on characters instead of words, avoiding the need to change things depending on locale.

Not sure on this, but what as @t-hamano said, if looking at `English` it is mostly used words, while some languages such as `Japanese` uses characters (in that some languages have 2 parts, to include space or to not include space). So to make it consistent, we should go with characters.

But worth noting the suggestion about checking for both words and characters. But still, what if words are less and characters satisfies teh condition, then above code will still throw the notice.

As per me we should follow the gutenberg approach, define both words and characters min length individually. Based on the locale use either words or characters for counting. And this should be conistent across all the exoeriments, this, Content Classification and Content Summarization?

dkotter: I'm open to any approach here, what I'm wanting is to ensure we approach this in a way that works across all of our features. Right now we are inconsistent with what features require a minimum content length and which ones don't. And the ones that do require it do it differently.

I would like to see a PR that standardizes this across all features and does it in a reusable way to avoid code duplication. At a high-level I think this means a client-side helper that all features can use. A server-side helper (with a filter around the content length) that all features can use. And then integrating that within all features, so a feature remains in a disabled state until the minimum length is hit.

As far as what the minimum length should be and if that should be characters, words or both, I'm open to suggestions there.

hbhalodia: Hi @dkotter @t-hamano, I have updated the PR with changes to make it first reusable accross all the features. 

I am closing this PR - https://github.com/WordPress/ai/pull/577, We can focus on to improve all this feature in this same PR.

Currently checking on if we can optimize to make use of both characters and words and display both as a suggestion shared by @t-hamano.

Thanks,

hbhalodia: Hi Team, I need the suggestion here. Based on current state of PR, everything is standarized, duplication removed and used helper functions.

All the features will either use `characters` if user's locale is the language that uses characters (Eg: Japanese), and will use the `words` if the users local is the language that use words to count (Eg: English).

Now, we have a question to what to show the user, should we update the text to show `characters or words` depending on the user's locale, or Should we introduce both minWords and minCharacter count and show user both and disable experiments if both of them does not satisfies the condition?

Currently, I am updating the PR with the first approach, to update text based on user's locale, will update it based on the feedback.

dkotter: > Now, we have a question to what to show the user, should we update the text to show characters or words depending on the user's locale

I'm saying this as someone that only speaks English so definitely want other opinions here but if the decision is to use either words _or_ characters depending on locale (and not standardize on characters only or standardize on checking both) I think the right approach is to update the message to show either characters or words, not both.

hbhalodia: Thanks, @dkotter I agree to check for both based on user's locale. I have updated this in the current PR.

includes/Experiments/Summarization/Summarization.php
```diff
@@ -106,10 +108,11 @@ public function enqueue_assets( string $hook_suffix ): void {
 		 * Filters the minimum content length required to enable summarization.
 		 *
 		 * @since 1.0.0
+		 * @deprecated x.x.x Use {@see 'wpai_min_content_length'} instead.
 		 *
 		 * @param int $min_content_length The minimum number of characters required. Default 100.
```
dkotter: This isn't always accurate now, as this number may be for characters or may be for words

includes/Experiments/Summarization/Summarization.php
```diff
@@ -106,10 +108,11 @@ public function enqueue_assets( string $hook_suffix ): void {
...
 		 */
-		$min_content_length = (int) apply_filters( 'wpai_summarization_min_content_length', 100 );
+		$min_content_length = (int) apply_filters( 'wpai_summarization_min_content_length', get_min_content_length( 'summarization', 100 ) );
```
dkotter: So I know 100 was the last value but that was when we looked at characters only. Now we'll look at words if the locale supports it so is 100 words the right length? Seems that may be too long

includes/Experiments/Content_Classification/Content_Classification.php
```diff
@@ -151,9 +153,10 @@ public function enqueue_assets( string $hook_suffix ): void {
 			'content_classification',
 			'ContentClassificationData',
 			array(
-				'enabled'        => $this->is_enabled(),
-				'strategy'       => $this->get_strategy(),
-				'maxSuggestions' => $this->get_max_suggestions(),
+				'enabled'          => $this->is_enabled(),
+				'strategy'         => $this->get_strategy(),
+				'maxSuggestions'   => $this->get_max_suggestions(),
+				'minContentLength' => get_min_content_length( 'content-classification', 150 ),
```
dkotter: Is 150 the right length here or should we lower that?

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
```diff
@@ -173,33 +184,49 @@ export default function ContentResizingToolbar( {
...
+				label: sprintf( label, magnitude ),
+				ariaLabel: sprintf( ariaLabel, magnitude ),
 			};
 		}
 
+		const label = isCharacterType
+			? /* translators: %d: Number of characters removed. */
+			  _n( '-%d character', '-%d characters', magnitude, 'ai' )
+			: /* translators: %d: Number of words removed. */
+			  _n( '-%d word', '-%d words', magnitude, 'ai' );
```
dkotter:
```suggestion
			  _n( '−%d word', '−%d words', magnitude, 'ai' );
```

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
```diff
@@ -173,33 +184,49 @@ export default function ContentResizingToolbar( {
...
-					magnitude
-				),
+				label: sprintf( label, magnitude ),
+				ariaLabel: sprintf( ariaLabel, magnitude ),
 			};
 		}
 
+		const label = isCharacterType
+			? /* translators: %d: Number of characters removed. */
+			  _n( '-%d character', '-%d characters', magnitude, 'ai' )
```
dkotter:
```suggestion
			  _n( '−%d character', '−%d characters', magnitude, 'ai' )
```

includes/Experiments/Summarization/Summarization.php
```diff
@@ -106,10 +108,11 @@ public function enqueue_assets( string $hook_suffix ): void {
 		 * Filters the minimum content length required to enable summarization.
 		 *
 		 * @since 1.0.0
+		 * @deprecated x.x.x Use {@see 'wpai_min_content_length'} instead.
```
dkotter, line 109: Should probably use `apply_filters_deprecated` here

dkotter (changes requested):
Left a few comments but overall this is looking good.

The only other thing I'd flag is this standardizes things for Content Classification, Content Resizing and Summarization but none of the other Features. I still think it would be ideal to standardize everything, which likely means closing out #479 and #545 and handling those in this PR, as well as looking at all other Features and ensure, where needed, they follow the same approach

includes/Experiments/Summarization/Summarization.php
hbhalodia: I am not sure on this what number would be the best here. If we go with less then what if locale supports characters, then it would be less either. So may be we can change to 50?

includes/Experiments/Content_Classification/Content_Classification.php
hbhalodia: I guess it's good thing to have more. Because if we have words as locale, higher the words AI may provide the better suggestions. For characters as well, it would provide better suggestions.

Here in the comment as well - https://github.com/WordPress/ai/pull/581#discussion_r3377996606, I guess we can stick with 100, that's because for words locale higher the words it would provide better summarization? yes?

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
hbhalodia: Looks same how its identified 😅.

Fixed.

includes/Experiments/Summarization/Summarization.php
hbhalodia: Fixed.

hbhalodia:
> Left a few comments but overall this is looking good.
> 
> The only other thing I'd flag is this standardizes things for Content Classification, Content Resizing and Summarization but none of the other Features. I still think it would be ideal to standardize everything, which likely means closing out #479 and #545 and handling those in this PR, as well as looking at all other Features and ensure, where needed, they follow the same approach

Thanks @dkotter, I am checking those and will update in the PR as needed.

hbhalodia:
## Summary

Fixes Content Resizing's **"Shorten"** action not detecting the length of **Japanese (and other character-based) text**, and generalizes the fix into a unified, locale-aware **minimum content length** system shared by every AI content experiment.

The root problem: content-length checks counted **words** (whitespace-delimited), which is meaningless for CJK languages where there are no spaces. This PR introduces locale-aware counting (words vs. characters, driven by WordPress core's word-count strategy) and a single, filterable source of truth for the minimum length, then applies it consistently across all content experiments — disabling the relevant action and showing a clear, locale-correct message when there isn't enough content.

## Problem / root cause

- Content Resizing's "Shorten" used a word count to decide whether there was enough text to shorten. In Japanese/Chinese/Korean locales (no word separators) the count was effectively wrong, so the gate misbehaved.
- Minimum-length logic was duplicated/hardcoded and inconsistent across experiments, with word-only assumptions baked in.

## Solution

1. **Locale-aware counting** — new `src/utils/word-count.ts` utilities that respect WordPress core's word-count strategy (`words` vs. `characters_excluding_spaces` / `characters_including_spaces`).
2. **Centralized, filterable minimum** — new `get_min_content_length()` PHP helper with a single `wpai_min_content_length` filter.
3. **Consistent UX** — each experiment disables its Generate/Regenerate action when content is too short and shows a message phrased in **words or characters** to match the locale.

## New shared infrastructure

- **`includes/helpers.php`** — new `get_min_content_length( string $feature_id, int $default )` that returns the per-feature minimum, filterable via `wpai_min_content_length` (`( $length, $feature_id )`).
- **`src/utils/word-count.ts`** — new utilities:
  - `getWordCountType()` — returns the locale's strategy via core's translatable `_x( 'words', … )` ("Do not translate" string), so CJK locales report a character strategy.
  - `getContentCount( content )` — counts using `@wordpress/wordcount` with that strategy.
  - `hasMinimumContent( content, min )` — the shared gate used everywhere.

## Per-experiment changes

All seven content experiments now localize a `minContentLength` (via `get_min_content_length()`) and gate their UI through `hasMinimumContent()` / `getWordCountType()`:

| Experiment | Default min | Notes |
|---|---|---|
| Content Classification | **100** | Hint now shows words **or** characters; suggest button gated. |
| Content Resizing | **100** | The original Japanese-text fix; "Shorten" now uses locale-aware count. |
| Editorial Notes | **100** | Gate retained; message standardized to words/characters. |
| Excerpt Generation | **100** | New gate; panel + inline buttons disabled with tooltip. |
| Meta Description | **100** | New gate; panel Generate/Regenerate + modal Generate gated. |
| Summarization | **100** | Now sourced from `get_min_content_length()`; see deprecation below. |
| Title Generation | **100** | New gate; standalone + toolbar buttons disabled with tooltip. |

PHP (each adds the `use function … get_min_content_length;` import + `minContentLength` in the localized data):
- `includes/Experiments/Content_Classification/Content_Classification.php`
- `includes/Experiments/Content_Resizing/Content_Resizing.php`
- `includes/Experiments/Editorial_Notes/Editorial_Notes.php`
- `includes/Experiments/Excerpt_Generation/Excerpt_Generation.php`
- `includes/Experiments/Meta_Description/Meta_Description.php`
- `includes/Experiments/Summarization/Summarization.php`
- `includes/Experiments/Title_Generation/Title_Generation.php`

Frontend (types + components/hooks reading `minContentLength`, computing `isContentTooShort`, disabling actions, and rendering words/characters messaging):
- Content Classification — `types.ts`, `useContentClassification.ts`, `SuggestionPanel.tsx`
- Content Resizing — `types.ts`, `ContentResizingToolbar.tsx`
- Editorial Notes — `EditorialNotesPlugin.tsx`, `useEditorialNotes.ts`
- Excerpt Generation — `types.ts`, `useExcerptGeneration.ts`, `ExcerptGeneration.tsx`, `ExcerptInlineButton.tsx`
- Meta Description — `types.ts`, `useMetaDescription.ts`, `MetaDescriptionPanel.tsx`, `MetaDescriptionModal.tsx`
- Summarization — `types.ts`
- Title Generation — `types.ts`, `components/TitleToolbar.tsx`

## Deprecation / backward compatibility

- **`Summarization.php`** — the legacy `wpai_summarization_min_content_length` filter is migrated to `apply_filters_deprecated( …, 'wpai_min_content_length' )`. The canonical `wpai_min_content_length` filter (applied inside `get_min_content_length()`) is now the supported extension point; the old filter still works for back-compat.

## Behavior

When content is below the threshold the action button is `disabled` (with `accessibleWhenDisabled`) and surfaces:

- *"… will be available when the post content has at least N words."* — word-based locales
- *"… at least N characters."* — character-based locales (e.g. Japanese), via `getWordCountType()`

Because of `accessibleWhenDisabled`, disabled buttons report state via `aria-disabled="true"` (focusable, tooltip announced) rather than the native `disabled` attribute.

## Testing

### Unit / Integration (PHPUnit) — 7 experiment test files
Each adds localization assertions:
- `test_enqueue_assets_localizes_default_min_content_length` → default value present in script data.
- `test_enqueue_assets_localizes_filtered_min_content_length` → value reflects a `wpai_min_content_length` filter.

Files under `tests/Integration/Includes/Experiments/…`: Content_Classification, Content_Resizing, Editorial_Notes, Excerpt_Generation, Meta_Description, Summarization, Title_Generation.

```bash
npm run wp-env:test start
npm run test:php tests/Integration/Includes/Experiments/Title_Generation/Title_GenerationTest.php
# …repeat per experiment, or run the whole suite:
npm run test:php
```

### End-to-end (Playwright) — 7 experiment spec files
Coverage added across `tests/e2e/specs/experiments/`:
- **Disabled-button tests** (short content → `aria-disabled="true"`) for title, excerpt, meta-description (and equivalents for the others).
- **Threshold tests** — e.g. Content Classification enables suggestions once content crosses the minimum; Content Resizing "Shorten" opens and renders the exact original/suggested text when content meets the minimum.
- Existing functional flows updated to seed sufficient content (`LONG_CONTENT`) so they remain green under the new gate.
- Removed the obsolete Excerpt "clean error notice on empty content" e2e test (that empty-content path is now prevented client-side).

```bash
npm run test:e2e -- specs/experiments/content-resizing.spec.js \
  specs/experiments/content-classification.spec.js \
  specs/experiments/title-generation.spec.js \
  specs/experiments/excerpt-generation.spec.js \
  specs/experiments/meta-description.spec.js
```

> e2e assertions assume each feature's default minimum (no `wpai_min_content_length` filter seeded in the test env).

## Configuration

```php
// Adjust the minimum per feature (defaults shown in the table above).
add_filter( 'wpai_min_content_length', function ( $length, $feature_id ) {
    // $feature_id: 'content-resizing' | 'content-classification' | 'summarization'
    //            | 'editorial-notes' | 'excerpt-generation' | 'meta-description'
    //            | 'title-generation'
    return $length;
}, 10, 2 );
```

## Notes / follow-ups

- `@since` / `@deprecated x.x.x` placeholders should be set to the target release version before merge.
- Optional: standardize Editorial Notes to the same `getSettings()` helper pattern used by the other experiments (currently reads the localized global inline; functionally equivalent).

hbhalodia:
Hi @dkotter, Below are some visuals for the updates,

## Title Generation

https://github.com/user-attachments/assets/e68d15fb-8249-4fd5-8978-bec27635b274

## Excerpt Generation

https://github.com/user-attachments/assets/751d92ac-3019-4baa-bfb0-6d5799d84494

## Meta Description Generation

https://github.com/user-attachments/assets/13e27c14-5801-48a8-98f8-2e86731dcac1

## Content Classification

https://github.com/user-attachments/assets/16f2d92b-1f32-44ed-9c49-5de3984d97af

Thanks,

dkotter (approved):

PR #582, opened 2026-05-19
Fix JS translation loading and make all text translatable

t-hamano:
## What?

Fixes internationalization (i18n) gaps across several experiments so that user-facing text is fully translatable.

## Why?

Three problems were found:

1. **JS translations were not working at all.** Strings in JS/TS were wrapped with `@wordpress/i18n` functions, but `wp_set_script_translations()` was never called for any enqueued script. As a result every JS-side string rendered in English regardless of the site locale.
2. **Translatable strings were built by hard concatenation.** In the Abilities Explorer help tab, each provider description was split into two translatable strings joined by a literal `<strong>` tag and `": "`. Translators could not see the full sentence and could not reorder the markup for their language.
3. **Some user-visible text was never wrapped in a translation function.** Toxicity/sentiment badge labels, badge states, an error notice, and JSON validation messages were hardcoded, so they could never be localized and were invisible to the i18n string scanner.

## How?

1. Call `wp_set_script_translations()` for every script enqueued through `Asset_Loader::enqueue_script()`. No path is passed because the plugin is distributed on WordPress.org, which auto-delivers translation JSON to `wp-content/languages/plugins/`.
2. Make each Abilities Explorer provider description a single translatable string with the `<strong>` tag inside, escaped with `wp_kses()` instead of `esc_html__()`.
3. Wrap the remaining hardcoded user-facing strings (badge states, error notice, JSON validation messages and status titles) with `__()`.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.7
Used for: Auditing the codebase for i18n issues and drafting the fixes. All changes were reviewed and verified by me.

## Testing Instructions

1. Switch the site to the Japanese (`ja`) locale and make sure the plugin's Japanese translations are installed.
2. Access http://localhost:8890/wp-admin/update-core.php and download any available translations.
3. Confirm JS-side strings now translate.

### Before

image

### After
image

## Changelog Entry

> Fixed - User-facing text in several experiments is now fully translatable, and JS-side translations are loaded at runtime.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-582-34ed4e6791c2787986564caa316f74686f5dbbe6.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/tests/Integration/Includes/Asset_LoaderTest.php
+++ b/tests/Integration/Includes/Asset_LoaderTest.php
@@ -134,7 +134,7 @@ class Asset_LoaderTest extends WP_UnitTestCase {
 	/**
 	 * @since 0.8.0
 	 */
-	public function test_enqueue_script_falls_back_to_empty_deps_when_missing(): void {
+	public function test_enqueue_script_succeeds_when_asset_file_omits_dependencies(): void {
 		$this->create_asset_file(
 			'no-deps',
 			array( 'version' => '1.0.0' )
@@ -143,7 +143,8 @@ class Asset_LoaderTest extends WP_UnitTestCase {
 		Asset_Loader::enqueue_script( 'no-deps', 'no-deps' );
 
 		$this->assertTrue( wp_script_is( 'ai_no-deps', 'enqueued' ) );
-		$this->assertSame( array(), wp_scripts()->registered['ai_no-deps']->deps );
+		// wp_set_script_translations() appends wp-i18n to the otherwise empty deps.
+		$this->assertSame( array( 'wp-i18n' ), wp_scripts()->registered['ai_no-deps']->deps );
 	}
 
 	/**
--- a/includes/Asset_Loader.php
+++ b/includes/Asset_Loader.php
@@ -78,8 +78,8 @@ final class Asset_Loader {
 	 * @param string $file_name The script file name without the .js extension.
 	 */
 	public static function enqueue_script( string $handle, string $file_name ): void {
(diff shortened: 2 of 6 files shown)


tests/Integration/Includes/Asset_LoaderTest.php
```diff
@@ -143,7 +143,8 @@ public function test_enqueue_script_falls_back_to_empty_deps_when_missing(): voi
 		Asset_Loader::enqueue_script( 'no-deps', 'no-deps' );
 
 		$this->assertTrue( wp_script_is( 'ai_no-deps', 'enqueued' ) );
-		$this->assertSame( array(), wp_scripts()->registered['ai_no-deps']->deps );
+		// wp_set_script_translations() appends wp-i18n to the otherwise empty deps.
+		$this->assertSame( array( 'wp-i18n' ), wp_scripts()->registered['ai_no-deps']->deps );
```
t-hamano, line 147: This change is correct, but I'm not entirely confident. There might be a more reasonable way to assert this.

dkotter (approved):

PR #586, opened 2026-05-19
Fix global AI enabled option migration

ekamran:
## What?

Fixes migration of the global AI enabled setting from older AI Experiments installs.

## Why?

Older versions saved the global toggle as `ai_experiments_enabled`.

The existing upgrade routine intended to migrate old experiment settings, but the global option used the wrong names. As a result, after upgrading from AI Experiments 0.5.0 to the current AI plugin, the global **Enable AI** setting could appear disabled even though it had been enabled before the upgrade.

## How?

- Migrates `ai_experiments_enabled` to the current `wpai_features_enabled` option.
- Keeps support for the singular typo option names created by earlier upgrade logic.
- Repairs the global enabled option before version checks, so already-upgraded sites can recover too.
- Avoids overwriting `wpai_features_enabled` if the current option already exists.
- Updates upgrade tests for both fresh legacy upgrades and already-upgraded repair cases.

## Testing Instructions

Automated checks run locally:

```bash
npm run test:php
npm run lint:php
npm run lint:php:stan
```

Manual verification:

1. Install AI Experiments 0.5.0.
2. Go to `Settings > AI Experiments`.
3. Enable experiments.
4. Upgrade to this branch.
5. Go to `Settings > AI`.
6. Confirm **Enable AI** remains enabled.

## Screenshots/Screencast

Before: AI Experiments 0.5.0 has experiments enabled.
Migration Issue before

After: Current AI settings preserve the global enabled state after upgrade.
Migration Issue after

## Use of AI Tools

AI assistance: Yes  
Tool(s): ChatGPT / Codex  

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-586-169dea71f52d6ff7f8528327d45b2ffdcabfe73e.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Admin/Upgrades.php
+++ b/includes/Admin/Upgrades.php
@@ -72,8 +72,6 @@ final class Upgrades {
 	 * @since 0.6.0
 	 */
 	public static function do_upgrades(): void {
-		self::repair_global_enabled_option();
-
 		$db_version = get_option( self::VERSION_OPTION_KEY, '' );
 
 		foreach ( self::UPGRADE_CLASSES as $upgrade_class ) {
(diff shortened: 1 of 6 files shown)


includes/Admin/Upgrades.php
```diff
@@ -72,6 +72,8 @@ public function init(): void {
 	 * @since 0.6.0
 	 */
 	public static function do_upgrades(): void {
+		self::repair_global_enabled_option();
```
dkotter: Thanks for catching this @ekamran. I'm wondering if we need this code or if the update to the migration script is good enough? While there's likely people that have upgraded past 0.6.0 and thus the upgrade changes above wouldn't impact them, I would expect they've already gone in and re-enabled the global AI option and thus this script isn't needed for them.

If we do think it's worth keeping this, I'd suggest this is moved to the `V1_0_0.php` migration class instead of having it here where it will run for every upgrade

ekamran: Thanks, that makes sense. I agree this should not run outside the versioned migration flow.

I still think the repair is worth keeping because we do not know how many installs upgraded through the broken `0.6.0` migration and never re-enabled the global setting. Some users may have noticed and fixed it manually, but others may just see AI features disabled after updating and assume that is the expected state.

So I think there are two cases worth covering:

1. Fresh upgrades from the old AI Experiments option name should be handled by fixing the `V0_6_0` migration.
2. Installs that already passed `0.6.0` with the incorrect option name should get a one-time repair in the next migration.

I’ll move the repair logic into `V1_0_0.php` so it only runs as part of a normal versioned upgrade, and keep the direct `V0_6_0` option-name correction for older installs.

ekamran: I moved the repair logic into the `V1_0_0` migration so it only runs as part of a versioned upgrade, while keeping the `V0_6_0` option-name fix for direct upgrades from the old AI Experiments versions.

I also updated the tests to cover both paths:
- direct legacy migration via `V0_6_0`
- repair for already-upgraded installs via `V1_0_0`
- no overwrite when `wpai_features_enabled` already exists

dkotter (approved):

PR #598, opened 2026-05-21
Fix: connector approval notice placement in request logs

### LINKED ISSUE #596: Connector Approvals request misplaced on Request Logs screen
### Description

When enabling the Connector Approvals and Request Logging experiments, the approval notice on the Request Log screen appears over top the header.  Not sure if the issue here is a DataForms/Views issue or a placement issue for the notice.  Ideally it shows beneath the page subhead of "Detailed logs of every AI request made by the WordPress AI plugin."

Image

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

Malayt04:
## What?

Closes: #596 

Fixes the connector approval notice appearing inside the page header on the AI Request Logs screen

## Why?

WordPress Core's `common.js` moves `.notice` elements to after the first `h1` in `.wrap`. The `@wordpress/admin-ui` `Page` component renders its `h1` inside a flex header, so the notice ends up as an inline flex child in the header row instead of in the content area.

## How?

Instead of displaying the notice via PHP (which triggers Core's relocation script), the PHP notice is suppressed on this screen and the pending approval data is passed to the React app. The notice is then rendered directly in the page content area, in the right place.

Changes:
- **`Admin_Notice.php`** — suppress PHP notice on the Request Logs screen
- **`AI_Request_Log_Page.php`** — pass `connectorApprovalNotice` data to JS; add `get_connector_approval_notice_data()` 
- **`types.ts`** — add `ConnectorApprovalNotice` interface and field to `LocalizedSettings`
- **`index.tsx`** — render the notice in the content canvas

### Use of AI Tools

AI assistance: Yes  
Tool(s): Antigravity 
Used for: Root cause analysis and implementation. Tested by me.

## Screenshots or screencast

# Before
Screenshot 2026-05-21 at 12 06 54 PM

# After
Screenshot 2026-05-21 at 12 07 14 PM

## Changelog Entry

> Fixed - Connector approval notice now correctly appears below the page subtitle on the AI Request Logs screen instead of overlapping the page header.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-598-687daef7f0b676a3a622564fd95d0c28cf5571bd.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Connector_Approval/Admin_Notice.php
+++ b/includes/Connector_Approval/Admin_Notice.php
@@ -125,7 +125,7 @@ final class Admin_Notice {
 		}
 
 		$screen = get_current_screen();
-		if ( $screen && 'tools_page_ai-connector-approval' === $screen->id ) {
+		if ( $screen && in_array( $screen->id, array( 'tools_page_ai-connector-approval', 'tools_page_ai-request-logs' ), true ) ) {
 			return;
 		}
 
--- a/includes/Logging/AI_Request_Log_Page.php
+++ b/includes/Logging/AI_Request_Log_Page.php
@@ -104,6 +104,7 @@ class AI_Request_Log_Page {
 				),
 				'connectorsUrl'    => admin_url( 'options-connectors.php' ),
 				'providerMetadata' => $this->get_provider_metadata(),
+				'connectorApprovalNotice' => $this->get_connector_approval_notice_data(),
 			)
 		);
 	}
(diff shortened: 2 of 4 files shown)


dkotter (changes requested):
I haven't tested but I'm assuming this approach works but it's a lot of changes just to support a better notice placement on one specific page.

If this is the root problem:

> WordPress Core's common.js moves .notice elements to after the first h1 in .wrap. The @wordpress/admin-ui Page component renders its h1 inside a flex header, so the notice ends up as an inline flex child in the header row instead of in the content area.

Is there a better/cleaner way to solve this by changing the markup of this Request Log page so this placement never happens?

Malayt04:
Thanks for the feedback! You're right that it's more changes than ideal for this task

The constraint is that the subhead "Detailed logs of every AI request" is rendered inside React by the @wordpress/admin-ui Page component. Because of that, any notice placed via PHP will always land either above the entire React app or inside the flex header (the current bug). There's no markup trick that gets it below the subtitle without being inside the React tree.

That said, if the exact placement requirement is flexible, there's a much simpler fix which makes the notice appear on the top of the page, above the entire header

1. Suppress the admin_notices version on this screen 
2. Output the notice HTML directly in render_page() before the React root 

Happy to switch to the simpler approach if you approve.

jeffpaul:
@Malayt04 an alternate PR to compare against where the notice is above could be helpful to consider which approach is most optimal given the current implementation specifics.

dkotter:
Closing in favor of #628

PR #604, opened 2026-05-22
Fix AI Status feature checklist state

ekamran:
## What?

Fixes the AI Status dashboard checklist so the "Enable a feature or experiment" step reflects whether an individual feature setting is enabled.

## Why?

The checklist currently uses the feature runtime enabled state for this step. That state also depends on the global Enable AI toggle, so when global AI is off the dashboard can show the feature step as incomplete even if a feature or experiment is already enabled in Settings > AI.

This makes the setup checklist confusing because the global AI step and the individual feature step are separate checklist items.

## How?

The dashboard widget now checks the saved individual feature options for the checklist step. The full status view still requires credentials, the global AI toggle, and at least one enabled feature setting.

## Testing Instructions

Manual testing:

1. Go to Settings > AI.
2. Enable at least one feature or experiment.
3. Turn the global Enable AI toggle off.
4. Go to Dashboard.
5. Confirm the AI Status checklist shows:
   - Configure an AI provider: complete, if a connector is configured.
   - Globally enable AI Features: incomplete.
   - Enable a feature or experiment: complete.

Automated checks run locally:

```bash
npm run test:php -- --filter AI_Status_WidgetTest
npm run lint:php -- includes/Admin/Dashboard/AI_Status_Widget.php tests/Integration/Includes/Dashboard/AI_Status_WidgetTest.php
npm run lint:php:stan -- includes/Admin/Dashboard/AI_Status_Widget.php tests/Integration/Includes/Dashboard/AI_Status_WidgetTest.php
npm run test:e2e -- tests/e2e/specs/admin/dashboard.spec.js
```

## Screenshots/Screencast

Before:

The Dashboard AI Status checklist showed "Enable a feature or experiment" as incomplete when global AI was off, even though an individual feature was enabled.

before

After:

The same checklist now shows "Enable a feature or experiment" as complete when an individual feature setting is enabled, while "Globally enable AI Features" remains incomplete.

after

## Use of AI Tools

AI assistance: Yes
Tool(s): ChatGPT / Codex

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-604-f7a0d6049c4a96eb51d90851eff10f875f2d9eeb.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Admin/Dashboard/AI_Status_Widget.php
+++ b/includes/Admin/Dashboard/AI_Status_Widget.php
@@ -62,14 +62,14 @@ class AI_Status_Widget {
 	 * @since 0.8.0
 	 */
 	public function render(): void {
-		$has_credentials = has_ai_credentials();
-		$global_enabled  = (bool) get_option( Settings_Registration::GLOBAL_OPTION, false );
-		$any_feature_on  = $this->has_any_enabled_feature();
+		$has_credentials         = has_ai_credentials();
+		$global_enabled          = (bool) get_option( Settings_Registration::GLOBAL_OPTION, false );
+		$feature_setting_enabled = $this->has_any_enabled_feature_setting();
 
-		if ( $has_credentials && $global_enabled && $any_feature_on ) {
+		if ( $has_credentials && $global_enabled && $feature_setting_enabled ) {
 			$this->render_status();
 		} else {
-			$this->render_getting_started( $has_credentials, $global_enabled, $any_feature_on );
+			$this->render_getting_started( $has_credentials, $global_enabled, $feature_setting_enabled );
 		}
 	}
 
@@ -78,11 +78,11 @@ class AI_Status_Widget {
 	 *
 	 * @since 0.8.0
 	 *
-	 * @param bool $has_credentials   Whether any AI provider credentials are configured.
-	 * @param bool $global_enabled    Whether the global features toggle is on.
-	 * @param bool $any_feature_on Whether at least one feature is enabled.
+	 * @param bool $has_credentials         Whether any AI provider credentials are configured.
+	 * @param bool $global_enabled          Whether the global features toggle is on.
+	 * @param bool $feature_setting_enabled Whether at least one feature setting is enabled.
 	 */
-	private function render_getting_started( bool $has_credentials, bool $global_enabled, bool $any_feature_on ): void {
+	private function render_getting_started( bool $has_credentials, bool $global_enabled, bool $feature_setting_enabled ): void {
 		$steps = array(
 			array(
 				'done'  => $has_credentials,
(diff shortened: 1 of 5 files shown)


includes/Admin/Dashboard/AI_Status_Widget.php
```diff
@@ -239,15 +239,16 @@ private function get_ai_connectors(): array {
...
 	 *
-	 * @return bool True if at least one feature is enabled.
+	 * @return bool True if at least one feature setting is enabled.
 	 */
-	private function has_any_enabled_feature(): bool {
+	private function has_any_enabled_feature_setting(): bool {
 		foreach ( $this->registry->get_all_features() as $feature ) {
-			if ( $feature->is_enabled() ) {
+			$option_name = 'wpai_feature_' . $feature::get_id() . '_enabled';
+			if ( (bool) get_option( $option_name, false ) ) {
```
dkotter: Probably a minor thing but in switching away from our `is_enabled` check, we do lose the filter that runs on each feature: `wpai_feature_{$this->id}_enabled`. So if someone is using that to manually enable or disable a feature, this check won't pick that up.

It seems the main concern here is our `is_enabled` check will first look at the global enable option and then the individual enable option. I think a better solution then is to break those up into separate methods so we end up with `is_enabled`, `is_globally_enabled` and `is_individually_enabled` (or something similar, not strongly opinionated on naming here).

The `is_enabled` check could then call these two new methods so it ends up functioning the same as it does today. And then here we could call the new `is_individually_enabled` so the filter still fires.

dkotter (changes requested):

includes/Admin/Dashboard/AI_Status_Widget.php
ekamran: Good catch, thanks. I updated this so the dashboard no longer reads the feature option directly.

The feature enabled logic is now split into `is_globally_enabled()` and `is_individually_enabled()`, while `is_enabled()` keeps the same effective behavior as before. The dashboard checklist now uses `is_individually_enabled()`, so the existing feature-level filters are still applied.

I also added tests for the individual feature filter behavior and for preserving the existing global + individual `is_enabled()` behavior.

dkotter (approved):

PR #612, opened 2026-05-22
Fix Abilities Explorer schema constraint validation

ekamran:
## What?

Fixes the Abilities Explorer test runner so the **Validate Input** button checks JSON schema constraints, not only basic field types.

This updates validation for:

- `minimum`
- `maximum`
- `enum`
- stricter `number` / `integer` handling

## Why?

Some ability input schemas include constraints that were displayed in the UI but not enforced by the validator.

For example, the Content Classification ability shows:

```json
"max_suggestions": {
  "type": "integer",
  "minimum": 1,
  "maximum": 10
}
```

Before this change, entering `"max_suggestions": 11` still showed the input as valid.

## How?

- Updates the PHP validation helper used by the Abilities Explorer AJAX handler.
- Updates the client-side validator used by the visible **Validate Input** button.
- Adds focused PHP coverage for integer, min/max, and enum validation.
- Adds E2E coverage for the visible Content Classification validation flow.

## Testing Instructions

1. Enable AI.
2. Enable the Abilities Explorer experiment.
3. Enable the Content Classification experiment.
4. Go to Tools > Abilities Explorer.
5. Open the test runner for `ai/content-classification`.
6. Set `max_suggestions` to `11`.
7. Click **Validate Input**.
8. Confirm the UI shows a validation error.
9. Change `max_suggestions` to `5`.
10. Click **Validate Input** again.
11. Confirm the UI shows the input is valid.

Automated checks run locally:

```bash
npm run lint:php -- includes/Experiments/Abilities_Explorer/Ability_Handler.php tests/Integration/Includes/Experiments/Abilities_Explorer/Ability_HandlerTest.php
npm run test:php -- --filter Ability_HandlerTest
npm run lint:php:stan -- includes/Experiments/Abilities_Explorer/Ability_Handler.php tests/Integration/Includes/Experiments/Abilities_Explorer/Ability_HandlerTest.php
npm run lint:js -- src/experiments/abilities-explorer/index.js tests/e2e/specs/experiments/abilities-explorer.spec.js
npm run test:e2e -- tests/e2e/specs/experiments/abilities-explorer.spec.js
npm run build
git diff --check
```

## Screenshots/Screencast

Before:
Before

After:
After

## Use of AI Tools

AI assistance: Yes

Tool(s): ChatGPT / Codex

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-612-0a4294ba15c7c89a52a6737aef8970386bd72dfd.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/Abilities_Explorer/Ability_Handler.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Handler.php
@@ -300,44 +300,9 @@ class Ability_Handler {
 					continue;
 				}
 
-				$value = $input[ $prop_name ];
-
-				if ( isset( $prop_schema['type'] ) ) {
-					$valid = self::validate_type( $value, $prop_schema['type'] );
-					if ( ! $valid ) {
-						$errors[] = sprintf(
-							'Field "%s" should be of type "%s"',
-							$prop_name,
-							$prop_schema['type']
-						);
-						continue;
-					}
-				}
-
-				if ( isset( $prop_schema['enum'] ) && is_array( $prop_schema['enum'] ) && ! in_array( $value, $prop_schema['enum'], true ) ) {
-					$errors[] = sprintf(
-						'Field "%s" must be one of: %s',
-						$prop_name,
-						implode( ', ', $prop_schema['enum'] )
-					);
-				}
-
-				if ( is_numeric( $value ) && isset( $prop_schema['minimum'] ) && $value < $prop_schema['minimum'] ) {
-					$errors[] = sprintf(
-						'Field "%s" must be at least %s',
-						$prop_name,
-						$prop_schema['minimum']
-					);
-				}
-
-				if ( ! is_numeric( $value ) || ! isset( $prop_schema['maximum'] ) || $value <= $prop_schema['maximum'] ) {
-					continue;
-				}
-
-				$errors[] = sprintf(
-					'Field "%s" must be at most %s',
-					$prop_name,
-					$prop_schema['maximum']
+				$errors = array_merge(
+					$errors,
+					self::validate_property( (string) $prop_name, $input[ $prop_name ], $prop_schema )
 				);
 			}
 		}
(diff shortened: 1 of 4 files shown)


src/experiments/abilities-explorer/index.js
```diff
@@ -296,30 +296,71 @@ import './index.scss';
...
 					function ( propName ) {
-						if ( propName in input ) {
-							const propSchema = schema.properties[ propName ];
-							const value = input[ propName ];
-
-							if ( propSchema.type ) {
-								const isValid = this.validateType(
-									value,
-									propSchema.type
+						if ( ! ( propName in input ) ) {
```
dkotter, line 303: So there is now a client-side Abilities API. I don't know off the top of my head but wondering if that can be used to validate things here rather than us maintaining our own validation?

tests/Integration/Includes/Experiments/Abilities_Explorer/Ability_HandlerTest.php
```diff
@@ -235,6 +235,79 @@ public function test_validate_input_validates_object_type() {
 		$this->assertFalse( $result['valid'] );
 	}
 
+	/**
+	 * Test validate_input validates integer type.
+	 *
+	 * @since n.e.x.t
```
dkotter: These should all be `@since x.x.x`

includes/Experiments/Abilities_Explorer/Ability_Handler.php
```diff
@@ -293,22 +293,51 @@ public static function validate_input( array $schema, array $input ): array {
...
+				if ( ! isset( $input[ $prop_name ] ) || ! is_array( $prop_schema ) ) {
 					continue;
 				}
 
-				$valid = self::validate_type( $input[ $prop_name ], $prop_schema['type'] );
-				if ( $valid ) {
+				$value = $input[ $prop_name ];
+
+				if ( isset( $prop_schema['type'] ) ) {
+					$valid = self::validate_type( $value, $prop_schema['type'] );
```
dkotter: Since we already have this `validate_type` method, wondering if it would be cleaner here to move all of this validation in there (or a newly named method) and that method can return the error messages?

dkotter (changes requested):

src/experiments/abilities-explorer/index.js
ekamran: Thanks for the review. I updated the PHP side to move the property-specific validation into a dedicated `validate_property()` helper, while keeping `validate_type()` focused on type checks. That keeps the behavior and existing messages the same, but makes `validate_input()` easier to follow.

I also updated the new test docblocks from `@since n.e.x.t` to `@since x.x.x`.

For the client-side Abilities API question: I checked this locally. Importing `@wordpress/abilities` here caused the built `ai_abilities_explorer` script to depend on the `wp-abilities` script handle, but that handle is not registered on this classic admin page. E2E caught that as a WordPress script dependency notice. Because of that, I left the existing local JS validation in place for now instead of adding script-loading changes to this PR.

Local checks run:

- `npm run test:php -- --filter Ability_HandlerTest`
- `npm run lint:php -- includes/Experiments/Abilities_Explorer/Ability_Handler.php tests/Integration/Includes/Experiments/Abilities_Explorer/Ability_HandlerTest.php`
- `npm run lint:php:stan -- includes/Experiments/Abilities_Explorer/Ability_Handler.php tests/Integration/Includes/Experiments/Abilities_Explorer/Ability_HandlerTest.php`
- `npm run build`
- `npm run test:e2e -- tests/e2e/specs/experiments/abilities-explorer.spec.js`

ekamran: I also considered another approach: making the **Validate Input** button call a validate-only AJAX endpoint and reuse `Ability_Handler::validate_input()` as the single source of truth.

That would avoid duplicating schema validation rules in JS, and it would also avoid importing `@wordpress/abilities`, so it would not introduce the `wp-abilities` script dependency issue on this classic admin page.

The tradeoff is that it expands the scope a bit: it would add a new AJAX action, require the usual nonce/capability handling, and make the validation button async with loading/error states. So I kept this PR smaller by extending the existing client-side validation and aligning the PHP validation path.

Happy to switch to the server-backed validation approach if you think that is cleaner for this screen.

dkotter: So you can import the client-side Abilities API without it becoming a script dependency, see how we do that at https://github.com/WordPress/ai/blob/35a4e1c0086628ce5717a5b8cd4abe2e426b7174/src/utils/run-ability.ts#L38.

I wouldn't introduce a server-side approach here but I do think long-term we'd be better off switching everything here (validation and requests) to use the client-side Abilities API, though that should be a separate PR.

Also seeing that the error messages we return here aren't properly translated, though that's an existing issue as well. I'm fine with merging this in as-is but definitely some follow-ups we can do here if desired

dkotter (approved):

PR #626, opened 2026-05-27
Fix WordPress 7.0 preview and test configs

### LINKED ISSUE #623: Update playground and blueprints to use WP 7.0
### Description

Playground appears to still be using 7.0 RC5, when it should be using the released version of 7.0.  I suspect this may be from:
https://github.com/WordPress/ai/blob/485c88463bf6b6ed6bcd60cc258393455315e478/.wordpress-org/blueprints/blueprint.json#L6
and:
https://github.com/WordPress/ai/blob/485c88463bf6b6ed6bcd60cc258393455315e478/.github/workflows/pr-playground-preview.yml#L85-L87

So perhaps updating there will ensure our Live Preview on WPORG and the Playground PR environments properly load 7.0?

Test on a PR or on WPORG live preview and see WP 7.0 RC5 loads instead of 7.0

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

mehul0810:
## What?

Closes https://github.com/WordPress/ai/issues/623

Updates the Playground, Plugin Check, PHPUnit, and wp-env configs now that WordPress 7.0 is available as the latest stable release.

## Why?

The AI plugin requires WP 7.0, but some configs still contained temporary pre-release or trunk settings that were added before WP 7.0 was released.

This could cause the AI plugin Playground preview to load WP 7.0 RC5 instead of the final WP 7.0 release. Now that 7.0 is released, these configs can use stable release defaults again while keeping explicit trunk coverage where it is still useful.

## How?

- Pins the WordPress.org Playground blueprint to WordPress `7.0`.
- Pins the PR Playground preview blueprint to WordPress `7.0`.
- Restores Plugin Check to run against `latest`.
- Restores the PHPUnit matrix to run against both `latest` and `trunk`.
- Restores `.wp-env` configs to `core: null`, which uses the latest production WordPress release.

### Use of AI Tools

AI assistance: Yes  
Tool(s): Codex  
Model(s): GPT-5  
Used for: Identifying the affected config files, preparing the local patch, and drafting validation steps. The changes were reviewed and confirmed by me.

## Testing Instructions
- Confirm blueprint.json after running the blueprint on WP Playground

## Screenshots or screencast

Not applicable. This is a configuration-only change.

## Changelog Entry

> Fixed - Playground and test configs now target the stable WordPress 7.0/latest release instead of temporary pre-release or trunk defaults where appropriate.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-626-b09a37b48b91c750a854ec056fd78c3ba07ce155.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/.github/workflows/pr-playground-preview.yml
+++ b/.github/workflows/pr-playground-preview.yml
@@ -83,7 +83,7 @@ jobs:
 
           const blueprint = {
             preferredVersions: {
-              wp: 'beta',
+              wp: '7.0',
             },
             steps: [
               {
--- a/.github/workflows/pr-playground-preview.yml
+++ b/.github/workflows/pr-playground-preview.yml
@@ -83,7 +83,7 @@ jobs:
 
           const blueprint = {
             preferredVersions: {
-              wp: '7.0',
+              wp: 'latest',
             },
             steps: [
               {
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -172,11 +172,11 @@ jobs:
       fail-fast: false
       matrix:
         php: ['8.4', '8.3', '8.2', '8.1', '8.0', '7.4']
-        wp: [trunk]
+        wp: [latest, trunk]
         coverage: [false]
         include:
           - php: '8.4'
-            wp: trunk
+            wp: latest
             coverage: true
     env:
       WP_ENV_PHP_VERSION: ${{ matrix.php }}
--- a/.wordpress-org/blueprints/blueprint.json
+++ b/.wordpress-org/blueprints/blueprint.json
@@ -3,7 +3,7 @@
 	"landingPage": "/wp-admin/options-general.php?page=ai-wp-admin",
 	"preferredVersions": {
 		"php": "7.4",
-		"wp": "7.0"
+		"wp": "latest"
 	},
 	"features": {
 		"networking": true
--- a/.wordpress-org/blueprints/blueprint.json
+++ b/.wordpress-org/blueprints/blueprint.json
@@ -3,7 +3,7 @@
 	"landingPage": "/wp-admin/options-general.php?page=ai-wp-admin",
 	"preferredVersions": {
 		"php": "7.4",
-		"wp": "beta"
+		"wp": "7.0"
 	},
 	"features": {
 		"networking": true
--- a/.github/workflows/plugin-check.yml
+++ b/.github/workflows/plugin-check.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Run plugin check
         uses: wordpress/plugin-check-action@18573eb38f13543484a1f095672dae3849a4fbb0 # v1.1.6
         with:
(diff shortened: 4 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -172,11 +172,11 @@ jobs:
       fail-fast: false
       matrix:
         php: ['8.4', '8.3', '8.2', '8.1', '8.0', '7.4']
-        wp: [trunk]
+        wp: [latest, trunk]
         coverage: [false]
         include:
           - php: '8.4'
-            wp: trunk
+            wp: latest
```
jeffpaul, line 179: @dkotter should this be trunk and latest?

.github/workflows/pr-playground-preview.yml
```diff
@@ -83,7 +83,7 @@ jobs:
 
           const blueprint = {
             preferredVersions: {
-              wp: 'beta',
+              wp: '7.0',
```
dkotter: We shouldn't specify a version here as this will get out of date whenever a new version is released. I think we'd want `latest`

.github/workflows/test.yml
dkotter: This is fine to just be `latest`. With the change above, our tests will run on both `latest` and `trunk`. This controls which version we run our full coverage report and there's no need to run that on both versions

.wordpress-org/blueprints/blueprint.json
```diff
@@ -3,7 +3,7 @@
 	"landingPage": "/wp-admin/options-general.php?page=ai-wp-admin",
 	"preferredVersions": {
 		"php": "7.4",
-		"wp": "beta"
+		"wp": "7.0"
```
dkotter: Similar to above, this should be `latest` I think

dkotter (changes requested):

.wordpress-org/blueprints/blueprint.json
dkotter: Changed in https://github.com/WordPress/ai/pull/626/changes/b09a37b48b91c750a854ec056fd78c3ba07ce155

.github/workflows/pr-playground-preview.yml
dkotter: Changed in https://github.com/WordPress/ai/pull/626/changes/b09a37b48b91c750a854ec056fd78c3ba07ce155

PR #628, opened 2026-05-27
fix: add inline class to connector approval admin notice for proper l…

### LINKED ISSUE #596: Connector Approvals request misplaced on Request Logs screen

Malayt04:
## What?

Closes: #596 

Fixes the connector approval notice overlapping the header on the AI Request Logs screen when both the Connector Approvals and AI Request Logging experiments are enabled.

## Why?

WordPress Core's `common.js` automatically relocates `.notice` elements to after the first `h1` in `.wrap`. The `@wordpress/admin-ui` `Page` component renders its `h1` inside a flex header container, so the notice ends up as an inline flex child in the header row.

## How?

Added the `inline` class to the notice output in `Admin_Notice::render()`. WordPress Core's `common.js` explicitly skips elements with the `inline` class when relocating notices, so the notice stays in its natural output position instead of being moved into the flex header.

### Use of AI Tools

AI assistance: Yes  
Tool(s): Antigravity  
Used for: Deciding most optimal approach 

### Before
Screenshot 2026-05-27 at 5 08 38 PM

### After
Screenshot 2026-05-27 at 5 09 13 PM

## Changelog Entry

> Fixed - Connector approval notice no longer overlaps the page header on the AI Request Logs screen.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22beta%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-628-4872fc61d02e0767b7c9d87eb96197c7c788df5d.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Connector_Approval/Admin_Notice.php
+++ b/includes/Connector_Approval/Admin_Notice.php
@@ -147,7 +147,7 @@ final class Admin_Notice {
 			self::DISMISS_NONCE
 		);
 
-		$class = 'notice notice-warning';
+		$class = 'notice notice-warning ai-connector-approval-notice';
 		if ( $screen && 'tools_page_ai-request-logs' === $screen->id ) {
 			$class .= ' inline';
 		}
--- a/includes/Connector_Approval/Admin_Notice.php
+++ b/includes/Connector_Approval/Admin_Notice.php
@@ -147,8 +147,14 @@ final class Admin_Notice {
 			self::DISMISS_NONCE
 		);
 
+		$class = 'notice notice-warning';
+		if ( $screen && 'tools_page_ai-request-logs' === $screen->id ) {
+			$class .= ' inline';
+		}
+
 		printf(
-			'<div class="notice notice-warning inline"><p>%s <a href="%s">%s</a> &middot; <a href="%s">%s</a></p></div>',
+			'<div class="%s"><p>%s <a href="%s">%s</a> &middot; <a href="%s">%s</a></p></div>',
+			esc_attr( $class ),
 			esc_html(
 				sprintf(
 					/* translators: %d: number of pending approval requests. */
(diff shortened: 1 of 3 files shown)


jeffpaul:
@dkotter this is probably a good visual fix, so as long as this appears less hacky solutionwise then perhaps we merge this instead of #598?

dkotter:
I think this is the better fix but my main concern here is we're adding that `inline` class to the notice on all pages, which changes the rendering of that admin notice for all pages. I think we should scope that to just the Request Logging page.

For example, here's the before and after on the Dashboard screen:

| Before | After |
| ---- | ---- |
| Screenshot 2026-05-27 at 8 33 04 AM  | Screenshot 2026-05-27 at 8 33 13 AM |

Malayt04:
I've updated the implementation to scope the .inline class conditionally so it is only applied on the Request Logs page. On other screens, the default relocation behavior remains completely untouched.

dkotter (dismissed):

dkotter (approved):

PR #650, opened 2026-06-03
Add bulk "Generate AI Summary" action to the posts list table

### LINKED ISSUE #614: Add support for bulk summary generation

Currently, the Content Summarization function is only accessible in the panel of the post editor. As a result, it is currently difficult to generate summaries for several existing posts at a time, as this requires opening the editor for each post and waiting for the summary to be generated.

Consider adding a button that will generate summary and save it to both post content and meta at the bottom of `/wp-admin/edit.php`, within the 'Bulk Actions' dropdown list.

prasadkarmalkar:
## What?

Closes #614

## Why?

AI summary generation is only available from within the block editor for a single post at a time. This PR adds a bulk workflow that lets users generate summaries for multiple posts in a single action directly from the list table.

## How?

References from the existing **Alt Text Generation** bulk action pattern:

1. **PHP (three new methods on `Summarization` experiment):**
   - `register_bulk_action()` — appends `"Generate AI Summary"` to the bulk actions menu.
   - `handle_bulk_action()` — validates capabilities (`edit_posts` + per-post `edit_post`), then redirects back to `edit.php` with `wpai_bulk_summary=1&wpai_post_ids=1,2,3` query args.
   - `maybe_enqueue_bulk_assets()` — detects the redirect on page reload, sanitizes post IDs, resolves the correct REST base from the current post type (so pages and custom post types work), and enqueues the bulk script with localized data (`window.aiSummarizationBulkData`).

2. **JavaScript (`src/experiments/summarization/bulk.ts`, new file):**
   - On `DOMContentLoaded`, reads `window.aiSummarizationBulkData` and processes posts sequentially.
   - Checks provider availability; shows an admin notice error if no AI connector is configured.
   - For each post: calls `runAbility('ai/summarization', { context: postId })` (the ability fetches post content from the DB itself), fetches raw block content via the REST API, strips any pre-existing summary block, prepends the new one, and saves both `content` and `meta.ai_generated_summary` in a single `apiFetch` POST.
   - Displays a live-updating progress notice and a final success/failure summary.
   - Cleans up `wpai_bulk_summary` and `wpai_post_ids` from the URL via `history.replaceState`.

3. **Webpack (`webpack.config.js`):** Adds `experiments/summarization-bulk` entry pointing to `bulk.ts`.

### Use of AI Tools

AI assistance: Yes
Tool(s): GitHub Copilot
Model(s): Claude Sonnet 4.6
Used for: Initial code skeleton. Final implementation reviewed and edited by me.

## Testing Instructions

**Prerequisites:** An AI connector must be configured in **Settings → Connectors**.

1. Go to **Posts → All Posts** (or **Pages → All Pages**).
2. Select two or more posts using the checkboxes.
3. Open the **Bulk Actions** dropdown — confirm **"Generate AI Summary"** is listed.
4. Select **"Generate AI Summary"** and click **Apply**.
5. The page reloads; a progress notice appears: _"Generating summaries: 0 / N…"_
6. Wait for processing to complete — the notice updates to show success/failure counts.
7. Open one of the processed posts in the block editor and confirm:
   - An **AI Summary** block group is prepended to the content.
   - The post meta `ai_generated_summary` is added
8. Run the bulk action again on the same posts — confirm the old summary block is replaced, not duplicated.
9. Repeat on the **Pages** list table to confirm post-type handling works.
10. **Negative test:** Disable the summarization experiment in **Settings → AI** and confirm the bulk action is no longer listed.

## Screenshots or screencast

https://github.com/user-attachments/assets/2dd245fb-4927-43ad-86a5-e683a21cc91d

## Changelog Entry

> Added - Bulk "Generate AI Summary" action to the posts and pages list table, enabling AI summary generation for multiple posts at once without opening the block editor.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-650-a6989417871c21e64f9edd7f709eed68f74228e0.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Experiments/Summarization/Summarization.php
```diff
@@ -119,6 +133,102 @@ public function enqueue_assets(): void {
 		);
 	}
 
+	/**
+	 * Adds the "Generate AI Summary" option to the posts list bulk actions menu.
```
dkotter:
```suggestion
	 * Adds the "Generate Summary" option to the posts list bulk actions menu.
```

includes/Experiments/Summarization/Summarization.php
```diff
@@ -119,6 +133,102 @@ public function enqueue_assets(): void {
...
+	 *
+	 * @param array<string, string> $actions The existing bulk actions.
+	 * @return array<string, string> The modified bulk actions.
+	 */
+	public function register_bulk_action( array $actions ): array {
+		if ( ! $this->is_enabled() ) {
+			return $actions;
+		}
+
+		$actions['wpai_generate_summary'] = __( 'Generate AI Summary', 'ai' );
```
dkotter:
```suggestion
		$actions['wpai_generate_summary'] = __( 'Generate Summary', 'ai' );
```

includes/Experiments/Summarization/Summarization.php
```diff
@@ -119,6 +133,102 @@ public function enqueue_assets(): void {
...
+
+		return $actions;
+	}
+
+	/**
+	 * Handles the "Generate AI Summary" bulk action by redirecting with selected post IDs.
```
dkotter:
```suggestion
	 * Handles the "Generate Summary" bulk action by redirecting with selected post IDs.
```

src/experiments/summarization/bulk.ts
```diff
@@ -0,0 +1,238 @@
...
+			// translators: 1: number processed so far, 2: total number of posts
+			__( 'Generating summaries: %1$d / %2$d\u2026', 'ai' ),
+			0,
+			total
+		)
+	);
+
+	for ( const id of postIds ) {
+		try {
+			// Generate summary the ability fetches the post content from the DB
```
dkotter:
```suggestion
			// The summarization ability fetches the post content from the DB
```

dkotter (changes requested):
Overall this looks good, left a few minor comments. In addition to those, we should look to get tests added here to cover these changes. Maybe some unit tests though likely won't provide much value but definitely want E2E tests to cover these changes

prasadkarmalkar:
Hi @dkotter , I have updated the above requested changes and also added e2e and integration test cases (Took reference from other test cases). Let me know if this is fine or need any changes.

includes/Experiments/Summarization/Summarization.php
```diff
@@ -54,6 +54,20 @@ public function register(): void {
...
 		add_action( 'enqueue_block_assets', array( $this, 'enqueue_block_assets' ) );
+
+		/**
+		 * Filters the post types that support the bulk "Generate Summary" action.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param string[] $post_types Post type slugs.
+		 */
+		$bulk_post_types = (array) apply_filters( 'wpai_summarization_bulk_post_types', array( 'post', 'page' ) );
```
dkotter: Thinking about this a bit more, any thoughts on making this more general for the entire feature? Right now this can lead to a difference between which post types can generate summaries in the editor vs generate summaries in bulk. It feels like we should maintain a single source of truth for which post types are supported in both cases, or just default to allowing all public post types in both scenarios

prasadkarmalkar: Yes, I checked all other experimental features as well. Currently almost all the features support all public post types. I think we should also move this to all public post types instead of adding a filter here for now. I checked how other features are validating the CPT and they are using `current_screen` action for it.

Here is a snippet I am thinking of using instead of direclty using filter. Let me know if this approach looks good. I will update the PR and test cases as well.

```
add_action( 'current_screen', array( $this, 'register_bulk_action_hooks' ) );

public function register_bulk_action_hooks( WP_Screen $screen ): void {
    if ( ! $screen || 'edit' !== $screen->base ) {
        return;
    }
    add_filter( "bulk_actions-{$screen->id}", array( $this, 'register_bulk_action' ) );
    add_filter( "handle_bulk_actions-{$screen->id}", array( $this, 'handle_bulk_action' ), 10, 3 );
}
```

dkotter: So this would likely work but not sure I love it. And thinking this through even more (sorry, should have thought of all of this initially), the underlying Summarization Ability only works on post types that are allowed in REST requests, so ideally we should add the same check here instead of looking at public post types (and likely some standardization we could do across the entire plugin for this).

I'd suggest something along these lines (note, untested), which ensures we only add support for post types that are allowed in REST requests and we only register the hooks when needed, instead of having those always register:

```php
add_action( 'load-edit.php', array( $this, 'register_bulk_action_hooks_for_screen' ) );

public function register_bulk_action_hooks_for_screen(): void {
    $post_type = isset( $_GET['post_type'] )
        ? sanitize_key( wp_unslash( $_GET['post_type'] ) )
        : 'post';

    if ( ! $this->post_type_supports_bulk_summarization( $post_type ) ) {
        return;
    }

    add_filter( "bulk_actions-edit-{$post_type}", array( $this, 'register_bulk_action' ) );
    add_filter( "handle_bulk_actions-edit-{$post_type}", array( $this, 'handle_bulk_action' ), 10, 3 );
}

private function post_type_supports_bulk_summarization( string $post_type ): bool {
    $post_type_obj = get_post_type_object( $post_type );

    return $post_type_obj
        && ! empty( $post_type_obj->show_in_rest )
        && ! empty( $post_type_obj->show_ui )
        && 'attachment' !== $post_type;
}
```

May even be worth making `post_type_supports_bulk_summarization` a new helper function in `helpers.php` where we pass in the feature name. This could then be used in other features in the future

prasadkarmalkar: Updated the code as per suggestion. Moved the check into a new post_type_supports_bulk_ai_action( string $post_type, string $feature_id ) helper in helpers.php rather than keeping it as a private method on the class. It uses a switch on $feature_id, so other features can be added there.
The load-edit.php hook approach you suggested is also in place. Updated tests in both SummarizationTest and HelpersTest to cover the new helper.

src/experiments/summarization/bulk.ts
```diff
@@ -0,0 +1,238 @@
...
+ * Serializes a summary string into a WordPress block markup string.
+ *
+ * Mirrors the block the editor inserts: a core/group block with the
+ * aiGeneratedSummary attribute set to true, containing one core/paragraph
+ * block per non-empty paragraph in the summary.
+ *
+ * @param summary The plain-text summary from the AI.
+ * @return Serialized block markup.
+ */
+function serializeSummaryBlock( summary: string ): string {
```
dkotter: This works but noting if we ever change the markup of this summary block, we'll need to ensure that changes in multiple places. If possible, would be great to have a single function that generates this markup that we can use everywhere. Right now we use `createBlock` in `useSummaryGeneration.ts` and then this file does it slightly differently. Is there a possibility to add a block creation method to `useSummaryGeneration.ts` that can be used everywhere?

src/experiments/summarization/bulk.ts
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Removes any existing AI-generated summary group block from serialized post content.
+ *
+ * @param content Serialized block content.
+ * @return Content with the summary block removed.
+ */
+function removeExistingSummaryBlock( content: string ): string {
+	return content
+		.replace(
+			/<!-- wp:group \{[^}]*"aiGeneratedSummary":true[^}]*\} -->[\s\S]*?<!-- \/wp:group -->\n*/,
```
dkotter: Seems like this regex could likely break, if for instance the attribute order changes. This is another place that would be great to standardize this in `useSummaryGeneration.ts` so we parse this block out the same in all contexts (if possible). Right now we have code in that file that parses content into blocks and then loops through all blocks to look at the block name and attributes that we may be able to use here

includes/helpers.php
```diff
@@ -686,6 +687,31 @@ function is_connector_plugin_active( array $connector_data ): bool {
...
+/**
+ * Determines whether a post type supports bulk AI actions for a given feature.
+ *
+ * @since x.x.x
+ *
+ * @param string $post_type  The post type slug to check.
+ * @param string $feature_id The feature identifier (e.g. 'summarization').
+ * @return bool True if the post type supports bulk AI actions for the feature.
+ */
+function post_type_supports_bulk_ai_action( string $post_type, string $feature_id ): bool {
```
dkotter: I would suggest removing `ai` from the naming here

includes/helpers.php
```diff
@@ -686,6 +687,31 @@ function is_connector_plugin_active( array $connector_data ): bool {
...
+	// Check if the post type is registered and supports REST API and UI.
+	$base_supported = $post_type_obj
+		&& ! empty( $post_type_obj->show_in_rest )
+		&& ! empty( $post_type_obj->show_ui );
+
+	switch ( $feature_id ) {
+		case Summarization::get_id():
+			return $base_supported && 'attachment' !== $post_type;
+		default:
+			return false;
```
dkotter: Any reason to not return `$base_supported` as our default here?

dkotter (changes requested):

includes/helpers.php
prasadkarmalkar: Fixed

includes/helpers.php
prasadkarmalkar: Used $base_supported as default return for any post type

src/experiments/summarization/bulk.ts
prasadkarmalkar: Created a utils.js for common utilities between `bulk.ts` and `useSummaryGeneration.ts`, and created a common `createSummaryBlock` function for it.

src/experiments/summarization/bulk.ts
prasadkarmalkar: Updated to use the flow as per useSummaryGeneration.ts and added the common function `findSummaryBlock` inside `utils.ts`

prasadkarmalkar: There is an issue with using `parse` existing content. Trying to find a solution as parse is giving an empty array .

prasadkarmalkar: Issue is fixed. as the `parse()/createBlock()` need block types registered first. Added a one-time `registerCoreBlocks()` call in bulk.ts

dkotter (changes requested):
Changes look good here but looks like the addition of the `@wordpress/block-library` is breaking most of our automated checks. Looks like it's pulling in a newer version of React maybe?

prasadkarmalkar:
@dkotter , Fixed the issue it was due to different version of node while installing the dependency on my local.

dkotter (changes requested):
Tested and is working well.

I think the last thing here is we recently added minimum content length requirements for most of our features and that doesn't seem to apply here for bulk generation. I would suggest we implement that for bulk generation to match individual generation

dkotter (dismissed):

dkotter (approved):

PR #653, opened 2026-06-03
Fix canceled meta description suggestion state

ekamran:
## What?

Fix the Meta Description modal so a canceled regenerated suggestion does not appear when reopening the modal through “Edit description”.

## Why?

When a post already had a saved meta description, regenerating a new suggestion and canceling it left that suggestion in component state.

After that, reopening “Edit description” could show the canceled suggestion in the textarea instead of the saved meta description shown in the sidebar.

## How?

The panel now tracks whether the modal is opened for editing the saved description or for generating a suggestion.

Suggestions are only passed into the modal while it is in generate mode, and stale suggestion state is cleared when leaving the modal.

## Testing Instructions

Automated checks run locally:

```bash
npm run lint:js -- src/experiments/meta-description/components/useMetaDescription.ts src/experiments/meta-description/components/MetaDescriptionPanel.tsx tests/e2e/specs/experiments/meta-description.spec.js
npm run typecheck
npm run build
WP_BASE_URL=http://localhost:8888 npm run test:e2e -- tests/e2e/specs/experiments/meta-description.spec.js
```

Manual verification:

1. Open a post with a saved meta description.
2. Click “Regenerate meta description”.
3. Wait for the generated suggestion.
4. Click “Cancel”.
5. Click “Edit description”.
6. Confirm the modal textarea shows the saved meta description, not the canceled suggestion.

## Screenshots/Screencast

Before:

BEFORE

After:

AFTER

## Use of AI Tools

AI assistance: Yes  
Tool(s): ChatGPT / Codex  

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-653-534bcca0514176f7f7a2126a9a7b60fc9920e725.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/meta-description/components/MetaDescriptionPanel.tsx
+++ b/src/experiments/meta-description/components/MetaDescriptionPanel.tsx
@@ -36,9 +36,6 @@ export default function MetaDescriptionPanel(): React.JSX.Element {
 
 	const [ isModalOpen, setIsModalOpen ] = useState( false );
 	const [ editableText, setEditableText ] = useState( '' );
-	const [ modalMode, setModalMode ] = useState< 'edit' | 'generate' >(
-		'generate'
-	);
 
 	const hasDescription =
 		currentDescription && currentDescription.trim().length > 0;
@@ -51,7 +48,6 @@ export default function MetaDescriptionPanel(): React.JSX.Element {
 			if ( ! ensureProviderAvailable() ) {
 				return;
 			}
-			setModalMode( 'generate' );
 			setIsModalOpen( true );
 			await generateDescription();
 			return;
@@ -62,7 +58,6 @@ export default function MetaDescriptionPanel(): React.JSX.Element {
 
 	const handleOpenEditModal = () => {
 		clearSuggestion();
-		setModalMode( 'edit' );
 		setEditableText( currentDescription );
 		setIsModalOpen( true );
 	};
@@ -72,13 +67,7 @@ export default function MetaDescriptionPanel(): React.JSX.Element {
 		if ( ! ensureProviderAvailable() ) {
 			return;
 		}
-		setModalMode( 'generate' );
 		setIsModalOpen( true );
-		await generateDescription();
-	};
-
-	const handleGenerate = async () => {
-		setModalMode( 'generate' );
 		await generateDescription();
 	};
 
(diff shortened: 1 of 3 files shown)


src/experiments/meta-description/components/MetaDescriptionPanel.tsx
```diff
@@ -31,10 +31,14 @@ export default function MetaDescriptionPanel(): React.JSX.Element {
 		ensureProviderAvailable,
 		generateDescription,
 		applyDescription,
+		clearSuggestion,
 	} = useMetaDescription();
 
 	const [ isModalOpen, setIsModalOpen ] = useState( false );
 	const [ editableText, setEditableText ] = useState( '' );
+	const [ modalMode, setModalMode ] = useState< 'edit' | 'generate' >(
```
dkotter: Is this needed? It seems like we only use the value of `modalMode` to determine if we should use the value of `suggestion` or instead use `null`. But with the other changes in here (running `clearSuggestion`), the value of `suggestion` will already be `null` when we open the modal and thus seems like this code isn't needed

dkotter (changes requested):

src/experiments/meta-description/components/MetaDescriptionPanel.tsx
ekamran: Good point, thanks. I simplified this by removing the extra `modalMode` state and the wrapper around `generateDescription`.

The fix now relies on clearing the stale suggestion before opening the edit flow and when closing the modal, so the existing `suggestion` state is enough to prevent the canceled generated value from appearing in “Edit description”.

I reran the focused checks locally after the change:

```bash
npm run lint:js -- src/experiments/meta-description/components/useMetaDescription.ts src/experiments/meta-description/components/MetaDescriptionPanel.tsx tests/e2e/specs/experiments/meta-description.spec.js
npm run typecheck
git diff --check
WP_BASE_URL=http://localhost:8888 npm run test:e2e -- tests/e2e/specs/experiments/meta-description.spec.js

dkotter (approved):

PR #658, opened 2026-06-03
Check REST visibility for content resizing posts

yusufhay:
## What?
Adds a REST visibility check to the Content Resizing ability permission callback when a post ID is supplied.

## Why?
Similar text abilities already prevent execution for posts whose post type is not exposed in REST. Content Resizing checked whether the user could edit the post, but did not apply that post type visibility check, leaving the ability inconsistent with the rest of the plugin.

## How?
After confirming the post exists and the current user can edit it, the permission callback now checks the post type and returns `false` if the post type is missing or not exposed in REST.

### Use of AI Tools
AI assistance: Yes
Tool(s): Codex / ChatGPT
Model(s): GPT-5
Used for: Comparing permission callbacks across similar abilities and drafting the consistency fix. I reviewed the change and PR description before submitting.

## Testing Instructions
1. Install and activate this PR build of the AI plugin.
2. Enable the Content Resizing experiment.
3. Open a normal REST-exposed post type, such as a post, and confirm content resizing still works for editable content.
4. Register or use a post type that is not exposed in REST.
5. Attempt to run the Content Resizing ability with a post ID from that post type.
6. Confirm permission is denied instead of allowing the ability to run.

## Screenshots or screencast
Not applicable. This PR does not change the UI.

## Changelog Entry

> Fixed - Restrict content resizing to REST-exposed post types when a post ID is provided.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-658-ef4fe4ce1cdda7bcd48d701e5451b37cabf3638e.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -169,6 +169,19 @@ protected function permission_callback( $args ) {
 					esc_html__( 'You do not have permission to run AI refinements on this post.', 'ai' )
 				);
 			}
+
+			// Ensure the post type is allowed in REST endpoints.
+			$post_type = get_post_type( $post_id );
+
+			if ( ! $post_type ) {
+				return false;
+			}
```
dkotter: We already have a check above that returns an error if the post can't be found. Is there ever a situation where a valid post is found but the post type isn't? Wondering if we can just remove these lines

dkotter (changes requested):

includes/Abilities/Content_Resizing/Content_Resizing.php
yusufhay: Good point. A valid post should already imply a valid post type here, so I removed the redundant `get_post_type()` false guard and the extra `! $post_type_obj` check. The permission callback now only checks whether the post type is exposed in REST.

dkotter (changes requested):
There are a handful of PHPCS and PHPStan errors here that need cleaned up before we merge this

dkotter (approved):

Issue #589, opened 2026-05-20
a11y: Focus is lost on many button actions


t-hamano:
## What problem does this address?

There are many issues where focus is lost when a button is clicked. I believe the reasons for focus loss and how to fix them are as follows:

- The button cannot be focused because the `disabled` prop is `true`. This can be addressed by adding the `accessibleWhenDisabled` prop.
- When the button is clicked, it disappears. In this case, we should explicitly focus on the related interactive element.

## What is your proposed solution?

The following is a list of potential issues detected by the AI. We need to evaluate whether these are actual problems and make corrections.

### A. `disabled` exists + no `accessibleWhenDisabled`

| Button | Location |
|---|---|
| "Yes, Purge All" | SettingsPanel.tsx:45 — Danger Zone of AI Request Logs settings |
| "Cancel" (delete confirmation) | SettingsPanel.tsx:54 — Same as above |
| "Generate featured image" | GenerateFeaturedImage.tsx:80 — Editor Featured Image panel |
| "Generate Summary" (ToolbarButton) | SummarizationBlockControls.tsx:49 — Block Controls |
| "Generate Summary" | SummarizationPlugin.tsx:84 — Post Status Info Panel |
| "Generate/Regenerate" (Toolbar) | TitleToolbar.tsx:208 — Title Toolbar |
| "Generate/Regenerate" (Modal) | TitleToolbar.tsx:245 — Title regeneration modal |
| "Suggest Tags/Categories" | SuggestionPanel.tsx:58 — Content Classification Panel |
| "Suggest again" | SuggestionPanel.tsx:150 — Same as above |
| ToggleControl (Connector approval) | ApprovalMatrixCard.tsx:98 — Connector Approval Matrix |

### B. Conditional rendering causing full disappearance

| Button | Location |
|---|---|
| "Generate/Regenerate Alt Text" | AltTextControls.tsx:239 — Image Block Inspector |
| "Generate/Regenerate Alt Text" | MediaEditorAltTextControl.tsx:113 — Media Editor |
| "Apply Editorial Updates" | EditorialUpdatesPlugin.tsx:28 — Post Status Info |
| "Generate/Regenerate excerpt" (panel) | ExcerptGeneration.tsx:42 — Excerpt Panel |
| "Generate/Regenerate excerpt" (inline) | ExcerptInlineButton.tsx:24 — Excerpt Link Area |
[...]

PR #663, opened 2026-06-04
[Content Resizing]: Fix focus loss when regenerating content resizing suggestions

yogeshbhutkar:
## What?
See https://github.com/WordPress/ai/issues/589

Fixes a focus-loss issue in the Content Resizing modal where clicking **Regenerate** after a successful generation could move focus behind the modal.

## Why?

The previous loading state replaced the modal content while regeneration was running. Since the focused `Regenerate` button disappeared from the DOM, keyboard focus could escape to the background page, leaving keyboard and screen reader users unable to tab back into the modal.

## How?

- Keeps the modal structure and action buttons mounted during regeneration.
- Renders an inline loading state inside the suggested-content panel instead of replacing the whole modal body.
- Disables the modal actions while loading, with `accessibleWhenDisabled` on `Regenerate` so focus remains stable.
- Adds a skeleton loading treatment with reduced-motion support.

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.5
Used for: Initial code skeleton; final implementation and tests were reviewed and edited by me.

## Testing Instructions
1. Create a new post.
2. Add a Paragraph block with enough text to resize.
3. Select the Paragraph block and open the **Resize Content** toolbar menu.
4. Choose **Shorten**, **Expand**, or **Rephrase**.
5. Wait for the suggested replacement modal to finish generating.
6. Click **Regenerate**.
7. While regeneration is in progress and after it completes, press `Tab` several times.

## Screencast

https://github.com/user-attachments/assets/b62902cb-2254-4f9d-b51c-5653a325bbe8

## Changelog Entry

> Fixed - Focus lost in generating state.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-663-bbf4d10971cfa1326f71f396145c978de0a4c622.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/content-resizing/components/ContentResizingToolbar.tsx
+++ b/src/experiments/content-resizing/components/ContentResizingToolbar.tsx
@@ -15,7 +15,13 @@ import {
 } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { store as blockEditorStore } from '@wordpress/block-editor';
-import { useState, useCallback, useMemo } from '@wordpress/element';
+import {
+	useState,
+	useCallback,
+	useMemo,
+	useRef,
+	useLayoutEffect,
+} from '@wordpress/element';
 import { __, _n, sprintf } from '@wordpress/i18n';
 import { store as noticesStore } from '@wordpress/notices';
 import { store as editorStore } from '@wordpress/editor';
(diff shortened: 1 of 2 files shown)


dkotter (changes requested):
Overall this looks good to me but it is a slight change from our current UI and there is one potential problem I see.

Right now when a suggestion is being generated, we open the modal and show just a loading state. In this PR, we open the modal, we show the original content and then show a loading state beneath that. On smaller screens or when the original content is long, the loading state isn't in the viewport so it's not clear anything is happening. Ideally we auto scroll to the bottom of the modal so the loading state is always immediately visible.

yogeshbhutkar:
Thanks for the review. This should be fixed with https://github.com/WordPress/ai/pull/663/commits/bbf4d10971cfa1326f71f396145c978de0a4c622.

https://github.com/user-attachments/assets/5eb0d057-9348-47cb-9481-7f5c45192f15

dkotter (approved):

PR #669, opened 2026-06-04
AI Request Logs: column “move left / move right” has no effect

### LINKED ISSUE #667: AI Request Logs: column “move left / move right” has no effect
### Description

On AI Request Logs, each table column header exposes DataViews controls to move the column left or right. Using those options does not change column order—the layout always snaps back to the default order.

### Expected Result
After “move left” or “move right”, columns stay in the new order (at least for the current session).

### Actual Result
Column order never changes from the default: Time → Operation → Provider → Tokens → Duration → Status.

### Cause
`In src/admin/ai-request-logs/components/LogsTable.tsx`, onChangeView runs sortFieldsByCanonicalOrder() on every view update. That re-sorts view.fields back to DEFAULT_VIEW_FIELDS, undoing DataViews’ reordering immediately after the user moves a column.

### Suggested fix
Persist nextView.fields as-is when the user reorders columns; only apply canonical ordering when initializing or when normalizing visibility, not on every onChangeView. Optionally persist column order in localStorage (filters/search already persist via logsQuery).

1) Install/activate the AI plugin with AI request logging enabled.
2) Generate at least one log entry so the table has data.
3) Open AI Request Logs in wp-admin.
4) Open the column header menu on any column (e.g. Provider).
5) Choose Move left or Move right.
6) Observe that column order does not change.
7) Repeat on another column—order remains the default.

Image

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

Trushiv04:
## What?
Closes #667

Fixes the AI Request Logs table so that "move left / move right" column reordering persists instead of snapping back to the default order.

## Why?
In `LogsTable.tsx`, `onChangeView` ran `sortFieldsByCanonicalOrder()` on every view update, which re-sorted the fields back to `DEFAULT_VIEW_FIELDS` immediately after the user reordered a column. As a result, column order changes had no effect.

## How?
Removed the canonical resort from `onChangeView` so the user's chosen field order (`nextView.fields`) is preserved. The initial field order is already set to `DEFAULT_VIEW_FIELDS` when state is initialized, so canonical ordering on the first render is unaffected. The now-unused `FIELD_ORDER` map and `sortFieldsByCanonicalOrder` helper were removed.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude
Used for: Locating the cause in onChangeView and drafting the fix. Final implementation and manual testing were done by me.

## Testing Instructions
1. Open **AI Request Logs** in wp-admin with at least one log entry
2. Open the column header menu on any column (e.g. Provider)
3. Choose "Move left" or "Move right"
4. Confirm the column order changes and persists (no snap-back to default)

## Screenshots or screencast
After the fix column reordering now persists:

https://github.com/user-attachments/assets/582c0217-844d-4238-b0cc-39d546b9d7a9

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-669-873de116ba839fab9632826dc6afa377248cc25d.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/admin/ai-request-logs/components/LogsTable.tsx
+++ b/src/admin/ai-request-logs/components/LogsTable.tsx
@@ -17,6 +17,7 @@ import { useCallback, useMemo, useState } from '@wordpress/element';
  * Internal dependencies
  */
 import type { ProviderMetadata } from '../../types/providers';
+import { DEFAULT_VIEW_FIELDS } from '../query';
 import type { FilterOptions, LogEntry, LogsQuery } from '../types';
 
 interface LogsTableProps {
@@ -41,15 +42,6 @@ interface ViewConfig {
 	fields: string[];
 	layout: NonNullable< ViewTable[ 'layout' ] >;
 }
-
-const DEFAULT_VIEW_FIELDS = [
-	'timestamp',
-	'operation',
-	'provider',
-	'tokens_total',
-	'duration_ms',
-	'status',
-];
 
 const formatTimestamp = ( timestamp: string ): string => {
 	const { formats } = getSettings();
@@ -241,6 +233,7 @@ const viewToQuery = ( view: View ): LogsQuery => {
 		userId: extractStringFilter( filters, 'user_id' ),
 		orderby: typeof sortField === 'string' ? sortField : 'timestamp',
 		order: sortDir,
+		fields: view.fields ?? [ ...DEFAULT_VIEW_FIELDS ],
 	};
 };
 
@@ -262,7 +255,7 @@ const queryToView = ( query: LogsQuery, viewConfig: ViewConfig ): View => {
 			field: query.orderby,
 			direction: query.order,
 		},
-		fields: viewConfig.fields,
+		fields: query.fields,
 		layout: viewConfig.layout,
 	};
 };
--- a/src/admin/ai-request-logs/components/LogsTable.tsx
+++ b/src/admin/ai-request-logs/components/LogsTable.tsx
@@ -51,17 +51,6 @@ const DEFAULT_VIEW_FIELDS = [
 	'status',
 ];
 
-const FIELD_ORDER = new Map(
-	DEFAULT_VIEW_FIELDS.map( ( id, index ) => [ id, index ] )
-);
-
-const sortFieldsByCanonicalOrder = ( ids: string[] ): string[] =>
-	[ ...ids ].sort( ( a, b ) => {
-		const ai = FIELD_ORDER.get( a ) ?? Infinity;
-		const bi = FIELD_ORDER.get( b ) ?? Infinity;
-		return ai - bi;
-	} );
-
 const formatTimestamp = ( timestamp: string ): string => {
 	const { formats } = getSettings();
 	return dateI18n( `${ formats.date } ${ formats.time }`, timestamp + 'Z' );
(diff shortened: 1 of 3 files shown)


dkotter (commented):
In testing, this does now allow you to change the order of these columns but unless I'm missing something, the order does not persist beyond the first page load. If I change the order and refresh, the order reverts back. Should we be trying to save the order for future page loads?

hbhalodia:
> In testing, this does now allow you to change the order of these columns but unless I'm missing something, the order does not persist beyond the first page load. If I change the order and refresh, the order reverts back. Should we be trying to save the order for future page loads?

Hi @dkotter, I guess yes that's the ask in issue ticket. If we notice the DataViews in the `Appearence --> Editor --> Pages`, the column order you change and then reloads it would be persistent for that user on page loads.

IMO, if we are providing this order, then we should same this somewhere else, we can remove the order field and make it fixed. The UX would be better either ways.

Trushiv04:
thanks! @dkotter You're right the previous version only kept the order for the current page view.

I've updated the PR to persist column order across page loads. The order is now saved to localStorage as part of `logsQuery` (the same mechanism that already persists filters and search), so it survives refreshes.

Changes:
- Added `fields` to the `LogsQuery` type and default query
- Added validation in `normalizeLogsQuery` so only known fields are restored (falls back to default order if the stored value is incomplete or corrupt)
- `viewToQuery` now captures the field order, and `queryToView` reads it back

Tested by reordering columns, refreshing, and confirming the order persists.

hbhalodia:
Hi @Trushiv04, Thanks for the PR. Can we also update the column visibility in the same PR? 

For example, please refer the screencast. When I update the column to hide, and refresh it again, the column is visible again. If we can land a fix within the scope of this PR would be good IMO. Because both are same issue and would use similar logic, so we can include this in PR.

## Before Fix (with this PR change)

https://github.com/user-attachments/assets/2302f7ea-aa0b-433f-a2ca-3f4bf2f7f4b9


## After Fix (Just visibility change)

https://github.com/user-attachments/assets/47504886-6eb4-4f98-a049-df158d134bb0

Can you please update the PR with a fix. This is because, if the user update any column to hide, then loads the page, the column order changes are lost they are not persisted.

I would add the same thing in issue description, so issue author can update the description.

Thanks,

Trushiv04:
Thanks @hbhalodia! You're right they're the same underlying issue, so it made sense to handle both here.

The root cause was in `normalizeFields`: it required the stored field list to match the full default length, otherwise it reset to all columns. When a user hid a column, the shorter list was treated as invalid and all columns came back on reload.

I've updated it so a shorter field list is treated as valid (the user may have intentionally hidden columns). It now only falls back to the default order when the stored value is missing or contains nothing valid.

So both column order and column visibility now persist across reloads.

Tested:
- Reorder columns → refresh → order persists
- Hide a column → refresh → column stays hidden

dkotter (approved):

jeffpaul:
@alexWinterjuice if you're able to confirm your WordPress.org username (and ideally link it with your GitHub account), then I can ensure its properly credited in the AI plugin release (and associated AI Contributor badge)... thanks!

PR #679, opened 2026-06-08
Gate image generation UI on provider support detection

t-hamano:
Related to https://core.trac.wordpress.org/ticket/64917#comment:14

> Ideally the UI would reveal that image generation isn't available _before_ the user types the prompt and clicks the button. Otherwise they waste time, risk losing the prompt, and hit a dead end while in the flow.

## What?

The server checks if any connectors supporting image generation are registered and passes this information to the client side. If no connectors supporting image generation are found, a warning is displayed without sending a REST request.

## Why?

It is unhelpful that users can attempt to generate images by entering prompts, even though there are no connectors supporting image generation.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: Drafting this pull request description. Implementation was reviewed by me.

## Testing Instructions

- Only register the Anthropic API key.
- Image Block:
  - Insert an Image block and click the Generate Image button.
  - A warning should be displayed.
- Media Library:
  - Access Media > Generate Image.
  - A warning should be displayed.
- Featured Image:
  - Open the post editor and click the Generate featured image button.
  - A warning should be displayed.

## Screenshot

Image Block:
image

Media Library:
image

Featured Image:
image

## Changelog Entry

> Changed - Gate the image generation UI on detected provider support for image generation.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-679-07fa198b7132106188035bd655a7050117f44d04.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -13,6 +13,7 @@ use Throwable;
 use WordPress\AI\Services\AI_Service;
 use WordPress\AI\Services\Guidelines;
 use WordPress\AiClient\AiClient;
+use WordPress\AiClient\Providers\Models\Enums\CapabilityEnum;
 
 /**
  * Purposely using return instead of exit here.
@@ -492,6 +493,47 @@ function has_ai_credentials(): bool {
 	 * @param array $connectors      The registered connectors.
 	 */
 	return (bool) apply_filters( 'wpai_has_ai_credentials', $has_credentials, $connectors );
+}
+
+/**
+ * Checks whether any configured connector exposes an image-generation-capable model.
+ *
+ * @since X.X.X
+ *
+ * @return bool True if at least one configured connector has an image-generation-capable model.
+ */
+function has_image_generation_support(): bool {
+	if ( ! class_exists( AiClient::class ) ) {
+		return false;
+	}
+
+	$registry   = AiClient::defaultRegistry();
+	$connectors = get_ai_connectors();
+
+	foreach ( array_keys( $connectors ) as $connector_id ) {
+		if ( ! has_connector_authentication( $connector_id ) ) {
+			continue;
+		}
+
+		try {
+			$provider_class = $registry->getProviderClassName( $connector_id );
+
+			/** @var \WordPress\AiClient\Providers\Contracts\ProviderInterface $provider_class */
+			$models = $provider_class::modelMetadataDirectory()->listModelMetadata();
+
+			foreach ( $models as $model ) {
+				foreach ( $model->getSupportedCapabilities() as $capability ) {
+					if ( CapabilityEnum::IMAGE_GENERATION === $capability->value ) {
+						return true;
+					}
+				}
+			}
+		} catch ( Throwable $e ) {
+			continue;
+		}
+	}
+
+	return false;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


includes/helpers.php
```diff
@@ -494,6 +495,47 @@ function has_ai_credentials(): bool {
...
+function has_image_generation_support(): bool {
+	if ( ! class_exists( AiClient::class ) ) {
+		return false;
+	}
+
+	$registry   = AiClient::defaultRegistry();
+	$connectors = get_ai_connectors();
+
+	foreach ( array_keys( $connectors ) as $connector_id ) {
+		if ( ! has_connector_authentication( $connector_id ) ) {
```
mindctrl, line 521: What do you think about using `is_connector_configured()` here instead? I think there are models that allow image generation without requiring auth.

dkotter:
This idea has been brought up multiple times and the conclusion we've come to in the past is to leave things as they are:

- If someone enables a Feature, we should always show the UI for that Feature even if they have no AI Connector in place to support that Feature
- In that scenario, when they try and use the Feature, we should output an error message letting them know they need a valid Connector in place

The discussion in the past has mostly focused on if we should completely remove the UI if there isn't a valid Connector while this PR approaches it differently, where the general UI is still there but you can't fully trigger it.

I'm fine with considering this approach but ideally these are things that should be talked through prior to trying to implement so we can ensure we're all on the same page first. For example, my opinion here is if we implement this for image generation, we should likely implement this for all Features to keep parity.

includes/helpers.php
t-hamano: The check for `has_connector_authentication()` is performed locally. However, when using `is_connector_configured()`, I think it actually sends a request to the external provider API.

https://github.com/WordPress/wordpress-develop/blob/55e33259af856fbac2e08caa4482f45b0d74f562/src/wp-includes/php-ai-client/src/Providers/ApiBasedImplementation/GenerateTextApiBasedProviderAvailability.php#L48-L56

Although I haven't investigated thoroughly yet, I believe we'll need to implement caching in the `is_connector_configured` helper function to likely suppress unnecessary token consumption and requests.

For now, leaving it as `has_connector_authentication()` might be acceptable.

t-hamano:
Thanks for the review.

The purpose of this PR is to prevent users from engaging in futile tasks. For example, users would be disappointed if they entered a long prompt text, clicked the Generate button, and then encountered an error.

I am happy to follow up if there are any other areas that need to be addressed.

dkotter (approved):

dkotter:
> The purpose of this PR is to prevent users from engaging in futile tasks. For example, users would be disappointed if they entered a long prompt text, clicked the Generate button, and then encountered an error.

Thanks, in thinking about it more I agree this is a better experience so I've merged this in now

> I am happy to follow up if there are any other areas that need to be addressed.

This may be the only feature that requires additional action after triggering (i.e. entering in a prompt). Most other features trigger immediately after you click a button so the error message we already show there is likely good enough. But happy for someone to review all of our other features and decide if changes are needed

PR #680, opened 2026-06-08
Fix: AI Request Logging: token count understated for thinking models (thoughtsTokenCount ignored)

### LINKED ISSUE #662: AI Request Logging: token count understated for thinking models (thoughtsTokenCount ignored)
## Description

The AI Request Logs show a lower token count than Google AI Studio 
for requests made with thinking models (e.g. `gemini-2.5-flash`, 
`gemini-3-flash-preview`).

Image

Image

## Expected behavior

Token count in the logs matches Google AI Studio.

## Actual behavior

The logs show fewer tokens. In the example above: **309** in WordPress vs **338 input + ~570 output** in Google AI Studio.

## Root cause

`Log_Data_Extractor::extract_token_usage()` ignores `thoughtsTokenCount` (internal reasoning tokens) from Google's `usageMetadata`. The SDK itself already handles this correctly in `GoogleTextGenerationModel.php:584`:

```php
($usage['candidatesTokenCount'] ?? 0) + ($usage['thoughtsTokenCount'] ?? 0)
```

But `Log_Data_Extractor.php:239–243` does not:

```php
$input  = $usage['promptTokenCount'] ?? null;
$output = $usage['candidatesTokenCount'] ?? null; // ← thoughtsTokenCount ignored
```

## Affected file

`includes/Logging/Log_Data_Extractor.php`

## Steps to reproduce

1. Enable the AI Request Logging experiment
2. Configure the Google provider with a thinking model (e.g. `gemini-3-flash-preview`)
3. Trigger a text generation (e.g. via the Title Generation experiment)
4. Compare the token count in **Tools → AI Request Logs** with the usage in **Google AI Studio → Usage**

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

prasadkarmalkar:
## What?

Closes #662

## Why?

Token count mismatched in AI Request Logging and in Google console logs

## How?

Updated Google token extraction in Log Data Extractor to compute output tokens as:
candidatesTokenCount + thoughtsTokenCount

## Testing Instructions

Enable the AI Request Logging experiment
Configure the Google provider with a thinking model (e.g. gemini-3-flash-preview)
Trigger a text generation (e.g. via the Title Generation experiment)
Compare the token count in Tools → AI Request Logs with the usage in Google AI Studio → Usage

## Screenshots or screencast

| After AI Logs | Gemini Logs |
| ------ | ----- |
|Screenshot 2026-06-08 at 1 59 15 PM|Screenshot 2026-06-08 at 2 32 15 PM|

## Changelog Entry

> Fixed: Google gemini token counts in request logs now match provider usage totals.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-680-8b12c88373f8d1a85057ee3fa711dd47e779d2d2.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Logging/Log_Data_Extractor.php
+++ b/includes/Logging/Log_Data_Extractor.php
@@ -239,7 +239,7 @@ class Log_Data_Extractor {
 		if ( isset( $response['usageMetadata'] ) && is_array( $response['usageMetadata'] ) ) {
 			$usage  = $response['usageMetadata'];
 			$input  = $usage['promptTokenCount'] ?? null;
-			$output = $usage['candidatesTokenCount'] ?? null;
+			$output = ($usage['candidatesTokenCount'] ?? 0) + ($usage['thoughtsTokenCount'] ?? 0);
 		}
 
 		/**
--- a/includes/Logging/Log_Data_Extractor.php
+++ b/includes/Logging/Log_Data_Extractor.php
@@ -239,7 +239,9 @@ class Log_Data_Extractor {
 		if ( isset( $response['usageMetadata'] ) && is_array( $response['usageMetadata'] ) ) {
 			$usage  = $response['usageMetadata'];
 			$input  = $usage['promptTokenCount'] ?? null;
-			$output = ( $usage['candidatesTokenCount'] ?? 0 ) + ( $usage['thoughtsTokenCount'] ?? 0 );
+			$output = isset( $usage['candidatesTokenCount'] ) || isset( $usage['thoughtsTokenCount'] )
+			? ( $usage['candidatesTokenCount'] ?? 0 ) + ( $usage['thoughtsTokenCount'] ?? 0 )
+			: null;
 		}
 
 		/**
--- a/includes/Logging/Log_Data_Extractor.php
+++ b/includes/Logging/Log_Data_Extractor.php
@@ -239,7 +239,7 @@ class Log_Data_Extractor {
 		if ( isset( $response['usageMetadata'] ) && is_array( $response['usageMetadata'] ) ) {
 			$usage  = $response['usageMetadata'];
 			$input  = $usage['promptTokenCount'] ?? null;
-			$output = ($usage['candidatesTokenCount'] ?? 0) + ($usage['thoughtsTokenCount'] ?? 0);
+			$output = ( $usage['candidatesTokenCount'] ?? 0 ) + ( $usage['thoughtsTokenCount'] ?? 0 );
 		}
 
 		/**
--- a/tests/Integration/Includes/Logging/Log_Data_ExtractorTest.php
+++ b/tests/Integration/Includes/Logging/Log_Data_ExtractorTest.php
@@ -390,12 +390,13 @@ class Log_Data_ExtractorTest extends WP_UnitTestCase {
 				'usageMetadata' => array(
 					'promptTokenCount'     => 300,
 					'candidatesTokenCount' => 150,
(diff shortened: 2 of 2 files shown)


includes/Logging/Log_Data_Extractor.php
```diff
@@ -239,7 +239,7 @@ public function extract_token_usage( array $response ): array {
 		if ( isset( $response['usageMetadata'] ) && is_array( $response['usageMetadata'] ) ) {
 			$usage  = $response['usageMetadata'];
 			$input  = $usage['promptTokenCount'] ?? null;
-			$output = $usage['candidatesTokenCount'] ?? null;
+			$output = ( $usage['candidatesTokenCount'] ?? 0 ) + ( $usage['thoughtsTokenCount'] ?? 0 );
```
dkotter: So previously we would set `null` here if `candidatesTokenCount` wasn't set. While that's an unlikely scenario, now if that isn't set, we return `0` instead. Not sure that will actually cause problems but may be best to keep original behavior here and ensure we return `null` if neither `candidatesTokenCount` or `thoughtsTokenCount` is set

tests/Integration/Includes/Logging/Log_Data_ExtractorTest.php
```diff
@@ -390,12 +390,13 @@ public function test_extract_token_usage_google_format(): void {
 				'usageMetadata' => array(
 					'promptTokenCount'     => 300,
 					'candidatesTokenCount' => 150,
+					'thoughtsTokenCount'   => 20,
```
dkotter, line 412: This test is great but it's just augmenting our existing test. Ideally we have a test for both situations, one where `thoughtsTokenCount` isn't passed and one where it is.

dkotter (changes requested):

includes/Logging/Log_Data_Extractor.php
prasadkarmalkar: Just to clarify, should the logic return null only if both `candidatesTokenCount` and `thoughtsTokenCount` are unset? If either one has a value, we'd keep that value instead of null?

```php
$output = ( isset( $usage['candidatesTokenCount'] ) || isset( $usage['thoughtsTokenCount'] ) )
  ? ( $usage['candidatesTokenCount'] ?? 0 ) + ( $usage['thoughtsTokenCount'] ?? 0 )
  : null;
```

This preserves the original behavior of returning `null` when neither is set while still correctly summing whichever tokens are available when at least one is present.

dkotter: Yes, that looks correct

tests/Integration/Includes/Logging/Log_Data_ExtractorTest.php
prasadkarmalkar: Added separate test cases for with thoughtsTokenCount and without both candidatesTokenCount and thoughtsTokenCount as well.

dkotter (approved):

PR #682, opened 2026-06-08
Fix[Editorial Notes]: Clarify Editorial Notes save flow and hide stale updates action

### LINKED ISSUE #678: Editorial Updates cannot match existing Notes after editor reload
### Description

When pending Editorial Notes already exist for a post, the editor can show the “Apply Editorial Updates” button, but clicking it fails with:

“No blocks found matching the existing Notes.”

This happens even though the post still has a block visible in the editor List View.

It looks like the Notes exist, but the saved/reloaded block no longer has the metadata needed for Editorial Updates to match the Note back to its block. The result is confusing because the UI offers an action that cannot apply the existing Notes.

1. Enable the AI plugin.
2. Enable the Editorial Notes and Editorial Updates experiments.
3. Configure an AI connector provider.
4. Create or open a post with block content.
5. Generate Editorial Notes for the post.
6. Save or reload the editor so the existing pending Notes remain.
7. Open List View and confirm the post still has a block.
8. Click “Apply Editorial Updates”.

The screenshot shows:

- the post still has a `Classic` block in List View
- “Apply Editorial Updates” is available
- clicking it shows “No blocks found matching the existing Notes.”

Image

### Environment info

Tested locally with wp-env.

- WordPress version: 7.0
- AI plugin: current `develop`
- Active plugins:
  - AI plugin
  - AI Provider for OpenAI
- Active theme: Twenty Twenty-One
- Browser: Chromium
- OS: macOS

Note: the OpenAI provider was active because this feature requires an AI connector to run.

- [x] Yes

- [x] Yes

- [ ] Block
- [x] Classic
- [ ] Not sure

yogeshbhutkar:
## What?
Closes https://github.com/WordPress/ai/issues/678

This PR improves the Editorial Updates flow in two related ways:
1. Adds a concise save prompt to the Editorial Notes snackbar after suggestions are generated.
2. Updates the `Apply Editorial Updates` button state so it only appears when there is at least one pending Note still linked to a block in the current editor state.

## Why?
Blocks store their related Note IDs in block metadata. If a user generates Notes and then refreshes before saving the post, the Notes themselves can persist, but the block metadata linking those Notes back to their blocks is lost. There is an upstream issue for this behavior, and the long-term fix should come from that decision (Ref: https://github.com/WordPress/gutenberg/issues/72717). Until then, the UI should avoid implying that orphaned Notes can still be applied to content.

## How?
The snackbar now nudges users to save after suggestions are added, helping preserve the block-to-Note links. The Editorial Updates button now derives its visibility from an actionable state: it checks the current blocks for linked Note IDs and only shows when one of those linked Notes is still pending.

Together, these changes reduce confusion by prompting the user to save at the right moment and hiding the update action when there is no linked block Note to apply.

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.5
Used for: Initial code skeleton; final implementation was reviewed and edited by me.

## Testing Instructions
1. Create or open a post with reviewable content.
2. Generate Editorial Notes and confirm the Snackbar prompts saving.
3. Refresh the editor without saving the post.
4. Confirm Apply Editorial Updates is hidden when the generated Notes are no longer linked to blocks.

## Screencast

https://github.com/user-attachments/assets/f482575f-1593-4ef3-9941-7a4a3f05b1c3

## Changelog Entry

> Fixed - Clarify the Editorial Notes save prompt and only show Editorial Updates when pending Notes are linked to current blocks.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-682-94e2f53a11bdadf3a3b29a7f25b343aae2b2a787.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/editorial-notes/components/EditorialNotesPlugin.tsx
+++ b/src/experiments/editorial-notes/components/EditorialNotesPlugin.tsx
@@ -85,6 +85,7 @@ export default function EditorialNotesPlugin() {
 								justifyContent: 'center',
 								width: '100%',
 							} }
+							accessibleWhenDisabled
 							__next40pxDefaultSize
 						>
 							{ buttonLabel }
--- a/src/experiments/editorial-notes/hooks/useEditorialNotes.ts
+++ b/src/experiments/editorial-notes/hooks/useEditorialNotes.ts
@@ -65,6 +65,15 @@ interface ReviewResult {
 	suggestions: Suggestion[];
 }
 
+interface ReviewSingleBlockResult {
+	suggestionCount: number;
+	didLinkNoteId: boolean;
+}
+
+interface CreateNoteResult {
+	didLinkNoteId: boolean;
+}
+
 interface NoteRecord {
 	id: number;
 	[ key: string ]: unknown;
@@ -79,7 +88,7 @@ interface NoteRecord {
  * @param resolvedNoteIds Set of Note IDs that have been resolved (approved).
  * @param noteContentById Map of Note ID → rendered Note content (pending only).
  * @param pendingNotes    All pending Notes for the post (for reply lookup).
- * @return The number of suggestions created, or 0 if the block was skipped.
+ * @return The suggestion count and whether a new Note thread was linked.
  */
 async function reviewSingleBlock(
 	block: Block,
@@ -88,19 +97,19 @@ async function reviewSingleBlock(
 	resolvedNoteIds: Set< number >,
 	noteContentById: Map< number, string >,
 	pendingNotes: ExistingNote[]
-): Promise< number > {
+): Promise< ReviewSingleBlockResult > {
 	// Look up any existing Note thread on this block.
 	const existingNoteId = block.attributes.metadata?.noteId ?? null;
 
 	// Skip blocks whose Note thread has been resolved.
 	if ( existingNoteId && resolvedNoteIds.has( existingNoteId ) ) {
-		return 0;
+		return { suggestionCount: 0, didLinkNoteId: false };
 	}
 
 	const blockText = getBlockText( block );
 
 	if ( blockText.length === 0 ) {
-		return 0;
+		return { suggestionCount: 0, didLinkNoteId: false };
 	}
 
 	// Collect pending Note texts for this block's thread as context.
@@ -148,11 +157,20 @@ async function reviewSingleBlock(
 	} );
 
 	if ( result?.suggestions && result.suggestions.length > 0 ) {
-		await createNote( block, postId, result.suggestions, existingNoteId );
-		return result.suggestions.length;
+		const { didLinkNoteId } = await createNote(
+			block,
+			postId,
+			result.suggestions,
+			existingNoteId
+		);
+
+		return {
+			suggestionCount: result.suggestions.length,
+			didLinkNoteId,
+		};
 	}
 
-	return 0;
+	return { suggestionCount: 0, didLinkNoteId: false };
 }
 
 /**
@@ -228,6 +246,7 @@ export function useEditorialNotes(): {
 			}
 
 			let totalSuggestions = 0;
+			let didLinkAnyNoteId = false;
 
 			// Process blocks in batches.
 			for (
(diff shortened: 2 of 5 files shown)


dkotter (commented):
Overall this looks good to me but probably worth pausing for a discussion on this.

Up to this point, we've more or less taken the approach that saving content should be a user initiated behavior. For instance, if I trigger Excerpt Generation, I still have to manually click the save button otherwise that excerpt is lost with a page refresh.

This PR changes that for Editorial Notes where we automatically trigger a save when a Note is added. While this ensures that Note and it's association aren't lost, this does mean the post is saved even though a user didn't intentionally click save. If they had made other changes to the post that they weren't ready to publish yet, this makes those changes live.

We may be fine with that though as mentioned, we've tried to stay away from that so far cc @jeffpaul

ekamran:
Thanks for looking into this. I agree this is the main tradeoff here.

From the issue side, the problem I wanted to report was that after reload the UI can still offer “Apply Editorial Updates”, but the existing Notes can no longer be matched back to the block. I was not assuming that automatically saving the whole post is necessarily the right fix.

Your point about user-initiated saving makes sense to me. If generating Editorial Notes triggers a full post save, that could also persist unrelated content changes the user was not ready to save yet.

So I think the safer direction may need a product/maintainer decision first. A few possible directions could be:

- keep post saving user-initiated, but make it clear that the post must be saved after generating Editorial Notes for the note/block association to survive reloads;
- store the Note-to-block association somewhere that does not require saving unrelated post content, if there is a reliable place for that;
- or avoid showing “Apply Editorial Updates” when pending Notes exist but no current block can be matched, and explain why.

The important part for me is avoiding the broken state where pending Notes exist, the action is shown, but applying updates only results in “No blocks found matching the existing Notes.”

yogeshbhutkar:
Thank you so much for all the inputs. I also noticed that the core Notes functionality itself seems to suffer from this issue (Ref: https://github.com/WordPress/gutenberg/issues/72717). If you create a note and refresh the page without saving, the note is stored, but its association is not persisted. Ideally, we could wait for the upstream resolution first, as it may fix this case as well. If not, we can follow/refer the approach agreed upon there for consistency.

For the other points, I agree that we can make the snackbar message clearer so users understand that they need to persist the changes. We could also explore showing the “Apply updates” button conditionally, based on whether the updates are actually applicable. This may require some pre-computation, but from a feasibility standpoint, it should be fine. 

If any other ideas come up while implementing this, I’d be more than happy to iterate on the approach in this PR.

ekamran:
Thanks for checking the upstream Notes issue. Since this seems tied to the core Notes behavior, I agree it’s safer not to introduce separate auto-save behavior in the AI plugin before that direction is settled.

From the AI plugin side, the main thing I’d want to avoid is the confusing state where pending Notes exist, “Apply Editorial Updates” is shown, but no matching blocks can be found. A clearer snackbar or only showing the action when updates are actually applicable both sound like reasonable interim directions to me.

dkotter:
> also noticed that the core Notes functionality itself seems to suffer from this issue (Ref: https://github.com/WordPress/gutenberg/issues/72717). If you create a note and refresh the page without saving, the note is stored, but its association is not persisted.

Yeah, noticed this myself while building this Feature and was one other reason why I left it like this, seemed best to match WordPress Core (even though easy to argue that's a bug).

> From the AI plugin side, the main thing I’d want to avoid is the confusing state where pending Notes exist, “Apply Editorial Updates” is shown, but no matching blocks can be found. A clearer snackbar or only showing the action when updates are actually applicable both sound like reasonable interim directions to me.

I agree, any way we can make this clearer I'm on board for

yogeshbhutkar:
This PR has been updated to clarify the snackbar messaging around saving changes. The “Apply Editorial Updates” button is now conditionally rendered based on the attached notes.

cc: @dkotter @ekamran

dkotter (approved):

PR #683, opened 2026-06-08
Add native vector search

artpi:
## Summary

This introduces native RAG search for hosts that can support it:

**New Infrastructure:**

- [x] Add an opt-in `rag-search` experiment backed by MariaDB native `VECTOR(1536)` storage and a cosine `VECTOR INDEX`. Gates activation on MariaDB 11.8+ and an authenticated OpenAI connector, using `text-embedding-3-small` embeddings.
- [x] Add fallback which stores embeddings in post meta and searches them in-memory. This should be performant enough for smaller sites
- [x] Add indexing lifecycle hooks, one-hour dirty-post cron scheduling, short batch indexing, transactional replace-by-post behavior where possible, cleanup hooks, and `wp ai rag` WP-CLI utilities.
- [ ] Move to embeddings supplied by WP AI CLient once that is available

Zrzut ekranu 2026-06-8 o 20 04 37

**User facing features**:

- [x] Augment the public search page to integrate the RAG results into the result
- [x] Introduce the "related posts" functionality just like in Jetpack - for SEO

Zrzut ekranu 2026-06-8 o 20 01 13

## Rationale
MariaDB 11.8 LTS ships a community-available vector index implementation, so WordPress can offer a lean RAG search experiment without introducing another runtime service or PHP dependency.

Unfortunately MySQL 9 does not have vector search capabilities, despite having vector store.

The feature is intentionally gated for supporting hosts: it only becomes operational when the site has both a supported MariaDB version and an authenticated OpenAI connector capable of producing the fixed 1536-dimensional embedding shape used by the first schema.

## Testing

Manual default `wp-env` smoke test:

```bash
source ~/.nvm/nvm.sh
nvm use 22.21.1
npm run wp-env -- start
npm run wp-env -- run cli wp ai rag status
```

Environment observed:

```text
WordPress: http://localhost:8888
MariaDB: 12.3.2-MariaDB-ubu2404
RAG available: yes
Index table: present
```

- Connect OpenAI key in the default wp-env site connectors screen
- Create few posts semantically similar to a specific topic but **not containing the keywords**. For example posts about tables being broken, but NOT having "furniture shaking" words.

Then ran the procesing manually since chunking happens every hour. So you can chunk them initially with wp-cli:

```bash
npm run wp-env -- run cli wp ai rag index --all --batch-size=50
npm run wp-env -- run cli wp ai rag status
```

Result:

```text
Processed 8 post(s): 8 clean, 0 error, 0 removed. 0 dirty post(s) remain.
RAG index rows: 8
Indexed posts: 8
```

Semantic search smoke test:

```text
http://localhost:8888/?s=how+do+I+stop+furniture+from+shaking
First result: RAG fixture ... cafe table leveling
```

A raw keyword lookup for `furniture` / `shaking` should return no matching posts, while the semantic search API returned the cafe-table post first.

Zrzut ekranu 2026-06-8 o 12 17 00

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-683-d2615bd620fd2bc6b8514132399f7c4afbc346e1.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/RAG_Search/RAG_Search.php
+++ b/includes/Experiments/RAG_Search/RAG_Search.php
@@ -120,7 +120,7 @@ class RAG_Search extends Abstract_Feature {
 		$fields = array(
 			array(
 				'id'      => 'augment_search',
-				'label'   => __( 'Augment WordPress search results', 'ai' ),
+				'label'   => __( 'Augment WordPress search page results with semantic search', 'ai' ),
 				'type'    => 'boolean',
 				'default' => self::DEFAULT_AUGMENT_SEARCH,
 			),
@@ -224,8 +224,8 @@ class RAG_Search extends Abstract_Feature {
 	 */
 	private function get_backend_field_elements( array $backends ): array {
 		$labels = array(
-			Availability::BACKEND_MARIADB => __( 'MariaDB vector index', 'ai' ),
-			Availability::BACKEND_MEMORY  => __( 'Memory exact scan', 'ai' ),
+			Availability::BACKEND_MARIADB => __( 'Optimal search method backed by MariaDB', 'ai' ),
+			Availability::BACKEND_MEMORY  => __( 'Fallback in-memory search backed by PHP', 'ai' ),
 		);
 
 		$elements = array();
--- a/includes/Experiments/RAG_Search/RAG_Search.php
+++ b/includes/Experiments/RAG_Search/RAG_Search.php
@@ -1,0 +1,151 @@
+<?php
+/**
+ * MariaDB vector RAG search experiment.
+ *
+ * @package WordPress\AI\Experiments\RAG_Search
+ */
+
+declare( strict_types=1 );
+
+namespace WordPress\AI\Experiments\RAG_Search;
+
+use WordPress\AI\Abstracts\Abstract_Feature;
+use WordPress\AI\CLI\RAG_Command;
+use WordPress\AI\Experiments\Experiment_Category;
+use WordPress\AI\RAG\Availability;
+use WordPress\AI\RAG\Index_Manager;
+use WordPress\AI\RAG\REST\RAG_Search_Controller;
+use WordPress\AI\RAG\Search_Augmenter;
+use WordPress\AI\Settings\Settings_Registration;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Adds semantic RAG search backed by MariaDB vector indexes.
+ *
+ * @since 1.1.0
+ */
+class RAG_Search extends Abstract_Feature {
+	/**
+	 * Default search augmentation setting.
+	 */
+	private const DEFAULT_AUGMENT_SEARCH = false;
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public static function get_id(): string {
+		return 'rag-search';
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function load_metadata(): array {
+		return array(
+			'label'       => __( 'RAG Search', 'ai' ),
+			'description' => __( 'Semantic search using OpenAI embeddings and native MariaDB 11.8+ vector indexes.', 'ai' ),
+			'category'    => Experiment_Category::ADMIN,
+			'capability'  => 'embedding_generation',
+		);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function register(): void {
+		$availability = new Availability();
+
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			\WP_CLI::add_command( 'ai rag', RAG_Command::class );
+		}
+
+		if ( ! $availability->is_available() ) {
+			add_action( 'admin_notices', array( $this, 'render_unavailable_notice' ) );
+			return;
+		}
+
+		$index_manager = new Index_Manager( $availability );
+		$index_manager->init();
+
+		( new RAG_Search_Controller( $availability ) )->init();
+
+		if ( ! $this->is_search_augmentation_enabled() ) {
+			return;
+		}
+
+		( new Search_Augmenter() )->init();
+	}
+
+	/**
+	 * Registers feature settings.
+	 *
+	 * @since 1.1.0
+	 */
+	public function register_settings(): void {
+		register_setting(
+			Settings_Registration::OPTION_GROUP,
+			static::get_field_option_name( 'augment_search' ),
+			array(
+				'type'              => 'boolean',
+				'default'           => self::DEFAULT_AUGMENT_SEARCH,
+				'sanitize_callback' => 'rest_sanitize_boolean',
+				'show_in_rest'      => true,
+			)
+		);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function get_settings_fields(): array {
+		return array(
+			array(
+				'id'      => 'augment_search',
+				'label'   => __( 'Augment WordPress search results', 'ai' ),
+				'type'    => 'boolean',
+				'default' => self::DEFAULT_AUGMENT_SEARCH,
+			),
+		);
+	}
+
+	/**
+	 * Returns whether front-end search should be augmented.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return bool True when enabled.
+	 */
+	public function is_search_augmentation_enabled(): bool {
+		return (bool) get_option( static::get_field_option_name( 'augment_search' ), self::DEFAULT_AUGMENT_SEARCH );
+	}
+
+	/**
+	 * Renders an admin notice when requirements are unmet.
+	 *
+	 * @since 1.1.0
+	 */
+	public function render_unavailable_notice(): void {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			return;
+		}
+
+		$availability = new Availability();
+		$message      = $availability->get_unavailable_reason();
+
+		if ( '' === $message ) {
+			return;
+		}
+
+		printf(
+			'<div class="notice notice-warning"><p>%s</p></div>',
+			esc_html(
+				sprintf(
+					/* translators: %s: Unavailable reason. */
+					__( 'RAG Search is enabled but inactive: %s', 'ai' ),
+					$message
+				)
+			)
+		);
+	}
+}
--- a/includes/Experiments/RAG_Search/RAG_Search.php
+++ b/includes/Experiments/RAG_Search/RAG_Search.php
@@ -15,6 +15,7 @@ use WordPress\AI\CLI\RAG_Command;
 use WordPress\AI\Experiments\Experiment_Category;
 use WordPress\AI\RAG\Availability;
 use WordPress\AI\RAG\Index_Manager;
+use WordPress\AI\RAG\REST\RAG_Maintenance_Controller;
 use WordPress\AI\RAG\REST\RAG_Search_Controller;
 use WordPress\AI\RAG\Related_Posts;
 use WordPress\AI\RAG\Search_Augmenter;
@@ -32,6 +33,11 @@ class RAG_Search extends Abstract_Feature {
 	 * Default search augmentation setting.
 	 */
 	private const DEFAULT_AUGMENT_SEARCH = false;
+
+	/**
+	 * Tracks WP-CLI command registration.
+	 */
+	private static bool $cli_registered = false;
 
 	/**
 	 * {@inheritDoc}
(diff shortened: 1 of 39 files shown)


JasonTheAdams:
This is neat! I really like the idea! Let's get some user-facing UI in here so folks can enjoy the benefits, so it isn't purely technical.

@jeffpaul @dkotter I'm curious what you think about having an experiment that introduces additional tables? Do we have any other experiments doing this?

I'm going to suggest we use this to push along the [Embeddings PR](https://github.com/WordPress/php-ai-client/pull/244) to have embeddings support in the AI Client in 7.1. I'm not keen on introducing experiments that have bespoke methods for calling AI and are coupled to specific providers. Not because that's generally "bad", but because a goal of this plugin is to be a good example to developers for using the AI Client, and this would be such a cool example.

Thanks, @artpi! 🙌

jeffpaul:
> Execution calls OpenAI embeddings directly for now because AI Client exposes provider/model discovery but does not yet expose embedding generation execution. The implementation keeps that boundary isolated in OpenAI_Embedding_Client so it can move to AI Client execution later.

This rationale makes sense to me, let's ensure there's a TODO in the code to remove the OAI requirement once the AI Client allows us to utilize any Connector with embeddings support.

@JasonTheAdams I'm fine with adding tables, though we may want to ensure there's cleanup when someone disables the experiment that either immediately removes those tables or offers to do so for the user.  If this ever looks to move out of experiment towards feature we can discuss more seriously about the table needs and structure.

dkotter:
> I'm not keen on introducing experiments that have bespoke methods for calling AI and are coupled to specific providers. Not because that's generally "bad", but because a goal of this plugin is to be a good example to developers for using the AI Client, and this would be such a cool example.

If I'm reading this correctly, is the idea that this PR should wait until embedding support is in the AI Client?

At a quick glance, my other concern here is the limitation with MariaDB 11.8. I'm not sure if there's numbers out there we can look at (in terms of what most WordPress sites run on) but I'm assuming this limits the number of people that would even be able to use this. Worth considering a fallback approach that would work for a larger percentage of users? Basically store in a custom table and then do our own calculations server-side? Performances issues on that to consider but in my testing, does scale decently well.

includes/Experiments/RAG_Search/RAG_Search.php
```diff
@@ -0,0 +1,259 @@
...
+	public function enqueue_editor_assets(): void {
+		Asset_Loader::enqueue_script( 'rag_search', 'experiments/rag-search' );
+	}
+
+	/**
+	 * Renders an admin notice when requirements are unmet.
+	 *
+	 * @since 1.1.0
+	 */
+	public function render_unavailable_notice(): void {
```
artpi, line 203: Do we need this now that we have a fallback?

includes/RAG/Index_Repository.php
```diff
@@ -0,0 +1,280 @@
...
+use WP_Post;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Stores and queries RAG chunks.
+ *
+ * @since 1.1.0
+ */
+class Index_Repository implements Index_Repository_Interface {
```
artpi: Schould this be renamed into MariaDB_Index_Repository?

includes/RAG/Availability.php
```diff
@@ -0,0 +1,333 @@
...
+class Availability {
+	/**
+	 * OpenAI connector identifier.
+	 */
+	public const OPENAI_CONNECTOR_ID = 'openai';
+
+	/**
+	 * MariaDB vector index backend.
+	 */
+	public const BACKEND_MARIADB = 'mariadb';
```
artpi, line 23: Should this constant live in the appropriate index repository classes?
Maybe availability shouldn't be concernings itself about details of each index repository?

includes/RAG/Availability.php
```diff
@@ -0,0 +1,333 @@
...
+
+	/**
+	 * Required MariaDB version for VECTOR INDEX support.
+	 */
+	private const MINIMUM_MARIADB_VERSION = '11.8.0';
+
+	/**
+	 * Required embedding model.
+	 */
+	private const EMBEDDING_MODEL = 'text-embedding-3-small';
```
artpi: Should this live in the embedding client class?
Lets move all details about how embedding works into the embedding client

includes/RAG/Availability.php
```diff
@@ -0,0 +1,333 @@
...
+	 *
+	 * @return bool True when all runtime requirements are met.
+	 */
+	public function is_available(): bool {
+		if ( null !== $this->available ) {
+			return $this->available;
+		}
+
+		if ( empty( $this->get_available_index_backends() ) ) {
+			$this->unavailable_reason = __( 'MariaDB 11.8 or newer is required when the compact memory fallback is disabled.', 'ai' );
```
artpi: this sounds like it should also live in tne index repository class

includes/RAG/Availability.php
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	/**
+	 * Checks whether the database is MariaDB 11.8+.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return bool True when VECTOR INDEX should be available.
+	 */
+	public function is_mariadb_vector_index_supported(): bool {
```
artpi, line 131: again sounds like a concern of the index repostory clas

includes/RAG/Availability.php
```diff
@@ -0,0 +1,333 @@
...
+
+	/**
+	 * Checks a raw database version string.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param string $version Raw SELECT VERSION() value.
+	 * @return bool True when supported.
+	 */
+	public function is_supported_mariadb_version( string $version ): bool {
```
artpi: index repository class concern

includes/RAG/Availability.php
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	/**
+	 * Returns the embedding vector dimension count.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return int Dimension count.
+	 */
+	public function get_embedding_dimensions(): int {
```
artpi: embedding client concern

includes/RAG/Availability.php
```diff
@@ -0,0 +1,333 @@
+<?php
```
artpi, line 1: this file should be agnostic of when each backend is available - it just gets a list of avialable backends and maybe a list of why some are unavailable

includes/RAG/Index_Manager.php
```diff
@@ -0,0 +1,880 @@
...
+	}
+
+	/**
+	 * Schedules the next indexing run.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param int $delay Delay in seconds.
+	 */
+	private function schedule_immediate( int $delay = 1 ): void {
```
artpi: Why do we have both schedule_delayed and schedule_immediate? lets combine into 1 menthod

includes/RAG/Index_Manager.php
```diff
@@ -0,0 +1,880 @@
...
+	 * @param \WP_Post $post Post object.
+	 * @return string Hash.
+	 */
+	private function build_content_hash( WP_Post $post ): string {
+		$data = array(
+			'post_id'           => (int) $post->ID,
+			'post_title'        => (string) $post->post_title,
+			'post_content'      => (string) $post->post_content,
+			'post_excerpt'      => (string) $post->post_excerpt,
+			'post_modified_gmt' => (string) $post->post_modified_gmt,
```
artpi: post modified i think does not make sense here because the content likely stayed the same

includes/RAG/Index_Repository.php
```diff
@@ -0,0 +1,280 @@
...
+	 * @return list<array<string, mixed>> Matching rows.
+	 */
+	public function search( array $embedding, array $args = array() ): array {
+		$this->validate_embedding( $embedding );
+
+		global $wpdb;
+
+		$defaults = array(
+			'limit'       => 20,
+			'model'       => 'text-embedding-3-small',
```
artpi: Should this come from configured embedding client?

includes/RAG/Index_Repository.php
```diff
@@ -0,0 +1,280 @@
...
+		$table_name  = $this->schema->get_table_name();
+		$vector_text = wp_json_encode( array_values( $embedding ) );
+		if ( ! is_string( $vector_text ) ) {
+			throw new RuntimeException( 'Failed to encode vector.' );
+		}
+
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- Dynamic table name targets the owned vector index table.
+		$sql = $wpdb->prepare(
+			"INSERT INTO {$table_name}
+				(post_id, post_type, post_status, chunk_id, chunk_index, chunk_offset, anchor, title, permalink, content, content_hash, embedding, embedding_model, embedding_dimensions, indexed_at)
```
artpi: If we have model, do we need dimensions?
do we need anchor and title? what for we should be able to infer everything from post id

includes/RAG/Index_Schema.php
```diff
@@ -0,0 +1,209 @@
...
+namespace WordPress\AI\RAG;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Handles database schema management for the RAG index table.
+ *
+ * @since 1.1.0
+ */
+class Index_Schema {
```
artpi: This index schema is only used in the "Index_Repository" which should be MariaDB_Index_Repository and is tied to that implementation? should it live there instaed?

includes/RAG/Memory_Index_Repository.php
```diff
@@ -0,0 +1,439 @@
...
+			throw new InvalidArgumentException( 'Chunk and embedding counts must match.' );
+		}
+
+		$records = array();
+		foreach ( $chunks as $index => $chunk ) {
+			$embedding = $embeddings[ $index ];
+			$this->validate_embedding( $embedding );
+
+			$records[] = array(
+				'post_id'              => (int) $post->ID,
```
artpi: Why are we storing post data in the meta of the post itself ???!!!!
We have this data from psost already!!
TITLE ?!

artpi: Can we store the absolute minimum necessary for these chunks?

includes/RAG/Post_Chunker.php
```diff
@@ -0,0 +1,228 @@
...
+
+	/**
+	 * Chunks a post for embedding.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @param \WP_Post $post Post object.
+	 * @return list<array{chunk_id:string, chunk_index:int, chunk_offset:int, anchor:string|null, title:string, permalink:string, content:string}> Chunk records.
+	 */
+	public function chunk_post( WP_Post $post ): array {
```
artpi, line 45: Would it make sense for every chunk to have titel and excerpt of the post? I am not sure what to do if excerpt is long... maybe only title?

jeffpaul:
While work here can and should continue, I'm going to mark this as `blocked` by https://github.com/WordPress/php-ai-client/pull/244 (and, ideally, subsequent support in connector plugins like https://github.com/WordPress/ai-provider-for-openai/pull/34, https://github.com/WordPress/ai-provider-for-google/pull/30, and https://github.com/Fueled/ai-provider-for-ollama/pull/69).

Best case scenario Is the embeddings work in the PHP client gets into a release there and then updated accordingly in the WP client as part of WP 7.1 so that the AI plugin can rely on what's natively exposed In core and not some hacky, non-recommended implementation approach.

dkotter:
@artpi Note we've now merged in https://github.com/WordPress/ai/pull/892 which brings over the embedding support from the PHP AI Client, allowing us to take advantage of that functionality prior to that being released in WordPress. I'd suggest refactoring this PR to take advantage of that

jeffpaul:
@artpi please see the new tracking issue for embedding integration in the AI plugin (#962) and whether you might be interested in coordinating with @priyanshuhaldar007 and @ColinM-sys who've also worked on #891 and #943 respectively on the smaller, more reviewable PRs on the path toward implementing embedding support... thanks!

PR #688, opened 2026-06-09
Update output schema for get-post-terms ability and sanitize image_url in Alt_Text_Generation

### LINKED ISSUE #668: Ability schemas are not JSON-Schema clean: object-valued sanitize_callback and a malformed ai/get-post-terms output schema break MCP tool conversion
## Summary

Two abilities ship input/output schemas that are not valid JSON Schema. `ai/alt-text-generation` puts an object-bound callable (`array( $this, 'method' )`) in `sanitize_callback`, and `ai/get-post-terms` has a structurally malformed `output_schema`. Any strict JSON-Schema consumer of these schemas breaks. Concretely, exposing them through `WordPress/mcp-adapter` causes a stack overflow (alt-text) and a validation failure (get-post-terms), so these abilities cannot become MCP tools.

## Environment

| | |
|---|---|
| `WordPress/ai` | 1.0.1 |
| WordPress | 7.0 (Abilities API in core) |
| PHP | 8.3 |
| Surfaced via | `wordpress/mcp-adapter` 0.5.0 |

## Problem 1: object-valued `sanitize_callback` in `ai/alt-text-generation`

`includes/Abilities/Image/Alt_Text_Generation.php`:

```php
'image_url' => array(
    'type'              => 'string',
    'sanitize_callback' => array( $this, 'sanitize_image_reference_input' ), // object-bound callable
    'description'       => '...',
),
```

The exported input schema therefore contains a live PHP object. A JSON-Schema consumer that walks the schema descends into that object graph. In `mcp-adapter` 0.5.0 this recurses until the Zend stack limit: `Maximum call stack size ... reached. Infinite recursion?`. The string-valued callbacks on `attachment_id` (`absint`) and `context` (`sanitize_textarea_field`) are fine; only the object-bound one is a problem.

## Problem 2: malformed `output_schema` in `ai/get-post-terms`

`includes/Abilities/Utilities/Posts.php`:

```php
'output_schema' => array(
    'type'       => 'object',
    'properties' => array(
        'type'  => 'array',          // 'properties' should be a map of name => sub-schema
        'items' => array( ... ),     // this looks like an array schema placed under 'properties'
    ),
),
```

`properties` must be an object mapping property names to sub-schemas. Here it holds `type` (string `"array"`) and `items`, which reads like an array schema was pasted into the `properties` slot. A strict validator reports `Expected array or object for key "type", got string`. The output is presumably meant to be an array of term objects, so the schema shape is wrong.

## Expected behavior

`input_schema` and `output_schema` should be valid JSON Schema. WordPress-only concerns such as sanitization belong in execution or validation, not as object-valued keys inside the exported schema. The `get-post-terms` output schema should describe an array of term objects correctly.

## Proposed fix

- For `alt-text-generation`: keep sanitization server-side, but do not emit an object-bound callable in the exported schema. Either reference a named function, or strip callable keys from the schema exposed via REST and MCP.
- For `get-post-terms`: correct the `output_schema` to a proper array-of-objects shape (for example `type: array` with `items: { type: object, properties: { term_id, name, slug, ... } }`), rather than nesting that under `properties`.
- Consider a unit test that asserts every registered `ai/*` ability schema round-trips through `wp_json_encode()` and validates as JSON Schema, so non-clean schemas are caught in CI.

## Context

Discovered while exposing the `WordPress/ai` abilities as MCP tools via `mcp-adapter`. A companion robustness issue is filed on `WordPress/mcp-adapter` (it should strip WordPress-only keys and guard against recursion regardless), but the schemas themselves should also be JSON-Schema clean. Happy to open a PR.

the-hercules:
## What?
Closes #668

Fixes invalid JSON Schema definitions for AI abilities exposed through the Abilities API, REST, and MCP adapters.

## Why?

Two ability schemas were not clean JSON Schema:

- `ai/alt-text-generation` exposed an object-bound PHP callable in `image_url.sanitize_callback`, which can cause strict schema consumers to recurse through a live PHP object graph.
- `ai/get-post-terms` had a malformed `output_schema`, with array schema keywords incorrectly nested under `properties`.

Strict consumers such as `wordpress/mcp-adapter` validate and traverse these schemas, so malformed or non-serializable schema values prevent these abilities from being exposed as MCP tools.

## How?

- Removed the object-bound `sanitize_callback` from the exported `image_url` input schema.
- Preserved `image_url` sanitization by running it inside the alt text ability execution path before resolving the image reference.
- Reworked `ai/get-post-terms` output schema into a valid array-of-objects schema.
- Converted returned `WP_Term` objects into associative arrays after the `wpai_get_post_terms` filter runs, so the ability output matches its JSON Schema.
- Updated `get_post_context()` to consume the new array-shaped term output.
- Added/updated integration tests for the clean term output schema, alt text sanitization behavior, and filter argument handling.

### Use of AI Tools

AI assistance: Yes  
Tool(s): Codex / ChatGPT  
Model(s): GPT-5.5  
Used for: Investigating the schema failures, drafting the implementation, and updating tests. Final changes were reviewed and adjusted by me.

## Changelog Entry

Ensure AI ability schemas and outputs are valid JSON Schema for strict REST and MCP consumers.

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-688-c97b6d3a962033ade952ff3fa78093a4d87a3483.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Image/Alt_Text_Generation.php
+++ b/includes/Abilities/Image/Alt_Text_Generation.php
@@ -67,9 +67,8 @@ class Alt_Text_Generation extends Abstract_Ability {
 					'description'       => esc_html__( 'The attachment ID of the image to generate alt text for.', 'ai' ),
 				),
 				'image_url'     => array(
-					'type'              => 'string',
-					'sanitize_callback' => array( $this, 'sanitize_image_reference_input' ),
-					'description'       => esc_html__( 'URL or data URI of the image to generate alt text for. Used if attachment_id is not provided.', 'ai' ),
+					'type'        => 'string',
+					'description' => esc_html__( 'URL or data URI of the image to generate alt text for. Used if attachment_id is not provided.', 'ai' ),
 				),
 				'context'       => array(
 					'type'              => 'string',
@@ -121,6 +120,8 @@ class Alt_Text_Generation extends Abstract_Ability {
 				'image_meta'    => '',
 			),
 		);
+
+		$args['image_url'] = $this->sanitize_image_reference_input( $args['image_url'] );
 
 		// Get the image reference.
 		$image_reference = $this->get_image_reference( $args );
--- a/includes/Abilities/Image/Alt_Text_Generation.php
+++ b/includes/Abilities/Image/Alt_Text_Generation.php
@@ -121,7 +121,9 @@ class Alt_Text_Generation extends Abstract_Ability {
 			),
 		);
 
-		$args['image_url'] = $this->sanitize_image_reference_input( $args['image_url'] );
+		if ( isset( $args['image_url'] ) ) {
+			$args['image_url'] = $this->sanitize_image_reference_input( $args['image_url'] );
+		}
 
 		// Get the image reference.
 		$image_reference = $this->get_image_reference( $args );
--- a/includes/Abilities/Utilities/Posts.php
+++ b/includes/Abilities/Utilities/Posts.php
@@ -206,53 +206,50 @@ class Posts {
 					'required'   => array( 'post_id' ),
 				),
 				'output_schema'       => array(
-					'type'        => 'object',
+					'type'        => 'array',
 					'description' => esc_html__( 'An array of WP_Term objects assigned to the post.', 'ai' ),
-					'properties'  => array(
-						'type'  => 'array',
-						'items' => array(
-							'type'  => 'array',
-							'items' => array(
-								'term_id'          => array(
-									'type'        => 'integer',
-									'description' => esc_html__( 'The ID of the term.', 'ai' ),
-								),
-								'name'             => array(
-									'type'        => 'string',
-									'description' => esc_html__( 'The name of the term.', 'ai' ),
-								),
-								'slug'             => array(
-									'type'        => 'string',
-									'description' => esc_html__( 'The slug of the term.', 'ai' ),
-								),
-								'term_group'       => array(
-									'type'        => 'integer',
-									'description' => esc_html__( 'The group ID of the term.', 'ai' ),
-								),
-								'term_taxonomy_id' => array(
-									'type'        => 'integer',
-									'description' => esc_html__( 'The taxonomy ID of the term.', 'ai' ),
-								),
-								'taxonomy'         => array(
-									'type'        => 'string',
-									'description' => esc_html__( 'The taxonomy name of the term.', 'ai' ),
-								),
-								'description'      => array(
-									'type'        => 'string',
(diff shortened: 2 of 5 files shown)


WouterP0lman:
Thanks for the quick turnaround here, @the-hercules. This matches what I ran into exactly.

I hit both problems while exposing the `WordPress/ai` abilities as MCP tools through `wordpress/mcp-adapter` on a local WP 7.0 + Abilities API setup:

- `ai/alt-text-generation` recursed to a stack overflow because `image_url.sanitize_callback` held an object-bound callable (`array( $this, 'method' )`) that the schema walker descended into.
- `ai/get-post-terms` failed conversion with `Expected array or object for key "type", got string` because of the malformed `output_schema`.

Your approach looks right to me: keeping sanitization in the execution path and reworking the term output into a valid array-of-objects schema. The schema round-trip test is a good guard against regressions.

includes/Abilities/Image/Alt_Text_Generation.php
```diff
@@ -122,6 +121,8 @@ protected function execute_callback( $input ) {
 			),
 		);
 
+		$args['image_url'] = $this->sanitize_image_reference_input( $args['image_url'] );
```
dkotter: The `image_url` param may not be passed (we either expect `attachment_id` or `image_url`) but we don't check that here. I don't think that will cause problems but probably better to only sanitize that value if it isn't `null`, something like:


```suggestion
		if ( isset( $args['image_url'] ) ) {
			$args['image_url'] = $this->sanitize_image_reference_input( $args['image_url'] );
		}
```

dkotter (changes requested):

includes/Abilities/Image/Alt_Text_Generation.php
the-hercules: @dkotter Yes keeping it as null more accurately represents “not provided,” and it avoids calling a sanitizer for a value that is intentionally absent. I have committed the required changes.

dkotter (approved):

Issue #40, opened 2025-08-31
WordPress Core Abilities


Jameswlepage:
As the Abilities API is developed for inclusion in WordPress Core, we need to define a default set of abilities that will be bundled with it. This proposal outlines a foundational set of abilities intended to be safe, useful for the majority of use cases, and non-destructive by default.

All ability names below follow the established `namespace/ability-name` convention enforced by the API. The `core` namespace is used to designate that they are part of the WordPress Core set.

#### Proposed Core Abilities

**Site and Settings**
- ✅ `core/get-site-info` (WordPress 6.9)
- ✅ `core/get-environment-info` (WordPress 6.9)
- `core/get-settings`
- `core/update-settings`

**Users**
- ✅ `core/get-user-info` (WordPress 6.9)
- `core/update-user-info`
- `core/get-user`
- `core/find-users`

**Posts and Pages**
- `core/find-posts`
- `core/get-post`
- `core/create-post`
- `core/update-post`
- `core/find-pages`
- `core/get-page`
- `core/create-page`
- `core/update-page`

**Media**
- `core/find-media-items`
- `core/get-media-item`
- `core/upload-media-item`
- `core/update-media-item`

**Comments**
- `core/find-comments`
- `core/get-comment`

**Taxonomy**
- `core/find-categories`
- `core/get-category`
- `core/find-tags`
- `core/get-tag`

**Menus**
- `core/get-menu-locations`
- `core/get-menu`

**Themes**
- `core/get-active-theme`
- `core/list-themes`

**Plugins**
- `core/list-plugins`
- `core/get-plugin`
- `core/activate-plugin`
- `core/deactivate-plugin`
- `core/update-plugin`

#### Out of Scope for this Proposal

- _Deleting_ content, users, or media.
- Installing or uninstalling plugins and themes.
- Theme switching.
- Creating or modifying menus.
- Any actions that cannot be easily reversed.

#### Open Questions

1.  **Plugin Management**: This proposal includes abilities to activate, deactivate, and update plugins. Given the potential for site instability, what guardrails are essential? Should these actions be limited to an allowlist of specific plugins by default to prevent unintended changes?

2.  **Destructive Actions**: This initial set deliberately excludes destructive actions like deleting posts or users. Should a future version include them? If so, what additional security measures, like a 'trash' or 'undo' ability, would be required to maintain a high degree of safety?

galatanovidiu:
For:

> **Posts and Pages**
> - core/find-posts
> - core/get-post
[...]

PR #691, opened 2026-06-09
Add a core/settings ability

jorgefilipecosta:
## Summary

Part of: https://github.com/WordPress/ai/issues/40

Adds the `core/settings` ability to the plugin, matching the equivalent WordPress core change in WordPress/wordpress-develop#12141. Read-only; returns settings flagged `show_in_abilities` as a **flat** `name => value` map (metadata in the output schema), filterable by `group` **or** `slugs` (mutually exclusive). Requires `manage_options`.

- **`Settings`** — registers `core/settings`. Kept near-identical to core's `WP_Settings_Abilities` (differences marked with `// Plugin:` comments). It registers at priority 11 and unregisters any core-provided copy first, so the plugin's version wins when both are present.
- **`Show_In_Abilities`** — seeds the `show_in_abilities` flag onto a curated set of core settings via the `register_setting_args` filter, so the ability returns data on a stock site before core ships the flag. Object-type-agnostic, for future reuse (post types, meta).

## Test plan

**Prerequisite:** the client-side Abilities API (`@wordpress/abilities`) is only loaded in the block editor once an AI experiment is enabled — the experiment's editor script declares the `@wordpress/abilities` + `@wordpress/core-abilities` modules as dependencies, which adds them to the editor's import map. So first enable an experiment such as **Excerpt Generation** and open a new post (`post-new.php`). Then, in the browser console:

```js
const { executeAbility } = await import( '@wordpress/abilities' );

await executeAbility( 'core/settings', {} );
await executeAbility( 'core/settings', { group: 'reading' } );
await executeAbility( 'core/settings', { settings: [ 'blogname' ] } );
```

- [ ] Flat `name => value` map with typed values
- [ ] `group` / `slugs` filters work; supplying both is rejected
- [ ] Non-admin (no `manage_options`) is denied

Integration tests added under `tests/Integration/Includes/Abilities/`. An end-to-end test (`tests/e2e/specs/abilities/core-settings.spec.js`) drives the ability through the client-side Abilities API from the editor (with an experiment enabled).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-691-a28f8819ce14ba36cc1f2f27490a3a586e05b49b.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

Core PR: https://github.com/WordPress/wordpress-develop/pull/12141


includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,328 @@
...
+	 * wp_register_core_ability_categories() to register the `site` category.
+	 *
+	 * @since 1.1.0
+	 */
+	public static function register_category(): void {
+		if ( wp_has_ability_category( self::CATEGORY ) ) {
+			return;
+		}
+
+		wp_register_ability_category(
```
gziolo: Site category is already registered in WP core, and AI plugin requires at least version 7.0 so this code is not needed.

https://github.com/WordPress/wordpress-develop/blob/1caffb527744a65090bdd546bee3d7f27d2ed332/src/wp-includes/abilities.php#L18-L24

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,328 @@
...
+	 * describing the value.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return array<string, array{option: string, group: string, default: mixed, schema: array<string, mixed>}> Settings keyed by exposed name.
+	 */
+	protected static function get_exposed_settings(): array {
+		$settings = array();
+
+		foreach ( get_registered_settings() as $option_name => $args ) {
```
gziolo, line 240: I noticed that `self::get_exposed_settings()` gets called twice - when compiling the input schema and when reading the settings. In effect, the same computation happens twice. More importantly, since `get_registered_settings()` runs filters internally, it might be worth reusing the initially computed structure to ensure that the input and output schemas match what's being processed and returned.

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,328 @@
...
+ * ability via the shared helpers (get_exposed_settings(), value_schema(), cast_value()).
+ *
+ * This class is kept almost identical to the WordPress core class `WP_Settings_Abilities`
+ * so the two implementations stay in sync. Differences from the core class are marked with
+ * `// Plugin:` comments. Additionally, all user-facing strings use esc_html__() with the
+ * 'ai' text domain rather than core's __().
+ *
+ * @internal This class should not be used outside the plugin and there is no guarantee of backwards compatibility.
+ *
+ * @since 1.1.0
```
gziolo: Every new docblock uses `@since 1.1.0`, but `CONTRIBUTING.md` mandates `@since x.x.x` for new code.
```suggestion
 * @since x.x.x
```

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,328 @@
...
+		$properties    = array();
+		foreach ( $settings as $exposed_name => $setting ) {
+			$properties[ $exposed_name ] = $setting['schema'];
+		}
+
+		wp_register_ability(
+			'core/settings',
+			array(
+				'label'               => esc_html__( 'Get Settings', 'ai' ),
+				'description'         => esc_html__( 'Returns WordPress settings as a flat map of setting name to value. By default returns all settings exposed to abilities, or optionally a subset filtered by settings group or by setting name.', 'ai' ),
```
gziolo: Looking at https://github.com/WordPress/wordpress-develop/blob/1caffb527744a65090bdd546bee3d7f27d2ed332/src/wp-includes/abilities.php#L18-L24, I don't see usage of `esc_html__`. It should be perfectly fine to use `__` instead.

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,328 @@
...
+		switch ( $type ) {
+			case 'boolean':
+				return (bool) $value;
+			case 'integer':
+				return (int) $value;
+			case 'number':
+				return (float) $value;
+			case 'array':
+			case 'object':
+				return is_array( $value ) ? $value : array();
```
gziolo, line 306: Nit: for an `object`-typed setting, a non-array value returns `array()`, which JSON-encodes as `[]` rather than `{}` and can fail `object` output-schema validation now that `execute()` validates the output. Edge case, but worth guarding.

tests/Integration/Includes/Abilities/Settings/SettingsTest.php
```diff
@@ -0,0 +1,254 @@
...
+	 *
+	 * @since 1.1.0
+	 */
+	public function test_filters_by_settings(): void {
+		$this->become_admin();
+		$this->register_ability();
+
+		$result = wp_get_ability( 'core/settings' )->execute( array( 'settings' => array( 'blogname', 'posts_per_page' ) ) );
+
+		$this->assertSame( array( 'blogname', 'posts_per_page' ), array_keys( $result ) );
```
gziolo: Nit: this asserts the exact key order, which depends on `get_registered_settings()` iteration order. Comparing sorted keys (or `assertEqualSets()`) would be less brittle if registration order ever changes.

gziolo:
@jeffpaul or @dkotter, what are the conditions for this new (and other planned) ability to get included in the AI plugin? Code-wise it's shaping up nicely according to what we discussed in https://github.com/WordPress/ai/issues/40 and in https://core.trac.wordpress.org/ticket/64605. The idea would be to land it in the plugin first for early testing, and then put it in WordPress core.

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,328 @@
...
+							'enum'        => $groups,
+							'description' => esc_html__( 'Return only settings that belong to this settings group.', 'ai' ),
+						),
+					),
+					'additionalProperties' => false,
+				),
+				array(
+					'title'                => esc_html__( 'Filter by name', 'ai' ),
+					'type'                 => 'object',
+					'required'             => array( 'settings' ),
```
gziolo: `fields` as a name would be more aligned with other existing core abilities.

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,328 @@
...
+	protected static function get_settings_input_schema( array $groups, array $setting_names ): array {
+		return array(
+			'type'    => 'object',
+			'default' => array(),
+			// Filter by group OR by name, but not both at once.
+			'oneOf'   => array(
+				array(
+					'title'                => esc_html__( 'All settings', 'ai' ),
+					'type'                 => 'object',
+					'additionalProperties' => false,
```
gziolo: This will need the default value, because otherwise validation on REST API will fail.

One way to go about it is to reshape filtering by field name, so it is marked as optional with the default value translating to all fields, like here:

https://github.com/WordPress/wordpress-develop/blob/1caffb527744a65090bdd546bee3d7f27d2ed332/src/wp-includes/abilities.php#L98-L107

gziolo: I missed the default value one level up, this `oneOf` configuration is a bit confusing. I'm really curious whether in practice, these top-level `type` and `default` will be processed as `oneOf` defines their types.

includes/Abilities/Show_In_Abilities.php
```diff
@@ -0,0 +1,104 @@
...
+	 * optional `name` and `schema` keys (mirroring the `show_in_rest` shape). This matches
+	 * the set marked natively by the core `core/settings` implementation.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @return array<string, bool|array<string, mixed>> Settings map keyed by option name.
+	 */
+	public static function settings_map(): array {
+		return array(
+			// General.
```
gziolo, line 88: Noting that some settings will get duplicated between `core/get-site-info` and `core/settings`.

It seems like we will have to follow up in WP core to replicate `show_in_abilities` check inside `core/get-site-info` once that concept is there.

More broadly, `show_in_abilities` might probably need more variety, for example:

```php
`show_in_abilities` => true,
`show_in_abilities` => false,
`show_in_abilities` => 'read',
`show_in_abilities` => 'write',
```

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Good catch, removed. Dropped both register_category() and the wp_abilities_api_categories_init hook in init().

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Fixed. Added a private static $exposed_settings that's computed once in register_get_settings() and reused by execute_get_settings(), so the input/output schema and the returned values all derive from a single walk of get_registered_settings(). Mirrored in the core PR.

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Replaced every @since 1.1.0 with @since x.x.x across the new files (Settings.php, Show_In_Abilities.php, and the two integration test files).

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Switched all ability strings from esc_html__() to __().

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Guarded, the object case now returns (object) ( is_array( $value ) ? $value : array() ) so it serializes as {}. I also applied the same []-vs-{} fix to the input schema's top-level default (now (object) array()). Mirrored in core.

tests/Integration/Includes/Abilities/Settings/SettingsTest.php
jorgefilipecosta: Switched to assertEqualSets(). Did the same in the equivalent core test.

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Renamed settings → fields to match.

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Reshaped and simplified things, the oneOf is gone, the input is now a single object with optional group and fields properties plus a top-level default, mirroring core/get-site-info. The two filters are now combinable (intersection) rather than mutually exclusive.

FWIW the previous oneOf did validate over REST (CI + the e2e exercised {}/group/name and passed  WP_Ability::normalize_input() applies the top-level default when input is null, and {} matched the "all settings" branch), but the flatter shape is simpler and removes the ambiguity.

includes/Abilities/Show_In_Abilities.php
jorgefilipecosta: Yah show_in_abilities will pass by some evolution in some cases like show_in_rest it may also need to pass schema information, and read/write distinction may also be a good approach. Would you want me to already include read/write distinction as part of this PR or do you think this is something for the future?

jorgefilipecosta:
Hi @gziolo, thank you a lot for the review I think I applied/answered all the feedback.

dkotter:
> @jeffpaul or @dkotter, what are the conditions for this new (and other planned) ability to get included in the AI plugin? Code-wise it's shaping up nicely according to what we discussed in #40 and in https://core.trac.wordpress.org/ticket/64605. The idea would be to land it in the plugin first for early testing, and then put it in WordPress core.

No real condition here, other than we typically like to utilize the things we're adding within the AI plugin (instead of just adding something others can use).

But that said, if the goal here is to help prove this concept out so we can get movement upstream, that's also a fine goal and works for me.

I've held off on reviewing things here as it seemed like it was till a work in progress but let me know when things are settled and I'm happy to review and get this moving along for our next release.

jorgefilipecosta:
Hi @dkotter, I think this one is ready for a review ☺️

includes/Abilities/Show_In_Abilities.php
gziolo: For this PR, we don't need to do anything yet, as `true` is more than enough. However, I can see how `core/manage-settings` might benefit from a more granular decision process of what should be editable. Some settings should not be changed in isolation, as that could break the site.

gziolo:
I went through all the changes applied, and everything's addressed. Thank you!

Two small, non-blocking follow-ups:

-  Document the "register setting before `wp_abilities_api_init`" contract. The exposed set is snapshotted at registration, so a setting flagged with `show_in_abilities` that registers after that hook won't show up. Reasonable constraint worth a sentence in the `show_in_abilities` docs so integrators know their `register_setting()` has to run before abilities init.

- Drop the fallback in `execute_get_settings()`.
  ```php
  $settings = self::$exposed_settings ?? self::get_exposed_settings();
  ```
  A registered ability always has the cache set, so the recompute branch is unreachable. `$settings = self::$exposed_settings;` is enough. If it's ever `null` you'd want to know and act on it.

tests/e2e-sample-settings/e2e-sample-settings.php
dkotter, line 1: I would suggest instead of adding another E2E test plugin, let's just add this code to our existing E2E test plugin and maybe rename that plugin to be more generic, like `e2e-testing` or something similar. That way we can use that for any future things like this instead of having to add a new plugin each time

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,309 @@
...
+	/**
+	 * Builds the input schema for the get ability: optional filters by group and/or name.
+	 *
+	 * Both `group` and `fields` are optional; supplying both narrows the response to their
+	 * intersection, and supplying neither returns every exposed setting.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string[] $groups      Available settings groups.
+	 * @param string[] $field_names Available exposed setting names.
```
dkotter:
```suggestion
	 * @param list[] $groups      Available settings groups.
	 * @param list[] $field_names Available exposed setting names.
```

dkotter (changes requested):
Looks good, just left a few minor comments

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Good point, I think in this case list<string> may even fit better the structure so I used list<string>. If you prefer list[] for consistency or other reason let me know and I will change.

tests/e2e-sample-settings/e2e-sample-settings.php
jorgefilipecosta: Good point, I applied the feedback and renamed the plugin 👍

jorgefilipecosta:
Hi @gziolo your feedback was applied.

> I went through all the changes applied, and everything's addressed. Thank you!
> 
> Two small, non-blocking follow-ups:
> 
> * Document the "register setting before `wp_abilities_api_init`" contract. The exposed set is snapshotted at registration, so a setting flagged with `show_in_abilities` that registers after that hook won't show up. Reasonable constraint worth a sentence in the `show_in_abilities` docs so integrators know their `register_setting()` has to run before abilities init.


It was documented on Show_In_Abilities class.


> * Drop the fallback in `execute_get_settings()`.
>   ```
>   $settings = self::$exposed_settings ?? self::get_exposed_settings();
>   ```

>     
>   A registered ability always has the cache set, so the recompute branch is unreachable. `$settings = self::$exposed_settings;` is enough. If it's ever `null` you'd want to know and act on it.

Fixed 👍

jorgefilipecosta:
Hi @dkotter all your feedback was applied. Thank your for the review!

includes/Abilities/Show_In_Abilities.php
```diff
@@ -0,0 +1,109 @@
...
+ * @since x.x.x
+ */
+class Show_In_Abilities {
+
+	/**
+	 * Registers the hooks that mark core objects as exposed to abilities.
+	 *
+	 * @since x.x.x
+	 */
+	public static function register(): void {
```
justlevine: Is there a reason for all these methods to be `public static` instead of `public` like the rest of the repo? If we don't want users to call them directly (we don't), then IMO we should scope them.

```suggestion
	public function register(): void {
```

includes/Abilities/Show_In_Abilities.php
```diff
@@ -0,0 +1,109 @@
...
+	 *
+	 * The value is whatever `show_in_abilities` should contain: `true`, or an array with
+	 * optional `name` and `schema` keys (mirroring the `show_in_rest` shape). This matches
+	 * the set marked natively by the core `core/settings` implementation.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array<string, bool|array<string, mixed>> Settings map keyed by option name.
+	 */
+	public static function settings_map(): array {
```
justlevine: As far as I see, we're not using this anywhere outside of `::mark_setting()`

```suggestion
	private static function settings_map(): array {
```

includes/Abilities/Show_In_Abilities.php
```diff
@@ -0,0 +1,109 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array<string, mixed> $args         The setting registration arguments.
+	 * @param array<string, mixed> $defaults     The default registration arguments.
+	 * @param string               $option_group The settings group.
+	 * @param string               $option_name  The option name.
+	 * @return array<string, mixed> The (possibly amended) registration arguments.
+	 */
+	public static function mark_setting( array $args, array $defaults, string $option_group, string $option_name ): array {
```
justlevine: Same as above. If it's just for the hook, then let's make it a `public` instance method instead of `static`

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,315 @@
...
+	 */
+	public static function register(): void {
+		self::register_get_settings();
+
+		/*
+		 * A future write-oriented ability can be registered here, reusing the shared
+		 * helpers below (get_exposed_settings(), value_schema(), cast_value()):
+		 *
+		 *     self::register_manage_settings();
+		 */
```
justlevine, line 83: If this ability is called `core/settings`, and we're still blocked on https://github.com/WordPress/wordpress-develop/pull/12074, then what's the name for the write-oriented version?

Or is the expectation to rename this one for the other to coexist semantically?

includes/Abilities/Settings/Settings.php
justlevine, line 1: Instead of hoping someone notices the `@internal` can we scope this?

- `public` instead of `public static` to kickoff the instance in `Main` or for any hook callbacks.
- `private` instead `public static` or `protected static` for methods that are only intended to run inside the instance.0
- Bonus: make the class itself `final`.

It's a trivial and non-breaking to open these up if a use case presents itself, but going the other direction is a breaking change.

includes/Abilities/Settings/Settings.php
jorgefilipecosta: I will follow up with the write versions after the read only versions. The plan for the write versions is to have a core/manage-settings and core/manage-content pair.

includes/Abilities/Show_In_Abilities.php
jorgefilipecosta: Applied feedback. Made the class final and moved it to instance methods, register() and mark_setting() are instance methods now (mark_setting stays public since it's the register_setting_args callback) and settings_map() is private. Main bootstraps it with ( new Show_In_Abilities() )->register(). In this case there is no core equivalent class so we can follow the normal plugin patterns.

includes/Abilities/Show_In_Abilities.php
jorgefilipecosta: Feedback applied 👍

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Good point. Made the class final and tightened the visibility: only the entry points that are actually called from outside stay public static ( init and registe), and the two callbacks (execute_get_settings and has_permission ). register_get_settings and the helpers are private static now, and CATEGORY is a private const.
I kept the methods static rather than instance though: this class is a near-verbatim mirror of core's WP_Settings_Abilities (the // Plugin: comments mark the few deltas). And the plan is for the ability classes in core to be static, we don't need instance information, as the only goal there is call the ability registration.

jorgefilipecosta:
Thank you a lot for the review @justlevine and raising good points I applied/answered your feedback 👍

includes/Abilities/Settings/Settings.php
justlevine: > this class is a near-verbatim mirror of core's WP_Settings_Abilities 

PS: my feedback would apply there too, and probably be more important there considering that unlike a plugin it's a lot harder for core to deal with back compat.

So personally, I'd still turn (here and and in core) `::register()` (and all the callbacks etc) from `public static` to `public` - at most keeping `::init()` a `public static` method for convenience. For me it's not about instance _information_ but about coupling, hidden dependencies, and intention. We don't want people using those methods outside this class instance - the only reason they're public at all  is because of how WP's callback system, so scoping here has value.

includes/Abilities/Show_In_Abilities.php
justlevine: Looks perfect, thanks!

includes/Abilities/Settings/Settings.php
jorgefilipecosta: Got it I changed both this PR and core 👍

jorgefilipecosta:
@justlevine I think I applied the remaining point of removing static from settings.

justlevine:
> @justlevine I think I applied the remaining point of removing static from settings.

Yup. Codewise LGTM, I have some upstream concerns regarding the naming aspects of `core/settings` and `core/manage-settings`, but that shouldn't stop merging in the Experiment here 👍

dkotter (approved):

includes/Abilities/Settings/Settings.php
```diff
@@ -0,0 +1,315 @@
...
+
+		$settings = $this->exposed_settings;
+		if ( null === $settings ) {
+			// The cache is populated in register_get_settings() before the ability is
+			// registered, so this is unreachable in practice; bail defensively otherwise.
+			return array();
+		}
+
+		$group  = isset( $input['group'] ) && is_string( $input['group'] ) ? $input['group'] : '';
+		$fields = isset( $input['fields'] ) && is_array( $input['fields'] ) ? $input['fields'] : array();
```
gziolo, line 158: I was checking WP CLI and their [`wp option list`](https://developer.wordpress.org/cli/commands/option/list/) command. Options are a lower-level concept than settings, but they have some things in common. I have some second thoughts about using `fields` in this specific context, as an individual setting has more fields than a value, for example: group. 

On one hand, the current approach aligns closely with `core/get-site-info` and how it operates, but that's a curated set of site settings. 

On the other hand, if we look at https://github.com/WordPress/ai/pull/739 and `core/content`, it's evident that `fields` are used there for individual rows rather than filtering which rows to pick.

PR #692, opened 2026-06-10
Feature: Cleanup plugin data on plugin uninstall

### LINKED ISSUE #690: Plugin does not clean up database table and options on uninstall
When the plugin is deleted, it leaves behind a custom database table (`wp_wpai_request_logs`) and all of its WordPress options (`wpai_*`) permanently. There is no `uninstall.php` file and no `register_uninstall_hook()` call anywhere in the plugin.

The log table stores request and response previews of AI operations user-authored content and AI outputs. Leaving this data behind after uninstall raises data privacy concerns and violates [WordPress plugin guidelines](https://developer.wordpress.org/plugins/plugin-basics/uninstall-methods/) which require plugins to clean up their own data on deletion.

The full list of data left behind:

- Custom table: `wp_wpai_request_logs`
- Options: `wpai_features_enabled`, `wpai_feature_{id}_enabled`, `wpai_feature_{id}_field_developer` (one set per feature), `wpai_version`, `wpai_failed_upgrade_message`, `wpai_connector_approvals`, `wpai_connector_approval_pending`, `wpai_request_logs_schema_version`

Add an `uninstall.php` file to the plugin root that drops the custom table and removes all `wpai_*` options.

add a **"Remove all data on uninstall"** checkbox to the plugin settings (unchecked by default) so users can choose whether to keep or wipe their data.

Note: cleanup should happen on **uninstall only**, not on deactivation, since deactivation is reversible and users may reactivate the plugin later.

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/690

## Why?

- PR provides the way to cleanup the plugin data on uninstall. User's/Developers can opt it out using the filter.

## How?

- Adds uninstall.php file, Cleanup class to remove tables, options and cron schedules.

### Use of AI Tools

- Yes
- Claude Code, Opus 4.8
- Used for implementation. Implementation was reviewed and tested by me.

## Testing Instructions

1. Activate AI plugin.
2. Go to Settings --> AI.
3. Enable some settings.
4. Uninstall the plugin via plugin list page.
5. Open the database and check for `wpai_*` options or check for table `wp_wpai_request_logs`. This should not exists.

## Screenshots or screencast

https://github.com/user-attachments/assets/89239788-b24e-4738-86ef-4f1589c79210

## Changelog Entry

> Added - Cleanup plugin data on uninstall.

---

# AI Summary

## Summary

Implements full data cleanup when the plugin is **deleted**, addressing #690 (no `uninstall.php`, custom table and `wpai_*` options left behind). Cleanup is **opt-out** and **destructive**,  It always runs when uninstall, user can opt out using the filter `wpai_remove_data_on_uninstall` . Deletion happens on **uninstall only** — never on deactivation.

## Behavior

- Deleting the plugin removes:
  - The custom table `{$wpdb->prefix}wpai_request_logs`.
  - All `wpai_*` options.
  - All plugin transients (`_transient_wpai_*` / `_transient_timeout_wpai_*` and their site-transient equivalents).
  - The scheduled `wpai_request_logs_cleanup` cron event.
- **Multisite:** cleanup runs per-site (`get_sites()` + `switch_to_blog()`), and the opt-in is evaluated per site so each site controls its own data.
- Cleanup runs from `uninstall.php` (the WordPress-recommended approach) rather than `register_uninstall_hook()`, keeping the logic out of the normal runtime path.
- Developers/Users can opt out for cleanup using the filter `wpai_remove_data_on_uninstall`. If the filter returns false, the cleanup would not happen on deletion.

### Backend (PHP)

- **`uninstall.php`** (new) — plugin-root uninstall entry point. Guards on `WP_UNINSTALL_PLUGIN`, loads only the autoloader (deliberately **not** `ai.php`, to avoid bootstrapping `Main`), and calls `Uninstall::uninstall()`.
- **`includes/Admin/Uninstall.php`** (new) — the cleanup service:
  - `uninstall()` — multisite-aware orchestration (loops sites; single-site otherwise).
  - `maybe_clean_current_site()` — no-ops unless `wpai_remove_data_on_uninstall` is `true` for the current site.
  - Drops the request-logs table, deletes `wpai_*` options (via `esc_like` + `LIKE`), deletes plugin transients, and clears the request-log cleanup cron.

### Tests

- **`tests/Integration/Includes/Admin/UninstallTest.php`** (new):
  - `test_uninstall_removes_data_by_default` — seeds the table + `wpai_*` options + a scheduled event, opts in, asserts all are removed and a non-`wpai_` option is preserved.
  - `test_uninstall_preserves_data_when_filtered_out` — asserts data is untouched when the opt-out is enabled by filter.
  - Flushes the object cache after `uninstall()` since the bulk `DELETE` runs via direct SQL (irrelevant during a real uninstall request, but the in-request `alloptions` cache would otherwise return stale values).

## Implementation notes

- **Deactivation is untouched** — only `uninstall.php` performs deletion, so deactivating remains fully reversible.
- **Option sweep uses `LIKE 'wpai\_%'`** (escaped via `esc_like`) so dynamic per-feature options (`wpai_feature_{id}_enabled`, `wpai_feature_{id}_field_developer`, etc.) are covered without enumerating them.

## Testing instructions

1. Activate AI plugin.
2. Go to Settings --> AI.
3. Enable some settings.
4. Uninstall the plugin via plugin list page.
5. Open the database and check for `wpai_*` options or check for table `wp_wpai_request_logs`. This should not exists.
6. Run the integration tests:
   ```bash
   npm run wp-env:test start
   npm run test:php tests/Integration/Includes/Admin/UninstallTest.php
   ```

---

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-692-f0dddcd445cbbd505d58d4a84ae3ca84dac33163.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Settings/Settings_Registration.php
```diff
@@ -93,6 +94,17 @@ public function register_settings(): void {
 			)
 		);
 
+		register_setting(
```
dkotter: So I know this was mentioned in the attached Issue but not sure I agree with having a setting for this. This seems like a lot to ask the average user to have to decide if they want this checked or not (most won't fully understand what it means). I'm open to differing opinions here but I'd suggest we remove the setting and just have the data deletion happen by default. cc / @jeffpaul 

New data settings section

dkotter (changes requested):

includes/Settings/Settings_Registration.php
jeffpaul: Also fine to avoid an additional setting here as well and default to data deletion.

hbhalodia: Hi @dkotter @jeffpaul, In my opinion, we should go with what other plugins do mostly. As I have from 10 plugins, almost 8-9 plugins provide the an user option to cleanup the data in either way. 

Either they provide the settings or when deactivate/uninstall they provide a modal popup to ask user choice to retain/remove the data.

In any how, we should respect the user's choice if they want to cleanup or not.

If we are not adding this, we can update this choice when user try to uninstall the plugin?

dkotter: I still feel like there's no need for a setting here. If someone is deliberately deleting the entire plugin, seems the right answer is to clean up after ourselves.

That said, I'm not opposed to adding a setting here if the consensus is that's what users want. I would argue that setting should be checked by default though, whereas this PR has that unchecked. I'd also suggest we put this in a new `Advanced` section instead of `Data` and maybe have that section collapsed by default

jeffpaul: I concur, for now let's remove a setting and if community feedback demands a setting to be able to disable this on uninstall then we can support that (would be fine if this PR had a hook to allow someone to disable that way for now)

jeffpaul: @hbhalodia are you able to update based on the feedback here so this could get into the next plugin release?

hbhalodia: Hi @jeffpaul, Sorry for the delay here, I was away for sometime due to medical reasons. I would update the PR to remove the settings as suggested.

> would be fine if this PR had a hook to allow someone to disable that way for now

Do we add a filter hook, if returns true, delete the data, if returns false do not delete the data on uninstall is the expectation for now?

hbhalodia: Hi @jeffpaul @dkotter, This is now ready to review.

- Removed the settings.
- Add the filter to bypass the data removal if needed.

Let me know if anything needs to be updated here.

includes/Admin/Uninstall.php
```diff
@@ -0,0 +1,173 @@
...
+			$wpdb->esc_like( '_transient_wpai_' ) . '%',
+			$wpdb->esc_like( '_transient_timeout_wpai_' ) . '%',
+			$wpdb->esc_like( '_site_transient_wpai_' ) . '%',
+			$wpdb->esc_like( '_site_transient_timeout_wpai_' ) . '%',
+		);
+
+		foreach ( $patterns as $like ) {
+			$wpdb->query( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+				$wpdb->prepare(
+					"DELETE FROM {$wpdb->options} WHERE option_name LIKE %s",
```
dkotter: This won't delete anything that is stored at the multisite level (those are stored in `sitemeta` table). That may be fine but worth validating we don't care about that

includes/Admin/Uninstall.php
```diff
@@ -0,0 +1,173 @@
...
+		 * @param bool $remove_data Whether to remove all plugin data. Default true.
+		 */
+		if ( ! (bool) apply_filters( 'wpai_remove_data_on_uninstall', true ) ) {
+			return;
+		}
+
+		self::drop_request_logs_table();
+		self::delete_options();
+		self::delete_transients();
+		self::clear_scheduled_events();
```
dkotter, line 116: Is there any meta we should clean up or is that all good to leave behind?

dkotter (changes requested):
Left a few comments but otherwise looks good. Would be great to get the PR description updated as it still references an old approach

hbhalodia:
> Would be great to get the PR description updated as it still references an old approach

Sure would update that and will check the feedbacks!

includes/Admin/Uninstall.php
hbhalodia: Thanks for flagging this! Yes we have some meta that needs the cleanup like, `wpai_meta_description`. Will check for other meta as well and update the cleanup and tests.

hbhalodia: Hi @dkotter, I have an doubt here. Should we also delete the comments that are generated by `WordPress AI` and also the meta for the comments? 

Screenshot 2026-07-14 at 10 29 37 AM

Ideally, we should delete those.


Screenshot 2026-07-14 at 10 29 47 AM

For comments that are added by `WordPress AI` as a author can be deleted, but I see https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/includes/Experiments/Editorial_Notes/Editorial_Notes.php#L137, it is passed through translatable function, so everytime it's not fixed that string is always `WordPress AI`. We may need to introduce a new distinguish criteria, maybe an email so we can delete the comments added by our plugin. 

Some of the meta keys are,

Comment meta's - `_wpai_toxicity_score`, `_wpai_sentiment`, `_wpai_analysis_status`, `_wpai_analyzed_at`, `ai_note`. 
Post meta - `wpai_meta_description`, `ai_generated_summary`, `ai_generated`.
User meta - `wpai_connector_approval_notice_dismissed`.

Also I see some inconsistency with this meta keys, We have everything prefixed with `wpai`, while `ai_note` and `ai_genetated_summary` are not prefixed, which makes it hard to find and delete using `like` operator. We may need to revisit this and update it to make it consistent.

**Update**

I see we do have now encryption in place as well, with `_secret_<secret_provider_name>`, and master key `_secrets_master_key` stored in the options table. We may also need to delete those as a part of cleanup.

includes/Admin/Uninstall.php
hbhalodia: I checked the plugin and did'nt find any refrence that writes to multisite level common tables, everything is site specific. Hence that cleanup is not needed.

includes/Admin/Uninstall.php
dkotter: Sorry, I should have been more clear here. I don't know if we should be deleting all meta, I was thinking more about any meta that no longer serves a purpose with the plugin deleted.

For example, `wpai_connector_approval_notice_dismissed` makes sense to delete.

But not sure I'd recommend deleting things like `wpai_meta_description` or `ai_generated_summary` as I could see sites still wanting to use/display that data.

I'm not sure all the meta we end up with on the site but worth doing a sweep through the plugin to put a list together and then I think we can decide what should or should not be deleted, rather than just deleting everything. Thoughts on that?

> I see we do have now encryption in place as well, with _secret_<secret_provider_name>, and master key _secrets_master_key stored in the options table. We may also need to delete those as a part of cleanup.

This should ideally already be cleaning itself up but if that's not working, that should be part of the encryption experiment and not part of this new uninstall routine. The way it's supposed to be working is if you disable the plugin and you have encryption turned on, all secrets should be unencrypted at that point so a site isn't left with inaccessible data. I know that was working at one point so just need to verify that still is the case. But I wouldn't delete that data as part of this.

> Also I see some inconsistency with this meta keys, We have everything prefixed with wpai, while ai_note and ai_genetated_summary are not prefixed, which makes it hard to find and delete using like operator. We may need to revisit this and update it to make it consistent.

I see you already have a separate issue and PR for this. I'm fine with standardizing on using `wpai` but agree that can be tackled separately

hbhalodia: > But not sure I'd recommend deleting things like wpai_meta_description or ai_generated_summary as I could see sites still wanting to use/display that data.

I am still unsure on this to delete it or not, but IMO, there are very less chance that users/developers might directly use the meta `wpai_meta_description` or `ai_generated_summary`. But there is one direction I am thinking. It might be good to not delete because, anyway we are not deleting the posts/page etc, For eg, if an image is generated by AI plugin, then on plugin deletion the image would still be there, so why not its meta should be present? Same goes for post, whatever stored in meta would eventually can be used with/without AI plugin.

> I'm not sure all the meta we end up with on the site but worth doing a sweep through the plugin to put a list together and then I think we can decide what should or should not be deleted, rather than just deleting everything. Thoughts on that?

That makes sense to me, I will go through the plugin to find all the meta list and will add it here in comment, then we can decide what goes to data removal and what not.

**Update:** Here is the list from the AI plugin for all the metas -- https://github.com/WordPress/ai/pull/692#issuecomment-4977076693

> This should ideally already be cleaning itself up but if that's not working, that should be part of the encryption experiment and not part of this new uninstall routine. The way it's supposed to be working is if you disable the plugin and you have encryption turned on, all secrets should be unencrypted at that point so a site isn't left with inaccessible data. I know that was working at one point so just need to verify that still is the case. But I wouldn't delete that data as part of this.

Thanks for this, I was not aware of that. I would verify and update the PR to remove the removal of the secrets.

Update: The above is working as expected, The only thing I noticed that main encryption master key is still present on deactivation. `_secrets_master_key`, Let me know if that needs to be a part of cleanup?

hbhalodia:
@dkotter, Here is the full plugin sweep to find all the metas that are owned and registered by the plugin. 

A complete audit of all meta the AI plugin **registers or writes**, grouped by object type, so we can decide what to remove on uninstall and what to leave alone.

## ✅ Post meta — owned by the plugin

| Meta key | Type | How registered/written | Location |
|---|---|---|---|
| `ai_generated_summary` | string, `show_in_rest` | `register_meta('post', …)` | `includes/Experiments/Summarization/Summarization.php:90` |
| `ai_generated` | integer, `show_in_rest` (attachment) | `register_post_meta('attachment', …)` | `includes/Features/Image_Generation/Image_Generation.php:135` |
| `wpai_meta_description` | string, `show_in_rest` | `register_post_meta(<public types>, …)` — **only when no SEO plugin is active** (`FALLBACK_META_KEY`) | `includes/Experiments/Meta_Description/Meta_Description.php:106`, `includes/Abilities/Meta_Description/SEO_Integration.php:29` |

## ✅ Comment meta — owned by the plugin

| Meta key | Type | How registered/written | Location |
|---|---|---|---|
| `ai_note` | boolean, `show_in_rest` | `register_meta('comment', …)` | `includes/Experiments/Editorial_Notes/Editorial_Notes.php:60` |
| `_wpai_toxicity_score` | — | `update_comment_meta` (not registered) | const: `includes/Experiments/Comment_Moderation/Comment_Moderation.php:39`; write: `includes/Abilities/Comment_Moderation/Comment_Analysis.php:141` |
| `_wpai_sentiment` | — | `update_comment_meta` | const: `Comment_Moderation.php:46`; write: `Comment_Analysis.php:142` |
| `_wpai_analysis_status` | — | `update_comment_meta` | const: `Comment_Moderation.php:53`; write: `Comment_Analysis.php:129/136/143` |
| `_wpai_analyzed_at` | — | `update_comment_meta` | const: `Comment_Moderation.php:60`; write: `Comment_Analysis.php:144` |

## ✅ User meta — owned by the plugin

| Meta key | How written | Location |
|---|---|---|
| `wpai_connector_approval_notice_dismissed` | `update_user_meta` | `includes/Connector_Approval/Admin_Notice.php:28` |

## Term meta — none

The plugin registers/writes **no term meta**. It uses the `wp_guideline_type` taxonomy and assigns categories via `wp_set_object_terms`, but those are Gutenberg/core-owned taxonomy relationships, not plugin term meta.

---

## ❌ Meta the plugin touches but does NOT own — **do not remove**

| Meta key(s) | Owner | Plugin's involvement |
|---|---|---|
| `_wp_attachment_image_alt` | WordPress core | Alt-text generation writes it (`includes/CLI/Alt_Text_Command.php:364`) |
| `_yoast_wpseo_metadesc`, `rank_math_description`, `_aioseo_description`, `_seopress_titles_desc` | Respective SEO plugins | Meta Description writes into the **active** SEO plugin's key (`includes/Abilities/Meta_Description/SEO_Integration.php:41-54`) |
| `_guideline_copy`, `_guideline_images`, `_guideline_site`, `_guideline_additional`, `_guideline_block_*` | **Gutenberg** (`wp_guideline` CPT) | Guidelines service only **reads** them (`includes/Services/Guidelines.php:88-92, 200`); never registered or written by this plugin |

**Note:** Generated by claude code, Opus 4.8. 

Thanks,

includes/Admin/Uninstall.php
dkotter: > Update: The above is working as expected, The only thing I noticed that main encryption master key is still present on deactivation. _secrets_master_key, Let me know if that needs to be a part of cleanup?

Good to hear that's working (especially since I built that piece 😄). I don't feel strongly on the `_secrets_master_key`. I guess probably should add it to our list to delete though won't cause any problems if we leave it behind

dkotter:
> Here is the full plugin sweep to find all the metas that are owned and registered by the plugin

Thanks @hbhalodia for pulling this together, this is super helpful!

The more I think about it, especially after looking through that list, I think likely fine to just leave all of that meta behind, other than maybe deleting `wpai_connector_approval_notice_dismissed`. While it's unlikely any of that data will be needed/used after the AI plugin is removed, there may be scenarios where a site still wants it.

And as you mention, all of that is associated with other objects that we aren't deleting (posts, attachments, comments) so seems like best to just leave that behind. And most of these are all fairly minimal amount of data so shouldn't cause any sort of database bloat if left.

But let me know if you feel differently or have opinions based on what you've seen other plugins do, whether they leave meta behind or not. Thanks again for all the work here!

includes/Admin/Uninstall.php
hbhalodia: > I guess probably should add it to our list to delete though won't cause any problems if we leave it behind

IMO, we should delete this as we are deleting the plugin and our plugin is the cause of adding that option key. Once we again add the plugin and activate it, it would repopulate the key (once settings of encryption is enabled) If I am not wrong. So I am in favor to remove this.

hbhalodia:
> > Here is the full plugin sweep to find all the metas that are owned and registered by the plugin
> 
> Thanks @hbhalodia for pulling this together, this is super helpful!
> 
> The more I think about it, especially after looking through that list, I think likely fine to just leave all of that meta behind, other than maybe deleting `wpai_connector_approval_notice_dismissed`. While it's unlikely any of that data will be needed/used after the AI plugin is removed, there may be scenarios where a site still wants it.
> 
> And as you mention, all of that is associated with other objects that we aren't deleting (posts, attachments, comments) so seems like best to just leave that behind. And most of these are all fairly minimal amount of data so shouldn't cause any sort of database bloat if left.
> 
> But let me know if you feel differently or have opinions based on what you've seen other plugins do, whether they leave meta behind or not. Thanks again for all the work here!

Thanks for the update here!

While it's hard to say what other plugins do, I feel like all the plugins have different scenarios to cover and they do differently. There is no any exact source of truth what needs to be done. Its the plugins choice what to do with there data, i.e., to either keep, to either delete, or to delete some part and keep some.

I have a strong opinion here that, since we are not deleting the objects like for example a generated comment, or a generated image, deleting there meta would not be beneficial and may loose some key sources for identification, but still all those meta are part of AI plugin and removing them should not cause any disturbance to those objects, that would function as is. The above statement does not sounds a complete favour on either remove or keep.

I am thinking to get an opinion here from more team members on what they feel should be good choice for uninstall, that is to remove the meta that are added by plugin or keep them as is. @jeffpaul @justlevine 

Also thinking to at least go with any one thing for now and we can iterate it on future based on the feedbacks we receive.

Thanks,

dkotter:
>  have a strong opinion here that, since we are not deleting the objects like for example a generated comment, or a generated image, deleting there meta would not be beneficial and may loose some key sources for identification, but still all those meta are part of AI plugin and removing them should not cause any disturbance to those objects, that would function as is. The above statement does not sounds a complete favour on either remove or keep.

I would say let's just leave that meta behind and as you mention, can always update that in the future to be deleted if we get requests for that. But definitely open to second opinions here.

hbhalodia:
> > have a strong opinion here that, since we are not deleting the objects like for example a generated comment, or a generated image, deleting there meta would not be beneficial and may loose some key sources for identification, but still all those meta are part of AI plugin and removing them should not cause any disturbance to those objects, that would function as is. The above statement does not sounds a complete favour on either remove or keep.
> 
> I would say let's just leave that meta behind and as you mention, can always update that in the future to be deleted if we get requests for that. But definitely open to second opinions here.

Sure, I will wait for the 2nd Opinion before moving forward to remove the meta removal from the PR.

It would be great if we have that earliest. 😅

includes/Admin/Uninstall.php
miyanialkesh7: > 1. Comments those are generated by `WordPress AI` and also they have meta for the comments
Screenshot 2026-07-14 at 10 29 37 AM

**Adding 2nd opnion**, we should not remove the comment meta if we are not removing the comments those are generated by AI plugin.

Point to note, many times we're adding many custom fields and images with the site those are generating reference links in database tables but we are not using it and those data are left as it is.

----

> Screenshot 2026-07-14 at 10 29 47 AM

**Adding 2nd opnion**, if we will not remove the comment meta then we must not remove the comments those are generated with AI plugin.

Those might be useful for the customers to showcase dummy or AI generated comments to achieve some keywords goal. If we will remove it, then they loss the work!

----

👉 Same kind of flow we should implement for all the `WordPress core` tables, and their key option/values which are less important.

Primarly add more focus on API key storage values, and custom database tables those are generated by `WordPress AI` plugin.

jeffpaul:
I concur with Darin, so @hbhalodia please proceed with any further changes needed here... thanks!

hbhalodia:
Thanks @jeffpaul, I would update the PR as needed. To confirm, we would be leaving the post,comment etc meta's behind and would only remove the options as a cleanup part as per discussions.

jeffpaul:
Yes, let's see if community feedback wants to see more/less cleaned up and can iterate in a future release/PR as necessary

hbhalodia:
Hi @jeffpaul @dkotter, This is now ready to be reviewed again. Updates,

1. Removed post, comment meta from the cleanup process.
2. Updated the test to ensure everything works as expected.

Thanks,

jeffpaul (commented):
Will defer to @dkotter's review before merging

dkotter (approved):

PR #696, opened 2026-06-10
Fix[Meta Description]: Improve Copy and Focus Behavior

yogeshbhutkar:
## What, Why, and How?

Related to https://github.com/WordPress/ai/issues/589

This PR addresses the following issues:

1. Disables the Accept button in the Meta Description modal while generation is in progress to prevent users from accidentally closing the modal and losing the ongoing generation state.
2. Fixes focus loss after accepting the first generated description. Since the modal initiator, the Generate button, disappears after the first generation, focus is moved to the Edit description button. Subsequent focus handling is managed by the modal as expected.
3. Visually separates the Cancel button and moves it to the right to emphasize that it can be a destructive action, as closing the modal discards the ongoing generation state.
4. Disables the text area while generation is in progress, since any edits made during generation would be overwritten by the generated content once it completes.
5. Fixes the disabled Copy button remaining clickable when `useCopyToClipboard` is used with accessibleWhenDisabled, by only applying the hook’s returned ref when the button is enabled.
6. Updates the copy feedback when the description changes. Previously, the Copied message remained visible for a fixed `4000ms` even after the text was edited. It now immediately resets to Copy to clipboard, giving users accurate feedback.

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.5
Used for: Initial code skeleton; final implementation was reviewed and edited by me.

## Testing Instructions
1. Open the Meta Description experiment in the editor.
2. Generate a meta description.
3. Verify the modal opens with generated text.
4. Verify focus is handled correctly.
5. Close the modal.
6. Verify focus is not lost.
7. Verify disabled copy does nothing.
8. Edit or regenerate the description.
9. Verify copy, apply, and cancel states.

## Screencast

https://github.com/user-attachments/assets/b8bfaaa9-49d3-45eb-8484-da56dc05e692

## Changelog Entry

> Fixed - Prevent accidental interactions and stale feedback in the Meta Description generation modal and improve focus handling.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-696-ffdbb3261d6c70301fbefa1522c9a5e35ffbd9fa.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/meta-description/components/MetaDescriptionModal.tsx
+++ b/src/experiments/meta-description/components/MetaDescriptionModal.tsx
@@ -34,18 +34,30 @@ function CopyButton( {
 	text: string;
 	disabled: boolean;
 } ): JSX.Element {
-	const [ showCopyConfirmation, setShowCopyConfirmation ] = useState( false );
+	const [ copiedText, setCopiedText ] = useState< string | null >( null );
+	const showCopyConfirmation = copiedText === text && text.length > 0;
+
 	const timeoutIdRef = useRef< ReturnType< typeof setTimeout > >();
-	const ref = useCopyToClipboard< HTMLButtonElement >( text, () => {
-		speak( __( 'Meta description copied to clipboard.', 'ai' ) );
-		setShowCopyConfirmation( true );
-		if ( timeoutIdRef.current ) {
-			clearTimeout( timeoutIdRef.current );
+	const pendingCopyTextRef = useRef< string | null >( null );
+
+	const ref = useCopyToClipboard< HTMLButtonElement >(
+		() => {
+			pendingCopyTextRef.current = text;
+			return text;
+		},
+		() => {
+			speak( __( 'Meta description copied to clipboard.', 'ai' ) );
+			setCopiedText( pendingCopyTextRef.current );
+
+			if ( timeoutIdRef.current ) {
+				clearTimeout( timeoutIdRef.current );
+			}
+
+			timeoutIdRef.current = setTimeout( () => {
+				setCopiedText( null );
+			}, 4000 );
 		}
-		timeoutIdRef.current = setTimeout( () => {
-			setShowCopyConfirmation( false );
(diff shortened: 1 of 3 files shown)


src/experiments/meta-description/components/MetaDescriptionModal.tsx
```diff
@@ -34,18 +34,30 @@ function CopyButton( {
...
+	const showCopyConfirmation = copiedText === text && text.length > 0;
+	const isCopyDisabled = disabled || showCopyConfirmation;
+
 	const timeoutIdRef = useRef< ReturnType< typeof setTimeout > >();
-	const ref = useCopyToClipboard< HTMLButtonElement >( text, () => {
-		speak( __( 'Meta description copied to clipboard.', 'ai' ) );
-		setShowCopyConfirmation( true );
-		if ( timeoutIdRef.current ) {
-			clearTimeout( timeoutIdRef.current );
+	const pendingCopyTextRef = useRef< string | null >( null );
```
dkotter: Is this new `ref` needed? I may be wrong but looking at the code it seems like things would work the same without this

yogeshbhutkar: This was kept for a theoretical race condition, but after testing it several times in practice, I believe we can safely remove it. I’ve updated this PR to reflect that change.

dkotter (approved):

PR #734, opened 2026-06-16
feat: add settings import/export functionality

### LINKED ISSUE #191: Add import/export support for AI settings and provider configuration
Provide a way to import and export AI settings and provider configurations to support portability across environments. This is especially useful for agencies, hosts, and multi-site workflows.

Security and credential handling must be carefully considered.

Acceptance criteria:

* Settings can be exported in a portable format
* Settings can be imported safely and predictably
* Sensitive credentials are handled securely
* Feature is clearly documented
* Consider support for env variables for settings/provider configs

coderGtm:
## What?

Closes #191 

This PR implements the ability to export and import non-sensitive AI configuration settings and introduces a dedicated AI section in WordPress Site Health.

## Why?
Allowing users to move AI configurations between environments (e.g., staging to production, multi-site etc.) improves the developer experience. By excluding sensitive credentials from the export, we ensure security while maintaining portability. The Site Health integration provides administrators with a quick overview of the plugin's status and credential health.

## How?
- **REST API**: Created `includes/REST/Settings_IO_Controller.php` to handle `GET /ai/v1/settings/export` and `POST /ai/v1/settings/import`.
- **Security**: Implemented `SENSITIVE_PATTERNS` filtering to ensure API keys, tokens, and secrets are never exported or overwritten during import.
- **Site Health**: Added `includes/Admin/Site_Health.php` to provide debug information (version, enabled features, credential status) and a status test for connectivity.
- **UI/UX**: 
    - Added "Export settings" and "Import settings" to the AI settings menu in `routes/ai-home/stage.tsx`.
    - Implemented a confirmation modal for imports.
- **Tests**: Added comprehensive integration tests in `tests/Integration/Includes/REST/Settings_IO_ControllerTest.php` and `tests/Integration/Includes/Admin/Site_HealthTest.php`.

### Use of AI Tools
AI assistance: Yes
Tool(s): GitHub Copilot
Model(s): Claude Sonnet 4.6, GPT-4o
Used for: Authoring the codebase exploration, UI components, and the integration test suite.

## Testing Instructions
1. Navigate to **AI Settings**.
2. Open the kebab menu (three vertical dots) in the top right.
3. Click **Export settings**. A JSON file containing your current toggles and developer configurations (but no API keys) should download.
4. Toggle several features to different states.
5. Click **Import settings** in the menu and select the previously downloaded file.
6. Confirm the prompt "This will overwrite existing AI settings."
7. **Verification**: The UI should update immediately to reflect the imported settings without a page refresh.
8. Navigate to **Tools > Site Health**.
9. In the **Status** tab, verify the "AI Plugin" test results.
10. In the **Info** tab, verify the "AI Plugin" section contains correct metadata about your setup.

## Screenshots or screencast
Menu Items:
Screenshot 2026-06-16 at 1 56 51 PM

Confirmation prompt before completing import:
Screenshot 2026-06-16 at 1 57 51 PM

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-734-4e628104c15c4a9991f2ffbe51aadb878bed51a6.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


coderGtm:
The failing JavaScript quality check is due to a conflict between eslint and typescript. While eslint prefers dot notation, typescript complains about it ([TS4111](https://www.typescriptlang.org/tsconfig/#noPropertyAccessFromIndexSignature)). Should I suppress the eslint error?

dkotter:
> The failing JavaScript quality check is due to a conflict between eslint and typescript. While eslint prefers dot notation, typescript complains about it ([TS4111](https://www.typescriptlang.org/tsconfig/#noPropertyAccessFromIndexSignature)). Should I suppress the eslint error?

We do have a number of these that we already ignore with inline statements so I think that's fine

includes/Admin/Site_Health.php
```diff
@@ -0,0 +1,222 @@
...
+			return array(
+				'label'       => __( 'AI credentials are configured', 'ai' ),
+				'status'      => 'good',
+				'badge'       => array(
+					'label' => __( 'AI', 'ai' ),
+					'color' => 'blue',
+				),
+				'description' => sprintf(
+					'<p>%s</p>',
+					__( 'At least one AI connector has credentials configured and is ready to use.', 'ai' )
```
dkotter, line 140: So this isn't entirely accurate. The `has_ai_credentials` function will tell us if a Provider has credentials set but doesn't tell us if the credentials are valid. We do have a new `is_connector_configured` function that we could use here instead

routes/ai-home/stage.tsx
```diff
@@ -258,6 +262,73 @@ interface InfoTipProps {
 	content: string;
 }
 
+/** Shape of an AI settings export payload (matches the PHP schema). */
```
dkotter: So this file is already really large so if we can extract out any of these new functions into individual component or hook files, that would be great

includes/REST/Settings_IO_Controller.php
```diff
@@ -0,0 +1,323 @@
...
+	 * Substrings that indicate a sensitive option name.
+	 *
+	 * Any setting whose name contains one of these strings (case-insensitive)
+	 * is excluded from exports and rejected during imports.
+	 *
+	 * @since x.x.x
+	 *
+	 * @var list<string>
+	 */
+	private const SENSITIVE_PATTERNS = array( 'api_key', 'token', 'secret', 'credential', 'password', 'auth' ); // phpcs:ignore SlevomatCodingStandard.Classes.DisallowMultiConstantDefinition
```
dkotter: Worth adding just `key` to this list?

dkotter: Also risk here if there's a setting named `author`, I think that gets matched due to `auth`

includes/REST/Settings_IO_Controller.php
```diff
@@ -0,0 +1,323 @@
...
+			if ( ! is_string( $option_name ) ) {
+				continue;
+			}
+
+			// Only import options that are registered and allowed.
+			if ( ! isset( $exportable[ $option_name ] ) ) {
+				continue;
+			}
+
+			update_option( $option_name, $value );
```
dkotter: Slight risk here since we don't run any sort of validation or sanitization

includes/Main.php
```diff
@@ -127,6 +128,9 @@ public function initialize_features(): void {
 				( new Dashboard_Widgets( $registry ) )->init();
 			}
 
+			// Register Site Health integration (always, not just in admin).
```
dkotter: Does this need to load everywhere or could we scope this to the admin only?

includes/REST/Settings_IO_Controller.php
dkotter: And would be nice to capture this in a test, ensuring if someone passes in data that is the wrong type (i.e. we expect a boolean for a certain field and someone passes in string) that the data is either sanitized properly or rejected

dkotter (changes requested):
Overall this looks really good, left a few comments. In addition, an E2E test here would be nice to verify the import/export options are available

coderGtm:
Thanks @dkotter for reviewing this. I will work on the feedback over the days and request a review again.

includes/Main.php
coderGtm: Site_Health's two filters (debug_information, site_status_tests) are only ever read from `wp-admin/includes/class-wp-debug-data.php` and `class-wp-site-health.php`, but those files are also `require_once`'d by:

- `WP_REST_Site_Health_Controller` (wp-includes/rest-api/endpoints/class-wp-rest-site-health-controller.php) powers the async Site Health status/info REST endpoints used by the JS on the Site Health page, and by the Health Check API used by mobile apps/monitoring tools. REST requests do not set is_admin() === true.

- `wp_site_health_scheduled_check` which is the weekly WP-Cron job that emails site owners a health summary. Cron requests also don't set is_admin() === true.

So a hard `is_admin()` gate would silently break the REST-based Site Health UI refresh and the weekly cron email. However, we can safely exclude plain front-end page loads, since none of those invocation paths ever touch the front end.

Hence, `Main::initialize_features()` now only instantiates `Site_Health` when `is_admin() || wp_doing_cron() || REST_REQUEST` is true.

includes/Admin/Site_Health.php
coderGtm: Well, I am not entirely convinced that Site Health checks should also check the validity of the credentials. But if you insist, I can implement it that way as I am not able to side either way on this.

includes/REST/Settings_IO_Controller.php
coderGtm: Agreed. I modified the detection algorithm to check all underscore delimited segments from the patterns. As a result, I removed api_key and added just key to the patterns as it can detect both.

Also, this new algorithm prevents incomplete word matches like the one you flagged (author).

jeffpaul:
@coderGtm looks like a merge conflict, then perhaps good for re-review on our end?

coderGtm:
@jeffpaul I am yet to add the e2e test and checking the validation and sanitization again. So once I do those and resolve the merge conflict, I'll manually request a re-review.

Thanks for checking it out.

coderGtm:
I have made the suggested improvements to this PR and it is ready for re-review.

cc: @jeffpaul @dkotter

includes/Admin/Site_Health.php
dkotter: I don't feel strongly on if we should check validity of credentials here but we can't accurately say `At least one AI connector has credentials configured and is ready to use` if we just check if credentials exist. I'd suggest we at least modify that message and just remove the `and is ready to use` part

includes/REST/Settings_IO_Controller.php
```diff
@@ -0,0 +1,388 @@
...
+	public function export_settings(): \WP_REST_Response {
+		$exportable = $this->get_exportable_option_names();
+		$settings   = array();
+		$providers  = array();
+
+		foreach ( $exportable as $option_name ) {
+			$value = get_option( $option_name );
+
+			// Skip options that have never been set.
+			if ( false === $value ) {
```
dkotter: Something to consider here, if a feature has been manually turned off, we don't export that value due to this check (I believe). So in a situation where someone is exporting from one environment and importing into another, if feature X is turned off in the former but turned on in the latter, the import won't turn that feature off.

This may be desired but if the goal here is to be able to export settings from one environment and import those exactly the same into another environment we may want to capture disabled feature state

routes/ai-home/hooks/use-settings-import-export.ts
```diff
@@ -0,0 +1,189 @@
...
+ */
+export function useSettingsImportExport(): UseSettingsImportExportReturn {
+	const fileInputRef = useRef< HTMLInputElement >( null );
+	const [ pendingImport, setPendingImport ] =
+		useState< ExportPayload | null >( null );
+	const [ isImporting, setIsImporting ] = useState( false );
+
+	const { createSuccessNotice, createErrorNotice } =
+		useDispatch( noticesStore );
+	const { invalidateResolution } = useDispatch( coreStore ) as any;
```
dkotter: I think we can remove the `as any` piece here as we've been trying to remove that across the plugin where possible

dkotter (changes requested):

includes/REST/Settings_IO_Controller.php
coderGtm: Makes sense. In fact, I think we should also include the unset value's default states because if those options are changed in the target env then to ensure parity between both envs, we need to capture everything.

dkotter (approved):

PR #748, opened 2026-06-19
Fix[Image Generation]: Detect support for connectors that authenticate without an API key (e.g. OAuth)

diarmuidie:
## What?
Follow up to #679.

`has_image_generation_support()` skipped connectors that authenticate without an API key (e.g. OAuth) because it only checked for API-key credentials. It now exposes a `wpai_has_image_generation_support` filter so those connectors can advertise support — without making any live API requests.

## Why?
#679 gates the image generation UI on `has_image_generation_support()`. That check relied solely on `has_connector_authentication()`, which by design only detects API-key credentials (environment variable, constant, or stored option).

As a result, a provider that is fully configured but authenticates via OAuth was never counted as supporting image generation, and the image generation UI stayed hidden even though the provider exposes an image-generation-capable model.

This mirrors how AI credential detection already lets connectors that do not rely on API key settings advertise availability through a filter (`wpai_has_ai_credentials`).

## How?
The connector loop still only inspects connectors that have API-key credentials, which is a request-free check. Connectors that authenticate without an API key now advertise support through a new `wpai_has_image_generation_support` filter, applied to the computed result:

```php
$result = (bool) apply_filters( 'wpai_has_image_generation_support', $has_support, $connectors );
```

This deliberately avoids `is_connector_configured()` / `isProviderConfigured()`, which issues a live API request to verify the connection. `has_image_generation_support()` runs on nearly every admin page load, so it must stay request-free.

A `$reset_cache` parameter allows the per-request memoization to be recomputed (used by the tests, and useful when connector configuration changes within a single request).

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: Investigating the regression, drafting the filter-based fix, and writing the integration tests; reviewed and edited by me.

## Testing Instructions
Automated: `tests/Integration/Includes/HelpersTest.php` covers `has_image_generation_support()` for the API-key path, the filter path (advertising and suppressing support), the missing-capability case, and the skip path for connectors without credentials.

Manual:
1. Configure an AI connector that authenticates **without** an API key (e.g. OAuth) and exposes an image-generation-capable model, with no API key set.
2. From the connector plugin, hook `wpai_has_image_generation_support` to advertise support (e.g. return `true`).
3. Verify the image generation UI is now available (it was hidden before this change).
4. Regression check: an API-key connector with image support still shows the UI; with no qualifying connectors and no filter, it stays hidden.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-748-3fbd4050041ce6708dfe1ffb569b1bf19688bbdf.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/helpers.php
```diff
@@ -525,7 +534,12 @@ function has_image_generation_support(): bool {
 	$connectors = get_ai_connectors();
 
 	foreach ( array_keys( $connectors ) as $connector_id ) {
-		if ( ! has_connector_authentication( $connector_id ) ) {
+		// A connector qualifies when it has API-key credentials, or when its
+		// provider reports itself configured through the registry. The latter
+		// covers connectors that authenticate without an API key (e.g. OAuth),
+		// consistent with how text generation honors connectors that do not
+		// rely on API key settings.
+		if ( ! has_connector_authentication( $connector_id ) && ! is_connector_configured( $connector_id ) ) {
```
dkotter: So the problem with using `is_connector_configured` is it uses `isProviderConfigured` which makes actual API requests to verify the connection. We try and stay away from using this as we don't want to make unnecessary API requests and right now, `has_image_generation_support` is run on basically every admin page load

dkotter (commented):
I've left a concern on the use of `is_connector_configured` but I'd also like more information on where you're encountering this, what AI Provider are you using? My understanding is the Connectors API in WordPress Core doesn't yet support OAuth and only supports API key authentication so trying to understand the real world use case here.

diarmuidie:
> I've left a concern on the use of `is_connector_configured` but I'd also like more information on where you're encountering this, what AI Provider are you using? My understanding is the Connectors API in WordPress Core doesn't yet support OAuth and only supports API key authentication so trying to understand the real world use case here.

Yes, you are correct. This is for an internal AI Gateway which uses oAuth. We handle the auth in our AI connector plugin so from the WordPress connector side it looks like a connector with `authentication.method = none` and as a result Image generation fails.

includes/helpers.php
diarmuidie: Argh. Good catch. I'll see if there is a better way to accomplish this...

dkotter:
> Yes, you are correct. This is for an internal AI Gateway which uses oAuth. We handle the auth in our AI connector plugin so from the WordPress connector side it looks like a connector with `authentication.method = none` and as a result Image generation fails.

So an easy approach would be to add a filter within the `has_image_generation_support` function so 3rd parties (in this case your internal AI Gateway) could then modify the result of that to be whatever they want (and could run whatever checks they want). We already do this in a few places, like `has_ai_credentials` and `has_valid_ai_credentials`

diarmuidie:
> So an easy approach would be to add a filter within the `has_image_generation_support` function so 3rd parties (in this case your internal AI Gateway) could then modify the result of that to be whatever they want (and could run whatever checks they want). We already do this in a few places, like `has_ai_credentials` and `has_valid_ai_credentials`

That's much nicer! @dkotter  I've implemented it in [a11055f](https://github.com/WordPress/ai/pull/748/commits/a11055f334b4df3c2aa3755d5091f8383f87942b)

includes/helpers.php
```diff
@@ -507,48 +507,62 @@ function has_ai_credentials(): bool {
...
 
-	$result = false;
+	/**
+	 * Filters whether image generation is supported.
+	 *
+	 * Allows third-party plugins to declare image generation support for
+	 * connectors that do not rely on API key settings (e.g. OAuth), without
+	 * triggering a live API request.
+	 *
+	 * @since 1.0.3
```
dkotter: All `@since` statements, both here and in the test file, should be set to `x.x.x`

dkotter (changes requested):

includes/helpers.php
diarmuidie: Resolved in commit [3fbd405](https://github.com/WordPress/ai/commit/3fbd4050041ce6708dfe1ffb569b1bf19688bbdf).

All `@since` tags introduced by this PR, in both `includes/helpers.php` and the test file, are now set to the `x.x.x` placeholder.

dkotter (approved):

PR #749, opened 2026-06-19
Feat/add role user access controls

### LINKED ISSUE #736: Expose role/user access controls per feature/experiment

Originally requested in https://wordpress.org/support/topic/limit-use-to-certain-roles/:

> How do we limit the AI features to certain roles or users? I don’t see any capabilities associated with AI in the user role editor screens. Need to limit access on sites with many editors.

Similarly to how we expose provider and model selection via the top right elipses menu, we could add another option in that menu to expose role and user selection as well.  This could effectively add a multi-select option for roles and users that would specifically allow folks to fine-tune which roles or specific users have access to certain features/experiments, with the default being the set of existing roles already supported (perhaps as hard-coded in the plugin).

Infinite-Null:
Closes: #736

## Description

This PR introduces **Feature-Level Access Controls exclusively for Editor Experiments**, allowing site administrators to restrict access to individual AI-powered editor features based on specific WordPress roles or individual users.

## Testing Instructions

1.  Navigate to **AI Settings** and enable **Enable AI**.
2.  Open the Developer Tools menu (three dots) and select **Access controls**.
3.  Under the "Editor Experiments" section, verify that the Roles and Users configuration panel appears beneath the feature toggles.
4.  Check a few roles (e.g., Editor, Author) for a specific feature and click **Save**. Verify the save is successful.
5.  In the Users field, search for a specific user and select them from the dropdown. Click **Save**.
6.  Refresh the page and verify that your selected roles and users persisted and display their correct names.
7.  Turn off "Access controls" from the Developer Tools menu, and verify that all selections are cleared.
8.  **Verify Enforcements:** Log in to the site as a user who *does not* possess the selected role and is *not* the selected user. Open the block editor and verify that the restricted experiment is **not** accessible or visible to them.
9.  Log in as the specifically selected user or a user with the granted role, and verify the feature **is** fully accessible.
10. Turn off "Access controls" from the Developer Tools menu, and verify that all selections are cleared and the feature becomes globally available again.

#### User: Ankit Shah | Role: Editor

https://github.com/user-attachments/assets/ea1ace20-8d2e-4b01-8a57-94796efb5687

## Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Sonnet 4.6

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-749-d3312b28b3bc688d0e8929ac4138b0d580522450.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
@Infinite-Null FYI some merge conflicts to resolve to keep this moving along

Infinite-Null:
Thanks for the heads up, @jeffpaul! I've resolved the merge conflicts now. Whenever you have a chance, could you please take another look? Thanks!

henryperkins:
Thanks for working on this. I maintain Flavor Agent, a third-party `Abstract_Feature` that uses the AI plugin’s feature framework while registering its editor assets, recommendation/apply abilities, and dedicated MCP server through separate bootstrap paths.

The current implementation appears to enforce access by calling `ai_current_user_can_access_feature()` inside each built-in Editor Experiment’s `register()` method. That handles those experiments, but I do not think it yet defines a complete contract for third-party features.

In Flavor Agent, for example, `Abstract_Feature::register()` owns the editor asset hook, while the feature’s Abilities and MCP server are registered elsewhere. An unauthorized user could therefore lose the editor UI while related programmatic surfaces remain registered unless the plugin independently knows which helper and option contract to call.

Would it make sense to:

1. Apply the access check centrally in the Feature Loader before invoking `register()`.
2. Treat `ai_current_user_can_access_feature( $feature_id )` as a documented, stable third-party API.
3. Add a test-only third-party `Abstract_Feature` fixture.
4. Document that separate Ability, REST, CLI, or MCP execution paths must use the same resolver in their permission boundary.
5. Include a test proving that an unauthorized direct Ability execution is denied rather than merely hidden from the settings/editor UI.

I can contribute the failing third-party integration test and the centralized enforcement change. The important acceptance criterion for me is that an unauthorized user receives neither the first-party feature surface nor an executable programmatic path, while existing WordPress capability checks remain in force.

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
```diff
@@ -59,6 +59,10 @@ protected function load_metadata(): array {
 	 * {@inheritDoc}
 	 */
 	public function register(): void {
+		if ( ! \WordPress\AI\ai_current_user_can_access_feature( $this->get_id() ) ) {
```
dkotter: I don't love forcing each experiment to run this same code. Since all experiments extend `Abstract_Feature`, could a check like this live there so any experiment that extends that class automatically gets this behavior? Or maybe even better in our Loader class?

tests/Integration/Includes/REST/Roles_Users_ControllerTest.php
```diff
@@ -0,0 +1,300 @@
...
+namespace WordPress\AI\Tests\Integration\Includes\REST;
+
+use WP_REST_Request;
+use WP_UnitTestCase;
+use WordPress\AI\REST\Roles_Users_Controller;
+
+/**
+ * Roles_Users_Controller test case.
+ *
+ * @since 0.9.0
```
dkotter: All `@since` statements in this PR need to be set to `x.x.x`

includes/REST/Roles_Users_Controller.php
```diff
@@ -0,0 +1,94 @@
...
+
+class Roles_Users_Controller {
+
+	private const API_NAMESPACE = 'ai/v1';
+
+	private const ROUTE = '/roles-users';
+
+	private const MAX_USERS = 10;
+
+	public function init(): void {
```
dkotter, line 55: Any reason to not have docblocks on all methods and constants in this class?

includes/REST/Roles_Users_Controller.php
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		$search         = (string) $request->get_param( 'search' );
+		$get_users_args = array(
+			'fields' => array( 'ID', 'display_name' ),
+			'number' => self::MAX_USERS,
+		);
+
+		if ( '' !== $search ) {
+			$get_users_args['search']         = '*' . $search . '*';
```
dkotter, line 119: Any escaping we should be applying here?

includes/helpers.php
```diff
@@ -740,3 +740,31 @@ function post_type_supports_bulk_action( string $post_type, string $feature_id )
...
 	}
 }
+
+/**
+ * Checks whether the current user has access to a given feature based on access control settings.
+ *
+ * If no roles or users are explicitly configured for the feature, it allows access by default.
+ * If there are configured roles/users, the current user must match at least one role or be explicitly listed.
+ *
+ * @since 0.1.0
```
dkotter:
```suggestion
 * @since x.x.x
```

includes/helpers.php
```diff
@@ -740,3 +740,31 @@ function post_type_supports_bulk_action( string $post_type, string $feature_id )
...
+ *
+ * @param string $feature_id The ID of the feature/experiment.
+ * @return bool True if the user has access, false otherwise.
+ */
+function ai_current_user_can_access_feature( string $feature_id ): bool {
```
dkotter: No need to use the `ai_` prefix here as this is a namespaced file

routes/ai-home/components/AccessControlSettings.tsx
```diff
@@ -0,0 +1,251 @@
...
+	const handleSave = useCallback( async () => {
+		await save();
+		setLocalRoles( null );
+		setLocalUsers( null );
+	}, [ save ] );
+
+	return (
+		<div
+			className="ai-access-control-mode-fields ai-feature-settings-form"
+			style={ { marginTop: 'var(--wpds-dimension-gap-md, 12px)' } }
```
dkotter: Ideally we don't use any inline styles here but move those to our CSS file

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
dkotter: For example, this PR doesn't apply this to Image Generation or Type Ahead

routes/ai-home/stage.tsx
```diff
@@ -820,6 +835,50 @@ function AISettingsPage() {
 		]
 	);
 
+	const handleToggleAccessControlMode = useCallback( async () => {
```
dkotter: It seems this resets things when the mode is turned off. That's not how this should work. Toggling it on/off just manages the UI but doesn't change what is saved. For instance, if I toggle `Model selection` off, any feature that had selected a model still recognizes that but the UI is no longer showing

dkotter: Also worth noting that we store the Model Selection toggle in browser localstorage just to determine if the UI shows for that particular person. But turning it on or off doesn't actually change any data, just whether the UI is visible. Right now if two users are on the site messing with things, one can easily toggle the UI on for themselves and wipe the data someone else was setting up

routes/ai-home/components/AccessControlSettings.tsx
```diff
@@ -0,0 +1,251 @@
...
+	}, [ suggestions ] );
+
+	const selectedUsersTokens = useMemo( () => {
+		return effectiveUsers.map(
+			( id ) => selectedUserMap.get( id ) ?? id.toString()
+		);
+	}, [ effectiveUsers, selectedUserMap ] );
+
+	// Seed selectedUserMap with every user returned from the API so that
+	// saved users always show their display name instead of a raw ID.
```
dkotter: I don't think this is always true. Our hard cap is 10 users so if you set more than 10 users, I think it will show an ID not a name

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
dkotter: The other issue to consider is that some features do other things in this `register` method beyond just loading the UI and the Ability. For instance, custom post meta is registered or caches are cleared. Those things will no longer run for certain users, if restriction is turned on, leading to weird edge cases that will be hard to debug. Ideally we figure out a way to prevent the actual experiment from running but still allow other pieces to function properly

routes/ai-home/components/AccessControlSettings.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+interface AccessControlSettingsProps {
+	featureId: string;
+}
+
+export function AccessControlSettings( {
+	featureId,
+}: AccessControlSettingsProps ): React.JSX.Element {
+	const { roles, isLoading, fetchError } = useRoles();
+	const { suggestions, isSearching, search } = useUserSearch();
```
dkotter: Both of these will be called for each feature that is on, even though the underlying request I think is the same. Ideally this data is hoisted out of this so it only runs once per page, not once per component

includes/helpers.php
```diff
@@ -740,3 +740,31 @@ function post_type_supports_bulk_action( string $post_type, string $feature_id )
...
+		return true;
+	}
+
+	$current_user = wp_get_current_user();
+
+	if ( in_array( $current_user->ID, $users, true ) ) {
+		return true;
+	}
+
+	return (bool) array_intersect( $current_user->roles, $roles );
```
dkotter: Ideally we add a custom filter around this for further customization

tests/Integration/Includes/REST/Roles_Users_ControllerTest.php
dkotter, line 1: Should add tests for the new helper function, `ai_current_user_can_access_feature`, to ensure it works as expected.

Also, most of the tests here are fairly shallow (and repetitive) in what they're testing, some just test `register_rest_route`'s own behavior rather than our code. Ideally these are cleaned up and made more robust, especially considering this feature will be seen as a security enhancement

tests/e2e/specs/admin/settings.spec.js
```diff
@@ -727,4 +729,485 @@ test.describe( 'Plugin settings', () => {
...
+
+		// Disable Access Controls.
+		await visitSettingsPage( admin );
+		await disableAccessControls( page );
+
+		// Disable the Content Summarization experiment.
+		await disableExperiment( admin, page, 'Content Summarization' );
+
+		// Delete the test user.
+		await requestUtils.deleteAllUsers();
```
dkotter: This seems like it will delete all users, not just the test user. Can we make this more specific?

tests/e2e/specs/admin/settings.spec.js
```diff
@@ -727,4 +729,485 @@ test.describe( 'Plugin settings', () => {
...
+			await expect( resetSaveButton ).not.toBeVisible( {
+				timeout: 10000,
+			} );
+		}
+		await expect( editorCheckboxInitial ).not.toBeChecked();
+
+		// Create a post.
+		const {
+			RequestUtils: WPRequestUtils,
+		} = require( '@wordpress/e2e-test-utils-playwright' );
```
dkotter: Why are we requiring this here?

tests/e2e/specs/admin/settings.spec.js
```diff
@@ -727,4 +729,485 @@ test.describe( 'Plugin settings', () => {
...
+			await roleSaveButton.click();
+			await expect( roleSaveButton ).not.toBeVisible( {
+				timeout: 10000,
+			} );
+		}
+
+		// Create a post as admin so the editor can open it.
+		const {
+			RequestUtils: WPRequestUtils,
+		} = require( '@wordpress/e2e-test-utils-playwright' );
```
dkotter: Same question

tests/e2e/specs/admin/settings.spec.js
```diff
@@ -727,4 +729,485 @@ test.describe( 'Plugin settings', () => {
...
+		);
+		await expect( addedToken ).toBeVisible( { timeout: 10000 } );
+
+		// Save the user access settings.
+		const userSaveButton = contentSummarizationAccessFormAgain.getByRole(
+			'button',
+			{ name: 'Save' }
+		);
+		await expect( userSaveButton ).toBeVisible( { timeout: 10000 } );
+		await userSaveButton.click();
```
dkotter, line 1177: Do we need to wait for save to complete before we proceed? Seems this could be brittle

includes/helpers.php
```diff
@@ -740,3 +740,31 @@ function post_type_supports_bulk_action( string $post_type, string $feature_id )
...
+	$roles = get_option( "wpai_feature_{$feature_id}_roles", array() );
+	$users = get_option( "wpai_feature_{$feature_id}_users", array() );
+
+	if ( empty( $roles ) && empty( $users ) ) {
+		return true;
+	}
+
+	$current_user = wp_get_current_user();
+
+	if ( in_array( $current_user->ID, $users, true ) ) {
```
dkotter:
```suggestion
	if ( in_array( $current_user->ID, array_map( 'intval', (array) $users ), true ) ) {
```

includes/helpers.php
```diff
@@ -740,3 +740,31 @@ function post_type_supports_bulk_action( string $post_type, string $feature_id )
...
+ * @since 0.1.0
+ *
+ * @param string $feature_id The ID of the feature/experiment.
+ * @return bool True if the user has access, false otherwise.
+ */
+function ai_current_user_can_access_feature( string $feature_id ): bool {
+	$roles = get_option( "wpai_feature_{$feature_id}_roles", array() );
+	$users = get_option( "wpai_feature_{$feature_id}_users", array() );
+
+	if ( empty( $roles ) && empty( $users ) ) {
```
dkotter, line 888: Do we also need an `is_array` check here on these?

includes/Settings/Settings_Registration.php
```diff
@@ -132,6 +134,40 @@ public function register_settings(): void {
 				)
 			);
 
+			register_setting(
```
dkotter, line 201: Might want to add a `sanitize_callback` to both of these to ensure we only save real user roles and real user IDs

routes/ai-home/components/AccessControlSettings.tsx
```diff
@@ -0,0 +1,251 @@
...
+			style={ { marginTop: 'var(--wpds-dimension-gap-md, 12px)' } }
+		>
+			{ isLoading && <Spinner /> }
+			{ ! isLoading && fetchError && (
+				<p className="ai-access-control-mode-field__error">
+					{ fetchError }
+				</p>
+			) }
+			{ ! isLoading && ! fetchError && (
+				<>
```
dkotter: Is this fragment necessary?

routes/ai-home/components/AccessControlSettings.tsx
```diff
@@ -0,0 +1,251 @@
...
+								<Button
+									variant="primary"
+									onClick={ handleSave }
+									disabled={ isSaving }
+									size="compact"
+									isBusy={ isSaving }
+									accessibleWhenDisabled
+								>
+									{ isSaving ? (
+										<Spinner />
```
dkotter: No need to display a spinner here, we can rely on `isBusy`

dkotter (changes requested):

jeffpaul:
@Infinite-Null please let us know when you're updated from comments/code review feedback for us to test this out again, thanks!

Infinite-Null:
> @Infinite-Null please let us know when you're updated from comments/code review feedback for us to test this out again, thanks!

Hey @jeffpaul, I am out till 18th August due to personal commitments. I will update the comments/code review after then. Thank you for understanding.

includes/Experiments/Alt_Text_Generation/Alt_Text_Generation.php
Infinite-Null: Because the access gate bails early in `register()`, meta registrations and cache hooks weren't running for restricted users.

To fix this, We can add a `register_infrastructure()` method to `Abstract_Feature` that runs before the access gate check. Experiments can override this to register post meta or global hooks that must run for all users.

Here’s how it looks in practice:

```php
    // Runs for all users regardless of access controls
    $this->register_infrastructure();

    if ( ! \WordPress\AI\current_user_can_access_feature( $this->get_id() ) ) {
        return;
    }

    // UI & abilities only load for authorized users
    add_action( 'wp_abilities_api_init', array( $this, 'register_abilities' ) );
    add_action( 'enqueue_block_editor_assets', array( $this, 'enqueue_assets' ) );
}

protected function register_infrastructure(): void {
    $this->register_post_meta();
}
```

cc: @dkotter @jeffpaul

includes/REST/Roles_Users_Controller.php
Infinite-Null: I think no manual escaping needed here. `WP_User_Query` (used internally by `get_users()`) already escapes the search param via `esc_like()` + `prepare()`. 


Ref: https://developer.wordpress.org/reference/functions/get_users/
Ref: https://developer.wordpress.org/reference/classes/wp_user_query/

`WP_User_Query::get_search_sql()`: the method that builds the LIKE clause via esc_like() + prepare(): https://developer.wordpress.org/reference/classes/wp_user_query/get_search_sql/
`WP_User_Query::prepare_query()`: method that invokes get_search_sql() as part of the query build: https://developer.wordpress.org/reference/classes/wp_user_query/prepare_query/

routes/ai-home/components/AccessControlSettings.tsx
Infinite-Null: Thank you for the feedback. I have updated the comment: 

```php
// Seed selectedUserMap with users returned from the API (capped at
// MAX_USERS i.e. 10 at a time). If more than
// MAX_USERS users are saved, any beyond the cap won't be included in
// this response and will fall back to showing their raw ID.
```

routes/ai-home/components/AccessControlSettings.tsx
Infinite-Null: Thank you for the feedback @dkotter, I have Hoisted `useRoles()` and `useUserSearch()` into a shared `RolesUsersProvider/context`, wrapped once at the top of `stage.tsx` instead of inside each AccessControlSettings instance.

tests/Integration/Includes/REST/Roles_Users_ControllerTest.php
Infinite-Null: > Should add tests for the new helper function, ai_current_user_can_access_feature, to ensure it works as expected.

Added the helper test to `tests/Integration/Includes/HelpersTest.php`

> Also, most of the tests here are fairly shallow (and repetitive) in what they're testing, some just test register_rest_route's own behavior rather than our code. Ideally these are cleaned up and made more robust, especially considering this feature will be seen as a security enhancement

Updated the tests to address the shallow/repetitive coverage

tests/e2e/specs/admin/settings.spec.js
Infinite-Null: Addressed this by swapping out `deleteAllUsers()` for a scoped `requestUtils.rest()` DELETE call targeting just `createdEditorUser.id`.

tests/e2e/specs/admin/settings.spec.js
Infinite-Null: Thank you for the feedback I have added an explicit wait for the success snackbar ('Access control settings saved.') after both the role-level and user-level Save button clicks

includes/Settings/Settings_Registration.php
Infinite-Null: Added `sanitize_callback` to both settings.

Infinite-Null:
I’ve addressed all the feedback and updated the PR accordingly. Thank you so much for taking the time to review it and for the detailed feedback.

Whenever you get a chance, could you please take another look.

cc: @jeffpaul @dkotter

jeffpaul:
1. I'm not seeing this as available on the Image Generation and Editing feature.
2. I think we can remove `Subscriber` as a role listed as none of these would be exposed for those folks anyway (perhaps `Contributor` as well?).
3. I think we should probably expose these for the Comment Moderation and Suggest Reply experiments as well.  Otherwise maybe add a note to the Admin Experiments text that notes that all experiments unless otherwise specified are scoped to admins-only?

Infinite-Null:
Hi @jeffpaul, Thanks for the feedback! I’ve updated the PR accordingly. I’ve also removed both Subscriber and Contributor from the listed roles.

Infinite-Null:
The failing tests seem to be unrelated to the changes introduced in this PR.

PR #757, opened 2026-06-22
fix(logging): capture generations that bypass the SDK HTTP transporter

### LINKED ISSUE #732: AI Request Logging only captures providers that use the SDK HTTP transporter; sidecar/custom-transport providers are invisible
## Summary

The AI Request Logging experiment captures requests by decorating the SDK's HTTP transporter. Any provider that performs its own HTTP (rather than routing generation through `AiClient`'s transporter) is therefore never logged — even though it's a first-class provider reachable via `wp_ai_client_prompt()` and shown in the Connectors UI. The Request Log silently under-reports, which is surprising given it presents as a log of *every* AI request.

## Root cause

`Logging_Integration::wrap_transporter()` installs the tap by swapping in a decorator:

```php
// includes/Logging/Logging_Integration.php
$registry = AiClient::defaultRegistry();
$registry->setHttpTransporter( new Logging_Http_Transporter( $registry->getHttpTransporter(), $log_manager ) );
```

Logging then happens entirely inside `Logging_Http_Transporter::send()`. A request is logged **iff** it flows through `HttpTransporter::send()`. Provider attribution compounds this — `Log_Data_Extractor::detect_provider()` derives the provider from the request **host**, so a provider talking to `127.0.0.1` couldn't be attributed even if its request were seen.

## Reproduction

1. Enable the **AI Request Logging** experiment.
2. Use a provider that doesn't route generation through the SDK transporter (e.g. one that proxies to a localhost sidecar or otherwise calls `wp_remote_request()` itself).
3. Run a generation via `wp_ai_client_prompt()`. It succeeds, but **no row** appears in **Tools → AI Request Log**. The same site logs OpenAI/Anthropic generations correctly.

## Impact

- Any provider with a custom transport is omitted from the log — observability, cost/token accounting, and debugging all under-count.
- Concrete case: a third-party `codex` provider that brokers requests through a localhost sidecar for ChatGPT-managed auth. It registers normally and serves generations, but is entirely absent from the Request Log. We've worked around it provider-side, but that means every such provider must reimplement logging and couple to `AI_Request_Log_Manager` internals — which is exactly what this issue asks to avoid.

## Proposed fix

Add a **provider-agnostic fallback** alongside the transporter decorator: have the experiment also listen to the core generation events, which fire for *every* provider regardless of transport — `wp_ai_client_before_generate_result` / `wp_ai_client_after_generate_result` (bridged from the SDK's `BeforeGenerateResultEvent` / `AfterGenerateResultEvent`, with the dispatcher registered in `wp-settings.php`).

The after-event exposes everything a row needs:

- `getModel()->providerMetadata()->getId()` — provider
- `getModel()->metadata()->getId()` — model
- `getResult()->getTokenUsage()` — tokens
- duration = before → after

To avoid double-logging transporter-based providers (which fire the event *and* pass through the decorator), the listener should only write when the transporter did **not** already log the current generation — e.g. a flag reset in the before-event and set by `Logging_Http_Transporter::send()`.

**Known limitation:** the SDK's after-event fires on success only (there is no error event), so this fallback captures *successful* non-transporter generations but not failed ones. That still closes the success case for all providers; failed custom-transport generations remain a separate gap that would need an SDK-level error event.

## Alternatives considered

- **Provider-side logging** (our current stopgap): works, but pushes coupling to `AI_Request_Log_Manager` into every provider and is easy to get wrong.
- **Hooking the WP HTTP API** (`pre_http_request`): too broad, noisy, and can't reliably attribute the provider.

## Prior art

#680 fixed a different "the logger misses data" case (thinking-token undercount) — same spirit of completeness.

i-anubhav-anand:
## What?
Closes #732

Adds a provider-agnostic logging fallback so the **AI Request Logging** experiment captures successful generations from providers that don't route through the SDK's HTTP transporter (e.g. a provider that proxies to a localhost sidecar).

## Why?
Today logging is installed purely by decorating the SDK HTTP transporter (`Logging_Http_Transporter`), and `Logging_Http_Transporter::send()` is the only place that calls `AI_Request_Log_Manager::log()`. A request is logged **iff** it flows through that transporter. Any first-class provider reachable via `wp_ai_client_prompt()` that uses a custom transport is therefore silently absent from **Tools → AI Request Log**, even though the log presents itself as a record of *every* AI request. The concrete case in #732 is a `codex` provider that brokers requests through a localhost sidecar.

## How?
WordPress core registers a PSR-14 dispatcher in `wp-settings.php` (`WordPress\AiClient\AiClient::setEventDispatcher( new WP_AI_Client_Event_Dispatcher() )`) that bridges the SDK's `BeforeGenerateResultEvent` / `AfterGenerateResultEvent` to the action hooks `wp_ai_client_before_generate_result` / `wp_ai_client_after_generate_result`. These fire from `PromptBuilder::generateResult()`, through which **all** `generate*()` calls funnel — for every provider, regardless of transport.

This PR adds `Logging_Event_Listener`, wired up in `Logging_Integration::init()`, which:

- On `wp_ai_client_after_generate_result`, writes a log row using the event:
  - provider = `getModel()->providerMetadata()->getId()`
  - model = `getModel()->metadata()->getId()`
  - tokens = `getResult()->getTokenUsage()` (prompt/completion)
  - duration = before-event → after-event
- Avoids double-logging transporter-based providers (which fire the event **and** pass through the decorator) with a per-generation flag: reset on the before-event, set by `Logging_Http_Transporter::send()`. When the transporter already logged the current generation, the after-event listener skips it.

**Known limitation (documented in the code):** the SDK's after-event fires on success only — there is no error event — so failed *custom-transport* generations remain uncaptured. Transporter-based providers still log errors as before. Closing the failure case for custom transports would require an SDK-level error event and is out of scope here.

The diff is surgical: +8 lines across the two existing files, plus one new listener class and one new test file.

### Use of AI Tools
AI assistance: Yes
Tool(s): AI coding assistant (agentic CLI)
Used for: investigating the SDK event/dispatcher flow in `wp-includes/php-ai-client` and `wp-includes/ai-client`, drafting the listener + tests, and writing this description. All code was written test-first, executed, and reviewed by me; I take responsibility for it.

## Testing Instructions
Automated (PHPUnit, via `wp-env`):

```
npm run wp-env:test start
npm run test:php -- --filter Logging
```

- New tests: `tests/Integration/Includes/Logging/Logging_Event_ListenerTest.php`
  - `test_logs_generation_that_bypasses_the_transporter` — a generation that never touches `Logging_Http_Transporter` produces one row with the right provider/model/tokens.
  - `test_does_not_double_log_transporter_based_generations` — a transporter-logged generation yields exactly one row (no event-based duplicate).
- Full Logging suite (`--filter Logging`) is green (120 tests); the full integration suite is green (838 tests, 34 pre-existing skips).

**Fail-before / pass-after** (source stashed, test kept):

```
# before (unmodified source):
1) ...Logging_Event_ListenerTest::test_logs_generation_that_bypasses_the_transporter
Expected the sidecar generation to produce one log row.
Failed asserting that 0 is identical to 1.

# after (with fix):
OK (2 tests, 7 assertions)
```

Manual: enable **AI Request Logging**, run a generation through a provider with a custom transport (localhost sidecar), and confirm a row now appears in **Tools → AI Request Log**.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-757-92ddff8b548b3dfc6099e146994bbe6cceed162f.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Logging/Logging_Event_Listener.php
+++ b/includes/Logging/Logging_Event_Listener.php
@@ -1,0 +1,152 @@
+<?php
+/**
+ * Provider-agnostic logging fallback driven by SDK generation events.
+ *
+ * @package WordPress\AI\Logging
+ */
+
+declare( strict_types=1 );
+
+namespace WordPress\AI\Logging;
+
+use WordPress\AiClient\Events\AfterGenerateResultEvent;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Logs successful generations that do not flow through the SDK HTTP transporter.
+ *
+ * The {@see Logging_Http_Transporter} only sees requests routed through the SDK's
+ * HTTP transporter. Providers with a custom transport (for example one that proxies
+ * to a localhost sidecar) never reach it and would otherwise be absent from the log.
+ *
+ * This listener taps the core generation lifecycle hooks, which fire for every
+ * provider regardless of transport:
+ *
+ * - `wp_ai_client_before_generate_result`
+ * - `wp_ai_client_after_generate_result`
+ *
+ * These are bridged from the SDK's `BeforeGenerateResultEvent` / `AfterGenerateResultEvent`
+ * by the dispatcher registered in WordPress core (`wp-settings.php`).
+ *
+ * To avoid double-logging transporter-based providers (which fire the event *and*
+ * pass through the decorator), a per-generation flag is reset on the before-event and
+ * set by {@see Logging_Http_Transporter::send()}; the after-event only writes a row
+ * when the transporter did not already log the current generation.
+ *
+ * Known limitation: the SDK's after-event fires on success only (there is no error
+ * event), so failed custom-transport generations are not captured here.
+ *
+ * @since 1.0.0
+ */
+class Logging_Event_Listener {
+
+	/**
+	 * Whether the transporter logged the current generation.
+	 *
+	 * @var bool
+	 */
+	private static bool $transporter_logged = false;
+
+	/**
+	 * The log manager instance.
+	 *
+	 * @var \WordPress\AI\Logging\AI_Request_Log_Manager
+	 */
+	private AI_Request_Log_Manager $log_manager;
+
+	/**
+	 * Timer for the in-progress generation.
+	 *
+	 * @var array{start: int}|null
+	 */
+	private ?array $timer = null;
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param \WordPress\AI\Logging\AI_Request_Log_Manager $log_manager The log manager.
+	 */
+	public function __construct( AI_Request_Log_Manager $log_manager ) {
+		$this->log_manager = $log_manager;
+	}
+
+	/**
+	 * Registers the generation lifecycle listeners.
+	 *
+	 * @since 1.0.0
+	 */
+	public function register(): void {
+		add_action( 'wp_ai_client_before_generate_result', array( $this, 'handle_before_generate' ) );
+		add_action( 'wp_ai_client_after_generate_result', array( $this, 'handle_after_generate' ) );
+	}
+
+	/**
+	 * Marks the current generation as already logged by the HTTP transporter.
+	 *
+	 * Called by {@see Logging_Http_Transporter::send()} so the after-event does not
+	 * write a duplicate row for transporter-based providers.
+	 *
+	 * @since 1.0.0
+	 */
+	public static function mark_transporter_logged(): void {
+		self::$transporter_logged = true;
+	}
+
+	/**
+	 * Resets per-generation state at the start of a generation.
+	 *
+	 * @since 1.0.0
+	 */
+	public function handle_before_generate(): void {
+		self::$transporter_logged = false;
+		$this->timer              = $this->log_manager->start_timer();
+	}
+
+	/**
+	 * Logs the generation result unless the transporter already logged it.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param object $event The after-generate event.
+	 */
+	public function handle_after_generate( object $event ): void {
+		$timer       = $this->timer;
+		$this->timer = null;
+
+		// The SDK transporter already captured this generation; avoid a duplicate row.
+		if ( self::$transporter_logged ) {
+			return;
+		}
+
+		if ( ! $event instanceof AfterGenerateResultEvent ) {
+			return;
+		}
+
+		$model      = $event->getModel();
+		$provider   = $model->providerMetadata()->getId();
+		$model_id   = $model->metadata()->getId();
+		$capability = $event->getCapability();
+		$capability = null !== $capability ? (string) $capability : '';
+		$tokens     = $event->getResult()->getTokenUsage();
+
+		$log_data = array(
+			'type'          => 'ai_client',
+			'operation'     => '' !== $capability ? $provider . ':' . $capability : $provider,
+			'provider'      => $provider,
+			'model'         => $model_id,
+			'tokens_input'  => $tokens->getPromptTokens(),
+			'tokens_output' => $tokens->getCompletionTokens(),
+			'status'        => 'success',
+			'context'       => array( 'logged_via' => 'generation_event' ),
+		);
+
+		if ( null !== $timer ) {
+			$log_data['duration_ms'] = $this->log_manager->end_timer( $timer );
+		}
+
+		$this->log_manager->log( $log_data );
+	}
+}
--- a/includes/Logging/Logging_Event_Listener.php
+++ b/includes/Logging/Logging_Event_Listener.php
@@ -31,22 +31,22 @@ defined( 'ABSPATH' ) || exit;
  *
  * To avoid double-logging transporter-based providers (which fire the event *and*
  * pass through the decorator), a per-generation flag is reset on the before-event and
- * set by {@see Logging_Http_Transporter::send()}; the after-event only writes a row
- * when the transporter did not already log the current generation.
+ * set via {@see Logging_Event_Listener::mark_generation_logged()}; the after-event only
+ * writes a row when the generation was not already logged.
  *
  * Known limitation: the SDK's after-event fires on success only (there is no error
  * event), so failed custom-transport generations are not captured here.
  *
- * @since 1.0.0
+ * @since x.x.x
  */
 class Logging_Event_Listener {
 
 	/**
-	 * Whether the transporter logged the current generation.
+	 * Whether another logger already recorded the current generation.
 	 *
 	 * @var bool
 	 */
-	private static bool $transporter_logged = false;
+	private static bool $generation_logged = false;
 
 	/**
 	 * The log manager instance.
@@ -65,7 +65,7 @@ class Logging_Event_Listener {
 	/**
 	 * Constructor.
 	 *
-	 * @since 1.0.0
+	 * @since x.x.x
 	 *
 	 * @param \WordPress\AI\Logging\AI_Request_Log_Manager $log_manager The log manager.
 	 */
@@ -76,7 +76,7 @@ class Logging_Event_Listener {
 	/**
 	 * Registers the generation lifecycle listeners.
 	 *
-	 * @since 1.0.0
+	 * @since x.x.x
 	 */
 	public function register(): void {
 		add_action( 'wp_ai_client_before_generate_result', array( $this, 'handle_before_generate' ) );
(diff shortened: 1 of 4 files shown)


includes/Logging/Logging_Event_Listener.php
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * To avoid double-logging transporter-based providers (which fire the event *and*
+ * pass through the decorator), a per-generation flag is reset on the before-event and
+ * set by {@see Logging_Http_Transporter::send()}; the after-event only writes a row
+ * when the transporter did not already log the current generation.
+ *
+ * Known limitation: the SDK's after-event fires on success only (there is no error
+ * event), so failed custom-transport generations are not captured here.
+ *
+ * @since 1.0.0
```
dkotter: See our contributor guide but all `@since` statements need to be set to `x.x.x`

includes/Logging/Logging_Event_Listener.php
```diff
@@ -0,0 +1,152 @@
...
+		$model      = $event->getModel();
+		$provider   = $model->providerMetadata()->getId();
+		$model_id   = $model->metadata()->getId();
+		$capability = $event->getCapability();
+		$capability = null !== $capability ? (string) $capability : '';
+		$tokens     = $event->getResult()->getTokenUsage();
+
+		$log_data = array(
+			'type'          => 'ai_client',
+			'operation'     => '' !== $capability ? $provider . ':' . $capability : $provider,
```
dkotter, line 139: I think this is different than what we log in our transporter. Ideally this matches for filtering.

includes/Logging/Logging_Event_Listener.php
```diff
@@ -0,0 +1,152 @@
...
+		if ( ! $event instanceof AfterGenerateResultEvent ) {
+			return;
+		}
+
+		$model      = $event->getModel();
+		$provider   = $model->providerMetadata()->getId();
+		$model_id   = $model->metadata()->getId();
+		$capability = $event->getCapability();
+		$capability = null !== $capability ? (string) $capability : '';
+		$tokens     = $event->getResult()->getTokenUsage();
```
dkotter, line 135: Is there ever a chance this returns `null` or empty, and thus would fatal when we call `$tokens->getPromptTokens()` below?

dkotter (changes requested):

dkotter:
@henryperkins I know you reported the original issue, do you by chance have some time to verify if this PR fixes things for your use case? Thanks!

jeffpaul:
@I-anubhav-anand are you able to make updates based on code review?

henryperkins:
@dkotter  Sorry for the delayed response here, and thanks for the ping — I tested this against ai-provider-for-codex v2.1 at 2827bb5, using WordPress 7.0.1, PHP 8.3.31, AI Client 1.3.1, and this PR at bb6a7a7.

I invoked the actual wp_ai_client_prompt()->using_model(...)->generate_text_result() path with a seeded user connection/model catalog. I intercepted only the loopback /v1/responses/text response to keep the test deterministic, so this did not use a live authenticated Codex account.

Results:

* At the PR base, with the provider’s own logging bridge disabled: 0 log rows.
* At this PR’s head, with that bridge disabled: exactly 1 row, with provider codex, model gpt-5-codex, tokens 7/3, status success, and logged_via=generation_event.
* At this PR’s head with the current provider unchanged: 2 success rows:
    * codex:responses/text, written by the provider’s existing RequestLogWriter
    * codex:text_generation, written by this PR’s event listener

The duplicate occurs because the provider already works around #732 by writing directly to the request log. This PR’s suppression flag only detects Logging_Http_Transporter, so it cannot tell that a custom provider has already logged the request.

I also tested a failed runtime response. The after-generation event did not fire, so this PR added no failure row. The current provider’s existing bridge did record one error row, consistent with the known limitation described in the PR.

So the event fallback does fix the original zero-log gap once the provider-side workaround is disabled or removed. With the current provider unchanged, however, successful requests are logged twice. The two paths also produce different operation values, which appears relevant to the existing filtering review comment.

i-anubhav-anand:
Thanks both — and thanks especially for the thorough live test @henryperkins, that's really useful.

To summarise where this leaves the PR:

- **The core fix works.** As Henry confirmed, with the provider's own logging bridge disabled the event listener takes the log from 0 rows → exactly 1 correct row (`provider: codex`, `model: gpt-5-codex`, tokens, `logged_via: generation_event`). That closes the original gap in #732 for any custom-transport provider.
- **But it double-logs when a provider's own bridge is still active.** With the codex provider unchanged, a successful generation produces two rows: `codex:responses/text` (the provider's own `RequestLogWriter`) and `codex:text_generation` (this listener). The suppression flag only detects `Logging_Http_Transporter`, so it has no way to know a custom provider already logged the request.
- **The two paths also produce different `operation` values** (`provider:capability` here vs `provider:basename(url_path)` in the transporter path via `Log_Data_Extractor`). That's the same inconsistency @dkotter flagged in review — Henry's repro just makes it concrete.

Plan for the follow-up commit:

1. `@since` tags → `x.x.x` per the contributor guide.
2. Add a null/empty guard before `getTokenUsage()->getPromptTokens()/getCompletionTokens()` (re: @dkotter's question — the SDK's after-event should always carry token usage on success, but I'll fail safe rather than rely on it).
3. Align the `operation` string format between the transporter path and the event-listener path so filtering is consistent regardless of which path wrote the row.
4. Generalise the suppression signal: rename `Logging_Event_Listener::mark_transporter_logged()` into a transport-neutral public API (e.g. "mark this generation as already logged") that any logger — not just the SDK transporter — can call.

On the duplicate specifically: this PR alone can't fully de-dupe a provider that logs through its own path, so item 4 is the hook that lets it. @henryperkins — since the codex bridge already gates on `class_exists()` for the AI plugin and resolves its sink through the `codex_provider_request_log_sink` filter, would you prefer to (a) call the new marker before the bridge's **success**-path write (keeping its **error**-path write, which this listener structurally can't capture since the SDK fires no error event), or (b) drop the bridge's success path entirely once this lands? Happy to shape the marker API around whichever you'd rather. I'll also note the residual duplicate risk explicitly in the PR description as a known limitation for providers that don't adopt it.

(Separately, minor: the bridge writes `type => 'text'`, but the REST `type` filter enum only accepts `'', 'ai_client', 'mcp_tool', 'ability'`, so codex rows aren't filterable by type — not a blocker for this PR, just flagging it as further motivation to make the event listener the canonical path.)

Will push the follow-up shortly.

i-anubhav-anand:
Pushed a follow-up commit addressing the review. Summary of what changed and one open question for @dkotter:

**1. `@since` tags** — all `@since` statements in the new files are now `x.x.x` per the contributor guide.

**2. Possible `null` token usage (@dkotter's question)** — I dug into the SDK's type declarations rather than add a guard, and the chain is non-nullable end to end:
- `AfterGenerateResultEvent::getResult(): GenerativeAiResult`
- `GenerativeAiResult::getTokenUsage(): TokenUsage`
- `TokenUsage::getPromptTokens(): int` / `getCompletionTokens(): int` (typed properties set in the constructor)

So `getTokenUsage()` can't return `null` on this path — and if the SDK ever violated its own contract, PHP would raise the `TypeError` inside the SDK before this listener ran. I've therefore left the direct calls in place rather than adding an unreachable guard; happy to add one anyway if you'd prefer belt-and-suspenders.

**3. De-duplication across custom loggers (re: @henryperkins's double-log finding)** — the suppression signal was transporter-specific (`mark_transporter_logged()`, keyed off `Logging_Http_Transporter`). I've renamed it to a transport-neutral public marker, `Logging_Event_Listener::mark_generation_logged()`, so **any** logger — not just the SDK transporter — can mark a generation as already recorded and have the after-event skip it. @henryperkins, since the codex bridge already gates on `class_exists()` for the AI plugin, it can call this marker right before its **success**-path write (keeping its **error**-path write, which this listener structurally can't capture), and the duplicate row goes away.

**Open question on the `operation` value (@dkotter)** — your review flagged that the event path's `operation` differs from the transporter's, and that's real, but the two are built from different data and can't be made identical: the transporter derives `provider:basename(url_path)` (the endpoint, e.g. `openai:chat/completions`) via `Log_Data_Extractor`, while the event listener only has the capability enum, giving `provider:capability` (e.g. `openai:text_generation`). The listener never sees the URL, so it can't reproduce the endpoint form. Which taxonomy would you like as the canonical one for the filter — endpoint-based or capability-based? Happy to normalise both paths to whichever you prefer; I left the current behaviour in place pending your call rather than guess.

I'll also expand the "known limitation" note in the description to cover the success-case duplicate for providers that don't adopt the marker.

henryperkins:
Thanks — and one clarification for the record: this was a deterministic WordPress/provider integration-path test with the loopback sidecar response mocked, not a live authenticated Codex generation.

My preference is (b), with a backward-compatible feature check:

* When the event-listener fallback is available and active, skip the provider’s success-path write and let the AI plugin provide the canonical success row.
* Keep the provider’s error-path write because the SDK currently emits no corresponding error event.
* On older AI plugin versions without the event listener, retain the existing success-path bridge.

That avoids coupling the provider’s normal success path to another logging marker and gives the request log consistent type, operation, and filtering behavior. A transport-neutral marker could still be useful for custom providers that intentionally retain their own success logger.

One caveat if the marker-based approach is retained: it should be called only after RequestLogWriter::record() returns true, not before the write. record() is best-effort and can fail; marking first could suppress the event listener even though the provider’s write failed, returning the request to zero log rows.

Once the follow-up commit is available, I’m happy to rerun the same matrix against it and then update the provider around the final API.

dkotter:
> The listener never sees the URL, so it can't reproduce the endpoint form. Which taxonomy would you like as the canonical one for the filter — endpoint-based or capability-based? Happy to normalise both paths to whichever you prefer; I left the current behaviour in place pending your call rather than guess.

We definitely want to standardize and if this new approach of logging only has access to the capability, that seems like a pretty easy answer to make that the standard for filtering. But let me know if I'm missing something there.

PR #761, opened 2026-06-23
Add save button to developer settings panel (Provider/Model selectors)

### LINKED ISSUE #763: Developer Settings: Model selections should require explicit Save (currently auto-save on change)
## Summary
In the Developer Settings (Provider/Model selectors), changing the Provider or Model dropdown saves the change immediately, there's no Save button. This is different from how the rest of the settings work, where you make changes and then click Save to confirm them.

Ref: https://github.com/WordPress/ai/issues/736#issuecomment-4735392961
> The Provider and Model settings don't do that right now but they probably should

## Current behavior
- Changing the Provider or Model selection saves it right away, with no confirmation step.

## Proposed behavior
- Changing Provider/Model should not save immediately.
- A **Save** button should appear once you've made a change.
- The change should only be saved when you click Save.

## Why this matters
This makes the Provider and Model selection Settings behave consistently with the rest of the settings, and avoids accidentally saving a change you didn't mean to make.

Infinite-Null:
## What?
Closes  #763

Adds an explicit **Save** button to the Developer Settings panel (Provider/Model selectors), instead of saving changes immediately on every selection.

## Why

Provider and Model changes were being saved as soon as you picked a new option, with no confirmation step. This was inconsistent with the rest of the settings UI, where changes are only saved when you click Save. A moderator flagged this during review.

## How to test

1.  Navigate to **AI Settings** and enable **Enable AI**.
2.  Open the Developer Tools menu (three dots) and select **Model selection**.
3. Change the Provider or Model.
4. Confirm a Save button appears and the change is **not** saved yet.
5. Click Save and confirm the change persists.
6. Confirm clicking on "Reset to default" shows a Save button, and Save still behaves as expected.

## Before / After

**Before:** Selecting a new Provider or Model saved it instantly.
**After:** Selecting a new Provider or Model shows a Save button; the change is only saved when you click it.

#### Before

https://github.com/user-attachments/assets/6680d5f2-b1e5-47d2-ad6c-2f99130da460

#### After

https://github.com/user-attachments/assets/9e25cde5-364e-4a83-bcde-984e0cc157c0

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-761-0c8acabb5acee4e1df782ffd5c172f176a15f81b.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/routes/ai-home/components/DeveloperSettings.tsx
+++ b/routes/ai-home/components/DeveloperSettings.tsx
@@ -4,7 +4,13 @@
 import { Button, Spinner } from '@wordpress/components';
 import { DataForm } from '@wordpress/dataviews';
 import type { Field, Form } from '@wordpress/dataviews';
-import { useCallback, useEffect, useMemo, useRef, useState } from '@wordpress/element';
+import {
+	useCallback,
+	useEffect,
+	useMemo,
+	useRef,
+	useState,
+} from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { Notice } from '@wordpress/ui';
 
@@ -46,7 +52,8 @@ export function DeveloperSettings( {
 	const { settings, update, clear, isSaving } =
 		useDeveloperFeatureSettings( featureId );
 
-	const [ draftSettings, setDraftSettings ] = useState< DeveloperSelection | null >( null );
+	const [ draftSettings, setDraftSettings ] =
+		useState< DeveloperSelection | null >( null );
 	const [ isSavingThis, setIsSavingThis ] = useState( false );
 
 	useEffect( () => {
@@ -62,7 +69,9 @@ export function DeveloperSettings( {
 	const currentSettings = draftSettings ?? settings;
 
 	const getModelElements = useCallback( () => {
-		const provider = providers.find( ( p ) => p.id === currentSettings.provider );
+		const provider = providers.find(
+			( p ) => p.id === currentSettings.provider
+		);
 		if ( ! provider ) {
 			return Promise.resolve( [] );
 		}
@@ -112,7 +121,10 @@ export function DeveloperSettings( {
 	const handleChange = useCallback(
 		( changes: Partial< DeveloperSelection > ) => {
 			if ( 'provider' in changes ) {
-				setDraftSettings( { ...currentSettings, provider: changes.provider ?? '', model: '' } );
+				setDraftSettings( {
+					provider: changes.provider ?? '',
+					model: '',
+				} );
 			} else {
 				setDraftSettings( { ...currentSettings, ...changes } );
 			}
@@ -131,7 +143,8 @@ export function DeveloperSettings( {
 		}
 	}, [ draftSettings, update, clear ] );
 
-	const hasSavedSelection = currentSettings.provider !== '' || currentSettings.model !== '';
+	const hasSavedSelection =
+		currentSettings.provider !== '' || currentSettings.model !== '';
 	const hasUnsavedChanges =
 		draftSettings !== null &&
 		( draftSettings.provider !== settings.provider ||
@@ -203,9 +216,10 @@ export function DeveloperSettings( {
 							<Button
 								variant="primary"
 								onClick={ handleSave }
-								disabled={ isSavingThis || !hasUnsavedChanges }
+								disabled={ isSavingThis || ! hasUnsavedChanges }
 								isBusy={ isSavingThis }
 								accessibleWhenDisabled
+								__next40pxDefaultSize
 							>
 								{ __( 'Save', 'ai' ) }
 							</Button>
@@ -221,12 +235,11 @@ export function DeveloperSettings( {
 										)
 										?.focus();
 									setDraftSettings( {
-										...currentSettings,
 										provider: '',
 										model: '',
 									} );
 								} }
-								disabled={ isSaving }
+								disabled={ isSavingThis }
 								accessibleWhenDisabled
 							>
 								{ __( 'Reset to default', 'ai' ) }
--- a/routes/ai-home/components/DeveloperSettings.tsx
+++ b/routes/ai-home/components/DeveloperSettings.tsx
@@ -135,16 +135,12 @@ export function DeveloperSettings( {
 	const handleSave = useCallback( () => {
 		if ( draftSettings ) {
 			setIsSavingThis( true );
-			if ( draftSettings.provider === '' && draftSettings.model === '' ) {
-				void clear();
-			} else {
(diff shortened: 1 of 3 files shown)


routes/ai-home/style.scss
```diff
@@ -88,7 +88,11 @@ body:has(.ai-settings-page) .components-popover {
 		margin-bottom: var(--wpds-dimension-gap-lg, 12px);
 	}
 
-	.ai-developer-mode-fields__reset-button {
-		margin-top: var(--wpds-dimension-gap-lg, 12px);
+	.ai-developer-mode-fields__actions {
+		display: flex;
+		align-items: center;
+		gap: var(--wpds-dimension-gap-lg, 16px);
```
dkotter, line 94: Looks like we're inconsistent with the fallback here, some places use 16px some use 12px. Should standardize that

routes/ai-home/components/DeveloperSettings.tsx
```diff
@@ -46,8 +52,26 @@ export function DeveloperSettings( {
 	const { settings, update, clear, isSaving } =
 		useDeveloperFeatureSettings( featureId );
 
+	const [ draftSettings, setDraftSettings ] =
+		useState< DeveloperSelection | null >( null );
+	const [ isSavingThis, setIsSavingThis ] = useState( false );
+
+	useEffect( () => {
+		if ( ! isSaving ) {
+			setIsSavingThis( false );
```
dkotter, line 61: So we already have `isSaving`, can you explain what the difference is between that and this new `isSavingThis`?

dkotter (changes requested):
Couple minor things but overall this looks good.

Ideally we get some E2E tests here though, to verify the new save button works and to also verify if you choose a provider/model and don't hit save, those values don't persist on page reload

routes/ai-home/components/DeveloperSettings.tsx
Infinite-Null: Hi @dkotter, Thank you for pointing that out. `isSaving` comes from the shared hook state, so when one experiment is saved, all experiment instances receive it as `true` and show their loading state. Here is a video demonstrating this behavior using the `isSaving` hook.

https://github.com/user-attachments/assets/29207328-b9a3-49f5-a387-094cd4ecd5a0


`isSavingThis` is local to each experiment component and tracks whether that specific instance initiated the save. This ensures only the experiment being saved shows the loading spinner.


Here is the video with our local state:

https://github.com/user-attachments/assets/e28c695e-8f19-4847-8a9c-f5ec5ee04e1b

routes/ai-home/style.scss
Infinite-Null: Thanks for pointing this out! Since `--wpds-dimension-gap-lg` is defined as `16px` in`:root`, that’s why I used `16px` as the fallback. While `.ai-developer-mode-fields__notice` is not part of this PR, I’ve updated that instance as well to use the same fallback value for consistency.

Screenshot 2026-06-24 at 5 17 53 PM

Infinite-Null:
> Couple minor things but overall this looks good.
> 
> Ideally we get some E2E tests here though, to verify the new save button works and to also verify if you choose a provider/model and don't hit save, those values don't persist on page reload

Hi @dkotter, thank you for the feedback! I’ve added the E2E tests to verify the save button behavior and that unsaved provider/model selections do not persist after a page reload. I’ve also made the necessary spacing fallback updates.

When you have a chance, could you please take another look? Thank you!

dkotter (commented):
One last thing I noticed in my testing that's worth discussing. If I choose a Provider (and/or a Model) and save, if I come back later and click the `Reset to default` link, I then have to save again for that to persist.

While I know that matches the initial behavior, it almost feels like a user taking the action of clicking `Reset` should be enough to trigger a save. Any thoughts on that?

Infinite-Null:
> One last thing I noticed in my testing that's worth discussing. If I choose a Provider (and/or a Model) and save, if I come back later and click the `Reset to default` link, I then have to save again for that to persist.
> 
> While I know that matches the initial behavior, it almost feels like a user taking the action of clicking `Reset` should be enough to trigger a save. Any thoughts on that?

Hi @dkotter, I appreciate you raising this. I had the same thought initially. One consideration I had, though, is that if someone clicks “Reset to default” by mistake, automatically saving it could be unexpected and would immediately overwrite their previously saved provider/model selection.

Keeping it as a pending change that requires Save also stays consistent with the rest of the form behavior, where changes are only persisted after explicitly saving. That said, I’m happy to adjust it if we feel Reset should be treated as an immediate action and reverted to the original behavior.

dkotter:
> One consideration I had, though, is that if someone clicks “Reset to default” by mistake, automatically saving it could be unexpected and would immediately overwrite their previously saved provider/model selection

While that's certainly possible, I would argue it's not likely someone will accidentally click that but if they do, not hard to reselect the provider and model.

On the flip side, when I was testing, I actually didn't realize I needed to hit save again after resetting and took me a bit to figure out why a feature was using the wrong model.

Now that's just my own experience and my own opinion but I'd argue for making the change to auto-save after resetting

Infinite-Null:
> > One consideration I had, though, is that if someone clicks “Reset to default” by mistake, automatically saving it could be unexpected and would immediately overwrite their previously saved provider/model selection
> 
> While that's certainly possible, I would argue it's not likely someone will accidentally click that but if they do, not hard to reselect the provider and model.
> 
> On the flip side, when I was testing, I actually didn't realize I needed to hit save again after resetting and took me a bit to figure out why a feature was using the wrong model.
> 
> Now that's just my own experience and my own opinion but I'd argue for making the change to auto-save after resetting

That makes sense, especially given the testing experience you described. I’ve updated the behavior so that clicking “Reset to default” now immediately resets and persists the settings without requiring an additional click on Save. I have also updated the tests accordingly. 

@dkotter when you have a moment, I’d appreciate it if you could take a look.

https://github.com/user-attachments/assets/37b5763c-4397-4619-af55-56d14e7ec176

dkotter (approved):

PR #764, opened 2026-06-23
Add a core/manage-settings ability

jorgefilipecosta:
## What

Adds a write-oriented `core/manage-settings` ability that updates WordPress settings, the write counterpart to the read-only `core/settings` ability added in #691.

Core's `WP_Settings_Abilities` already reserves a slot for this (`register_manage_settings()`) but does not yet implement it. The plugin implements it in the same `Settings` class, reusing the exposed-settings snapshot and the shared helpers (`get_exposed_settings()`, `value_schema()`, `cast_value()`).

- **Input**: a map of setting name → new value (at least one). The input/output schemas reuse each setting's own value schema, with `additionalProperties: false`.
- **Output**: a map of each updated setting name → its stored value.
- **Permission**: `manage_options` (shared with `core/settings`).
- **Annotations**: `readonly: false`, `destructive: false`, `idempotent: true`.

## Atomic, all-or-nothing

The Abilities API validates the input against the registered input schema before `execute_manage_settings()` runs, so an invalid or unknown value aborts the whole call before any `update_option()` fires — matching `WP_REST_Settings_Controller::update_item()`, which rejects the entire request via `has_valid_params()` when any value is invalid. Accepted values are sanitized against their schema, stored, then read back and cast for the response.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-764-68d5383052b097c39e5ea8390aa82d7fd648011d.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Settings/Settings.php
+++ b/includes/Abilities/Settings/Settings.php
@@ -17,10 +17,13 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Class - Settings
  *
- * Registers the read-only `core/settings` ability, which returns WordPress settings as a
- * flat map of setting name to value. Only settings flagged with `show_in_abilities` are
- * exposed. It is structured to also back a future write-oriented `core/manage-settings`
- * ability via the shared helpers (get_exposed_settings(), value_schema(), cast_value()).
+ * Registers the read-only `core/settings` ability and the write-oriented `core/manage-settings`
+ * ability. Both operate on the same settings — those flagged with `show_in_abilities` — as a flat
+ * map of setting name to value: `core/settings` reads them and `core/manage-settings` updates them,
+ * sharing the helpers get_exposed_settings(), value_schema() and cast_value().
+ *
+ * Plugin: WordPress core's WP_Settings_Abilities currently reserves `core/manage-settings` but does
+ * not implement it; the plugin ships the write ability ahead of core.
  *
  * This class is kept almost identical to the WordPress core class `WP_Settings_Abilities`
  * so the two implementations stay in sync. Differences from the core class are marked with
@@ -74,13 +77,7 @@ final class Settings {
 	 */
 	public function register(): void {
 		$this->register_get_settings();
-
-		/*
-		 * A future write-oriented ability can be registered here, reusing the shared
-		 * helpers below (get_exposed_settings(), value_schema(), cast_value()):
-		 *
-		 *     $this->register_manage_settings();
-		 */
+		$this->register_manage_settings();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/Abilities/Settings/Settings.php
```diff
@@ -136,6 +130,62 @@ private function register_get_settings(): void {
...
+	 *
+	 * @since x.x.x
+	 */
+	private function register_manage_settings(): void {
+		// Plugin: unregister any core-provided copy first so the plugin's version wins.
+		if ( wp_has_ability( 'core/manage-settings' ) ) {
+			wp_unregister_ability( 'core/manage-settings' );
+		}
+
+		$settings   = (array) $this->exposed_settings;
```
gziolo, line 150: Nit, if there are no exposed settings for some reason, then maybe this ability should never get registered.

includes/Abilities/Settings/Settings.php
```diff
@@ -74,13 +74,7 @@ public function init(): void {
 	 */
 	public function register(): void {
 		$this->register_get_settings();
```
gziolo, line 76: Maybe https://github.com/WordPress/ai/blob/68d5383052b097c39e5ea8390aa82d7fd648011d/includes/Abilities/Settings/Settings.php#L91-L92

should run first in `register()` and bail out if there are no settings exposed.

includes/Abilities/Settings/Settings.php
```diff
@@ -136,6 +130,62 @@ private function register_get_settings(): void {
...
+					'type'                 => 'object',
+					'description'          => __( 'A map of setting name to the new value to store. At least one setting is required.', 'ai' ),
+					'properties'           => $properties,
+					'minProperties'        => 1,
+					'additionalProperties' => false,
+				),
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'description'          => __( 'A map of each updated setting name to its new value.', 'ai' ),
+					'properties'           => $properties,
```
gziolo, line 172: Nit, for symmetry, it could use the same schema:
```suggestion
					'properties'           => $properties,
					'minProperties'        => 1,
```

includes/Abilities/Settings/Settings.php
```diff
@@ -175,6 +225,53 @@ public function execute_get_settings( $input = array() ): array {
...
+			if ( ! is_string( $exposed_name ) || ! isset( $settings[ $exposed_name ] ) ) {
+				// `additionalProperties: false` already rejects unknown keys upstream; guard defensively.
+				continue;
+			}
+
+			$setting = $settings[ $exposed_name ];
+
+			// Sanitize against the declared schema before storing; update_option() additionally
+			// runs the setting's own registered sanitize_callback.
+			$value = rest_sanitize_value_from_schema( $value, $setting['schema'], $exposed_name );
```
gziolo, line 262: I'm curious why sanitization is necessary here. Is it applied automatically for the REST API endpoint? I don't see explicit calls in the update method:

https://github.com/WordPress/wordpress-develop/blob/a2d90ce260d5d20c4a5a0ab56b9f1c06053af060/src/wp-includes/rest-api/endpoints/class-wp-rest-settings-controller.php#L145-L208

There is also a special case where passing `null` effectively removes the option from the database so it uses the default value provided.

jeffpaul:
I'd like to see #863 land alongside this specific PR (and any others that aren't explicitly read abilities), to help ensure folks enabling these sorts of abilities are doing so knowingly.

PR #770, opened 2026-06-24
Prompt template extension points

### LINKED ISSUE #192: Add extension points for custom prompt templates
Introduce extension points that allow developers to define, override, or extend prompt templates used by AI experiments. This enables customization without forking or rewriting experiments.

Acceptance criteria:

* Prompt templates can be extended or overridden via hooks
* Changes are scoped and predictable
* Works across multiple experiments
* Behavior is documented for developers

the-hercules:
# Add prompt template extension points

## Summary

Introduces consistent, scoped extension points so developers can customize the prompts
and model configuration of AI abilities **without forking** the plugin. Adds three
per-ability filter hooks plus a developer guide.

Closes #192 

## What changed

A single slug helper on `Abstract_Ability` derives a hook-safe slug from each ability
name (`ai/title-generation` → `title_generation`), and three filters are wired through
every AI ability:

- **`wpai_{slug}_system_instruction`** — override/extend the system instruction for one
  ability. Added once in `Abstract_Ability::get_system_instruction()`, running **after**
  the existing global `wpai_system_instruction` filter.
- **`wpai_{slug}_prompt`** — modify the assembled user prompt before it is sent. Added to
  the 10 abilities that did not already expose one.
- **`wpai_{slug}_prompt_builder`** — adjust temperature, model/provider preference, request
  options, etc. Added to all 12 prompt-building abilities, **after** the default model
  preference is applied and **before** generation support is verified.

Abilities covered: Title, Excerpt, Summarization, Content Resizing, Meta Description,
Editorial Updates, Editorial Notes, Alt Text, Image Prompt Generation, Image Generation,
Comment Analysis, Content Classification.

Where abilities carry extra context, it is passed as additional filter arguments
(e.g. `$length` for summarization, `$block_type` for editorial notes, `$action` for content
resizing).

## Backward compatibility

- The global `wpai_system_instruction` filter is unchanged; the new scoped filter runs after it.
- `meta_description` and `content_classification` already shipped a `wpai_{slug}_prompt`
  filter with richer signatures (`$content`/`$title`, taxonomy args). Those are **left exactly
  as-is** — no duplicate hook is registered — so existing integrations keep working. They are
  documented alongside the new hooks.
- `wpai_meta_description_result_temperature` still applies; the new meta-description builder
  filter runs after it.
- Every new hook is additive and a no-op unless a callback is attached, so default behavior is
  byte-for-byte unchanged.

## Design notes / decisions

A few decisions were made in the absence of issue feedback (happy to adjust):

1. **Scoped system-instruction hook included.** The issue asks for per-ability system
   instruction control; the global filter alone would require name-branching.
2. **Image Generation included** for symmetry. Its `_prompt` is the image prompt (not post
   content), and editorial guidelines are appended after the filter — documented as such.
3. **Builder filter fires after `set_provider_model_preference()`** so it can intentionally
   override the developer-mode model choice without breaking the default path.
4. Structured-output abilities (`comment_analysis`, `content_classification`,
   `editorial_notes`) document that the builder should be extended, not replaced, to preserve
   JSON response schemas.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-770-ca265c16011b6642e74ba2b7c89c5614376667ea.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter (commented):
While the changes in this PR work, I think it may be worth discussing or exploring an alternative approach. Right now each individual Ability is responsible for ensuring content is passed through a filter. So as new Abilities are added, we'll want to ensure they get these same filters.

I'm wondering if it would be better / easier to maintain if these filters were automatically added. As an example, if we introduce a `get_prompt_builder` method on the `Abstract_Ability` class, the filter could live there and get automatically applied when each individual Ability uses that method. That would reduce the amount of code changes in this PR and would ensure any new Abilities in the future automatically get those filters

the-hercules:
Hi @dkotter 
Thanks for the review! You make a great point about maintainability and relying on developers to remember to manually add `apply_filters` into every new ability is definitely a recipe for them getting missed over time.

I looked into moving `get_prompt_builder()` entirely into `Abstract_Ability` so it's handled 100% automatically as you suggested. The tricky part is that different abilities configure the `Prompt_Builder` in very specific ways *before* the filter is meant to run. For example:
- `Summarization` sets `->using_temperature( 0.9 )`
- `Content_Classification` sets `->as_json_response( ... )`
- `Generate_Image` sets `RequestOptions` with a 90s timeout and attaches reference images.

If `Abstract_Ability::get_prompt_builder()` creates the builder and immediately fires the filter hook, individual abilities would have to append their configurations *afterwards*. That would defeat the purpose of the hooks, because third-party developers wouldn't be able to override things like temperature if the ability hardcodes it after the filter runs.

**Proposed Compromise**
Instead of a single `get_prompt_builder()` that builds from scratch, what if we centralize the filter logic into a helper method in `Abstract_Ability`? 

We could introduce a `filter_prompt_builder()` method on the base class:

```php
protected function filter_prompt_builder( \WP_AI_Client_Prompt_Builder $builder, string $feature_class, array $fallback_models = array(), ...$filter_args ) {
    // 1. Apply model preferences
    $builder = $this->set_provider_model_preference( $builder, $feature_class, $fallback_models );
    
    // 2. Automatically apply the builder filter
    return apply_filters( "wpai_{$this->get_ability_slug()}_prompt_builder", $builder, ...$filter_args );
}
```

Then, an individual ability simply configures its specific settings and passes the builder through the helper:
```php
$prompt_builder = wp_ai_client_prompt( $prompt )
    ->using_system_instruction( $this->get_system_instruction() )
    ->using_temperature( 0.7 );

// One standardized call that applies the model fallbacks AND fires the filter hook
$prompt_builder = $this->filter_prompt_builder( $prompt_builder, Title_Generation_Experiment::class, array(), $prompt );
```

We could do something similar for the prompt string itself with a `filter_prompt()` helper. 

This drastically reduces the boilerplate across the abilities and centralizes the hook logic. While new abilities will still have to call `$this->filter_prompt_builder()`, they already have to call `$this->set_provider_model_preference()` anyway, so combining them makes it very hard to forget the filters.

Let me know if this approach sounds good to you, and I'll update the PR!

dkotter:
That's a decently thought out approach and I'm not opposed to it. One other solution to consider after looking at things again is to add this filter in the existing `ensure_*_supported` methods, as I believe every Ability calls that and it's expected any new Ability added in the future will call that. So this wouldn't require a new method to be called and if we run the filter first and then run the supported check, we also ensure any changes someone makes to the prompt builder are supported by a configured model.

So I'd suggest still adding a new `filter_prompt_builder` method that runs the filter and then use that in the supported methods:

```php
private function filter_prompt_builder( $prompt_builder ) {
      $ability_slug = $this->get_ability_slug();

      $filtered = apply_filters( "wpai_prompt_builder_{$ability_slug}", $prompt_builder, $this );

      // Guard against a filter returning a non-builder (null, WP_Error, etc.).
      if ( ! $filtered instanceof \WP_AI_Client_Prompt_Builder ) {
              return $prompt_builder;
      }

      return $filtered;
}

protected function ensure_text_generation_supported( $prompt_builder, string $message ) {
      $prompt_builder = $this->filter_prompt_builder( $prompt_builder );

      if ( ! $prompt_builder->is_supported_for_text_generation() ) {
              return new WP_Error( 'unsupported_model', $message );
      }

      return $prompt_builder;
}
```

the-hercules:
@dkotter 
That’s a good idea to reduce boilerplate, the process of trying to handle the filter automatically inside `ensure_*_supported()`. 

However, I'd lean toward keeping a dedicated `filter_prompt_builder()` helper instead, for two main reasons:

**1. Loss of Explicit Context Variables**
Right now, some abilities pass highly useful, explicit variables to the filter. For example, `Summarization` passes the `$length` argument alongside the prompt:
`apply_filters( "wpai_..._prompt_builder", $builder, $prompt, $length );`

This allows a third-party developer to easily check if the requested summary is 'short' or 'long' and adjust the builder accordingly. If we bury the filter inside `ensure_text_generation_supported( $builder, $message )`, the hook loses access to those specific variables. 

While we could pass `$this`, the Ability classes are stateless during a request (they don't store `$length` as a class property). This means a developer inspecting `$this` still cannot access the `$length` for the current execution. The only way to fix this would be to update the signature of `ensure_text_generation_supported()` to accept `...$filter_args`—but if we do that, we have to manually update all 12 ability files to pass those arguments down anyway, which defeats the boilerplate-reduction goal of the suggestion!

**2. Separation of Concerns**
A method named `ensure_*_supported` strongly signals that it's strictly a validation check. Firing a powerful, mutating developer hook inside a validation method is a side-effect that might surprise future contributors. Keeping the hook in a dedicated `filter_prompt_builder()` helper makes the intent of the code much clearer.

Since new abilities will have to call `$this->set_provider_model_preference()` anyway, wrapping that inside a dedicated `$this->filter_prompt_builder()` still centralizes the logic and makes it very hard to forget.

Let me know what you think.

dkotter:
@the-hercules This all sounds like valid arguments to me so happy to review and test things once changes are up. Thanks!

the-hercules:
@dkotter I’ve committed the changes you requested. Could you please review them when you get a chance?

dkotter (approved):

PR #772, opened 2026-06-24
Restore classification suggestion pill to original position on API failure

### LINKED ISSUE #771: Suggested Tags/Categories when selected are removed from view irrespective of end result being a failure. (Content Classification exp)
### Description

The Content Classification Experiment feature provides AI suggestions for tags/categories to add to a post.

**Expected**: When a tag/category is selected (by clicking on it), if it fails to be added, it should be visible again for re-attempting to add it.

**What currently happens**: When a tag/category is selected, the frontend will always remove it first and will not add it back to the list if an error occurs in the api call resolving it.

**Why this is a bug**: It leads to a missing AI suggestion which could have still been utilised by re-attempting/manually entering it. It causes poor User Experience.

Ensure the AI features are enabled and Content Classification experiment is enabled.

1. Open the post editor.
2. Make a new post with atleast 150 words of content.
2. Open the Post sidebar.
3. Click **Suggest Tags** (or Categories) to get suggestions. Wait for suggestions to appear.
4. Turn off your internet connection or perform a simulation of it via developer console.
5. Click one of the suggested tag pills.
6. **Result**: The pill disappears immediately. When the network request fails, the pill never returns to the list and is lost forever.

Other possible reproduction steps can be: an unresponsive backend, a nonce expiry, an error in the backend which gets shown but will thus also lead to pill disappearance.

https://github.com/user-attachments/assets/0a12f391-22b6-46cb-98db-5e5c733e29e2

### Environment info

- WordPress v7.0 , AI Plugin Version 1.0.2 - dev branch , Twenty Twenty-Five Theme
- Opera GX browser
- using Macbook with MacOS Tahoe v26

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

Intenzi:
## What?

Closes #771 

This PR visually adds back the AI suggested/generated tags/categories pills which fail to be added to the post upon performing the click action that visually removes them. (Content Classification Experiment)

## Why?

It is expected that when a tag/category is selected (by clicking on it), if it fails to be added, it should be visible again for re-attempting to add it.
It resolves the visual issue and provides better User Experience.

## How?

I have implemented the PR in the following manner:

- Modified `addTermToPost` method to return a success boolean flag upon completion.
- This is then utilised inside `handleAccept` callback which calls `addTermToPost`
- If the API call was not successful i.e. `addTermToPost` returned `false` for success, we restore the suggested tag/category back to the suggestions state list via `setSuggestions()`. It also takes into account the index of the suggestion to try restore at the same position from which the pill was removed.

I have then taken account of an additional edge case:
+Lengthy API calls can lead to the user performing further actions such as `Suggest again` and `Dismiss all`, in which case we should not be restoring the pill after API failure is received.

- This has been implemented by storing a `generationCounts` variable that stores an integer value for the current "session" of pills displayed. 
- Session here implies the state after clicking on "Suggest again" or "Dismiss all" or the initial "Suggest Categories/Tags" button.
- I track this state as an integer value which gets incremented upon the above mentioned actions. 
- The old session integer is captured in a closure by the `.then()` callback as `clickGenerationCount` variable.
- Once the `addTermToPost` method fails it compares the present `generationCounts` against the old `clickGenerationCount` to determine whether we have a stale state for adding back the suggestion pill.

- I have refactored suggestion re-insertion logic into a separate method of `insertSuggestionAt()` to reduce nested code.

+I have implemented e2e tests that validate the resolution of the same.
- The tests intercept the `/wp/v2/tags/` API calls via `page.route()` to return a simulated failure.
- The tests utilise manually deferred API responses to simulate/enact a slow network request that only resolves to failure when the test explicitly triggers it (after the user clicks on Suggest again/Dismiss all for the edge case for example).

### Use of AI Tools

AI assistance: Yes
Tool(s): Antigravity
Model(s): Gemini 3.5 Flash (Low)
Used for: Identifying the relevant possible fixes. Implemented the E2E tests using the same. Final implementation decision and tests were reviewed and edited by me.

## Testing Instructions

1. Open the post editor.
2. Make a new post with atleast 150 words of content.
2. Open the Post sidebar.
3. Click **Suggest Tags** (or Categories) to get suggestions. Wait for suggestions to appear.
4. Turn off your internet connection or perform a simulation of it via developer console.
5. Click one of the suggested tag pills.
6. **Result**: The pill reappears. When the network request fails, the pill returns to the list and is re-useable.

## Screenshots or screencast

https://github.com/user-attachments/assets/6ec6c7cd-ab1f-425a-ae20-57d454463335

## Changelog Entry

> Fixed - Restored tag and category suggestion pills to their original positions if the backend term assignment API fails, resolving stale state race conditions.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-772-e45cb3f26f94e1f9f8356f356f37b1e0160e8426.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


Intenzi:
I haven't implemented for one edge case, I'd like to discuss on clarification around its approach decisions and whether we want to address it at all:

Edge case: Restoring pills when the API call fails *after* the component has unmounted.

If a user clicks to accept a suggestion (starting the asynchronous `addTermToPost` call) and immediately switches sidebar tabs (unmounting the classification panel) -- any subsequent API failure will occur when the component is no longer in the DOM. Since the component is unmounted: calling the React state setter `setSuggestions()` is discarded and the restored pill is lost forever (it is never synced back to the module-level `suggestionsCache` that was implemented by me in #769 ).

#### Potential Solution & React Anti-Patterns
To solve this, my thought process was to simply track the mount status using a ref:
1. Initialize `const isMounted = useRef(true)` and set it to `false` in a `useEffect` cleanup.
2. In the `.then()` handler, if `isMounted.current` is `false`, update the module-level `suggestionsCache[taxonomy]` instead of calling `setSuggestions()` while keeping the setSuggestion modification as-is for if the component isMounted.

However, using `isMounted` is widely considered an anti-pattern in modern React. It suggests that the component is trying to manage side-effects and global state synchronization that belong outside of its own lifecycle.

#### Questions for Discussion
To resolve this cleanly, we might need a more holistic reapproach to how API calls and state are managed, or is the module-level cache enough? Is it acceptable to leave it as-is for now, or is decoupling the state management the right next step for this plugin's architecture?

I am thinking that we could use WordPress Data Store to register a redux store that manages the state of the suggestions per taxonomy. I'm hesitant to however introduce that level of architectural overhead for an edge case fix.

Looking forward to your thoughts on the same.

dkotter:
> To resolve this cleanly, we might need a more holistic reapproach to how API calls and state are managed, or is the module-level cache enough? Is it acceptable to leave it as-is for now, or is decoupling the state management the right next step for this plugin's architecture?
> 
> I am thinking that we could use WordPress Data Store to register a redux store that manages the state of the suggestions per taxonomy. I'm hesitant to however introduce that level of architectural overhead for an edge case fix.

I would say I'm totally fine leaving this as-is for now. Seems we may be trying to fix something that won't be encountered in real user interactions. If we get reports of people running into this and wanting a fix, we can then look to solve this.

Intenzi:
I agree, keeping it simple and leaving it as-is for now makes the most sense. PR is ready for review.

dkotter (approved):

PR #774, opened 2026-06-25
New Ability: core/read-users

jorgefilipecosta:
## What?
- Adds a read-only `core/read-users` ability with single-user lookup by `id`, `user_email`, `user_login`, or `user_nicename`.
- Single-user lookup modes return the user object directly; collection mode returns `users`, `total`, and `total_pages`.
- Adds collection mode with optional `roles`, `has_published_posts`, `include`, `page`, `per_page`, and `fields`; omitted REST input defaults to empty collection mode.
- Uses lean default fields when `fields` is omitted, while keeping sensitive identifiers and fields opt-in and permission-gated.
- Uses WP_User-shaped field names, including `user_login`, `user_email`, `user_nicename`, `user_url`, and `user_registered`, to align with Core’s existing `core/get-user-info` ability.
- Gates sensitive identifiers and fields against Core REST users behavior, omits inaccessible fields per user, excludes capabilities/meta, and respects `get_option( 'show_avatars' )`.
- Adds PHPUnit and browser-level Abilities API coverage.

## Testing
- `php -l includes/Abilities/Users/Users.php`
- `php -l tests/Integration/Includes/Abilities/Users/UsersTest.php`
- `vendor/bin/phpcs --standard=phpcs.xml.dist`
- `vendor/bin/phpstan analyse --memory-limit=1G`
- `npx wp-scripts lint-js tests/e2e/specs/abilities/core-read-users.spec.js`
- `npm run lint:js`
- `npm run typecheck`
- `npm run build`
- `npm run test:php -- --filter UsersTest`
- `npm run test:php`
- `npm run test:e2e -- tests/e2e/specs/abilities/core-read-users.spec.js --project=chromium`

## Changelog entry

Added - New core/read-users Ability; retrieves a single user by ID, email, login, or nicename, or a filtered and paginated users collection, with sensitive fields opt-in and permission-gated.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-774-4bb161ebb7638f82f354c4bd18e6f80e1623bbee.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Users/Users.php
+++ b/includes/Abilities/Users/Users.php
@@ -750,6 +750,7 @@ final class Users {
 			),
 			'user_registered' => array(
 				'type'        => 'string',
+				'format'      => 'date-time',
 				'description' => __( 'Registration date for the user. Present when the current user can view it.', 'ai' ),
 			),
 			'roles'           => array(
@@ -861,7 +862,7 @@ final class Users {
 				$data['locale'] = (string) get_user_locale( $user );
 			}
 			if ( $fields_requested( 'user_registered' ) ) {
-				$data['user_registered'] = (string) $user->user_registered;
+				$data['user_registered'] = gmdate( 'c', strtotime( $user->user_registered ) );
 			}
 		}
 
--- a/includes/Abilities/Users/Users.php
+++ b/includes/Abilities/Users/Users.php
@@ -1,0 +1,886 @@
+<?php
+/**
+ * The `core/users` WordPress Ability.
+ *
+ * @package WordPress\AI
+ *
+ * @since x.x.x
+ */
+
+declare( strict_types=1 );
+
+namespace WordPress\AI\Abilities\Users;
+
+use WP_Error;
+use WP_User;
+use WP_User_Query;
+
+// Exit if accessed directly.
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class - Users
+ *
+ * Registers the read-only `core/users` ability, which retrieves one or more
+ * readable WordPress users. Supports fetching a single readable user by ID,
+ * email, username, or slug, or querying a paginated collection optionally
+ * filtered by roles or published-post authorship. Field-level access is enforced
+ * per user by omitting fields the current user cannot view.
+ *
+ * This class is kept almost identical to the WordPress core class `WP_Users_Abilities`
+ * so the two implementations stay in sync. Differences from the core class are marked with
+ * `// Plugin:` comments. Additionally, all user-facing strings use the 'ai' text domain.
+ *
+ * Plugin: the class is final and instance-based (with private helpers), matching the
+ * plugin's other ability classes (e.g. `Settings`) and core's `WP_Settings_Abilities`.
+ *
+ * @internal This class should not be used outside the plugin and there is no guarantee of backwards compatibility.
+ *
+ * @since x.x.x
+ */
+final class Users {
+
+	/**
+	 * The ability category used for user abilities.
+	 *
+	 * @since x.x.x
+	 * @var string
+	 */
+	private const CATEGORY = 'user';
+
+	/**
+	 * Default number of users returned per page in collection mode.
+	 *
+	 * @since x.x.x
+	 * @var int
+	 */
+	private const DEFAULT_PER_PAGE = 10;
+
+	/**
+	 * Maximum number of users returned per page in collection mode.
+	 *
+	 * @since x.x.x
+	 * @var int
+	 */
+	private const MAX_PER_PAGE = 100;
+
+	/**
+	 * Public/read-context user fields.
+	 *
+	 * @since x.x.x
+	 * @var string[]
+	 */
+	private array $read_fields = array(
+		'id',
+		'display_name',
+		'description',
+		'url',
+		'link',
+		'slug',
+	);
+
+	/**
+	 * Fields that expose edit-context user data.
+	 *
+	 * @since x.x.x
+	 * @var string[]
+	 */
+	private array $sensitive_fields = array(
+		'username',
+		'email',
+		'first_name',
+		'last_name',
+		'nickname',
+		'locale',
+		'registered_date',
+	);
+
+	/**
+	 * Hooks the ability into the Abilities API.
+	 *
+	 * Plugin: this method has no equivalent in the core class. In core, register() is
+	 * invoked directly from wp_register_core_abilities() (already on the
+	 * `wp_abilities_api_init` hook). The plugin instead hooks register() slightly later
+	 * (priority 11) so it can override any core-provided copy, and registers the category
+	 * as a fallback in case core has not.
+	 *
+	 * @since x.x.x
+	 */
+	public function init(): void {
+		add_action( 'wp_abilities_api_categories_init', array( $this, 'register_category' ), 11 );
+		add_action( 'wp_abilities_api_init', array( $this, 'register' ), 11 );
+	}
+
+	/**
+	 * Registers the `user` ability category if it is not already registered.
+	 *
+	 * Plugin: this method has no equivalent in the core class; core relies on
+	 * wp_register_core_ability_categories() to register the `user` category.
+	 *
+	 * @since x.x.x
+	 */
+	public function register_category(): void {
+		if ( wp_has_ability_category( self::CATEGORY ) ) {
+			return;
+		}
+
+		wp_register_ability_category(
+			self::CATEGORY,
+			array(
+				'label'       => __( 'Users', 'ai' ),
+				'description' => __( 'Abilities that retrieve or manage WordPress users.', 'ai' ),
+			)
+		);
+	}
+
+	/**
+	 * Registers all user abilities.
+	 *
+	 * Must run on the `wp_abilities_api_init` hook.
+	 *
+	 * @since x.x.x
+	 */
+	public function register(): void {
+		$this->register_get_users();
+	}
+
+	/**
+	 * Registers the read-only `core/users` ability.
+	 *
+	 * @since x.x.x
+	 */
+	private function register_get_users(): void {
+		// Plugin: unregister any core-provided copy first so the plugin's version wins.
+		if ( wp_has_ability( 'core/users' ) ) {
+			wp_unregister_ability( 'core/users' );
+		}
+
+		wp_register_ability(
+			'core/users',
+			array(
+				'label'               => __( 'Get Users', 'ai' ),
+				'description'         => __( 'Retrieves one or more readable WordPress users. Fetch a single readable user by ID, email, username, or slug, or query a paginated collection optionally filtered by roles or published-post authorship.', 'ai' ),
+				'category'            => self::CATEGORY,
+				'input_schema'        => $this->get_users_input_schema(),
+				'output_schema'       => $this->get_users_output_schema(),
+				'execute_callback'    => array( $this, 'execute_get_users' ),
+				'permission_callback' => array( $this, 'check_permission' ),
+				'meta'                => array(
+					'annotations'  => array(
+						'readonly'    => true,
+						'destructive' => false,
+						'idempotent'  => true,
+					),
+					'show_in_rest' => true,
+					'pagination'   => true,
+				),
+			)
+		);
+	}
+
+	/**
+	 * Permission callback for the `core/users` ability.
+	 *
+	 * Implements defense in depth: this gate decides whether the request may proceed at
+	 * all, while the per-user read checks in {@see self::execute_get_users()} are the
+	 * authoritative, row-level enforcement.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param mixed $input Optional. The ability input. Default empty array.
+	 * @return bool True if the request may proceed, false otherwise.
+	 */
+	public function check_permission( $input = array() ): bool {
+		$input = is_array( $input ) ? $input : array();
+
+		if ( ! is_user_logged_in() ) {
+			return false;
+		}
+
+		if ( ! empty( $input['roles'] ) && ! current_user_can( 'list_users' ) ) {
+			return false;
+		}
+
+		$lookup_type = $this->get_lookup_type( $input );
+		if ( '' === $lookup_type ) {
+			return true;
+		}
+
+		$user = $this->find_user( $input );
+		if ( ! $user instanceof WP_User || ! $this->is_user_member_of_site( $user ) ) {
+			return false;
+		}
+
+		return $this->can_read_user_for_lookup( $user, $lookup_type );
+	}
+
+	/**
+	 * Executes the `core/users` ability.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param mixed $input Optional. The ability input. Default empty array.
+	 * @return array<string, mixed>|\WP_Error A map with a `users` list, or a WP_Error on failure.
+	 */
+	public function execute_get_users( $input = array() ) {
+		$input  = is_array( $input ) ? $input : array();
+		$fields = $this->normalize_fields( $input );
+
+		$lookup_type = $this->get_lookup_type( $input );
+		if ( '' !== $lookup_type ) {
+			$user = $this->find_user( $input );
+			if ( ! $user instanceof WP_User
+				|| ! $this->is_user_member_of_site( $user )
+				|| ! $this->can_read_user_for_lookup( $user, $lookup_type )
+			) {
+				return $this->not_found_error();
+			}
+
+			return array(
+				'users'       => array( $this->format_user( $user, $fields ) ),
+				'total'       => 1,
+				'total_pages' => 1,
+			);
+		}
+
+		$per_page = $this->normalize_per_page( $input );
+		$page     = isset( $input['page'] ) ? max( 1, $this->input_int( $input['page'] ) ) : 1;
+
+		$query_args = array(
+			'number'      => $per_page,
+			'offset'      => ( $page - 1 ) * $per_page,
+			'count_total' => true,
+		);
+
+		if ( ! empty( $input['roles'] ) && current_user_can( 'list_users' ) ) {
+			$query_args['role__in'] = $this->normalize_string_list( $input['roles'] );
+		}
+
+		if ( current_user_can( 'list_users' ) ) {
+			$has_published_posts = $this->normalize_has_published_posts( $input );
+			if ( null !== $has_published_posts ) {
+				$query_args['has_published_posts'] = $has_published_posts;
+			}
+		} else {
+			$query_args['has_published_posts'] = $this->get_public_author_post_types();
+		}
+
+		$query = new WP_User_Query( $query_args );
+
+		$users = array();
+		foreach ( $query->get_results() as $user ) {
+			if ( ! $user instanceof WP_User || ! $this->is_user_member_of_site( $user ) || ! $this->can_read_user( $user ) ) {
+				continue;
+			}
+
+			$users[] = $this->format_user( $user, $fields );
+		}
+
+		$total_users = (int) $query->get_total();
+
+		return array(
+			'users'       => $users,
+			'total'       => $total_users,
+			'total_pages' => $per_page > 0 ? (int) ceil( $total_users / $per_page ) : 0,
+		);
+	}
+
+	/**
+	 * Casts a raw input value to a non-negative integer.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param mixed $value The raw input value.
+	 * @return int The value as a non-negative integer, or 0 when not scalar.
+	 */
+	private function input_int( $value ): int {
+		return is_scalar( $value ) ? absint( $value ) : 0;
+	}
+
+	/**
+	 * Determines the single-user lookup type represented by the input.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array<mixed> $input The ability input.
+	 * @return string The lookup type, or an empty string for collection mode.
+	 */
+	private function get_lookup_type( array $input ): string {
+		foreach ( array( 'id', 'email', 'username', 'slug' ) as $key ) {
+			if ( array_key_exists( $key, $input ) ) {
+				return $key;
+			}
+		}
+
+		return '';
+	}
+
+	/**
+	 * Finds a user by one of the supported unique input identifiers.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array<mixed> $input The ability input.
+	 * @return \WP_User|null User object, or null when not found.
+	 */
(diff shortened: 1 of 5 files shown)


includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+	 * Returns the supported field list in output order.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string[] Supported field names.
+	 */
+	private function get_fields(): array {
+		$include_avatars = (bool) get_option( 'show_avatars' );
+
+		if ( null !== $this->fields && $include_avatars === $this->fields_include_avatars ) {
```
gziolo: My understanding is that fields get cached here, to avoid repeating the same computation on every access. It isn't clear to me why `get_option( 'show_avatars' )` is called each time, and in addition, getting saved in `$this->fields_include_avatars`. It looks like it would be perfectly fine to avoid creating the property that remembers the value of the `show_avatars` option, and run this check inline once when the list of fields get cached. Then reading cache would just work fine.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+	 * Returns the default field list in output order.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string[] Default field names.
+	 */
+	private function get_default_fields(): array {
+		$fields = $this->default_fields;
+
+		if ( get_option( 'show_avatars' ) ) {
```
gziolo: Instead of calling the option, it'd be possible to remove `show_avatars` from the list if it isn't listed there. In practice, a diff between default_fields and fields would also work in case other fields are removed for some reason.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+				'type'        => 'array',
+				'description' => __( 'Roles assigned to the user. Present when the current user can view them.', 'ai' ),
+				'items'       => array(
+					'type' => 'string',
+					'enum' => $this->get_role_names(),
+				),
+			),
+		);
+
+		if ( get_option( 'show_avatars' ) ) {
```
gziolo: Well, maybe it wouldn't be such a bad idea to access this option once, and cache as local property 😄

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+	}
+
+	/**
+	 * Returns the supported field list in output order.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string[] Supported field names.
+	 */
+	private function get_fields(): array {
```
gziolo: So this returns the list of field names, not the field definitions, that is included in the `get_users_output_schema()`.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+	 * request any subset, and restricted fields are omitted when unavailable.
+	 * Single-user mode returns the user object directly, while collection mode returns
+	 * a paginated wrapper.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array<string, mixed> The output JSON Schema.
+	 */
+	private function get_users_output_schema(): array {
+		$user_properties = array(
```
gziolo: There could be a better connection between `get_fields()` and `$user_properties` defined here. It could help to call `get_option( 'show_avatars' )` only in one method.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+	 * @return string The lookup type, or an empty string for collection mode.
+	 */
+	private function get_lookup_type( array $input ): string {
+		foreach ( array( 'id', 'user_email', 'user_login', 'user_nicename' ) as $key ) {
+			if ( array_key_exists( $key, $input ) ) {
+				return $key;
+			}
+		}
+
+		return '';
```
gziolo: Should the `collection` type be explicitly named? The code is difficult to follow otherwise.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+		if ( ! is_user_logged_in() ) {
+			return false;
+		}
+
+		if ( ! empty( $input['roles'] ) && ! current_user_can( 'list_users' ) ) {
+			return false;
+		}
+
+		$lookup_type = $this->get_lookup_type( $input );
+		if ( '' === $lookup_type ) {
```
gziolo: What is the expected behavior when fields are requested that the user can't access? Should we block the entire request here as well?

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+			$data['link'] = (string) get_author_posts_url( $user_id, $user->user_nicename );
+		}
+		if ( $fields_requested( 'user_nicename' ) ) {
+			$data['user_nicename'] = (string) $user->user_nicename;
+		}
+		if ( $fields_requested( 'avatar_urls' ) && get_option( 'show_avatars' ) ) {
+			$data['avatar_urls'] = rest_get_avatar_urls( $user );
+		}
+
+		if ( $can_view_sensitive ) {
```
gziolo, line 1018: To my earlier point, we might decide that this check is not needed here if we deny access when the user tries to access fields they have no access to. In that case, `$field_requested` would contain only fields the user can see. The consequences are that this ensures that the fields requested get included in the response.

gziolo:
@jorgefilipecosta, great work so far. I left my initial feedback to evaluate. The logic is quite complex, or rather sensitive, as it has several security concerns that we need to ensure are addressed correctly from the start.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+
+			if ( array() === $public_post_types ) {
+				return array(
+					'users'       => array(),
+					'total'       => 0,
+					'total_pages' => 0,
+				);
+			}
+
+			$query_args['has_published_posts'] = $public_post_types;
```
gziolo: In collection mode, when the current user cannot list users, this line always adds `has_published_posts` to the query. That makes the database query the real filter, instead of the per-user `can_read_user()` check below.

The side effect: a logged-in user who has no published posts cannot find themselves here. For example, a subscriber who calls `{ include: [own_id] }` (or sends an empty request) gets back an empty list, even though `can_read_user()` would let them read their own account (the `is_current_user` case). The same happens for a user you are allowed to edit but who has not published anything.

This also does not match the `include` field description, which says results still respect read permissions. Single-user mode does let you read yourself, so the two modes behave differently here. Is excluding self on purpose, or should we keep self (and editable users) readable? Either way it would be good to add a test, since this case is not covered yet.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+	 * @return bool Whether the user is publicly visible as an author.
+	 */
+	private function is_public_author( WP_User $user ): bool {
+		$post_types = $this->get_public_post_types();
+		if ( array() === $post_types ) {
+			return false;
+		}
+
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.count_user_posts_count_user_posts -- Public-author checks only consider public post types.
+		return count_user_posts( (int) $user->ID, $post_types ) > 0;
```
gziolo, line 491: `count_user_posts()` runs a `COUNT` query on the posts table. In collection mode this method is reached once for every user on the page (through `can_read_user()` in the loop above), so a page of 100 users can fire around 100 extra `COUNT` queries.

For callers without `list_users`, the main query already limits results to authors with published posts (the `has_published_posts` argument). So this per-user check repeats work the query has already done. Could we skip the public-author check in collection mode and trust the query, and keep this check only for the single-user path?

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+			'required'             => array( 'users', 'total', 'total_pages' ),
+			'properties'           => array(
+				'users'       => array(
+					'type'        => 'array',
+					'description' => __( 'The readable users matching the collection request.', 'ai' ),
+					'items'       => $user_schema,
+				),
+				'total'       => array(
+					'type'        => 'integer',
+					'description' => __( 'Total number of users matching the query, across all pages, after applying the permission filter to the query. Surfaced over REST as the X-WP-Total header.', 'ai' ),
```
gziolo: Small wording fix: this says `total` is sent as the `X-WP-Total` header (and the next field mentions `X-WP-TotalPages`). The endpoint that runs an ability only returns the value inside the JSON body; it does not set these headers. Only the endpoints that *list* abilities set them. So a client that looks for `X-WP-Total` in the response will not find it here. Could we remove the header mention, or set the headers if that was the goal?

gziolo (commented):
A few more points from my read, on top of the inline notes above. They are about behavior and one small docs fix, so none of them block the PR.

gziolo:
I added three more comments discovered with the help of Claude Code review.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,1035 @@
...
+				'execute_callback'    => array( $this, 'execute_get_users' ),
+				'permission_callback' => array( $this, 'check_permission' ),
+				'meta'                => array(
+					'annotations'  => array(
+						'readonly'    => true,
+						'destructive' => false,
+						'idempotent'  => true,
+					),
+					'show_in_rest' => true,
+					'pagination'   => true,
```
gziolo: @jorgefilipecosta, what role does the `pagination` meta play?

includes/Abilities/Users/Users.php
gziolo: Decision: we keep omitting the fields a user cannot see, instead of failing the whole request.

Reasons:
1. It matches WordPress core's users behavior, which this class stays in sync with.
2. In collection mode the same field can be visible for some users and hidden for others. A single reject would break valid requests (for example, "give me the emails of everyone I manage").
3. A per-field reject would also leak which users the caller can edit.

The output schema documents this. Each sensitive field says "Present when the current user can view it."

Fields that are not supported at all (an unknown name, or `avatar_urls` when avatars are off) are still rejected by schema validation. New tests now cover that case.

Resolving this thread.

includes/Abilities/Users/Users.php
gziolo: We chose to keep the per-user omit instead of denying the request (see the field access thread above). So this check is still needed and stays. Resolving this thread.

includes/Abilities/Users/Users.php
gziolo: Decision: we keep the current behavior. In collection mode, callers who cannot list users see only public authors, so a user with no published posts does not appear there, even through `include`. This matches WordPress core, where you read your own account through a single-user lookup (like the REST `/users/me` endpoint), not the collection.

One thing that helped the decision: for a caller without `list_users`, "a user you can edit" is only yourself, since editing other users needs `edit_users`, which implies `list_users`. So this is really about self.

Changes (commit 2adf4b1a):
- Added a test that pins the mode difference. A subscriber who includes their own ID gets an empty collection, while a single-user lookup by ID returns them.
- Added a code comment at the query that explains the intentional exclusion.
- Clarified the `include` field description so it no longer implies self is always included.

Resolving this thread.

includes/Abilities/Users/Users.php
gziolo: I removed it for now, as it isn't supported in any place in WP core as of today.

gziolo (dismissed):
Thanks @jorgefilipecosta, this is great work. I only added the final touches on top of my own review notes, so there is nothing left on your side to pick up. Every thread is resolved now.

Final touches from my side:
- Consolidated the user fields into a single source of truth, so the field list, output schema, and defaults all come from one place.
- Simplified field normalization and added tests that reject unknown or unavailable fields.
- Let the query handle row-level access in collection mode, which drops a per-user COUNT query on each result.
- Removed the pagination meta.

Two behavior questions I settled, both to match WordPress core:
1. Fields the current user cannot see are omitted per user, not rejected. This keeps collection mode usable and matches core REST.
2. A caller who cannot list users reads their own account through a single-user lookup, not the collection. There is a test for it now.

PHPUnit, PHPCS, and PHPStan all pass. Approving. Thank you for the careful work on a sensitive ability.

jorgefilipecosta:
Thank you a lot for the review and applying some suggestions @gziolo. It seems the first version of the ability is ready.
@jeffpaul, @dkotter is there any further changes that we should do before merging? Or it is fine to merge and iterate after? Thank you in advance for any insights!

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,986 @@
...
+		if ( null !== $this->user_properties ) {
+			return $this->user_properties;
+		}
+
+		$user_properties = array(
+			'id'              => array(
+				'type'        => 'integer',
+				'description' => __( 'The user ID.', 'ai' ),
+			),
+			'display_name'    => array(
```
peterwilsoncc: Suggest matching this to rest api: `name`. I have the same suggestion for the other fields too so won't write this in for all.

It's also not a hill I will die on.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,986 @@
...
+	 * @return string[] Public post type names.
+	 */
+	private function get_public_post_types(): array {
+		if ( null !== $this->public_post_types ) {
+			return $this->public_post_types;
+		}
+
+		$post_types = array();
+
+		foreach ( get_post_types( array( 'public' => true ), 'names' ) as $post_type ) {
```
peterwilsoncc: It's a real gotcha but `public` does not actually mean the post type is public. For example, the [post type in this gist is not publicly viewable](https://gist.github.com/peterwilsoncc/16df069cd23d95be6e2ca5a6a0ee99ee).

I also don't think you'll need the foreach loop with this and it's a little overengineering.

```suggestion
		$publicly_viewable_post_types = array_filter( get_post_types(), 'is_post_type_viewable' );
		foreach ( $publicly_viewable_post_types as $post_type ) {
```

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,986 @@
...
+		$query_args = array(
+			'number'      => $per_page,
+			'offset'      => ( $page - 1 ) * $per_page,
+			'count_total' => true,
+		);
+
+		$include = $this->normalize_include( $input );
+		if ( array() !== $include ) {
+			$query_args['include'] = $include;
+			$query_args['orderby'] = 'include';
```
peterwilsoncc: Just in case we add additional cache sharing to user queries (per my comment on the content abilitiy).

```suggestion
```

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,986 @@
...
+			),
+			'display_name'    => array(
+				'type'        => 'string',
+				'description' => __( 'The display name for the user.', 'ai' ),
+			),
+			'description'     => array(
+				'type'        => 'string',
+				'description' => __( 'Description of the user.', 'ai' ),
+			),
+			'user_url'        => array(
```
peterwilsoncc: Drop the `user_` prefix, it's a user by defintion in this context.

Elsewhere too.

```suggestion
			'url'        => array(
```

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,986 @@
...
+	 * Returns public post types.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string[] Public post type names.
+	 */
+	private function get_public_post_types(): array {
+		if ( null !== $this->public_post_types ) {
+			return $this->public_post_types;
+		}
```
peterwilsoncc: I'd avoid this pattern as post types can be unregistered and re-registered with different arguments, as a result the public post types may be changed between the registration and execution of the ability.

includes/Abilities/Users/Users.php
```diff
@@ -0,0 +1,986 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array<string, mixed> User field definitions.
+	 */
+	private function get_user_properties(): array {
+		if ( null !== $this->user_properties ) {
+			return $this->user_properties;
+		}
+
```
peterwilsoncc: As above re the possibility things may change.

peterwilsoncc (commented):
Greg asked me to pop in with a review, I've added a few comments from my first pass below.

dkotter:
> Thank you a lot for the review and applying some suggestions @gziolo. It seems the first version of the ability is ready. @jeffpaul, @dkotter is there any further changes that we should do before merging? Or it is fine to merge and iterate after? Thank you in advance for any insights!

Looks like a lint failure that will need cleaned up but otherwise I'm okay proceeding, though may be worth waiting for @peterwilsoncc to chime back in from his initial review

includes/Abilities/Users/Users.php
jorgefilipecosta: Adopted the REST API field names across the board in b8451bbf: display_name → name, and the same for the rest.

includes/Abilities/Users/Users.php
jorgefilipecosta: Good catch. Switched to array_filter( get_post_types(), 'is_post_type_viewable' ) and dropped the foreach in be58db87

includes/Abilities/Users/Users.php
jorgefilipecosta: Removed it in e01b49dc.

includes/Abilities/Users/Users.php
jorgefilipecosta: Done in b8451bbf — user_url → url, and dropped the prefix on the other fields too 👍

includes/Abilities/Users/Users.php
jorgefilipecosta: Good point. Dropped the cache, get_public_post_types() now resolves on every call in be58db87.

includes/Abilities/Users/Users.php
jorgefilipecosta: Same here, removed the cache so the properties are rebuilt each call in be58db87.

jorgefilipecosta:
Hi @gziolo and @peterwilsoncc thank you a lot for your reviews everything was addressed.

jorgefilipecosta:
> > Thank you a lot for the review and applying some suggestions @gziolo. It seems the first version of the ability is ready. @jeffpaul, @dkotter is there any further changes that we should do before merging? Or it is fine to merge and iterate after? Thank you in advance for any insights!
> 
> Looks like a lint failure that will need cleaned up but otherwise I'm okay proceeding, though may be worth waiting for @peterwilsoncc to chime back in from his initial review

Hi @dkotter, sorry that failure was temporary during the development it is addressed.

includes/Abilities/Users/Users.php
gziolo: The fun part is that these names are inconsistent in the project. In WP 6.9 we shipped `core/get-user-info` that has field names closer to [`wp user list`](https://developer.wordpress.org/cli/commands/user/list/):

https://github.com/WordPress/wordpress-develop/blob/e60ffe877e9e725b88baa6fd661634bad3be356c/src/wp-includes/abilities.php#L142-L165

I further extended the list of fields in this release cycle, so it looks like this:

https://github.com/WordPress/wordpress-develop/blob/fb532bafdb4490269b23cdcf8c8e082845410247/src/wp-includes/abilities.php#L145-L204

I'm perfectly fine following the REST API as a precedent for naming uniformly across all abilities to avoid repeating the same decisions. 

What would be important is to align on names for newly introduced fields for `core/get-user-info` and using the same titles and descriptions everywhere as this definition ideally should be shared. In fact, it's a strong indicator that in the long run `core/get-user-info` should be a special case for `core/read-user` that preselect the current user.

includes/Abilities/Users/Users.php
gziolo: Thanks @peterwilsoncc. Your point reached further than the first fix.

`get_public_post_types()` was switched to `is_post_type_viewable()`, but `has_published_posts: true` was still handed straight to `WP_User_Query`. Core resolves that itself as `get_post_types( array( "public" => true ) )`, which is the exact set you warned about. So a boolean filter still matched an author whose only published post used a post type like the one in your gist, even though the ability rejected that post type in the input enum, in collection mode, and in the single-user lookup.

Fixed in 4bb161eb. The ability now resolves the post types itself for both the boolean form and the explicit list, so the two mean the same thing.

That also covers your other comment about things changing. Nothing is resolved from a snapshot anymore, so post types registered or unregistered after the schema was built are handled.

There is a new test that registers the post type from your gist and checks all three paths.

gziolo (approved):
Thank you @jorgefilipecosta and @peterwilsoncc. I pushed six follow-up commits. Here is what each one does:

- `00f98114` — Bug fix. `email` and `avatar_urls` can now be `null`. WordPress can store a user with no email, and `get_avatar_url()` can return `false`. Both failed output validation, and in collection mode one bad row failed the whole page.
- `110b37f3` — Docs and a test. `is_public_author()` also counts private posts for a caller who can read them, so a single-user lookup and a collection can disagree. This matches the REST users controller, so we keep it and pin it with a test.
- `a8caabfb` — Test fix. Newer WordPress versions sanitize an invalid email in a different way.
- `d70c24b6` — Docs only. The input enums are a snapshot taken when the ability is registered. This is on purpose, so the schema stays a stable contract.
- `00f9fd6c` — Docs and a test. The `include` order is not used for sorting. The test fails if `orderby => include` ever comes back.
- `4bb161eb` — Bug fix. `has_published_posts: true` was passed to `WP_User_Query`, which resolves it with `public => true`. That is the gotcha @peterwilsoncc described. The ability now resolves the post types itself, so `true` and an explicit list mean the same thing.

PHPUnit, PHPCS, and PHPStan all pass.

The `core/get-user-info` field naming question is a follow-up, not a blocker. In the long run `core/get-user-info` could become a preset of this ability for the current user.

Approving. Thank you both for the careful work on a sensitive ability.

jorgefilipecosta:
Thank you a lot for all the insights I think this needs a last approval from @dkotter or @jeffpaul so we can merge and continue to iterate.

dkotter (approved):

PR #777, opened 2026-06-25
dev: Update Composer deps and remediate PHPStan smells

justlevine:
## What?

Bumps Composer deps to their latest versions, and remediates the resulting PHPStan errors.

More specifically, it makes `php-stubs/wordpress-stubs@7.0` a direct dependency aliased to WP 6.9.4, and then adds various inline ignores, where issues with the lack of FQCN cause type mismatches.

Now-unnecessary ignores have been removed from the PHPStan config.

## Why?

- Broadly, it's better to use the latest stubs - and suppress the type errors - than outdated ones, and type-hinting with inline-ignore annotations for improper hints are better than `ignoreErrors` because it allows the rest of the related code to still be typehinted correctly.

- While https://github.com/php-stubs/wp-cli-stubs/pull/24 fixed the use of SemVer to allow v7.x, releases are tied to WP-CLI and therefore it hasn't hit downstream. Aliasing to 6.9.4 solves this temporarily.

### Use of AI Tools

None.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
|        |       |

## Changelog Entry

> Developer - Bump Composer dev-dependencies and remediate new PHPStan smells

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-777-26e27614f36987ccbbe768f0e5243db2d1e921ad.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,7 @@
 	"require-dev": {
 		"automattic/vipwpcs": "^3.0",
 		"dealerdirect/phpcodesniffer-composer-installer": "^1.2.1",
+		"php-stubs/wordpress-stubs": "7.0 as 6.9.4",
 		"php-stubs/wp-cli-stubs": "^2.12",
 		"phpcompatibility/phpcompatibility-wp": "^3.0.0-alpha",
 		"phpstan/extension-installer": "^1.3",
@@ -21,7 +22,7 @@
 		"slevomat/coding-standard": "^8.0",
 		"szepeviktor/phpstan-wordpress": "^2.0.2",
 		"wp-coding-standards/wpcs": "^3.2.0",
-		"wp-phpunit/wp-phpunit": "^6.9",
+		"wp-phpunit/wp-phpunit": "^7.0",
 		"yoast/phpunit-polyfills": "^4.0"
 	},
 	"autoload": {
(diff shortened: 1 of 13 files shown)


jeffpaul:
@justlevine is this still in progress or can it come out of draft for review?

justlevine:
Per https://github.com/WordPress/php-ai-client/issues/257#issuecomment-4896294644, the FQCN bug has been patched upstream, so wordpress-stubs will be fixed when 7.1 generates.

I think that changes the math here, and I should move the ignores out of the file and into `ignoreErrors` it's not worth the file churn. cc @dkotter in case you disagree

dkotter:
> I think that changes the math here, and I should move the ignores out of the file and into `ignoreErrors` it's not worth the file churn. cc @dkotter in case you disagree

That makes sense to me 👍🏻

jeffpaul:
@justlevine heads up that I'm punting this to the next release, so you've got some time (especially to see what lands in WP 7.1 beta 1) before this ideally gets wrapped up

PR #787, opened 2026-06-26
Fix Abilities Explorer input schema validation

dkotter:
## What?

Closes https://wordpress.org/support/topic/bug-ability_handlervalidate_input-wont-accept-scalar-inputs/

Ensure when we validate the input schema in the Abilities Explorer page we allow both array and scalar inputs

## Why?

Abilities can have input schemas that are just scalar (like a single post ID) and also support input schemas that are arrays. Our validation we run in the Abilities Explorer was only allowing array schemas so any Abilities that had scalar schemas were rejected.

## How?

Remove the array type hint on both `invoke_ability` and `validate_input` so scalar schemas can be used.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 4.8
Used for: Validating bug, suggesting a fix and executing that fix. Final review and testing by me

## Testing Instructions

1. Enable the Abilities Explorer feature
2. Go to `Tools > Abilities Explorer`
3. Choose an Ability from the list and click on it (like `Get Environment Info`)
4. Click the `Test Ability` button
5. From here, click the `Validate Input` button to ensure the schema validates correctly
6. Click the `Invoke Ability` and ensure the Ability is triggered successfully

## Changelog Entry

> Fixed - Ensure scalar input schemas are allowed in the Abilities Explorer validation.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-787-b20616f2ad6c9815d6a6817e8a55d957ad5f75b0.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (dismissed):
Tested via P{layground and using test steps things worked as expected.

squarestar:
The `mark_setting()` method in `Show_In_Abilities.php` is hooked to the `register_setting_args` filter and has a strict array type declaration for `$args`. One plugin on my system (WPBakery Page Builder) calls register_setting() with null args, which causes a fatal error. Maybe this method needs to account for issues in plugins? A defensive solution could be:

```
public function mark_setting( ?array $args, array $defaults, string $option_group, string $option_name ): array {
    $args = $args ?? [];
```

dkotter:
> The `mark_setting()` method in `Show_In_Abilities.php` is hooked to the `register_setting_args` filter and has a strict array type declaration for `$args`. One plugin on my system (WPBakery Page Builder) calls register_setting() with null args, which causes a fatal error.

Thanks for the flag there. I'd argue that's an issue in that plugin as they shouldn't be returning `null` for that, they should be using an empty array, but that is a good flag. I've pushed b20616f2ad6c9815d6a6817e8a55d957ad5f75b0 to remove the typing on these params to avoid these issues, plus adding an `is_array` check

jeffpaul:
@squarestar if you can [connect your GitHub and WordPress.org accounts](https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/), then I can formally credit you on the release announcement post and grant the AI Contributor badge.

squarestar:
Thanks but that WP account isn't the one I'd want connected to this GH account. I've connected my WP squarestar account but I used a different one to open the topic on wp.org. No worries if you don't credit me.

jeffpaul:
@squarestar what's your preferred WPORG account to credit for work on this AI plugin release then?

squarestar:
https://profiles.wordpress.org/squarestar/

Thanks!

PR #794, opened 2026-06-26
Fix Editorial Note generation spinner scope - to active block in block settings menu

### LINKED ISSUE #792: Editorial Note generation spinner shows on all blocks' settings menu
### Description

In the Editorial Note experiment:
- When a single content block's "Generate Editorial Notes" option is selected, it changes to Reviewing text with a spinner incorrectly for *all blocks*.
- If I head over to another content block's option menu in the meanwhile, it will also show Reviewing even though we have selected the review generation for a specific separate content block only.

This leads to assumption that all blocks are undergoing an Editorial review.

1. Enable Editorial Note experiment.
2. Open the post editor and insert two paragraph blocks with enough content to run AI Editorial reviews.
3. Select the first block, click the "Options" (three dots) button in the block toolbar, and click "Generate Editorial Note".
4. While the review is running (spinner showing on the first block), select the second block and click its "Options" button in the block toolbar.
5. Observe the menu item for Editorial Notes on the second block:
6. Actual Behavior: It incorrectly shows "Reviewing…" with a spinner, suggesting the second block is also being reviewed.
7. Expected Behavior: It should display the "Generate Editorial Note" label but disabled, along with helper info text indicating that "Another block is currently being reviewed."

https://github.com/user-attachments/assets/16663f82-61d2-412b-ba6e-1dd7843809df

### Environment info

WordPress v7.0 , AI Plugin Version 1.0.2 - dev branch , Twenty Twenty-Five Theme
Opera GX browser
using Macbook with MacOS Tahoe v26

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

Intenzi:
## What?

Closes #792 

For the "Generate Editorial Note" option in Editorial Note experiment:
- Shows the "Reviewing..." loading spinner only on the actual clicked block, instead of showing it on every content block's option menu.
- Displays an additional info for the other blocks stating that a review process is ongoing for another block.

## Why?
It solves a visual bug where running an AI review on one block makes it look like *every* block is being reviewed. 

## How?

This is address by tracking the specific block ID under review instead of a global true/false state, other blocks correctly show as disabled with an explanation (Another block is currently being reviewed) rather than showing a confusing loading spinner.

- Instead of using a simple `isReviewing` boolean state, we now track `reviewingClientId` which stores the ID of the block currently under review / `null` if none under review.
- This allows to check for `isReviewing` i.e. any block is under review by evaluating `!!reviewingClientId` i.e. truthy if a block ID is stored and false if `null`.
- By comparing `reviewingClientId` to the current block's ID, we only show the "Reviewing..." text and loading spinner conditionally to the under review block only.
- And for non equality, where we know a review is still occurring, the menu item remains as "Generate Editorial Note" and displays an additional info label of "Another block is currently being reviewed." while staying disabled.

- Added an e2e test to verify the same.

### Use of AI Tools

AI assistance: Yes
Tool(s): Antigravity IDE
Model(s): Gemini 3.5 Flash (Low)
Used for: Identifying the relevant files for the fix, suitable resolutions and E2E test implementation. Final implementation and tests were reviewed and edited by me.

## Testing Instructions

1. Enable Editorial Note experiment.
2. Open the post editor and insert two paragraph blocks with enough content to run AI Editorial reviews.
3. Select the first block, click the "Options" (three dots) button in the block toolbar, and click "Generate Editorial Note".
4. While the review is running (spinner showing on the first block), select the second block and click its "Options" button in the block toolbar.
5. Observe the menu item for Editorial Notes on the second block:
6. It should display the "Generate Editorial Note" label but disabled, along with helper info text indicating that "Another block is currently being reviewed."

## Screenshots or screencast

https://github.com/user-attachments/assets/6341889d-e0a7-47fd-9c8a-56e16f3c944f

## Changelog Entry

> Fixed - Scope the Editorial Note generation loading spinner only to the block currently being reviewed.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-794-1af82c6eb1e7333b70b56b71cf32831ba9ae004b.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
@Intenzi any reason we're blocking other blocks from kicking off a single block editorial review?  Can multiple editorial reviews on single blocks NOT run at the same time?

Intenzi:
@jeffpaul Yeah, thats definitely something we could look into as an enhancement! The decision to block concurrent reviews is existing behaviour that we inherited, we were already utilising `reviewBlocked` to prevent other blocks from being reviewed simultaneously. The goal of this PR was to fix the UX of that existing decision ^-^

src/experiments/editorial-notes/components/EditorialNotesPlugin.tsx
```diff
@@ -177,22 +178,35 @@ export default function EditorialNotesPlugin() {
...
 							disabled={
-								isReviewingBlock || isBlockReviewDisabled
+								isAnyBlockReviewing || isBlockReviewDisabled
 							}
 							onClick={ () => {
 								if ( clientId ) {
 									reviewBlock( clientId );
 								}
 							} }
+							{ ...( isAnyBlockReviewing && ! isThisBlockReviewing
```
dkotter: I believe `info` accepts `undefined` so this could likely be simplified to something like:

```ts
info={
    isAnyBlockReviewing && ! isThisBlockReviewing
        ? __( 'Another block is currently being reviewed.', 'ai' )
        : undefined
}
```

tests/e2e/specs/experiments/editorial-notes.spec.js
```diff
@@ -303,4 +303,99 @@ test.describe( 'AI Editorial Notes Experiment', () => {
...
+		} );
+
+		// Set up a deferred promise to intercept and hold the Ability request.
+		let resolveRequest;
+		const requestPromise = new Promise( ( resolve ) => {
+			resolveRequest = resolve;
+		} );
+
+		// Intercept the modern WP Abilities REST endpoint. This route was updated upstream from
+		// wp/abilities/v1/run to wp-abilities/v1/abilities. We target the run path specifically.
```
dkotter:
```suggestion
		// Intercept the modern WP Abilities REST endpoint.
```

dkotter (changes requested):

src/experiments/editorial-notes/components/EditorialNotesPlugin.tsx
Intenzi: I tried implementing this, it turns out the project has `exactOptionalPropertyTypes: true` enabled in `tsconfig.json`. Because of that, TypeScript raises a compile error if we explicitly pass `undefined` to `info` instead of omitting the property entirely.

I've updated it to use a shorter conditional spread pattern instead.

dkotter (approved):

PR #802, opened 2026-06-29
Fix[Experiments]: Adopt character based count to gate features

### LINKED ISSUE #767: AI feature buttons remain disabled for sufficient non-space-delimited content in word-based locales
### Description

PR https://github.com/WordPress/ai/pull/581 introduced locale-aware minimum content length checks across AI features. That helped align gating AI Features with user’s locale: CJK and similar locales can use a character-based count, while other locales use a word-based count via WordPress’ `_x( 'words', 'Word count type. Do not translate!' )` language-pack behavior.

However, one important drawback is that the gate is based on the WP/user locale, not the language of the actual post content. In multilingual publishing workflows, those can differ. For example, a site or user profile may be set to English while the author writes Japanese, Chinese, Korean, Thai, or similar content. In that case, `_x()` continues to resolve to words, but the content naturally needs a character-based count. The result is that AI feature buttons can remain disabled even when the post has enough meaningful content.

Image

Gutenberg (upstream) may use the locale-aware strategy for displaying word count at the sidebar, which is informational. Here, the same strategy controls whether a feature is available, so the consequence is more significant.

I’d like to use this issue to discuss the best approach here. My current preference is to standardize these gates on `characters_excluding_spaces`. That would make the checks deterministic, avoid false negatives for CJK (or similar) content under word-count based locales, and keep the threshold semantics consistent across users.

There is also a related consistency concern today: a `minContentLength` value of 15 means “15 characters” for character-count locales, but “15 words” for word-count locales. While expressiveness is not directly comparable across languages, using the same scalar for different units makes the gate inconsistent, especially in multilingual setups. 

|EN|JA|
|-|-|
|Image|Image|

If this is already tracked elsewhere, or if this behavior is intentional for now, please feel free to close this issue.

1. Set the language to English.
2. Create a new post with sufficient CJK, or similar non-space-delimited content.
3. Observe that the AI feature buttons remain disabled, even though the post contains enough meaningful content.

_No response_

### Environment info

- AI Plugin: Latest (develop)

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

yogeshbhutkar:
## What, Why, and How?

Closes https://github.com/WordPress/ai/issues/767

This PR standardizes AI feature content-length checks on character counts excluding spaces, instead of switching between words and characters based on locale.

This fixes cases where AI feature buttons stay disabled for sufficient CJK or similar non-space-delimited content when the site or user locale uses word-based counting. It also keeps `minContentLength` consistent by making the unit deterministic across users and languages.

The character thresholds are based on the previous word thresholds, using 5 characters as an approximate average word length. So a previous limit of 15 words now becomes 75 (15*5) characters. This keeps the limits calculated and proportional, rather than choosing new values arbitrarily. 

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.5
Used for: Code reviews

## Testing Instructions

1. Create a new post.
2. Confirm gated AI features show the minimum length message using character count.
3. Keep the site/editor locale set to English.
4. Add Japanese content to the post.
5. Confirm the gate is removed once the content reaches the required character length.
6. Use Content Resizing and confirm the changeset shows character counts correctly instead of “no change”.
7. Confirm the Shorten action works with sufficient Japanese content in an English locale.

## Screenshots

|Before|After|
|-|-|
|before-01|after-01|
|before-02|after-02|
|before-03|after-03|

## Changelog Entry

> Changed - Use character-based count to gate AI Experiments.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-802-cc285a8b3e6d72611507ea10cc402fdeb5af69c8.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Content_Resizing/Content_Resizing.php
+++ b/includes/Abilities/Content_Resizing/Content_Resizing.php
@@ -13,7 +13,8 @@ use WP_Error;
 use WordPress\AI\Abstracts\Abstract_Ability;
 use WordPress\AI\Experiments\Content_Resizing\Content_Resizing as Content_Resizing_Experiment;
 
-use function WordPress\AI\count_words;
+use function WordPress\AI\count_characters_excluding_spaces;
+use function WordPress\AI\get_min_content_length;
 
 /**
  * Content resizing WordPress Ability.
@@ -32,13 +33,13 @@ class Content_Resizing extends Abstract_Ability {
 	protected const ACTION_DEFAULT = 'rephrase';
 
 	/**
-	 * The minimum word count for the shorten action.
+	 * The default minimum content length in characters for the shorten action.
 	 *
-	 * @since 0.9.0
+	 * @since x.x.x
 	 *
 	 * @var int
 	 */
-	protected const SHORTEN_MIN_WORDS = 5;
+	public const DEFAULT_MIN_CONTENT_LENGTH = 25;
 
 	/**
 	 * {@inheritDoc}
@@ -105,17 +106,22 @@ class Content_Resizing extends Abstract_Ability {
 			);
 		}
 
+		$min_content_length = get_min_content_length(
+			'content-resizing',
+			self::DEFAULT_MIN_CONTENT_LENGTH
+		);
+
 		// "shorten" action requires a minimum word count.
 		if (
 			'shorten' === $args['action'] &&
-			count_words( wp_strip_all_tags( $content ) ) < self::SHORTEN_MIN_WORDS
+			count_characters_excluding_spaces( $content ) < $min_content_length
 		) {
 			return new WP_Error(
 				'content_too_short',
 				sprintf(
-					/* translators: %d: Minimum word count. */
-					esc_html__( 'A minimum of %d words is required to shorten the content.', 'ai' ),
-					self::SHORTEN_MIN_WORDS
+					/* translators: %d: Minimum content length in characters. */
+					esc_html__( 'A minimum of %d characters is required to shorten the content.', 'ai' ),
+					$min_content_length
 				)
 			);
 		}
--- a/includes/Abilities/Content_Resizing/Content_Resizing.php
+++ b/includes/Abilities/Content_Resizing/Content_Resizing.php
@@ -111,7 +111,7 @@ class Content_Resizing extends Abstract_Ability {
 			self::DEFAULT_MIN_CONTENT_LENGTH
 		);
 
-		// "shorten" action requires a minimum word count.
+		// "shorten" action requires a minimum character count.
 		if (
 			'shorten' === $args['action'] &&
 			count_characters_excluding_spaces( $content ) < $min_content_length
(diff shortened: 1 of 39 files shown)


includes/Abilities/Content_Resizing/Content_Resizing.php
```diff
@@ -105,17 +106,22 @@ protected function execute_callback( $input ) {
 			);
 		}
 
+		$min_content_length = get_min_content_length(
+			'content-resizing',
+			self::DEFAULT_MIN_CONTENT_LENGTH
+		);
+
 		// "shorten" action requires a minimum word count.
```
dkotter: This comment is no longer accurate

src/experiments/excerpt-generation/components/useExcerptGeneration.ts
dkotter: This default now no longer matches what's on the PHP side. I think we have a few other places that do the same thing (`TitleToolbar.tsx`, `useContentClassification.ts`, `useSummaryGeneration.ts`, and `useMetaDescription.ts`) so ideally those are all updated to be consistent

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
dkotter, line 1: With the changes here, I think we'll need to update the content resizing documentation as it still mentions word counts and now we do only character counts

includes/helpers.php
```diff
@@ -77,31 +77,39 @@ function normalize_content( string $content ): string {
...
  *
- * @param string $text The text to count words in.
- * @return int The number of words.
+ * @param string $text The text to count characters in.
+ * @return int The number of non-whitespace characters.
  */
-function count_words( string $text ): int {
-	$text = trim( $text );
-	if ( '' === $text ) {
+function count_characters_excluding_spaces( string $text ): int {
```
dkotter, line 90: Ideally we should get a unit test for this new function

dkotter (changes requested):

dkotter (approved):

jeffpaul:
@hasanmehmood if you can [connect your GitHub and WordPress.org accounts](https://make.wordpress.org/core/handbook/tutorials/linking-your-github-and-w-org-profiles/), then I can more formally credit you in the release post there and grant you the AI Contributor badge.

PR #820, opened 2026-07-02
feat: update Type Ahead experiment implementation

### LINKED ISSUE #816: Type-Ahead experiment loads wp-editor on the front end, intermittently breaking WooCommerce block checkout (sample Beanie/Cap cart)
## Summary

The Type-Ahead experiment registers its assets on `enqueue_block_assets`, which fires on the **front end** as well as in the editor, with no `is_admin()` guard:

https://github.com/WordPress/ai/blob/9db55ee59c642f006ea9795b30bd44ee3d66d9a4/includes/Experiments/Type_Ahead/Type_Ahead.php#L70-L73

The built `experiments/type-ahead` script declares `wp-editor` as a dependency (see the generated `build-scripts/experiments/type-ahead.asset.php`), so **every front-end page load pulls in the entire block-editor stack** (`editor.min.js` plus ~20 dependency scripts) and — critically — **registers the `core/editor` data store on the front end**.

## Impact 1: intermittently corrupts WooCommerce block checkout

WooCommerce's Store API cart resolver decides "am I in the editor?" via `!! select( 'core/editor' )`:

https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/data/utils/is-editor.ts

and, when true, seeds the cart store with hard-coded preview data instead of fetching the real cart (`data/cart/resolvers.ts`):

```ts
if ( isEditor() ) {
	dispatch.receiveCart( previewCart );
	return;
}
```

With this plugin active, whether `editor.min.js` has executed before the checkout block mounts is a script-execution vs. lazy-chunk race, so shoppers **intermittently see WooCommerce's sample order summary — "Beanie ×2 / Cap ×1, total $49.20" — instead of their real cart on `/checkout`**, and the real cart is never fetched for that page view. The Mini-Cart block is exposed to the same store poisoning on every front-end page. (One can argue WooCommerce's `isEditor()` heuristic is fragile too, but an editor-only feature loading `wp-editor` on the front end is the actionable defect here.)

We debugged this in production for a while before tracing it to this plugin — happy to provide more detail if useful.

## Impact 2: front-end payload

The full editor bundle ships on every front-end page for a feature that can only run inside the block editor ("Ghost text suggestions while writing paragraphs in the block editor", category `EDITOR`).

## Also worth noting

`enqueue_assets()` enqueues the script even when the experiment is **disabled** — `enabled` is only passed through `localize_script()` as data:

https://github.com/WordPress/ai/blob/9db55ee59c642f006ea9795b30bd44ee3d66d9a4/includes/Experiments/Type_Ahead/Type_Ahead.php#L96-L115

## Steps to reproduce

1. WordPress + WooCommerce (block checkout) + AI plugin 1.1.0 active. The Type-Ahead experiment can be OFF — the script enqueues regardless.
2. Visit any front-end page and run `!! wp.data.select( 'core/editor' )` in the console → returns `true` (expected: `false`). `wp-includes/js/dist/editor.min.js` is present in the page's script list.
3. Add any product to the cart, open `/checkout`, and hard-refresh repeatedly → the Order Summary intermittently renders the `previewCart` sample data (Beanie/Cap) depending on the race.

## Suggested fix

The script is editor-only, so either:

- hook it on `enqueue_block_editor_assets` instead of `enqueue_block_assets`, or
- bail out of `enqueue_assets()` early when `! is_admin()`;
- and ideally skip enqueueing entirely when the experiment is disabled.

The Summarization experiment already handles this correctly: only its (iframe-relevant) **stylesheet** rides `enqueue_block_assets`, while its script stays behind a `get_current_screen()` check.

## Environment

- ai 1.1.0 (with ai-provider-for-anthropic 1.0.3 / ai-provider-for-openai 1.0.3)
- WordPress 7.0, PHP 8.3, WooCommerce 10.9.1
- The hook is unchanged on current trunk (9db55ee).

SohamPatel46:
## What?

Closes #816

This PR restricts the Type-Ahead experiment assets to load only within the block editor, and ensures they are not enqueued at all if the experiment is disabled.

## Why?

Previously, the Type-Ahead experiment was hooked to `enqueue_block_assets`, which executes on the front end as well as in the editor. Because its script declares `wp-editor` as a dependency, this was unintentionally pulling the entire block-editor stack into every front-end page load. This extra payload was unnecessary, and it also intermittently corrupted the WooCommerce block checkout by poisoning the cart store with `core/editor` preview data. Additionally, the assets were enqueued even when the experiment was disabled.

## How?

1. Updated `includes/Experiments/Type_Ahead/Type_Ahead.php` to hook into `enqueue_block_editor_assets` instead of `enqueue_block_assets`.
2. Added an early return `if ( ! $this->is_enabled() ) { return; }` to `enqueue_assets()` so assets aren't enqueued when the experiment is off.
3. Updated the integration tests in `Type_AheadTest.php` to verify the correct hook (`enqueue_block_editor_assets`).

### Use of AI Tools

AI assistance: Yes
Tool(s): Gemini AI via Antigravity IDE
Model(s): Gemini 3.1 Pro
Used for: Identifying the issue root cause, modifying code for the hook update.

## Testing Instructions

1. Ensure the AI plugin is active.
2. Toggle the Type-Ahead experiment ON.
3. Open any front-end page and open the browser console.
4. Run `!! wp.data.select( 'core/editor' )` in the console. It should now return `false` (or throw an error if `wp.data` is completely absent), verifying the editor bundle is no longer polluting the front end.
5. Toggle the Type-Ahead experiment OFF. 
6. Open the block editor and verify its specific scripts (`experiments/type-ahead`) don't load.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| Screenshot 2026-07-03 at 12 34 58 AM | Screenshot 2026-07-03 at 12 35 19 AM |
| Screenshot 2026-07-03 at 12 34 45 AM | Screenshot 2026-07-03 at 12 34 21 AM |

## Changelog Entry

> Fixed - Prevent Type-Ahead assets from loading on the front end when the experiment is disabled.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-820-001b39e6dd6318de3a762084afb07e90b6a9ab03.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Experiments/Type_Ahead/Type_Ahead.php
```diff
@@ -94,6 +94,10 @@ public function register_abilities(): void {
 	 * @since 1.1.0
 	 */
 	public function enqueue_assets(): void {
+		if ( ! $this->is_enabled() ) {
+			return;
+		}
```
dkotter: I don't think this is needed as the `register` method only fires when `is_enabled` is true, so this just runs the same check again

tests/Integration/Includes/Experiments/Type_Ahead/Type_AheadTest.php
```diff
@@ -102,8 +102,24 @@ public function test_register_hooks_actions() {
...
-			'enqueue_assets should be hooked to enqueue_block_assets'
+			has_action( 'enqueue_block_editor_assets', array( $experiment, 'enqueue_assets' ) ),
+			'enqueue_assets should be hooked to enqueue_block_editor_assets'
 		);
 	}
+
+	/**
+	 * Tests enqueue_assets() bails when experiment is disabled.
+	 *
+	 * @since 1.1.1
```
dkotter:
```suggestion
	 * @since x.x.x
```

dkotter (changes requested):
Left a couple minor comments. Will also need to ensure the documentation is updated with the new hook: https://github.com/WordPress/ai/blob/56cef56c1fa90e436d943bc732880fa200cf0b7c/docs/experiments/type-ahead.md?plain=1#L38

includes/Experiments/Type_Ahead/Type_Ahead.php
```diff
@@ -69,7 +69,7 @@ protected function load_metadata(): array {
 	 */
 	public function register(): void {
 		add_action( 'wp_abilities_api_init', array( $this, 'register_abilities' ) );
-		add_action( 'enqueue_block_assets', array( $this, 'enqueue_assets' ) );
+		add_action( 'enqueue_block_editor_assets', array( $this, 'enqueue_assets' ) );
```
dkotter, line 72: Tested this and while the JS loads properly, the CSS we load isn't working anymore so that needs fixed

dkotter (changes requested):

includes/Experiments/Type_Ahead/Type_Ahead.php
```diff
@@ -113,6 +113,15 @@ public function enqueue_assets(): void {
 		);
 	}
 
+	/**
+	 * Enqueues the block stylesheet for the editor iframe and the front end.
```
dkotter, line 117: So the CSS we load here is not meant for the front-end, only the admin. Ideally we scope this to only load in the admin

dkotter (changes requested):

jeffpaul:
@SohamPatel46 are you able to work on updates from the code review above?

dkotter (dismissed):

dkotter (approved):

jeffpaul:
@xuanji86 if you can confirm your WPORG account and [connect it to your GitHub profile](https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/), then I can properly credit you in the AI plugin release and grant you the AI Contributor badge.

PR #830, opened 2026-07-04
Fix/815 grant ai plugin access

### LINKED ISSUE #815: Connector Approvals doesn't immediately flag need to grant AI plugin access to available provider(s)
### Description

Originally reported by @vfontjr on WPORG support forum ("[Enabling Experimental Features Breaks Open AI Connector](https://wordpress.org/support/topic/enabling-experimental-features-breaks-open-ai-connector/)").

When you first enable Connector Approvals there isn't an admin notice showing you need to approve the AI plugin to use an AI provider plugin.  You only first realize something Is amiss when you try to use AI functionality and get an error message.

Lightly related to #660.

1. Connect AI provider
2. Enable AI experiments (i.e. Title Gen), including Connector Approvals
3. Refresh page
4. See that no admin notice requests AI plugin access to connected AI provider
5. go into hello world post and try to generate a title
6. See that error shows no available connector
7. return to WP admin and see that admin notice now shows AI plugin needs access to a provider

_No response_

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

roberiacono:
## What?
Closes #815 

It displays an admin notice when Connector Approval is enabled in Settings page, informing you that the AI plugin must be approved before it can use an AI provider plugin.

## Why?
When you first enable Connector Approvals there isn't an admin notice showing you need to approve the AI plugin to use an AI provider plugin. You only first realize something Is amiss when you try to use AI functionality and get an error message.

## How?
in `includes/Settings/Settings_Page.php` I add $data['hasApprovedConnector'] which I use in `routes/ai-home/components/FeatureToggle.tsx` to display routes/ai-home/components/ConnectorApprovalNotice.tsx.
$data['hasApprovedConnector'] use `( new Approvals_Store() )->get_approvals()` to check if there is at least one approved connector.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Sonnet 5
Used for: Initial code skeleton and suggestions; final implementation were reviewed and edited by me. Used also for tests.

## Testing Instructions

- On `wp-admin/tools.php?page=ai-connector-approval` disable all approved connectors.
- Enable Connector Approval on Settings page
- You should now see the admin notice.

## Screenshots or screencast
connector approval

## Changelog Entry
Add - adds an admin notice when Connector Approvals are enabled, prompting admins to approve the AI plugin for use with an AI provider, preventing later runtime errors.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-830-f878888b372fac4bf6e1f23f14997d19d6f46f2b.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


tests/bootstrap.php
```diff
@@ -33,6 +33,13 @@
 // Give access to tests_add_filter() function.
 require_once $_test_root . '/includes/functions.php';
 
+// The wp-build generated render callback is never loaded during tests
+// (the WPAI_IS_TEST guard in Requirements skips build/build.php). Stub it
+// so Settings_Page::init() registers its script module data filter.
+if ( ! function_exists( 'ai_ai_wp_admin_render_page' ) ) {
+	function ai_ai_wp_admin_render_page(): void {}
+}
```
dkotter: Is this needed to get the `test_settings_page_script_module_data_includes_connector_fields` test to run? I've just never run into issues myself with tests on the settings page so curious on this

dkotter (commented):
So overall this works but not sure I love having to pass around the `connectorApprovalUrl` and `hasApprovedConnector` data to all features, especially when the Approval experiment isn't even turned on.

Wondering if there's a better way to handle that, maybe even go super simple here and either update the Connector Approval description to make it clear that turning it on will block all plugins, including the AI plugin. Or continue showing that inline notice this PR adds but not worry about checking if we have approved connectors

tests/bootstrap.php
roberiacono: Yes, I think the stub is needed.

I verified it directly by running the test with and without it:

With the stub: test_settings_page_script_module_data_includes_connector_fields passes (8 assertions).
With the stub removed: it fails at assertNotFalse( has_filter( 'script_module_data_ai-wp-admin' ) ), has_filter() returns false.

This is because `Settings_Page::init()` on https://github.com/WordPress/ai/blob/30fc728391e0fe40d20aa8b70b64a5126fde2935/includes/Settings/Settings_Page.php#L58 branches on `function_exists( 'ai_ai_wp_admin_render_page' )`.  That function is defined by `build/pages/ai/page-wp-admin.php` which is defined in `build/build.php`.

https://github.com/WordPress/ai/blob/30fc728391e0fe40d20aa8b70b64a5126fde2935/includes/Requirements.php#L106 explicitly skips requiring build/build.php when WPAI_IS_TEST is true, so PHPUnit runs never define that function.

Without the stub, init() never registers the `script_module_data_ai-wp-admin` filter that exposes `connectorApprovalUrl`/`hasApprovedConnector`/etc. , which is exactly what this test asserts on.

roberiacono:
Maybe we can add a REST API endpoint for connector approval state, so only the feature that needs it pays the cost.
Actually, I see there is already `connector-approvals` API endpoint that maybe we can use https://github.com/WordPress/ai/blob/30fc728391e0fe40d20aa8b70b64a5126fde2935/includes/Connector_Approval/REST_Controller.php#L69-L72

dkotter:
> Maybe we can add a REST API endpoint for connector approval state, so only the feature that needs it pays the cost.
> Actually, I see there is already `connector-approvals` API endpoint that maybe we can use

I wouldn't introduce a new endpoint for something as simple as this but if that endpoint works I'd be fine with that approach. Just noting again that I'd lean towards a much simpler solution here where we just default to always making it clear that turning on this experiment blocks all AI interactions, including the AI plugin itself.

roberiacono:
in commit [1b2a0aad](https://github.com/WordPress/ai/pull/830/commits/1b2a0aad5635613a43dd37eb0c013c73c040b666) I only improved the description.

In [2a68664](https://github.com/WordPress/ai/pull/830/commits/2a68664329d89fab5b39c10570c21dc2a1a36a7f) I added the API call and admin notice. The `/connector-approvals` endpoint is fetched when the page loads and whenever the toggle state changes to `checked`.

routes/ai-home/components/FeatureToggle.tsx
```diff
@@ -61,6 +118,25 @@ export function FeatureToggle( {
...
+							className="ai-developer-mode-fields ai-feature-settings-form"
+							direction="column"
+							gap="md"
+						>
+							<Stack direction="column" gap="md">
+								{ isCheckingApprovals && <Spinner /> }
+								{ ! isCheckingApprovals &&
+									approvalState &&
+									approvalState.approvals.length === 0 && (
+										<ConnectorApprovalNotice />
```
dkotter, line 146: Maybe I'm doing something wrong but I don't ever see this notice show up. Here's the steps I took:

1. I already had Connector Approvals turned on with connectors approved. I removed approval and verified things were blocked
2. Went back to the AI settings page and no notice shows
3. I toggled Connector Approvals off and then back on
4. I see a slight flash of a loading icon but no notice ever shows

roberiacono: Thanks for your fixes!

I've followed your steps, and everything looks good on my end.

FireShot Capture 222 - AI ‹ RIACO Shop — WordPress -  localhost

Peek 2026-07-09 21-57



1. Ran npm run build
2. Opened the DevTools and disabled the cache
3. Followed the steps you described

Maybe I'm missing something.

dkotter: Okay, tested again and narrowed it down.

If I have a connector that I have previously approved but then I disable that connector plugin (in my case I had approved Ollama but had since disabled the plugin), we still store that as being approved (so if in the future they activate the plugin again, it will work).

But I think the check we're doing here just looks for any approved connector, not if there's an active approved connector. Probably more of an edge case but likely a good thing to fix up if we can.

roberiacono:
Added a new variable called `hasApprovedConnector` that checks whether there is at least one active connector that has approved the plugin. If not, display the `ConnectorApprovalNotice`.

dkotter (approved):

PR #841, opened 2026-07-07
CI: Do not run `build-plugin-zip` when PR metadata changes

yogeshbhutkar:
## What, Why, and How?

While working on other PRs, I noticed that build-plugin-zip reruns for PR metadata-only changes, such as title or description edits. Each run takes about three minutes, and since the generated artifact depends on the PR code rather than PR metadata, repeated edits can waste CI time. This change limits the workflow to code-relevant PR events (opened, synchronize, and reopened), matching [Gutenberg’s corresponding workflow](https://github.com/WordPress/gutenberg/blob/trunk/.github/workflows/build-plugin-zip.yml), which uses the default pull_request activity types.

P.S. edited also runs when the PR base branch changes. If we decide that base-branch changes should still rebuild the plugin ZIP (which I don't think is required), we can keep edited and add an if guard so only that specific edit path triggers the workflow.

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.5
Used for: Code review

## Testing Instructions
- CI should be green.
- Changing PR metadata should not trigger `build-plugin-zip`.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-841-a5bcb0e8f9401085dee0a0020cfe920257da34dd.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/.github/workflows/build-plugin-zip.yml
+++ b/.github/workflows/build-plugin-zip.yml
@@ -2,7 +2,7 @@ name: Build Plugin Zip
 
 on:
   pull_request:
-    types: ['opened', 'synchronize', 'reopened', 'edited']
+    types: ['opened', 'synchronize', 'reopened']
 
 # Cancels all previous workflow runs for pull requests that have not completed.
 concurrency:


dkotter (commented):
I'm open to thoughts here but this workflow is intentionally run on the `edited` flow as the purpose of this workflow is to build a final plugin zip which then triggers our [PR Playground Preview](https://github.com/WordPress/ai/blob/develop/.github/workflows/pr-playground-preview.yml) workflow.

That workflow will add a link to the PR description that will open that PR in a Playground instance. That workflow uses the [WordPress Playground PR Preview Action](https://github.com/WordPress/action-wp-playground-pr-preview) and it recommends using the `edited` action. The main reason for that is if someone edits the PR and removes the Playground link, the workflow will automatically trigger and add it back.

yogeshbhutkar:
This makes sense to me. Thank you for your time. Closing this PR as this is a calculated decision.

PR #739, opened 2026-06-16
Add core/read-content ability

jorgefilipecosta:
## Summary

Part of: https://github.com/WordPress/ai/issues/40

Adds the read-only `core/read-content` ability to the plugin, mirroring the companion WordPress Core implementation so the two stay in sync. It overrides any core-provided copy.

The ability has explicit modes:
- single post by `id`, with optional `post_type` guard, returns the post directly
- single post by `post_type` and `slug` returns the post directly
- query mode returns `{ posts, total, total_pages }` and supports `status`, `author`, `parent`, `include`, `fields`, `page`, and `per_page`

Defaults are lean: `id`, `post_type`, `status`, `date`, `slug`, and `title_rendered`. Heavier fields are opt-in via `fields`; raw fields require edit access, while rendered fields can be requested for readable posts.

Companion Core PR: https://github.com/WordPress/wordpress-develop/pull/12195

## Security

Uses a coarse capability gate plus per-post read/edit checks. Missing, unreadable, mismatched, and unexposed posts return the same not-found response.

## Tests

PHPUnit integration tests cover permissions, fields, single-post modes, query mode, and `include`. The E2E spec `tests/e2e/specs/abilities/core-read-content.spec.js` exercises the client-side ability modes and `include`.

## Manual testing

Paste this in the browser console on an editor/admin page after enabling AI. It creates two published posts and one draft, then exercises query mode, draft query mode, query `include`, single-post by `id`, and single-post by `post_type` + `slug`. Each ability call logs `{ input, output }`.

```js
const { ready } = await import( '@wordpress/core-abilities' );
if ( ready ) {
	await ready;
}

const { executeAbility } = await import( '@wordpress/abilities' );
const apiFetch =
	window.wp?.apiFetch || ( await import( '@wordpress/api-fetch' ) ).default;

const run = async ( input ) => {
	try {
		const output = await executeAbility( 'core/read-content', input );
		console.log( { input, output } );
		return output;
	} catch ( error ) {
		const output = {
			error: {
				code: error?.code,
				message: error?.message,
			},
		};
		console.log( { input, output } );
		return output;
	}
};

const suffix = Date.now();
const seededPosts = await Promise.all(
	[ 'one', 'two' ].map( ( label ) =>
		apiFetch( {
			path: '/wp/v2/posts',
			method: 'POST',
			data: {
				title: `core/read-content manual test ${ label }`,
				slug: `core-read-content-manual-${ suffix }-${ label }`,
				status: 'publish',
[...]

PR #852, opened 2026-07-10
Respect explicit and core-owned show_in_abilities setting values

gziolo:
## What?

Follow up to #739.

`Show_In_Abilities::mark_setting()` adds the `show_in_abilities` flag to a curated list of core settings, so the `core/read-settings` ability returns data on a stock site. Its docblock says it respects a value that is already there, and only fills the flag in when it is absent. Today it does neither reliably. This PR makes the code match that promise.

These two changes started inside #739, which is about `core/read-content`. They have nothing to do with post types, so they are moved here to keep both pull requests focused.

## Why?

There are two separate problems.

**An explicit opt-out is ignored.** The check uses `empty( $args['show_in_abilities'] )`. A value of `false` is empty, so a site that deliberately opts a curated setting out has that choice overwritten and the setting is exposed anyway. An empty array is treated the same way.

**Core's own default would be overridden.** `register_setting()` passes the filter the defaults it is about to merge in, and the method ignores them. That array is core's statement of which arguments it understands. It already lists `show_in_rest`. When core ships `show_in_abilities`, it will pick the default, and each setting will opt in through `register_setting()`, the way `blogname` opts in to `show_in_rest` today. Without a guard, this polyfill would keep forcing the curated settings on and override the default core chose. That list includes `admin_email` and `siteurl`.

## How?

Check for the key instead of for a truthy value, so `false` and `array()` are respected.

Read the `$defaults` argument the filter already receives, and return early when `show_in_abilities` is in it. Core then owns the flag. This matches what `mark_post_type()` and `mark_registered_post_types()` do for post types in #739.

Nothing changes on any WordPress that exists today, because core does not pass the argument yet. On a live site the same 18 curated settings are still flagged.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: Finding the two problems, writing the fix and the tests, and running the checks. Every change was reviewed and directed by me.

## Testing Instructions

1. Run the unit tests: `npm run test:php -- --filter Show_In_AbilitiesTest`. All tests should pass.
2. Confirm each fix is really covered. Change the guard back to `empty( $args['show_in_abilities'] )` and only `test_respects_explicit_false_setting_value` fails. Remove the `core_declares_setting_flag()` call and only `test_stands_down_once_core_declares_the_flag` fails.
3. Confirm there is no regression to `core/read-settings`. With the plugin active, run:

```
wp eval 'rest_get_server(); global $wp_registered_settings; echo count( array_filter( $wp_registered_settings, fn( $a ) => ! empty( $a["show_in_abilities"] ) ) );'
```

It should print `18`, the size of the curated list.

There is also a tripwire test, `test_core_does_not_yet_declare_the_setting_flag`. It reads the defaults core really passes to the filter. It will fail on the day core starts shipping the argument, which is when someone should come back and remove this polyfill.

## Changelog Entry

> Fixed - Respect an explicit `show_in_abilities` value on curated settings, and leave the flag to WordPress core once core declares it.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-852-ca8d66c6f4b9ad98306525a5fa1881a3fcbf8bf7.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Show_In_Abilities.php
+++ b/includes/Abilities/Show_In_Abilities.php
@@ -46,12 +46,33 @@ final class Show_In_Abilities {
 	}
 
 	/**
+	 * Checks whether WordPress core declares the setting flag natively.
+	 *
+	 * `register_setting()` applies the `register_setting_args` filter before merging the
+	 * caller's arguments over its defaults, so the defaults array is core's own statement of
+	 * which arguments it understands. Once `show_in_abilities` appears there, core owns the
+	 * flag: it picks the default and each setting opts in through `register_setting()`, the
+	 * way `show_in_rest` already works.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param mixed $defaults The default registration arguments.
+	 * @return bool True when core declares `show_in_abilities` as a setting argument.
+	 */
+	private function core_declares_setting_flag( $defaults ): bool {
+		return is_array( $defaults ) && array_key_exists( 'show_in_abilities', $defaults );
+	}
+
+	/**
 	 * Adds the `show_in_abilities` flag to curated core settings as they are registered.
 	 *
-	 * Respects an explicit `show_in_abilities` value already present on the setting (for
-	 * example once core ships it natively), only filling it in when absent.
+	 * Respects an explicit `show_in_abilities` value already present in the registration
+	 * arguments, including an explicit `false` opt-out, only filling it in when the key is
+	 * absent entirely. Does nothing once core declares the flag natively, so the polyfill
+	 * never overrides a default core chose.
 	 *
 	 * @since 1.1.0
+	 * @since x.x.x Respects an explicit falsy value, and stands down once core declares the flag.
 	 *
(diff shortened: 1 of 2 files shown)


jorgefilipecosta (approved):
LGTM 👍

dkotter (approved):

PR #855, opened 2026-07-10
Markdown feeds experiment

### LINKED ISSUE #845: New Experiment: Markdown feeds (powered by `html-to-md`)

Markdown output can make content easier for AI tools, agents, search/indexing workflows, static exports, and reader experiences that prefer structured plaintext.

The prior implementation in #194 explored much of the product surface, including Markdown feed output, singular Markdown views, autodiscovery links, `Accept: text/markdown` support, settings toggles, filters, and tests.  This new experiment should revisit that direction with https://github.com/dmsnell/html-to-md as the conversion layer rather than maintaining a separate converter.

Create a new Markdown Feeds experiment that exposes content as Markdown, building on the prior work in #194 and using https://github.com/dmsnell/html-to-md for HTML-to-Markdown conversion.

Potential scope to consider:
- Add a new Markdown Feeds experiment, disabled by default
- Provide Markdown output for feed requests
- Support Markdown output for singular content
- Prefer `?format=md` as the primary URL strategy
- Keep `.md` URL support optional (likely via an experiment option that's hidden by default behind the work being done in #842), given prior concerns around WordPress permalink structure
- Consider `Accept: text/markdown` negotiation as an optional setting
- Use `dmsnell/html-to-md` for conversion
- Include filters for modifying or extending feed and singular Markdown sections
- Add tests covering feed output, singular output, visibility rules, URL handling, and conversion behavior

_No response_

- [x] Yes

dkotter:
## What?

Closes #845

Adds a new experiment, Markdown Feeds, that when enabled introduces new markdown feeds for singular content and main archive feeds.

## Why?

Having the ability to return content in markdown is useful for many situations, particularly for AI agents and machine readers.

## How?

- Introduce a new `Markdown_Feeds` experiment class that registers needed hooks
- Introduce new feed renderer classes, `Markdown_Singular_Renderer` and `Markdown_Feed_Renderer`, that will render markdown for singular feeds and normal feeds
- Introduce a `Markdown_Converter` class that provides a wrapper for converting HTML to Markdown
- Pull in the [HTML to MD](https://github.com/dmsnell/html-to-md) package to use for the actual conversion. Decided to pull this in to our custom `includes/Vendor` directory (alongside our encryption package) instead of including this at build time via composer. Ensure the file names match our autoloading process and add custom namespaces to avoid conflicts with other plugins using the same tool
- Adds a new setting that is off by default but when turned on tries to automatically serve the markdown feed if the request sends `Accept: text/markdown`
- An autodiscovery link tag for the Markdown feed is printed in the `<head>` tag as well
- Two filters added to change the contents of the feeds: `wpai_markdown_singular_sections` and `wpai_markdown_feed_item_sections`

To access a normal markdown feed, take the feed URL and append `/markdown` or `?feed=markdown`, depending on if pretty permalinks are in use. Examples: `https://example.com/feed/markdown`, `https://example.com/category/news/feed/markdown`

To access a singular markdown feed, access the URL and append `?output_format=markdown`, example: `https://example.com/this-is-an-article?output_format=markdown`.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Fable 5, Opus 4.8
Used for: Putting together a plan and executing on that plan. Plan reviewed and modified by me and all code was reviewed and tested by me

## Testing Instructions

1. Pull this PR down
2. Go to `Settings > AI` and enable Markdown Feeds
3. Ensure you have pretty permalinks on and have some content created
4. Go to your homepage and append `/feed/markdown`
5. Ensure a markdown feed is shown
6. Go to a single piece of content and append `?format=md`
7. Ensure a markdown feed is shown
8. Inspect the source of a post or page and ensure you see a `link` tag pointing to the markdown feed

## Screenshots or screencast

Markdown Feeds settings

Example of a singular markdown feed

## Changelog Entry

> Added - New experiment, Markdown Feeds, that creates markdown feeds at `/feed/markdown/` (available in every feed context) and of individual posts and pages via `?output_format=markdown`

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-855-90ba1d97b8316082ece482dae8511b4e04ddaf22.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter:
I've seen an increase in chatter recently around markdown content and how it's better to support this using content negotiation headers only and not support having separate URLs that serve markdown. This experiment adds both, with content negotiation being opt-in.

I had debated changing that logic here to either strip out the stand-alone markdown URLs or make that opt-in and have content negotiation be the default mode. But in doing some more research, seems most sites that have adopted markdown accessible content have taken both approaches (content negotiation and separate markdown URLs). The main one against this is [Vercel](https://vercel.com/blog/making-agent-friendly-pages-with-content-negotiation), which does content negotiation only. Also seems like it's [inconsistent among AI tools](https://acceptmarkdown.com/status), some send those headers, some do not.

The biggest issue, particularly in WordPress, with content negotiation headers is that depending on the cache setup, you can end up with [weird caching bugs](https://github.com/Automattic/wordpress-activitypub/issues/783). And some setups don't yet support the content negotiation headers. That's ultimately why we put that as opt-in.

So after thinking about it more and doing more research, I think the approach this PR takes is likely the right one for now, supporting separate markdown URLs by default and allowing those that want / those that know it won't interfere with caching, to opt-in to content negotiation.

That said, there were two things worth discussing:

1. Right now our single post markdown URL uses `?format=md`. WordPress.org uses `?output_format=md` (see https://wordpress.org/news/2026/08/mary-lou/?output_format=md). Should we standardize on `output_format` so we match?
2. Instead of using query params for feeds, we use pretty permalinks (`/feed/markdown/`). We can cut some code out if we switch to query params instead (probably matching `?output_format=markdown`). In addition to reducing code slightly, it would also match what we do at the single post level as described above

cc / @jeffpaul

jeffpaul:
On 1 and 2 above, my question would be "what do we expect RAG-based tools to look for first onf markdown feed formatting/access?" and aligning to that as our base case (even if that's NOT what WPORG does).

cc/ @justlevine @JasonTheAdams @dd32

jeffpaul:
Discussed in today's WP AI Contributor call, let's move forward with both the headers and a query param for output matching WPORG to keep things simple.

dkotter:
> Discussed in today's WP AI Contributor call, let's move forward with both the headers and a query param for output matching WPORG to keep things simple.

This has been updated so we now use `output_format=md` as the query param

PR #856, opened 2026-07-10
Register initial settings before core/read-settings snapshots them

jorgefilipecosta:
## What?

Follow up to WordPress/wordpress-develop#12141.

Ensure `core/read-settings` registers WordPress's initial settings before it snapshots the settings exposed through the Abilities API. Add regression coverage for an ability-first request where `rest_api_init` has not fired.

## Why?

The Abilities API registry initializes lazily on first use. Its `wp_abilities_api_init` hook is not ordered relative to `rest_api_init`, where WordPress registers its initial settings, and REST initialization may not run at all during cron or WP-CLI requests.

When `core/read-settings` registers first, it currently snapshots an empty settings registry. Because that snapshot supplies the input schema, output schema, and execution callback for the remainder of the request, the ability cannot expose normal Core settings such as `blogname`.

## How?

Before computing the exposed-settings snapshot, `Settings::register()` calls `register_initial_settings()` when `rest_api_init`:

- has not fired, or
- is currently firing and may not yet have reached Core's settings callback.

Once `rest_api_init` has completed, the guard does nothing. Re-registering settings later during the normal REST callback is harmless.

The regression test clears the registered-settings global and the `rest_api_init` action count, registers the ability under the previously broken ordering, and verifies that `blogname` is present in both the output schema and execution result.

This does not change the public ability name, schemas, filters, permissions, annotations, or exposed-settings list.

### Use of AI Tools

AI assistance: Yes  
Tool(s): Codex  
Model(s): GPT-5  
Used for: Plugin/Core parity auditing, implementation, regression testing, and verification. All changes and test results were reviewed.

## Testing Instructions

- `npm run build`
- `npm run test:php -- --filter SettingsTest` (10 tests, 35 assertions)
- `npm run test:php` (1010 tests, 2779 assertions, 34 skipped)
- `npm run lint:php`
- `vendor/bin/phpstan analyse --memory-limit=2G --debug`
- WP-CLI no-REST check:
  - `rest_api_init`: `0`
  - `core/read-settings` output schema contains `blogname`: `true`

The targeted browser E2E was not used for this regression because loading the REST/browser client initializes the path that the bug specifically bypasses.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-856-0c0e2c5c919f18cb36a142e4139354fec36137f1.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Settings/Settings.php
+++ b/includes/Abilities/Settings/Settings.php
@@ -22,6 +22,14 @@ defined( 'ABSPATH' ) || exit;
  * exposed. It is structured to also back a future write-oriented `core/manage-settings`
  * ability via the shared helpers (get_exposed_settings(), value_schema(), cast_value()).
  *
+ * The exposed settings are captured when the ability registers on `wp_abilities_api_init`.
+ * That hook fires lazily on first use of the abilities registry, which is not ordered
+ * relative to `rest_api_init` (where core registers its own settings) and can happen
+ * without it entirely, e.g. on cron or WP-CLI. register() therefore ensures core's
+ * initial settings are registered before the snapshot is computed. Other plugin settings
+ * flagged with `show_in_abilities` must be registered before the abilities registry is
+ * first used in a request; registering them on `init` is reliable.
+ *
  * This class is kept almost identical to the WordPress core class `WP_Settings_Abilities`
  * so the two implementations stay in sync. Differences from the core class are marked with
  * `// Plugin:` comments. Additionally, all user-facing strings use the 'ai' text domain.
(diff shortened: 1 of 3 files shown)


gziolo (approved):
**Approve** ✅

Solid fix. It solves a real problem — on cron or WP-CLI, `rest_api_init` never fires, so the ability was snapshotting an empty registry and couldn't expose things like `blogname`. Small, well-scoped change, and the regression test genuinely fails without it. Bonus: it keeps parity with core's `WP_Settings_Abilities`.

One trade-off worth naming: on cron/WP-CLI, using the abilities registry now registers all core settings globally for the rest of the request. That mirrors what is proposed for WP core, so it's fine — just a slightly broader effect than the snapshot alone.

Nothing blocking. Thanks for the clear write-up. 👍

dkotter (approved):

PR #861, opened 2026-07-13
Link Editorial Updates success notice to visual revisions

### LINKED ISSUE #507: Iterate on Editorial Updates end flow to Visual Revisions

When you get to the end of the Refine from Notes flow, a toast message provides a link to the classic revisions screen to see the changes.

Let's update that to link to the new Visual Revisions screen coming with WP 7.0 (as the AI plugin requires 7.0 as a minimum version) so that folks can benefit from that improve revisions review experience.

zeus2611:
## What?
Closes #507

Updates the "Review in Revisions" action on the Editorial Updates success snackbar to open WordPress 7.0's in-editor visual revisions UI instead of always linking to the classic `revision.php` screen.

## Why?
The AI plugin requires WP 7.0+, which ships with an improved in-editor visual revisions experience. The classic revisions screen is a worse review experience for folks applying AI-refined edits and should only be used as a fallback.

## How?
- Added `openVisualRevisions()`, which opens the Document Settings sidebar (if collapsed) and clicks Core's own rendered "Revisions" button (`PrivatePostLastRevision` in `@wordpress/editor`) to enter the visual revisions view.
- This isn't reachable via a public store action — `setCurrentRevisionId` is a private action gated behind `@wordpress/private-apis`, whose opt-in allow-list is hardcoded to WP core module names and throws for any other caller. Simulating a real click on Core's own button is the supported workaround; WordPress's own `@wordpress/e2e-test-utils-playwright` package does the same thing for the Settings sidebar toggle.
- Added `getRevisionReviewAction()` to branch between the visual path and the classic `revision.php` fallback, gated on the editor's `disableVisualRevisions` setting (true when classic metaboxes are active on the post, per WP core).
- `openVisualRevisions()` itself also falls back to the classic URL if the button can't be found within a short polling window, so the action never silently no-ops.

### Use of AI Tools
AI assistance: Implementation was planned with assistance from Claude Code (Anthropic). All code was reviewed, tested, and validated.

## Testing Instructions
1. Enable the **Editorial Updates** experiment.
2. Create/open a post, add a paragraph block, attach a pending Editorial Note to it.
3. Click **Apply Editorial Updates** in the Document sidebar.
4. After refinement completes, click **Review in Revisions** on the success snackbar.
5. Expected: the visual revisions diff opens in-editor (no page navigation).
6. To test the fallback: in the browser console, run `wp.data.dispatch('core/editor').updateEditorSettings({ disableVisualRevisions: true })`, then repeat steps 3–4. Expected: navigates to `wp-admin/revision.php?revision=<id>`.

## Screenshots or screencast
| Before | After |
|------|------|
| Screenshot 2026-07-13 at 7 58 50 PM | Screenshot 2026-07-13 at 7 57 33 PM |

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-861-b84408500186a839f3765911d8e079921bc11833.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


tests/Integration/Includes/Experiments/Editorial_Updates/Editorial_UpdatesTest.php
```diff
@@ -135,4 +135,23 @@ public function test_enqueue_assets_runs_without_error() {
...
 		$this->assertTrue( true, 'enqueue_assets() should run without throwing an exception' );
 	}
+
+	/**
+	 * Tests that enqueue_assets() localizes the expected data keys.
+	 *
+	 * admin_url is required for the classic revisions fallback when visual
+	 * revisions are disabled (e.g. when classic metaboxes are active).
+	 *
+	 * @since 1.1.0
```
dkotter:
```suggestion
	 * @since x.x.x
```

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
```diff
@@ -65,6 +65,136 @@ async function resolveNote( noteId: number ): Promise< void > {
...
+}: {
+	adminUrl: string | undefined;
+	lastRevisionId: number;
+} ): void {
+	const MAX_ATTEMPTS = 20;
+	const INTERVAL_MS = 100;
+
+	// The revisions button only exists in the DOM once the Document
+	// Settings sidebar is mounted, so open it first if it's collapsed.
+	const settingsToggle = findButtonByAccessibleName( 'Settings' );
```
dkotter: What about sites that aren't in English? Seems if we search for the literal string `Settings` this will never work for them

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
```diff
@@ -372,16 +502,17 @@ export function useEditorialUpdates(): {
...
-					lastRevisionId && adminUrl
-						? [
-								{
-									label: __( 'Review in Revisions', 'ai' ),
-									url: `${ adminUrl }revision.php?revision=${ lastRevisionId }`,
-								},
-						  ]
-						: [];
+				const disableVisualRevisions = !! (
+					select( editorStore ) as any
```
dkotter: Do we need the `as any` here? I believe we removed this from other places (though likely missed some) since the store is typed properly

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
```diff
@@ -65,6 +65,136 @@ async function resolveNote( noteId: number ): Promise< void > {
...
+	// The revisions button only exists in the DOM once the Document
+	// Settings sidebar is mounted, so open it first if it's collapsed.
+	const settingsToggle = findButtonByAccessibleName( 'Settings' );
+	if ( settingsToggle?.getAttribute( 'aria-expanded' ) === 'false' ) {
+		settingsToggle.click();
+	}
+
+	let attempts = 0;
+	const tryClick = () => {
+		const revisionsButton = document.querySelector< HTMLButtonElement >(
```
dkotter, line 134: To the comment above about looking at the name `Settings`, can we not do something similar that we're doing here where we use CSS selectors instead?

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
```diff
@@ -65,6 +65,136 @@ async function resolveNote( noteId: number ): Promise< void > {
 	} );
 }
 
+/**
+ * Finds a button-like element by its accessible name (aria-label or
+ * trimmed text content), optionally scoped to a container.
+ */
+function findButtonByAccessibleName(
```
dkotter: For any new functions we add here, should ensure we have proper `@param` and `@return`

tests/e2e/specs/experiments/editorial-updates.spec.js
```diff
@@ -220,6 +220,38 @@ test.describe( 'Editorial Updates Experiment', () => {
...
+				.getEditorSettings()?.disableVisualRevisions;
+			if ( disableVisualRevisions ) {
+				return (
+					typeof action.url === 'string' &&
+					action.url.includes( 'revision.php' )
+				);
+			}
+			return typeof action.onClick === 'function' && ! action.url;
+		} );
+		expect( hasRevisionAction ).toBe( true );
```
dkotter: So we ensure the link exists but doesn't seem we ever test that clicking on the link works as expected and loads the proper revisions view

dkotter (changes requested):

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
zeus2611: Switched to locating the sidebar toggle via its `aria-controls="edit-post:document"` attribute instead — that comes from the complementary area identifier Core registers for the Document Settings sidebar and is the same regardless of site language. Removed the accessible-name helper since it's now unused.

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
zeus2611: Removed it — `select(editorStore).savePost()` and `.getCurrentPostLastRevisionId()` are typed fine without a cast. `getEditorSettings()` still needed a narrow cast since `disableVisualRevisions` isn't in its public return type, so I scoped that cast to just the one property instead of casting the whole call.

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
zeus2611: Done, see reply on the "Settings" comment above — same fix covers both.

src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
zeus2611: Added full JSDoc to navigateToClassicRevisions, openVisualRevisions, and getRevisionReviewAction.

tests/e2e/specs/experiments/editorial-updates.spec.js
zeus2611: Reworked the assertion to click the real "Review in Revisions" button and wait for `.editor-revisions-header` to become visible — that's the container Core only renders while reviewing a revision, so this now proves the visual revisions view actually opens rather than just checking the action shape.

tests/Integration/Includes/Experiments/Editorial_Updates/Editorial_UpdatesTest.php
zeus2611: Updated to use the x.x.x placeholder.

dkotter (approved):

PR #862, opened 2026-07-13
Make image generation request timeout configurable

### LINKED ISSUE #853: Generate_Image Ability uses a hardcoded timeout value
During a testing session of the Image Generation experiment, I noticed that the Generate_Image Ability uses a hard coded timeout value: https://github.com/WordPress/ai/blob/cf1bf7191a0b35580ae3ed47637f2e90bcf009ce/includes/Abilities/Image/Generate_Image.php#L250

This means that if a developer wanted to extend the timeout for Image generation using the wp_ai_client_default_request_timeout [filter](https://github.com/WordPress/wordpress-develop/blob/b5ba245eddb49465699c4f2fbd403463e4e529ca/src/wp-includes/ai-client/class-wp-ai-client-prompt-builder.php#L202), it would not take affect.

I'd like to suggest that instead of hard coding the image generation time out, the AI plugin uses the filter to increase the time out for image generation to 90 seconds. That way the filter can still be used if needed.

ishitaj34:
## What?

Closes #853 

This PR makes the request timeout for the image generation ability configurable via the `wp_ai_client_default_request_timeout` filter, while retaining 90 seconds as the default timeout.

## Why?

The timeout was hardcoded to 90 seconds which completely bypassed the `wp_ai_client_default_request_timeout` global filter, meaning developers could not customise or extend the timeout for slower image generation requests.

## How?

- Replaced hardcoded `$request_options->setTimeout( 90 )` call with a filtered value: 

```php
$request_options->setTimeout( apply_filters( 'wp_ai_client_default_request_timeout', 90 ) )
```

- Added integration tests `test_request_timeout_defaults_to_90` and `test_request_timeout_can_be_filtered` to verify the defaults and overridden filter values.
- Created a helper subclass `Testable_Generate_Image` to bypass environment checks so the tests can run anywhere.

### Use of AI Tools

AI assistance: Yes
Tool(s): Gemini 3.5 Flash) 
Used for: Drafting integration test assertions and preparing validation commands.

## Testing Instructions

Run the automated tests:

```
npm run test:php -- --filter=Image_GenerationTest
```

## Changelog Entry

> Changed - Make image generation request timeout configurable via the `wp_ai_client_default_request_timeout` filter.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-862-e9d79e5a7f2124eb5db4bcf427fd9c2cef9e19b0.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Abilities/Image/Generate_Image.php
```diff
@@ -247,7 +247,8 @@ protected function generate_image( string $prompt, ?string $reference_image = nu
 	 */
 	private function get_prompt_builder( string $prompt, ?string $reference_image = null ) {
 		$request_options = new RequestOptions();
-		$request_options->setTimeout( 90 );
+		$timeout         = apply_filters( 'wpai_default_request_timeout', 90 );
+		$request_options->setTimeout( $timeout );
```
dkotter: Couple thoughts here:

- Right now this filter is only used for image generation yet the filter name, `wpai_default_request_timeout`, is generic. If we want to keep the filter here, I'd suggest we make that specific to the image generation feature
- That said, I could see benefit in using this to set the timeout for other features as well. In that case, I'd suggest adding a new helper function in `helpers.php` that you pass in the feature ID and the default timeout value and it then passes that through a filter (see `get_min_content_length` as an example). Other features could then use this to get a timeout value and 3rd parties can then filter the timeout based on the feature being used

tests/Integration/Includes/Abilities/Image_GenerationTest.php
```diff
@@ -380,4 +380,106 @@ public function test_meta_returns_expected_structure() {
 		$this->assertArrayHasKey( 'show_in_rest', $meta, 'Meta should have show_in_rest' );
 		$this->assertTrue( $meta['show_in_rest'], 'show_in_rest should be true' );
 	}
+
+	/**
+	 * Test that the request timeout defaults to 90 seconds.
+	 *
+	 * @since 0.8.1
```
dkotter: Why are these statements set to `0.8.1`? These should all be `x.x.x` and they'll be updated during release

tests/Integration/Includes/Abilities/Image_GenerationTest.php
```diff
@@ -380,4 +380,106 @@ public function test_meta_returns_expected_structure() {
...
+	 */
+	public function test_request_timeout_defaults_to_90() {
+		if ( ! function_exists( 'wp_ai_client_prompt' ) ) {
```
dkotter: Does this ever not exist here?

dkotter (changes requested):

includes/Abilities/Image/Generate_Image.php
ishitaj34: Thanks for the feedback! 
I went ahead with the helper function approach and added a `get_default_request_timeout( string $feature_id, int $default_timeout )` helper in `helpers.php`.

tests/Integration/Includes/Abilities/Image_GenerationTest.php
ishitaj34: Thanks for pointing that out. I wasn't aware of the release process using placeholders. I have updated all of the `@since` statements to use the `x.x.x` placeholder.

tests/Integration/Includes/Abilities/Image_GenerationTest.php
ishitaj34: Yeah, I had added it as a safeguard at first wrt test environments, but I've gone ahead and removed it since it's redundant.

dkotter (approved):

PR #867, opened 2026-07-14
Fix: Bug Inconsistency: Standardize post meta key naming with the `wpai_` prefix

### LINKED ISSUE #866: Bug Inconsistency: Standardize post meta key naming with the `wpai_` prefix
### Description

The plugin's options are almost universally namespaced with the `wpai_` prefix (`wpai_features_enabled`, `wpai_feature_*`, `wpai_version`, etc.), but the post meta keys are inconsistent:

| Current meta key | Prefix | Where registered |
|---|---|---|
| `wpai_meta_description` | ✅ `wpai_` | `SEO_Integration::FALLBACK_META_KEY` |
| `ai_generated` | ❌ `ai_` | Image Generation feature |
| `ai_generated_summary` | ❌ `ai_` | Summarization experiment |
| `ai_note` | ❌ `ai_` | Comment Generation experiment |

`wpai_meta_description` already follows the convention, so it can serve as the reference. The two `ai_`-prefixed keys (`ai_generated`, `ai_generated_summary`) should be renamed so all plugin-owned meta consistently uses the `wpai_` prefix.

This makes plugin data easier to identify, query, and clean up (e.g. the uninstall routine can then rely on a single `wpai_` prefix sweep).

### Proposed rename

| Old key | New key |
|---|---|
| `ai_generated` | `wpai_generated` |
| `ai_generated_summary` | `wpai_generated_summary` |
| `ai_note` | `wpai_note` |

(Exact new names are open for discussion — the goal is simply a consistent `wpai_` prefix.)

## Important: this is a breaking change and needs a data migration

Renaming a meta key changes:
1. The row stored in `wp_postmeta` (existing content keeps the **old** key).
2. The row stored in `wp_commentmeta` (existing content keeps the **old** key).
3. The REST API field name (since these are registered with `show_in_rest`), which the editor JS reads/writes.

So the rename must ship with a **migration** that copies existing values from the old key to the new key (and removes the old one) for all affected posts, comments meta, plus a lockstep update of every PHP and JS reference. Without a migration, existing AI-generated flags and saved summaries would be orphaned.

### 1. Meta registration
- **`includes/Features/Image_Generation/Image_Generation.php`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/includes/Features/Image_Generation/Image_Generation.php#L134-L144
- **`includes/Experiments/Summarization/Summarization.php`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/includes/Experiments/Summarization/Summarization.php#L89-L99
- **`includes/Experiments/Editorial_Notes/Editorial_Notes.php`** - 
https://github.com/WordPress/ai/blob/656fbde488b162ef2526e090e25b9bc887b567a2/includes/Experiments/Editorial_Notes/Editorial_Notes.php#L60-L63
https://github.com/WordPress/ai/blob/656fbde488b162ef2526e090e25b9bc887b567a2/includes/Experiments/Editorial_Notes/Editorial_Notes.php#L117-L119

### 2. PHP read/write of the meta
- **`includes/Abilities/Image/Import_Base64_Image.php`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/includes/Abilities/Image/Import_Base64_Image.php#L297-L299

### 3. Editor JavaScript / TypeScript (must change in lockstep with REST field)
- **`src/experiments/summarization/bulk.ts`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/src/experiments/summarization/bulk.ts#L165-L171
- **`src/experiments/summarization/functions/useSummaryGeneration.ts`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/src/experiments/summarization/functions/useSummaryGeneration.ts#L82-L87
- **`src/features/image-generation/components/AILabel.tsx`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/src/features/image-generation/components/AILabel.tsx#L51
- **`src/experiments/editorial-notes/hooks/useEditorialNotes.ts`** - 
https://github.com/WordPress/ai/blob/656fbde488b162ef2526e090e25b9bc887b567a2/src/experiments/editorial-notes/hooks/useEditorialNotes.ts#L477-L482

### 4. New migration
- **`includes/Admin/Upgrades/V_x_x_x.php`** (new) — extend `Abstract_Upgrade`, copy `ai_generated` → `wpai_generated` and `ai_generated_summary` → `wpai_generated_summary` for existing posts, then delete the old keys.
- **`includes/Admin/Upgrades.php`** (line ~51, `UPGRADE_CLASSES`) — register the new upgrade class (follows the existing `V0_5_0` / `V0_6_0` / `V1_0_0` pattern).

### 5. Uninstall cleanup
- **`includes/Admin/Uninstall.php`** (`delete_meta()`) — update the meta key patterns; after the rename, both keys fall under the `wpai_` prefix so the cleanup can be simplified accordingly.

### 6. Tests
- **`tests/Integration/Includes/Features/Image_Generation/Image_GenerationTest.php`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/tests/Integration/Includes/Features/Image_Generation/Image_GenerationTest.php#L131-L132
- **`tests/Integration/Includes/Abilities/Image_ImportTest.php`** - https://github.com/WordPress/ai/blob/8c3606a3f88b726237c5fd44beaba801b6c10d97/tests/Integration/Includes/Abilities/Image_ImportTest.php#L261
- **`tests/Integration/Includes/Experiments/Editorial_Notes/Editorial_NotesTest.php`** and **`wp-content/plugins/ai/tests/e2e/specs/experiments/editorial-updates.spec.js`**
- Add a test for the new migration (old key → new key, old key removed).

## Acceptance criteria
- [ ] `ai_generated`, `ai_generated_summary` and `ai_note` renamed to `wpai_`-prefixed keys across PHP and JS.
- [ ] A migration moves existing meta from old keys to new keys and removes the old keys, for all affected posts/attachments.
- [ ] REST + editor behavior (AI image label, generated summaries, bulk summarize) works unchanged after the rename.
- [ ] Uninstall cleanup and tests updated.

1. Enable AI Plugin.
2. Generate AI summary on any post/page with content.
3. Check the DB, and notice the meta stored for that post ID has key `ai_generated_summary`.
4. Generate the meta description.
5. Check for the same post ID and would have key `wpai_meta_description`.

- None

### Environment info

- WordPress 7.0.1.
- AI plugin develop branch.
- No extra plugin/theme activated.

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

### Usage of AI tool

- Claude Opus 4.8 to generate the description and PR title.
- Used to make it more polished and detailed.

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/866

## Why?

- PR fixes the inconsistency in meta keys for the post metas.
- PR also adds the migration/upgrade file to convert from old meta key to new meta key

## How?

- PR renames the meta key from old to new meta key.
- PR updates the references from old to new meta key.
- PR adds the upgrade file that supports the migration from old meta to new meta key.

### Use of AI Tools

- Yes, Claude Code.
- Opus 4.8
- Used for implementation of the issue based on details added by me on the issue. Implementation was then reviewed and tested locally by me.

## Testing Instructions

- Activate AI plugin.
- Generate the summary of any post/page.
- Check the DB for the meta key, it would be new meta key.
- For migration, Activate the plugin with the version mentioned, the old meta key should be now replaced with new meta key.

## Screenshots or screencast
- None

## Changelog Entry

> Changed - Updated all meta keys to use the `wpai` prefix instead of just `ai`. Note this changes the prefix on `ai_generated`, `ai_generated_summary` and `ai_note` meta so if you are directly using those, update to using the `wpai_` names.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-867-f23a7447e888b888b1f5990683ceb14f9aef29b5.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Admin/Upgrades/V1_3_0.php
```diff
@@ -0,0 +1,84 @@
...
+	 *
+	 * Migrates post and comment meta keys from the legacy `ai_` prefix to the
+	 * `wpai_` prefix.
+	 *
+	 * @since x.x.x
+	 */
+	protected function upgrade(): void {
+		$this->rename_post_meta_key( 'ai_generated', 'wpai_generated' );
+		$this->rename_post_meta_key( 'ai_generated_summary', 'wpai_generated_summary' );
+		$this->rename_comment_meta_key( 'ai_note', 'wpai_note' );
```
dkotter, line 54: Since we're using direct DB queries to update meta, may need a cache flush after this runs in order to ensure any cached values are flushed out. Likely not a problem as we're not changing data, just keys but may be better safe than sorry

includes/Admin/Upgrades/V1_3_0.php
```diff
@@ -0,0 +1,84 @@
...
+ * Upgrade routine for standardizing meta keys on the `wpai_` prefix.
+ *
+ * Renames the legacy `ai_generated` and `ai_generated_summary` post meta keys and
+ * the `ai_note` comment meta key to their `wpai_`-prefixed equivalents so all
+ * plugin-owned meta shares a consistent namespace.
+ *
+ * @since x.x.x
+ * @internal
+ */
+class V1_3_0 extends Abstract_Upgrade {
```
dkotter, line 26: This is likely to go in v1.3.0 but there's a possibility of there being a 1.2.1 version. I'm fine leaving this as 1.3.0 but may hold this PR until we're sure it's going to be part of v1.3.0

includes/Admin/Upgrades/V1_3_0.php
```diff
@@ -0,0 +1,84 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $old_key The existing meta key.
+	 * @param string $new_key The meta key to migrate to.
+	 */
+	private function rename_post_meta_key( string $old_key, string $new_key ): void {
+		global $wpdb;
+
+		$wpdb->update( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
dkotter: Do we need safeguards here to ensure the new meta key doesn't already exist before updating? That's a very unlikely scenario but if someone happens to update the plugin and generate new data before this migration fully completes, you could end up with two pieces of meta under the same key

dkotter (changes requested):
Left a few comments but overall this looks good. I do see there's still references to the old meta keys in some of our documentation so also need another sweep across the entire plugin to ensure all references are updated

includes/Admin/Upgrades/V1_3_0.php
hbhalodia: Sure, Will flush the cache to ensure everything works expected.

includes/Admin/Upgrades/V1_3_0.php
hbhalodia: Yes, that would be fine. I will update PR based on the milestone we finalize for this PR to land.

includes/Admin/Upgrades/V1_3_0.php
hbhalodia: Yeah, that's very unlikely but a possibility. I am not sure how often it would be, because Upgrade would run either on activation, or on current version is greater than what stored, and on admin_init hook, which would probably run first. Also the upgrade is not an background process or a cron job, so which would be very unlikely that it runs in background and user goes to post and generate the meta. 

While it may be something like if user cancels the process midway, but ultimately this happens on admin_init, so even if cancels midway, it would resume and since it would run in sequence the upgrade would complete first IMO.

Still We can update the query something as below to ensure we end up only having a single source of truth,

```SQL
UPDATE wp_postmeta AS pm
LEFT JOIN wp_postmeta AS existing
    ON existing.post_id = pm.post_id
    AND existing.meta_key = 'new_key'
SET pm.meta_key = 'new_key'
WHERE pm.meta_key = 'old_key'
    AND existing.meta_id IS NULL;
```

The above SQL would ensure that we update the meta IDs that does not have new key, After that we run the delete operation on the old key to remove the orphaned ones as below,

```SQL
DELETE FROM wp_postmeta
WHERE meta_key = 'old_key';
```

This would guard against the scenario explained. I am updating the PR with this, let me know if we need to revert and just use normal update from `old to new`.

Thanks,

hbhalodia:
> Left a few comments but overall this looks good. I do see there's still references to the old meta keys in some of our documentation so also need another sweep across the entire plugin to ensure all references are updated

This is now done in commit - https://github.com/WordPress/ai/pull/867/commits/2d53db7522ac4eb9ebce43a05e7530d0195a454b

Thanks,

includes/Admin/Upgrades/V1_3_0.php
dkotter: `wp_cache_flush` works and is likely fine since this is a one-time thing but does clear out the entire cache, whereas we only need to worry about the post and comment cache. We could modify things to track which post IDs and comment IDs were updated and then just clear those caches (potentially using `clean_post_cache/clean_comment_cache`). Any thoughts on that or is that not needed?

includes/Experiments/Editorial_Notes/Editorial_Notes.php
```diff
@@ -59,7 +59,7 @@ public function register(): void {
 
 		register_meta(
 			'comment',
-			'ai_note',
+			'wpai_note',
```
dkotter, line 62: Something I hadn't thought about until now, these meta keys we're renaming are set as `show_in_rest => true`. This means this data is exposed via REST and there could be others that are consuming/using that data. This rename would break their integrations.

I would guess there's a low (potentially zero) amount of users doing this but any thoughts on what is a breaking change here? At the very least, we'll need to call this out as a breaking change in our changelog but wondering if there's anything else we should consider here

hbhalodia: > I would guess there's a low (potentially zero) amount of users doing this but any thoughts on what is a breaking change here?

Yes, there could be breaking change, for eg; Users using/depend on Rest API for data would break, since `ai_generated` key would no longer be return in response, it would be `wpai_generated`. So user who access it as `meta.ai_generated` would get undefined, so it would silently break that part.

Another example could be user external writes request with body params as `{"meta": {"ai_generated": ""}}` would break, because rest does not knows about this key, so it would silently drops this key. So until noticed it won't be visible.

May be we have 2 pathways here,

1. Document it as a breaking change in newer version. Simplest but may have backward compatibility issue. But since amount of users would potentially be zero doing this would work.
2. Keep the meta key and register newer one side by side, drop the other one after few releases and add the deprecation notice or similar using the older keys could be an option.

IMO, we should go with the 1st, just document it as a breaking change in the release doc.

includes/Admin/Upgrades/V1_3_0.php
hbhalodia: I do not think, `clean_post_cache/clean_comment_cache` is needed, as `wp_cache_flush` would wipe every cache on the site. So IMO it's not needed as post/comment and every object cache would be flushed.

dkotter: Yes, `wp_cache_flush` will clear the cache we need here but also deletes all other caches. So on sites heavily reliant on cached data for performance, running `wp_cache_flush` can have a negative impact on the site. Likely won't be a problem here but we could be more specific with the cache flush functions to only clear what we need

includes/Experiments/Editorial_Notes/Editorial_Notes.php
dkotter: Yeah, I think I'm fine with option 1 here. But probably worth us discussing @jeffpaul prior to merging this in

includes/Admin/Upgrades/V1_3_0.php
hbhalodia: Thanks, also this is a one off thing, so we can go with wp_cache_flush here.

includes/Experiments/Editorial_Notes/Editorial_Notes.php
hbhalodia: Hi @dkotter @jeffpaul, Any updates on this? Let's get this updated before more users uses the `AI` plugin, so better to update these inconsistencies.

Thanks,

dkotter: I've not had a chance to discuss with Jeff but I'm still fine with option 1 here so let's move forward with that

Edit: just saw Jeff left a [comment](https://github.com/WordPress/ai/issues/866#issuecomment-5187147941) on the Issue. I'm fine with the renaming

hbhalodia: Thanks, @dkotter I think this would be added at the time of release right? So no any changes in PR.

Let me know if there is any other feedback on this?

dkotter (approved):

dkotter:
Note that E2E and Plugin Check tests aren't running right know due to WordPress 7.0.4 not being properly tagged on GitHub:

PR #881, opened 2026-07-17
Feature: New Experiment Abilities Toggle

### LINKED ISSUE #863: New Experiment: Abilities toggle

As we're already adding new abilities (e.g. #691, #774) and considering abilities outside of just `read` capabilities (e.g. #764), I would like to see us wrap these at least within a single "Ability APIs" sort of Admin Experiment.  Read-only abilities are one thing to have automatically exposed on sites, but ones that could be destructive feel like it should be purposely enabled by a site owner before those are available.

For now, let's at least wrap available Ability APIs from the AI plugin in an experiment so that they're purposely enabled for usage/testing.

There's a potential future state where we have so many abilities across CRUD operations that it would be not as easy to just "enable all abilities" sort of a settings toggle and perhaps then consider updating the Abilities Explorer to be a broader experiment where within that you can enable individual abilities or types of abilities or whatever.

- [x] Yes

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/863

## Why?

- Need to provide an option to hide the custom abilities behind the toggle experiment button.

## How?

- Added new Admin Custom experiment, Which gates the plugin core abilities behing a single opt-in toggle, so enabling it exposes all of them at once via the abilities.

### Use of AI Tools

- Yes, Cluade Code, Opus 5.
- Used for inital implementation related to gating abilities and adding a custom ability experiment.
- Optimization and structuring related changes.
- Implementation was reviewed by me.

## Testing Instructions

1. Checkout this branch and go to AI plugin settings.
2. Scroll to Admin Experiments --> Custom Abilities.
3. Before enabling check the number of abilities registered from `Abilities Explorer` page.
4. Enable the settings, check again.
5. You would see when toggled is on, the custom abilities are registered else they are not.

## Changelog Entry

> Added - New Experiment: Custom Abilities. Gates the plugin's custom WordPress Abilities (`ai/get-post-details`, `ai/get-post-terms`, `core/read-settings`, `core/read-users`, `core/read-content`) behind a single opt-in toggle, so enabling it exposes all of them at once via the Abilities API. Note for anyone that is using these Abilities, you'll need to enable this new experiment first for those to be available.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-881-ddd4f0125a406c6fde1cbd53fa24d73629170766.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

--- 

## AI Summary

Implements #863 — wraps the plugin's "always-on" custom WordPress Abilities behind a single opt-in experiment. Previously these abilities were registered unconditionally on every request and exposed via the Abilities API / MCP. Now a site owner enables the **Custom Abilities** experiment and gains access to all of them at once; when it's off, none are registered.

## Motivation

Per #863, ability APIs (some of which go beyond read-only operations) shouldn't be automatically exposed — they should require purposeful enablement. This gates the current custom abilities behind an experiment toggle, giving admins explicit control and establishing a scalable pattern for future abilities.

### New: `Custom Abilities` experiment
- A standard experiment (`custom-abilities`, Admin category) nested under the global **Enable AI** toggle. When enabled, it registers every gated ability together — no per-ability toggles.
- `Show_In_Abilities` (infrastructure, not a user-facing ability) runs once before registration when any enabled ability needs core objects exposed to the Abilities API.

### New: gated-ability framework (scalable)
- `Abstract_Gated_Ability` — minimal contract: `register()` + `requires_core_object_exposure()`.
- Four thin wrappers around the existing ability classes:
  - `Post_Utilities` → `ai/get-post-details`, `ai/get-post-terms`
  - `Read_Settings` → `core/read-settings`
  - `Read_Users` → `core/read-users`
  - `Read_Content` → `core/read-content`
- `Gated_Abilities` registry + `wpai_gated_abilities` filter for third-party extensibility.
- **Adding a new gated ability = one small class + one registry line** (or via the filter). No closures — each ability is independently unit-testable.

### Removed always-on registration
- `Main.php` no longer registers `Posts`, `Show_In_Abilities`, `Settings`, `Users`, and `Content` abilities directly; they're now owned by the experiment.

### Decoupling guard (no internal regressions)
- Extracted `Posts::get_post_details()` / `Posts::get_post_terms()` static methods; the abilities' `execute_callback`s delegate to them.
- `get_post_context()` now calls these static methods directly instead of via `wp_get_ability()`, so the six generation features that rely on post context (excerpt, title, meta-description, image-prompt, summarization, content-classification) keep working regardless of whether the Posts ability is toggled on. The toggle only controls external (REST/MCP) exposure.

## Files

**Added**
- `includes/Experiments/Custom_Abilities/Custom_Abilities.php`
- `includes/Abstracts/Abstract_Gated_Ability.php`
- `includes/Abilities/Gated/Gated_Abilities.php`
- `includes/Abilities/Gated/Post_Utilities.php`
- `includes/Abilities/Gated/Read_Settings.php`
- `includes/Abilities/Gated/Read_Users.php`
- `includes/Abilities/Gated/Read_Content.php`

**Modified**
- `includes/Main.php` — removed always-on ability registration + unused imports
- `includes/Experiments/Experiments.php` — registered `Custom_Abilities`
- `includes/helpers.php` — `get_post_context()` uses the new static methods
- `includes/Abilities/Utilities/Posts.php` — extracted static methods; callbacks delegate

## Behavior

- **Opt-in, no migration.** After update, these abilities are OFF until an admin enables *Enable AI → Custom Abilities*. This is intentional per #863 (explicit enablement).
- Internal plugin features are unaffected thanks to the `get_post_context()` decoupling.

## Notes for reviewers

- `@since x.x.x` docblock tags are placeholders to be stamped at release.
- No JS/build changes — this is PHP-only. The experiment renders as a standard toggle on the settings screen.
- Third parties can register their own gated abilities via the `wpai_gated_abilities` filter.


jeffpaul (dismissed):
Tests well via Playground, will await @dkotter's review before merging

includes/Experiments/Custom_Abilities/Custom_Abilities.php
```diff
@@ -0,0 +1,86 @@
...
+		return 'custom-abilities';
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function load_metadata(): array {
+		return array(
+			'label'       => __( 'Custom Abilities', 'ai' ),
+			'description' => __( 'Register the plugin\'s custom WordPress Abilities (post details & terms, settings, users, and content) for use via the Abilities API and MCP.', 'ai' ),
```
dkotter: I'd probably remove the specific abilities here as this list will certainly change and we'll have to remember to keep this up-to-date
```suggestion
			'description' => __( 'Register the plugin\'s custom WordPress Abilities for use via the Abilities API and MCP.', 'ai' ),
```

includes/Abilities/Gated/Gated_Abilities.php
```diff
@@ -0,0 +1,102 @@
...
+		 *
+		 * @since x.x.x
+		 *
+		 * @param array<class-string<\WordPress\AI\Abstracts\Abstract_Gated_Ability>> $classes Gated ability class names.
+		 */
+		$items = apply_filters( 'wpai_gated_abilities', self::GATED_ABILITY_CLASSES );
+
+		$abilities = array();
+		foreach ( array_unique( (array) $items ) as $item ) {
+			if ( ! is_string( $item ) ) {
```
dkotter: Because we pass `$items` through a filter (`wpai_gated_abilities`), someone else could filter this to return data we don't expect. Right now we run `array_unique` first and then check if each item is a string. But if someone uses the filter and returns a multidimensional array (as an example), I think this will throw an error (since `array_unique` casts entries to strings). Ideally we filter out any non-string items first and then do the unique check

includes/helpers.php
```diff
@@ -124,56 +125,52 @@ function count_characters_excluding_spaces( string $text ): int {
 function get_post_context( int $post_id ): array {
 	$context = array();
 
-	// Get the post details using the get-post-details ability.
-	$details_ability = wp_get_ability( 'ai/get-post-details' );
-	if ( $details_ability ) {
-		$details = $details_ability->execute( array( 'post_id' => $post_id ) );
+	// Get the post details directly (not via the ability) so the context is
+	// available even when the get-post-details ability is gated off.
+	$details = Posts::get_post_details( $post_id );
```
dkotter, line 137: One thing we should note here is that by previously running `WP_Ability::execute`, that would trigger the permission callback. Now anyone calling `get_post_context` directly won't run through that permission check. This is likely fine but should be noted somewhere to anyone using this knows they need to have their own permission check

includes/Abilities/Utilities/Posts.php
```diff
@@ -359,6 +236,153 @@ public function permission_callback( array $args ) {
...
+	 *
+	 * Shared by the `ai/get-post-details` ability and internal callers such as
+	 * get_post_context(), so the data remains available even when the ability
+	 * itself is gated off and not registered.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param int           $post_id The ID of the post to get the details of.
+	 * @param array<string> $fields  The fields to return. Defaults to all supported fields.
+	 * @return array<string, string>|\WP_Error The post details, or WP_Error if the post doesn't exist.
```
dkotter: Might be worth changing this to `mixed` since we pass results through a filter and thus can't be 100% sure others will use that filter properly

dkotter (changes requested):
Left some comments that need looked at, in addition to a few other things:

1. We should look to add documentation here matching our other experiments
2. It doesn't appear we have E2E tests specific to this new experiment, ensuring the experiment can be turned on/off and abilities are available/not available depending on that state
3. Technically this is a breaking change for any site that is relying on these Abilities. Once they upgrade, they'll need to enable this new experiment before those Abilities are available. That's likely fine but will need to ensure that is called out when we put together the release (cc @jeffpaul). Could also make this experiment enabled by default though that goes against how everything else is set up right now

hbhalodia:
Thanks for the review @dkotter, I will update the PR to address those.

includes/Experiments/Custom_Abilities/Custom_Abilities.php
hbhalodia: Fixed in - 75d86d9f4f87ff6620431a074a614bdbfa77c1b7

includes/helpers.php
hbhalodia: Thanks, Fixed in - 75d86d9f4f87ff6620431a074a614bdbfa77c1b7

includes/Abilities/Gated/Gated_Abilities.php
hbhalodia: Thanks, Fixed in commit - 75d86d9f4f87ff6620431a074a614bdbfa77c1b7

includes/Abilities/Utilities/Posts.php
hbhalodia: Thanks, Address this for both terms and posts. Fixed in - 75d86d9f4f87ff6620431a074a614bdbfa77c1b7

hbhalodia:
> Left some comments that need looked at, in addition to a few other things:
> 
> 1. We should look to add documentation here matching our other experiments

Added the documentation in the commit - 6e4af353bb2c0f39bd487e958689fef9c566abae
Have created the write up with AI who was having the context regarding the change. Have reviewed myself, let me know if anything needs to be added or re-written. Used the context from other experiments.

> 2. It doesn't appear we have E2E tests specific to this new experiment, ensuring the experiment can be turned on/off and abilities are available/not available depending on that state

Added the E2E tests for the experiment in commit - d37e0795140ef2d1c0d9634148e9804219cc1feb


> 3. Technically this is a breaking change for any site that is relying on these Abilities. Once they upgrade, they'll need to enable this new experiment before those Abilities are available. That's likely fine but will need to ensure that is called out when we put together the release (cc @jeffpaul). Could also make this experiment enabled by default though that goes against how everything else is set up right now

I guess it would be added while we do the release, so have not added anything here.

Thanks,

README.md
dkotter, line 1: Will want to add this same thing to the `readme.txt`

docs/experiments/custom-abilities.md
```diff
@@ -0,0 +1,96 @@
...
+2. **Verify registration:**
+   - Open the Abilities Explorer (or query the REST `abilities` endpoint) and confirm the plugin's gated abilities are listed.
+3. **Verify gating:**
+   - Disable the experiment and confirm those abilities are no longer registered.
+4. **Verify extensibility:**
+   - Add a class via the `wpai_gated_abilities` filter and confirm it registers when the experiment is enabled.
+
+### Automated Testing
+
+- `tests/Integration/Includes/Experiments/Custom_Abilities/Custom_AbilitiesTest.php`
```
dkotter: I've been removing this documentation section from new docs as it doesn't really provide any value

dkotter (changes requested):
Couple more minor things I've noted.

And I don't know if related to this or just a general bug in our plugin but I'm finding that if you toggle the experiment on and off a few times, we end up losing credentials, which is really weird. May not be related to this PR but figured I'd flag it:

https://github.com/user-attachments/assets/64263a80-3732-4b44-ac1c-f1d58e0da8a1

jeffpaul:
Let's not enable these abilities by default, but instead call out in the changelogs and with an upgrade notice that they'll need to be enabled for anyone making use of abilities previously.

dkotter:
> And I don't know if related to this or just a general bug in our plugin but I'm finding that if you toggle the experiment on and off a few times, we end up losing credentials, which is really weird. May not be related to this PR but figured I'd flag it

Quick followup to this bug. I have verified the same behavior on `develop` so this is not related to this PR.

In diving in, seems this is an upstream issue within WordPress Core. Any request to `/wp/v2/settings` will trigger core to re-validate all Connector API keys, even if you aren't specifically saving an API key. So in our case, we make a request to that endpoint when we save our plugin settings.

If anything goes wrong in that validation (in my case, seems like occasionally requests to the OpenAI models endpoint is timing out at 5 seconds), Core sets the API key to a blank string.

We can look to fix on our end by introducing a custom endpoint to handle saving our settings, bypassing the `/wp/v2/settings` all together. But also seems ideal to fix this upstream to avoid having Core re-validate credentials on every request, even if credentials aren't being passed in.

docs/experiments/custom-abilities.md
dkotter: Done in https://github.com/WordPress/ai/pull/881/changes/ddd4f0125a406c6fde1cbd53fa24d73629170766

README.md
dkotter: Done in https://github.com/WordPress/ai/pull/881/changes/ddd4f0125a406c6fde1cbd53fa24d73629170766

dkotter (approved):

dkotter:
Note that E2E and Plugin Check tests aren't running right know due to WordPress 7.0.4 not being properly tagged on GitHub:

README.md
hbhalodia: Thanks, had missed it, was on my plate but somehow missed.

hbhalodia:
> Couple more minor things I've noted.
> 
> And I don't know if related to this or just a general bug in our plugin but I'm finding that if you toggle the experiment on and off a few times, we end up losing credentials, which is really weird. May not be related to this PR but figured I'd flag it:
> 
>  missing-credentials.mov

Thanks @dkotter, I noticed that too as well in earlier days while I was contributing to AI plugin, but it happen so randomly that I was not able to reproduce with the exact steps again and again, hence have'nt logged the ticket 😅. 

Will create the ticket on core or either GB and will add the reproduction steps.

Thanks for exact reproduction steps.

Update: Raised the [trac ticket](https://core.trac.wordpress.org/ticket/65867) and [PR](https://github.com/WordPress/wordpress-develop/pull/13031).

Cc: @jeffpaul

PR #884, opened 2026-07-18
Abilities Explorer: support custom providers in the filter dropdown and statistics

### LINKED ISSUE #883: Abilities Explorer: provider filter dropdown doesn't include custom providers

It is currently difficult to filter the Abilities Explorer by a custom provider.

`Ability_Handler::detect_provider()` already supports custom providers via an ability's `meta['provider']`, and `get_provider_label()` deliberately passes unknown provider names through — so a plugin can label its abilities with its own provider name: the Provider column shows it, and the `?provider=` query arg filters correctly. But the provider `<select>` in `Ability_Table::extra_tablenav()` is hardcoded to all/Core/Plugins/Theme, so custom providers can never be selected in the UI.

**Proposal:** build the provider options dynamically from the fetched ability list, mirroring the dynamic category filter added in #344 / #355 (`get_unique_categories()`): known origins first in canonical order, custom providers appended alphabetically. Sites without custom providers would see an unchanged dropdown.

Possible follow-up (out of scope here): the stat cards on the Explorer overview also hardcode Core/Plugins/Theme counts, so abilities with a custom provider currently leave the "Plugins" bucket.

I have a small patch ready and will open a PR referencing this issue.

Build the provider `<select>` options dynamically from the already-fetched ability list, exactly mirroring the dynamic category filter (`get_unique_categories()` from #344 / #355):

- Add `Ability_Table::get_unique_providers()` — known origins (Core, Plugin, Theme) first in canonical order, then any custom providers (from `meta['provider']`) appended alphabetically.
- Replace the three hardcoded `<option>` tags in `extra_tablenav()` with a loop over it, labelling options via the existing `Ability_Handler::get_provider_label()` (which also makes the dropdown label consistent with the Provider column).

Sites without custom providers see an unchanged dropdown. Filtering already works for custom providers via the existing `?provider=` handling — only the options list is missing. PR incoming referencing this issue.

We ship a plugin that registers 80+ abilities via the Abilities API. Users browsing the Explorer on a site with 112 abilities have no way to see just ours (or just any one plugin's) — the only workaround is hand-crafting the ?provider= URL, which nobody discovers. Any plugin registering multiple abilities benefits the same way.

- [x] Yes

azizulhasan:
## What?

Closes #883

Makes the Abilities Explorer treat custom providers (set via an ability's `meta['provider']`) as first-class citizens:

1. The provider filter dropdown is built dynamically, so custom providers become selectable filter options.
2. The overview statistics count abilities by their **origin** (Core/Plugin/Theme), so an ability with a custom provider label still counts in its origin bucket instead of disappearing from the stat cards.

## Why?

`Ability_Handler::detect_provider()` already supports custom providers via `meta['provider']`, and `get_provider_label()` deliberately passes unknown provider names through — the Provider column shows them and the `?provider=` query arg filters correctly. But the provider `<select>` in `Ability_Table::extra_tablenav()` was hardcoded to all/Core/Plugins/Theme, so custom providers could never be selected in the UI. See #883.

The stat cards had the mirrored problem: `get_statistics()` only counted the three hardcoded buckets, so any ability with a custom provider label silently vanished from every card (only Total stayed correct). Rendering a card per custom provider would not scale — a site with ten such plugins would get ten extra cards — so the cards stay Core/Plugins/Theme and the counting becomes origin-based instead.

## How?

Follows the same approach as the dynamic category filter added in #355 (`get_unique_categories()`):

- New `Ability_Table::get_unique_providers()` — known origins (Core, Plugin, Theme) first in canonical order, custom providers appended alphabetically.
- The three hardcoded `<option>` tags become a loop, labelled via the existing `Ability_Handler::get_provider_label()`. This also makes the dropdown label ("Plugin") consistent with the Provider column, which used the singular while the old dropdown said "Plugins".
- New `Ability_Handler::detect_origin()` holds the prefix-based Core/Plugin/Theme detection; `detect_provider()` returns `meta['provider']` when set and falls back to the origin. Formatted abilities expose a new `origin` field and `get_statistics()` counts by it — the `by_provider` array keys are unchanged, so `Admin_Page` needs no changes.
- Integration tests: new `Ability_TableTest` covers `get_unique_providers()` with and without a custom provider (this also covers the line the Codecov report flagged), and a new `Ability_HandlerTest` case asserts a custom-provider ability still increments its origin bucket in the statistics.

Sites without custom providers see no change anywhere (same four dropdown options, same counts). The accessible label added in #642 is preserved.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Fable 5
Used for: drafting the implementation and this PR description under my direction; I reviewed the code and tested it on a live site before submitting.

## Testing Instructions

1. On a site with the plugin active, visit Tools → Abilities Explorer: the provider dropdown shows all/Core/Plugin/Theme and the stat cards show the same counts as before.
2. Register an ability whose meta includes `'provider' => 'My Plugin'` (e.g. via `wp_register_ability()`).
3. Reload the Explorer: the dropdown now contains a "My Plugin" option; selecting it and clicking Filter lists exactly the abilities with that provider.
4. The stat cards are unchanged by the custom label: the ability still counts under Plugins (previously it dropped out of all three cards).

Tested on WordPress 7.1-beta1 with 112 registered abilities, 82 of them carrying a custom provider: the dropdown gains the custom option and filters to exactly those 82, and the stat cards read Core 6 / Plugins 106 / Theme 0 — identical to the counts before the custom labels were applied.

## Changelog Entry

> Changed - The Abilities Explorer provider filter dropdown is now built dynamically and includes custom providers, and the overview statistics count abilities by origin so custom-provider abilities remain in their Core/Plugins/Theme bucket.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-884-40d0a75a105645dc36915a9c0d7236752307e0fa.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/Abilities_Explorer/Ability_Handler.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Handler.php
@@ -95,6 +95,7 @@ class Ability_Handler {
 			'name'          => $ability->get_label(),
 			'description'   => $ability->get_description(),
 			'provider'      => self::detect_provider( $name, $meta ),
+			'origin'        => self::detect_origin( $name ),
 			'category'      => self::get_ability_category( $ability ),
 			'input_schema'  => $ability->get_input_schema(),
 			'output_schema' => $ability->get_output_schema(),
@@ -193,6 +194,22 @@ class Ability_Handler {
 			return $meta['provider'];
 		}
 
+		return self::detect_origin( $name );
+	}
+
+	/**
+	 * Detects the origin (Core, Plugin, or Theme) of an ability from its name.
+	 *
+	 * Unlike the provider, which can be overridden with a custom label via the
+	 * ability's `meta['provider']`, the origin always resolves to one of the
+	 * three known buckets, so it is suitable for aggregate statistics.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $name Ability name (slug).
+	 * @return string Origin type: 'Core', 'Plugin', or 'Theme'.
+	 */
+	private static function detect_origin( string $name ): string {
 		// Detect based on name prefix (namespace/ability format).
 		$parts = explode( '/', $name );
 		if ( count( $parts ) === 2 ) {
@@ -415,16 +432,13 @@ class Ability_Handler {
 		);
 
 		foreach ( $abilities as $ability ) {
-			// Count by provider.
-			if ( ! isset( $ability['provider'] ) ) {
+			// Count by origin so abilities with a custom provider label still
+			// land in their Core/Plugin/Theme bucket.
+			if ( ! isset( $ability['origin'], $stats['by_provider'][ $ability['origin'] ] ) ) {
 				continue;
 			}
 
-			if ( ! isset( $stats['by_provider'][ $ability['provider'] ] ) ) {
-				continue;
-			}
-
-			++$stats['by_provider'][ $ability['provider'] ];
+			++$stats['by_provider'][ $ability['origin'] ];
 		}
 
 		return $stats;
--- a/includes/Experiments/Abilities_Explorer/Ability_Table.php
+++ b/includes/Experiments/Abilities_Explorer/Ability_Table.php
@@ -110,9 +110,17 @@ class Ability_Table extends \WP_List_Table {
 		// Apply provider filter.
 		$provider_filter = isset( $_REQUEST['provider'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['provider'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		if ( ! empty( $provider_filter ) && 'all' !== $provider_filter ) {
-			$abilities = array_filter(
+			// Known origins filter by origin, so abilities carrying a custom
+			// provider label still appear under their Core/Plugin/Theme bucket
+			// (consistent with the statistics); custom providers filter by
+			// their exact label.
+			$known_origins = array( 'Core', 'Plugin', 'Theme' );
+			$abilities     = array_filter(
 				$abilities,
-				static function ( $ability ) use ( $provider_filter ) {
+				static function ( $ability ) use ( $provider_filter, $known_origins ) {
+					if ( in_array( $provider_filter, $known_origins, true ) ) {
+						return ( $ability['origin'] ?? '' ) === $provider_filter;
+					}
 					return $ability['provider'] === $provider_filter;
 				}
 			);
@@ -265,7 +273,7 @@ class Ability_Table extends \WP_List_Table {
 	 */
 	public function column_provider( $item ): string {
 		$provider = $item['provider'];
-		$class    = 'ability-provider ability-provider-' . strtolower( $provider );
+		$class    = 'ability-provider ability-provider-' . sanitize_title( $provider );
 
 		return sprintf(
 			'<span class="%s">%s</span>',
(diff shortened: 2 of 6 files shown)


dkotter (changes requested):
This PR works as described though there are a few pre-existing issues that we should look to fix here I think:

1. We don't properly sanitize the provider name when we use it as a class. So if I set a custom provider as `WordPress AI`, the class we add ends up being `ability-provider-wordpress ai` when it should instead be `ability-provider-wordpress-ai`
2. We don't properly style the custom provider pill:

Image

Image

And just as a personal opinion, not sure what the benefit is of this custom `provider` meta. I know that's a pre-existing thing but could be worth a discussion on maybe just removing that.

Seems odd to me to have that show in the Providers dropdown and show as a Provider pill but doesn't actually show in the top statistics. And then it's weird to have custom providers bucketed in the `Plugins` section of the statistics but if I filter by `Plugin`, it doesn't show. Just various inconsistencies like those that may be worth fixing or just removing the support all together.

azizulhasan:
Thanks for the review @dkotter — all three points are addressed in 40d0a75:

1. **Sanitization** — the pill class is now built with `sanitize_title()`, so a provider of `WordPress AI` produces `ability-provider-wordpress-ai`. Fixed in both `Ability_Table::column_provider()` and the detail view in `Admin_Page.php`, which had the same issue.

2. **Custom pill styling** — the neutral pill styles (background/border/color that `-plugin`/`-theme` used) moved onto the base `.ability-provider` class, so any custom provider now renders as a proper pill; `-core` still overrides with its blue variant. Since custom provider names are arbitrary, styling the base class is the only approach that covers every possible label.

3. **Consistency** — the known-origin filters (Core/Plugin/Theme) now match by *origin*, so filtering by "Plugin" includes abilities that carry a custom provider label — consistent with the statistics, which count by origin. Filtering by a custom provider matches its exact label. New integration test `test_provider_filter_matches_origin_for_known_providers` covers both directions.

On keeping vs. removing the support: this PR now cleanly splits the two concepts that were previously conflated in one field — *origin* (Core/Plugin/Theme, derived from the ability name) drives the stat buckets and the known filters, while the `meta['provider']` *label* only affects the pill text and adds a filter option. That's what caused the inconsistencies you listed, and they shouldn't be reproducible on this branch: on a test install with ~82 custom-labelled abilities, filtering by "Plugin" now returns 106 items (matching the "Plugins" stat card exactly), and filtering by the custom label returns only those 82.

The value I see in the label is attribution — on a busy site the Explorer can list 100+ abilities, and "which plugin exposed this capability to the AI" is a real trust/debugging question that a wall of identical "Plugin" pills can't answer. `get_provider_label()` has passed custom labels through since 0.4.0, so plugins may already rely on it. That said, it's your call — if you'd prefer to drop the label pass-through entirely rather than support it, I'm happy to rework the PR in that direction instead.

dkotter (approved):

PR #886, opened 2026-07-20
Fix: Meta Description: Issue with Yoast plugin for Meta Description experiment

### LINKED ISSUE #874: Meta Description: Issue with Yoast plugin for Meta Description experiment
### Description

Initially reported here: https://wordpress.org/support/topic/generate-meta-description/

When the Yoast SEO plugin is active, the WordPress AI plugin uses the `_yoast_wpseo_metadesc` meta key to store the meta description — the same key Yoast SEO uses for its meta description.

There are two issues here, and both appear to originate from the Yoast plugin itself.

**1. The issue reported in the [support topic above](https://wordpress.org/support/topic/generate-meta-description/).**

- The meta description works for the `post` post type but does not work for `page` or any other custom post type (CPT).
- This appears to be related to Yoast, since `post` and `page` share the same meta key and are registered in the same way. So if it works for the `post` post type, it should work for pages by default as well.

**2. The meta description stops working even for posts when the `Yoast AI` feature is disabled.**

- While testing scenario (1), I first installed and activated the plugin, then found a setting called `Yoast AI`. I disabled it because I did not want it.
- After disabling it and re-checking scenario (1), I found that the meta description for posts was no longer working. Following extensive debugging, I re-enabled the `Yoast AI` feature, and scenario (1) became reproducible again — the meta description started working for the `post` post type once more.
- This may be because enabling Yoast AI adds a **Generate Description with AI** button. That button appears to populate the Yoast meta description box via a dispatch, and those values are then used on save. This would explain why the meta description works when Yoast AI is enabled but not when it is disabled.
- Please see the screencast below.

- Please see the Screencast for reference.

https://github.com/user-attachments/assets/60e31452-b121-40aa-9959-5166bcba21c7

### Environment info

- Yoast latest Plugin.
- WordPress AI develop branch.
- No any other 3rd Party plugin.

- [x] Yes

- [ ] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

**Note:** Usage of AI

- To check through Yoast Plugin and it's pathways on how plugin stores meta description and finding the exact refrences to be added.
- Claude Code using Opus 4.8.

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/874

## Why?

- PR resolves the issue where, Yoast meta description is only saved for posts and no other post types.
- It also resolves the issue where, `Yoast AI` is disable, still meta description is saved via yoast.
- Read more on issue mentioned.

## How?

- Adds the `seo-adapter.ts` file, which register the adapter and use the required SEO adapters to work with.
- If the Yoast plugin is active, the adapter uses Yoast method and way to store/fetch the meta description, else fallback to default what's being performed.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 4.8
Used for: Initial code implementation and final implementation and testing was done by me.

## Testing Instructions

1. Open both post and page.
2. Make sure to activate yoast plugin.
3. Use Meta Description experiment to generated meta description.
4. See both post and page stores the meta description correctly to Yoast meta and our custom meta description box.

## Screenshots or screencast

https://github.com/user-attachments/assets/8be352b0-e95c-4ee2-99d4-5ed4b1e140c3

## Changelog Entry

> Fixed - Meta Description suggestions applied on pages and custom post types were lost on save when Yoast SEO was active.

---

# AI Summary

## What?

Routes the Meta Description **Apply** action through the active SEO plugin's own editor store instead of always writing to `core/editor` post meta. Introduces a small, slug-keyed adapter layer and ships a **Yoast** adapter, so an applied AI meta description updates the SEO plugin UI and persists on **all post types** (posts, pages, and CPTs).

## Why?

`useMetaDescription.applyDescription` wrote the suggestion via `editPost( { meta: { [ metaKey ]: text } } )` (the `core/editor` store). That only works when the meta key is REST-registered for the current post type:

- **No SEO plugin:** the plugin registers its fallback key for all REST post types, so this already works everywhere.
- **Yoast active:** Yoast exposes `_yoast_wpseo_metadesc` to REST **only for the `post` post type** (`register_meta( … 'object_subtype' => 'post' )`). On a **page or CPT** the key isn't REST-registered, so `editPost` is dropped on save and the Yoast snippet field (bound to the `yoast-seo/editor` store) never reflects it — the applied description is lost. It also only appeared to work on posts when Yoast's own AI feature happened to be enabled, because that feature mounts the hook that mirrors `core/editor` meta into the Yoast store.

The fix targets the SEO plugin's own store, which is the source of truth for its UI and its all-post-type save path.

## How?

- **New `src/experiments/meta-description/seo-adapters.ts`** — a `SeoEditorAdapter` abstraction (`apply` / `read`) selected by the active plugin slug (`window.aiMetaDescriptionData.seoPlugin`):
  - `defaultAdapter` — unchanged behaviour: `editPost` on the meta key; reads from `core/editor` meta. Used for the fallback key and any not-yet-adapted SEO plugin.
  - `yoastAdapter` — `apply` dispatches `updateData( { description } )` to `yoast-seo/editor` (drives the UI + Yoast's metabox save on every post type) **and** calls `editPost` to mark the editor dirty so the Save button enables (and to persist via REST on `post`). `read` uses `getSnippetEditorDescription()` with a fallback to core meta if the Yoast store isn't ready.
  - `getAdapter()` with an `ai.metaDescription.seoAdapters` JS filter, mirroring the existing PHP `wpai_meta_description_seo_plugins` filter for extensibility.
- **`components/useMetaDescription.ts`** — selects the adapter by slug, routes `applyDescription` through `adapter.apply()`, and derives `currentDescription` from `adapter.read()` (fixing the display / "already set" check on Yoast pages, where `core/editor` meta doesn't carry the key).
- **PHP unchanged** — `SEO_Integration` remains the single source of truth for detection and the meta key. Which plugin is active and its key stay server-side; *how* to write to the editor store is client-side, per-plugin behaviour.

**Scope:** ships the Yoast adapter only. Rank Math / All in One SEO / SEOPress continue to use the default (`editPost`) adapter until each is verified — dedicated adapters can be added incrementally (AIOSEO in particular uses a non-`@wordpress/data` store).

## Testing Instructions

Preconditions: an AI connector with a text-generation model configured, and the Meta Description experiment enabled.

1. **Yoast active — Page (the bug):** Activate Yoast SEO. Edit a **Page** with ≥250 characters of content. Open the **Meta Description** panel → **Generate** → **Apply**.
   - The Yoast "Meta description" field/snippet preview updates to the applied text.
   - The Save/Update button becomes enabled even with no other edit.
   - **Save → reload:** the value persists; the front end outputs `<meta name="description" …>` with the text.
2. **Yoast active — Post:** Repeat step 1 on a **Post**; confirm it still works.
3. **Yoast active — CPT:** Repeat on a public custom post type that shows the Yoast metabox.
4. **Yoast AI feature on/off:** Repeat step 1 with Yoast's own AI feature toggled both ways — behaviour should now be identical.
5. **No SEO plugin (regression):** Deactivate all SEO plugins. Generate → Apply on a Post and a Page; confirm the value persists in the fallback meta and renders in the head (default path unchanged).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-886-bfba0be3cc66bc483224a1aba92803e763809d92.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/meta-description/components/useMetaDescription.ts
+++ b/src/experiments/meta-description/components/useMetaDescription.ts
@@ -7,7 +7,7 @@
  */
 import { dispatch, useDispatch, useSelect } from '@wordpress/data';
 import { store as editorStore } from '@wordpress/editor';
-import { useState, useCallback } from '@wordpress/element';
+import { useState, useCallback, useMemo } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
 import { store as noticesStore } from '@wordpress/notices';
 
@@ -17,6 +17,7 @@ import { store as noticesStore } from '@wordpress/notices';
 import { runAbility } from '../../../utils/run-ability';
 import { ensureProvider } from '../../../utils/provider-status';
 import { hasMinimumContent } from '../../../utils/character-count';
+import { getAdapter } from '../seo-adapters';
 import type {
 	MetaDescriptionAbilityInput,
 	MetaDescriptionAbilityResponse,
@@ -52,7 +53,12 @@ interface UseMetaDescriptionReturn {
 export function useMetaDescription(): UseMetaDescriptionReturn {
 	const localized = getLocalized();
 	const metaKey = localized?.metaKey ?? 'wpai_meta_description';
-	const hasSeoPlugin = Boolean( localized?.seoPlugin );
+	const seoPlugin = localized?.seoPlugin ?? null;
+	const hasSeoPlugin = Boolean( seoPlugin );
+
+	// The active SEO plugin decides how the description is written to and read from the
+	// editor. The default adapter uses core/editor post meta; Yoast targets its own store.
+	const adapter = useMemo( () => getAdapter( seoPlugin ), [ seoPlugin ] );
 
 	const { editPost } = useDispatch( editorStore );
 	const { removeNotice, createErrorNotice } = dispatch( noticesStore );
@@ -66,19 +72,23 @@ export function useMetaDescription(): UseMetaDescriptionReturn {
 		[]
 	);
 
-	const { postId, content, title, meta } = useSelect( ( select ) => {
-		const editor = select( editorStore );
-		const currentMeta = editor.getEditedPostAttribute( 'meta' ) as
-			| Record< string, string >
-			| undefined;
+	const { postId, content, title, meta, currentDescription } = useSelect(
+		( select ) => {
+			const editor = select( editorStore );
+			const currentMeta = editor.getEditedPostAttribute( 'meta' ) as
+				| Record< string, string >
+				| undefined;
 
-		return {
-			postId: editor.getCurrentPostId() as number,
-			content: editor.getEditedPostContent(),
-			title: editor.getEditedPostAttribute( 'title' ) as string,
-			meta: currentMeta,
-		};
-	}, [] );
+			return {
+				postId: editor.getCurrentPostId() as number,
+				content: editor.getEditedPostContent(),
+				title: editor.getEditedPostAttribute( 'title' ) as string,
+				meta: currentMeta,
+				currentDescription: adapter.read( select, { metaKey } ),
+			};
+		},
+		[ adapter, metaKey ]
+	);
 
 	const minContentLength =
 		localized?.minContentLength ?? MINIMUM_CONTENT_COUNT_DEFAULT;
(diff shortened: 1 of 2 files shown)


dkotter (approved):

dkotter:
Note we'll want to credit @carlos-widgilabs on this as well, which wasn't captured by prop bot cc / @jeffpaul

PR #887, opened 2026-07-21
Experiment: Add Internal link suggestions

### LINKED ISSUE #875: New Experiment: Suggest internal links within post content

Authors and editors often lack full context of everything already published on a site, so valuable opportunities to link new posts to related existing content go unused.  Internal links are one of the higher-leverage, lower-effort levers for on-site engagement and SEO (they pass link equity, surface related content to readers, and help establish topical authority), but finding good linking opportunities by hand doesn't scale, especially on larger sites.

Add a new experiment that suggests up to 5 (filterable) internal links per post, pointing to other existing content on the same site.  Links are suggested only, never auto-inserted so that the author/editor reviews and accepts/rejects each one, consistent with how every other experiment in this plugin requires explicit user action.

I considered a "related posts" block/widget at the end of a post as an alternative, but this would have lower engagement than in-content contextual links, and doesn't help with the deeper problem of missing internal links inside the body copy.

Two entry points into the same underlying suggestion ability:

1. Pre-publish flow (primary)

Suggestions surface as part of the pre-publish panel: run against the current draft content, return up to 5 candidate links (anchor text >> target post/page), and let the user accept or dismiss each before publishing.

2. On-demand flow, ahead of pre-publish (secondary)

A user can manually trigger link suggestions at any point while drafting, using the same pattern already established by the Editorial Notes experiment (which reviews content block-by-block and adds Notes with suggestions for accessibility/readability/grammar/SEO).  This experiment would do the same for internal links:
WP 6.9–7.0: attach the note to the relevant paragraph block, since inline/text-range Notes aren't available yet.
WP 7.1+: attach the note to the specific word(s)/phrase once inline Notes ship, anchoring the suggestion to the exact anchor text rather than the whole paragraph.

Since Editorial Notes already solves "review content block-by-block, write a Note" and Editorial Updates already solves "apply a Note's suggestion to content," this experiment could be a new ability (ai/internal-link-suggestions or similar) plus registration as a consumer of those existing patterns, rather than new editor plumbing.  The only downside there is it requires those experiments to be enabled to fully function, so not optimal.

### Scope for v1

* Up to 5 suggested links per post; count filterable.
* Candidate targets: other posts/pages on the same site (taxonomy archives and cross-network/multisite targets are reasonable future iterations).
* Anchor text pulled from the post's own content, no invented anchors or URLs.
* Suggestions are review-and-accept only; nothing is auto-published or auto-inserted.
* Filterable: link count, eligible post types as suggestion targets, eligible post types for suggestions to appear on.

### Open questions

* Should this reuse the Editorial Notes/Editorial Updates code paths directly, or register as its own experiment that happens to write Notes the same way?   _[I think the later, but open to discussion here.]_
* Does the suggestion ability need its own lightweight content/link index or is per-post context sufficient to send to the connector at suggestion time?
* Should the Notes-based flow auto-detect and degrade gracefully between paragraph-level (6.9–7.0) and inline (7.1+) Notes based on the installed core version, or should that just follow whatever core version the site is running with no separate toggle?  _[I think the former, as that allows us to remove the 6.9-7.0 fallback once 7.1 becomes our min WP version]_
* Do we want a short "reason" string noting why the AI thinks this is a good suggestion to help the editor evaluate relevance before accepting?

_No response_

- [x] Yes

Infinite-Null:
Closes #875

# Internal Link Suggestions Experiment

## Description

This PR implements the **Internal Link Suggestions** experiment, which uses AI to suggest contextual internal links within post content by analyzing the current draft and matching relevant phrases to published posts or pages on the site.

## Testing Instructions

1. Ensure an AI provider is connected and enabled under **Settings → AI**, and the **Internal Link Suggestions** experiment is active under `Editor Experiments` in `/wp-admin/options-general.php?page=ai-wp-admin`.
2. Create or edit a post with several paragraphs of content. Ensure you have other published posts/pages on your site to link to.
3. Open the Post sidebar and click **Suggest Internal Links**.
4. Verify that relevant internal link suggestions appear with context snippets.
6. Click **Accept** on a suggestion to verify that the suggestion is accepted.
7. Click **Dismiss** on a suggestion to verify it is not accepted.

## Screenshots or screencast

https://github.com/user-attachments/assets/2f0e89c1-33e6-4542-9266-9c2231812685

## Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 4.6

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-887-6f38815c646685f9ff29c9818aed0949f7b5cd37.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+				),
+				'post_id'          => array(
+					'type'              => 'integer',
+					'sanitize_callback' => 'absint',
+					'description'       => esc_html__( 'ID of the post being edited.', 'ai' ),
+				),
+				'max_suggestions'  => array(
+					'type'              => 'integer',
+					'sanitize_callback' => 'absint',
+					'description'       => esc_html__( 'Maximum number of link suggestions to return (1–10).', 'ai' ),
```
dkotter, line 88: If we limit this from 1 to 10, our schema should set that here with `minimum` and `maximum` attributes

includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @param mixed $input The input arguments to the ability.
+	 * @return bool|\WP_Error True if the user has permission, WP_Error otherwise.
+	 */
+	protected function permission_callback( $input ) {
+		$post_id = isset( $input['post_id'] ) ? absint( $input['post_id'] ) : 0;
+
+		if ( ! $post_id ) {
```
dkotter, line 226: Our schema says that a post ID is required so I don't think this check will ever be needed, unless I'm missing something

includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+			array(
+				'post_content'     => '',
+				'post_id'          => 0,
+				'max_suggestions'  => self::DEFAULT_MAX_SUGGESTIONS,
+				'excluded_anchors' => array(),
+			)
+		);
+
+		$post_content     = wp_kses_post( (string) $args['post_content'] );
+		$post_id          = absint( $args['post_id'] );
```
dkotter, line 162: I know we validate post ID in our permission callback but to match our other abilities, probably best to ensure this is a valid post ID here

includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+	 *
+	 * @param string $prompt The assembled prompt.
+	 * @return \WP_AI_Client_Prompt_Builder|\WP_Error The prompt builder, or WP_Error on failure.
+	 */
+	private function get_prompt_builder( string $prompt ) {
+		$prompt_builder = wp_ai_client_prompt( $prompt )
+			->using_system_instruction( $this->get_system_instruction() )
+			->as_json_response( $this->suggestions_schema() );
+
+		$prompt_builder = $this->set_provider_model_preference( $prompt_builder, Internal_Links_Experiment::class );
```
dkotter: We have two new methods we'll want to integrate here: `$this->filter_prompt` and `$this->filter_prompt_builder`.

The latter replaces this `$this->set_provider_model_preference` call

includes/Abilities/Internal_Links/system-instruction.php
```diff
@@ -0,0 +1,49 @@
...
+2. **Match to the site index.** Each suggestion must reference a URL from the <site-index> list. Do NOT invent URLs.
+3. **Relevance first.** Only suggest a link when the target page is genuinely relevant to the anchor phrase in context. Avoid superficial keyword matches.
+4. **No duplicates.** Do not suggest the same anchor text or the same URL more than once.
+5. **Respect the cap.** Return at most the number of suggestions specified in <max-suggestions>.
+6. **Context sentence.** For each suggestion, copy the sentence or clause from the post that contains the anchor text into the `context` field. This helps the editor understand placement.
+7. **Quality over quantity.** If fewer than <max-suggestions> high-quality links exist, return fewer. An empty array is valid if no good matches exist.
+8. **Skip already-linked text.** If an `<already-linked>` list is provided, do NOT suggest any anchor text that appears in that list. Those phrases are already hyperlinked in the post.
+
+## Output format
```
dkotter: Is this needed? We already provide the output format we expect when making a request so seems like this is unnecessary and wastes tokens

includes/Experiments/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,117 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Internal Links experiment.
+ *
+ * Uses AI to suggest contextual internal links within a post by analysing
+ * the current draft and identifying relevant published posts or pages on
+ * the same site. All suggestions require editor review before being applied.
```
dkotter: I'd probably just remove this

includes/Experiments/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,117 @@
...
+		return 'internal-links';
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function load_metadata(): array {
+		return array(
+			'label'       => __( 'Internal Link Suggestions', 'ai' ),
+			'description' => __( 'Uses AI to suggest relevant internal links within post content, using existing text as anchor text. All suggestions require editor review before being applied. Requires an AI connector that includes support for text generation models.', 'ai' ),
```
dkotter: No need to call out `uses AI` as most everything we ship uses AI.

includes/Experiments/Experiments.php
```diff
@@ -43,6 +43,7 @@ final class Experiments {
 		\WordPress\AI\Experiments\Meta_Description\Meta_Description::class,
 		\WordPress\AI\Experiments\Title_Generation\Title_Generation::class,
 		\WordPress\AI\Experiments\Type_Ahead\Type_Ahead::class,
+		\WordPress\AI\Experiments\Internal_Links\Internal_Links::class,
```
dkotter: We'll want to sort this alphabetically

src/experiments/internal-links/components/InternalLinksPlugin.tsx
```diff
@@ -0,0 +1,101 @@
...
+		? sprintf(
+				/* translators: %d: minimum number of characters required. */
+				__(
+					'Internal Link Suggestions will be available when the post content has at least %d characters.',
+					'ai'
+				),
+				minContentLength
+		  )
+		: __(
+				'Analyses this post and suggests relevant internal links using existing text as anchor text.',
```
dkotter: This shows `post` on all post types, for instance on a page. Maybe change `post` to just `content`?

includes/Experiments/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,117 @@
...
+			'category'    => Experiment_Category::EDITOR,
+		);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function register(): void {
+		add_action( 'wp_abilities_api_init', array( $this, 'register_abilities' ) );
+		add_action( 'enqueue_block_editor_assets', array( $this, 'enqueue_assets' ) );
```
dkotter: Let's change the default priority here so we can ensure the `Generate Editorial Notes` and `Apply Editorial Updates` buttons stay next to each other if all of those are turned on. For example, in the Summarization experiment, we set the priority to 5 to help with this

src/experiments/internal-links/hooks/useInternalLinks.ts
```diff
@@ -0,0 +1,285 @@
...
+		const alreadyLinked = getLinkedAnchorTexts( content );
+		const excludedAnchors = [ ...alreadyLinked ];
+
+		try {
+			const result = await runAbility< SuggestionResponse >(
+				'ai/internal-links',
+				{
+					post_content: content,
+					post_id: postId,
+					max_suggestions: maxSuggestions,
```
dkotter, line 219: `maxSuggestions` comes over as a string so we'll need to cast that to an integer otherwise we get a schema validation error: `Error: Ability "ai/internal-links" has invalid input. Reason: input[max_suggestions] is not of type integer`

src/experiments/internal-links/hooks/useInternalLinks.ts
```diff
@@ -0,0 +1,285 @@
...
+/**
+ * Applies an internal link suggestion to the block editor.
+ *
+ * Finds the first block whose plain-text content contains the anchor text
+ * and wraps that first occurrence in an HTML <a> tag.
+ *
+ * @param suggestion The accepted link suggestion.
+ * @param blocks     All blocks in the editor.
+ */
+function applyLinkToBlock( suggestion: LinkSuggestion, blocks: Block[] ): void {
```
dkotter, line 125: This does remove any inline formatting. For instance, if I have a string that is bold and that gets set as a suggestion, applying the suggestion removes the bold from the text. Ideal would be to ensure we consider rich text instead of plaintext where possible

dkotter: Also, if this function doesn't apply the link correctly, we still show a success message. Ideally we ensure the replacement actually works before showing a success and if it doesn't work, we show a proper error message

includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string                             $raw             Raw JSON string from the AI.
+	 * @param string                             $plain_text      Plain-text post content.
+	 * @param list<array{url: string, title: string}> $site_index List of linkable posts.
+	 * @param int                                $max_suggestions Maximum number of suggestions.
+	 * @return list<array{anchor_text: string, url: string, title: string, context: string}>
+	 */
+	private function parse_and_validate_response( string $raw, string $plain_text, array $site_index, int $max_suggestions ): array {
```
dkotter, line 421: Should we be checking the `excluded_anchors` here as well? If an LLM doesn't follow our instructions properly and returns a link for an already linked phrase, we'll end up replacing that link. Since we already validate the response here, could add additional validation for that

includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+				'update_post_meta_cache' => false,
+				'update_post_term_cache' => false,
+				'fields'                 => 'ids',
+			)
+		);
+
+		$index = array();
+
+		foreach ( $query->posts as $id ) {
+			$title = get_the_title( $id );
```
dkotter, line 329: Since we only want the IDs, this is a new DB request to get the title (and potentially permalink). I think if we run `_prime_post_caches( $query->posts, false, false );` before the loop that will solve it or might be fine to just request all fields instead of IDs and still get better performance.

includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+	/**
+	 * Maximum number of posts to include in the site index sent to the AI.
+	 *
+	 * Kept low to stay token-efficient.
+	 *
+	 * @since x.x.x
+	 *
+	 * @var int
+	 */
+	private const SITE_INDEX_LIMIT = 200;
```
dkotter, line 44: Should we allow this to be filtered?

src/experiments/internal-links/components/InternalLinksPlugin.tsx
```diff
@@ -0,0 +1,101 @@
...
+						{ buttonDescription }
+					</span>
+				</FlexItem>
+
+				{ suggestions.length > 0 && (
+					<FlexItem>
+						<p className="description ai-internal-links__suggestions-header">
+							{ sprintf(
+								/* translators: %d: number of suggestions found. */
+								__( '%d suggestion(s) found.', 'ai' ),
```
dkotter: Should use `_n` here for plurals

src/experiments/internal-links/hooks/useInternalLinks.ts
```diff
@@ -0,0 +1,285 @@
...
+		const regex = new RegExp( `(${ escapedAnchor })`, '' );
+		const updatedHtml = rawContent.replace(
+			regex,
+			`<a href="${ url }">${ anchorText }</a>`
+		);
+
+		const attributeKey =
+			'content' in block.attributes ? 'content' : 'value';
+
+		( dispatch( blockEditorStore ) as any ).updateBlockAttributes(
```
dkotter: I think we can remove the `as any` from here

src/experiments/internal-links/hooks/useInternalLinks.ts
```diff
@@ -0,0 +1,285 @@
...
+
+	const minContentLength: number =
+		( window as any ).aiInternalLinksData?.minContentLength ??
+		MINIMUM_CONTENT_COUNT_DEFAULT;
+
+	const maxSuggestions: number =
+		( window as any ).aiInternalLinksData?.maxSuggestions ?? 5;
+
+	const { content, postId } = useSelect( ( selectStore ) => {
+		const editor = selectStore( editorStore ) as any;
```
dkotter: I think we can remove the `as any` from here

src/experiments/internal-links/hooks/useInternalLinks.ts
```diff
@@ -0,0 +1,285 @@
...
+				}
+			);
+		} finally {
+			setIsLoading( false );
+		}
+	};
+
+	const acceptSuggestion = ( suggestion: LinkSuggestion ) => {
+		const blocks = (
+			select( blockEditorStore ) as any
```
dkotter: I think we can remove the `as any` from here

includes/Abilities/Internal_Links/Internal_Links.php
```diff
@@ -0,0 +1,488 @@
...
+
+		// Convert HTML to plain text for anchor text matching.
+		$plain_text = normalize_content( wp_strip_all_tags( $post_content ) );
+
+		if ( empty( trim( $plain_text ) ) ) {
+			return array( 'suggestions' => array() );
+		}
+
+		// Build the list of linkable posts/pages from this site.
+		$site_index = $this->build_site_index( $post_id );
```
dkotter, line 187: In testing, I'm not getting very good results based on this approach. I think we may want to pause here to decide if this is an approach we should continue on or not. My suggestion would be to wait until we have embeddings support and we can use those to find relationships, which I think will scale better and give us much better results.

dkotter (changes requested):
Testing works though results aren't great (which I've left a comment around this). Probably worth a discussion on the approach here prior to proceeding.

In addition, adding documentation to match how our other experiments are documented would be great

includes/Abilities/Internal_Links/Internal_Links.php
jeffpaul: Embeddings is almost certainly what I had imagined leveraging when I opened the related issue, so feels good to update to leverage that (those?).

dugyen:
While testing this PR, I ran into the two currently-failing "Run E2E tests" checks and traced them to stale assertions in the e2e spec rather than app bugs — a rebrand commit changed the UI copy from "the post content" to "the content", and the suggestions-found header now uses `_n()` for proper pluralization ("1 suggestion found." for the singular case), but the spec still asserted the old strings.

Opened a small fix against this PR's branch: https://github.com/Infinite-Null/ai/pull/4

(The PHP 8.3 WP trunk failure looks like an unrelated transient composer/GitHub API 504 during dependency download, not a code issue.)

dugyen:
Ran a fuller review over this PR's diff (8 independent finder passes + direct source verification of every candidate). 10 findings, most severe first. Findings #1–#3 substantially overlap with what @dkotter already flagged in review (the `excluded_anchors` gap and the "success shown even when the replace fails" issue) — including them here with exact repro steps in case they're useful for the fix.

**1. `excluded_anchors` sanitized and sent to the prompt, but never re-checked when validating the AI's response** — [Internal_Links.php#L210](https://github.com/WordPress/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/includes/Abilities/Internal_Links/Internal_Links.php#L210) (call site) / [#L421](https://github.com/WordPress/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/includes/Abilities/Internal_Links/Internal_Links.php#L421) (`parse_and_validate_response()` signature).
Every other rule (verbatim anchor, valid URL, no duplicate anchor/URL) is defensively re-checked in `parse_and_validate_response()`, but `excluded_anchors` isn't even passed in. If the model ignores rule 8 in `system-instruction.php` and returns an already-linked phrase with a valid site-index URL, nothing filters it back out — it's shown to the editor as a fresh suggestion.

**2. Anchor match is checked against plain text, but the replace runs against raw HTML** — [useInternalLinks.ts#L135](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/src/experiments/internal-links/hooks/useInternalLinks.ts#L135).
`applyLinkToBlock` tests `plainContent.includes(anchorText)` (tags stripped) but inserts the link via `rawContent.replace(regex, ...)` against the raw HTML. E.g. for `Our <strong>customer support</strong> team is great.`, plain text contains `"customer support team"` so the check passes, but the raw HTML has `</strong>` splitting that exact substring, so the regex never matches — `updatedHtml === rawContent` and no link is ever inserted.

**3. Success notice shown even when the link insert silently failed** — [useInternalLinks.ts#L249](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/src/experiments/internal-links/hooks/useInternalLinks.ts#L249).
`acceptSuggestion` calls `applyLinkToBlock` (returns `void`, no success/failure signal), then unconditionally removes the suggestion from the list and shows "Internal link applied. Save the post to keep the change." Compounds #2: the editor believes the link was added and saves, only to find out later (or never) that it wasn't.

**4. Inline attribute-key pick misses the existing image/pullquote special-case** — [useInternalLinks.ts#L149](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/src/experiments/internal-links/hooks/useInternalLinks.ts#L149).
Picks the write attribute via `'content' in block.attributes ? 'content' : 'value'`, reimplementing (with different semantics than the read at L123's `content ?? value` fallback) what `getEditableTextAttribute()` in `src/utils/blocks.ts` already does, including its `core/image` → `alt` special case. A suggestion whose anchor text lives in an image's alt text will write to the wrong attribute or no-op instead.

**5. Bypasses the plugin's standard prompt/model-preference filters** — [Internal_Links.php#L193](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/includes/Abilities/Internal_Links/Internal_Links.php#L193), [get_prompt_builder() L386-402](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/includes/Abilities/Internal_Links/Internal_Links.php#L386-L402).
`execute_callback()` never routes the prompt through `->filter_prompt()`, and `get_prompt_builder()` inlines its own model-preference logic instead of `Abstract_Ability::filter_prompt_builder()` — the pattern every sibling ability (Editorial_Notes, Meta_Description, Excerpt_Generation, Summarization, ...) uses. Any code hooking the standard `wpai_{slug}_prompt` / `wpai_{slug}_prompt_builder` filters silently never fires for this ability.

**6. N+1 queries building the site index** — [Internal_Links.php#L313](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/includes/Abilities/Internal_Links/Internal_Links.php#L313) (same spot @dkotter flagged).
`build_site_index()` runs `WP_Query` with `'fields' => 'ids'` (doesn't warm the post cache) with no `_prime_post_caches()` call, then loops `get_the_title()`/`get_permalink()` per ID — up to `SITE_INDEX_LIMIT` (200) extra individual DB lookups per "Suggest Internal Links" click.

**7. Loose falsy/`empty()` checks silently drop literal \"0\" values** — [Internal_Links.php#L332](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/includes/Abilities/Internal_Links/Internal_Links.php#L332) and [L445-447](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/includes/Abilities/Internal_Links/Internal_Links.php#L445-L447).
`if ( ! \$title || ! \$url )` and `empty( \$item['anchor_text'] ) || ...` both treat the string `"0"` as absent. A page titled exactly "0" is silently excluded from the site index; a suggestion with anchor/url/title exactly "0" is silently dropped even though the `is_string()` checks two lines later would otherwise accept it. Edge cases, but silent.

**8. Duplicates existing `utils/blocks.ts` helpers** — [useInternalLinks.ts#L51](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/src/experiments/internal-links/hooks/useInternalLinks.ts#L51) (`toPlainString`) and [L106](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/src/experiments/internal-links/hooks/useInternalLinks.ts#L106) (`flattenAll`).
Both reimplement `toPlainText()`/`flattenBlocks()` from `src/utils/blocks.ts`, already used the same way by sibling experiments (e.g. editorial-updates). A future fix to the shared helpers (their own comment notes RichText's shape has changed before) won't propagate here.

**9. `as any` casts hide a real type/runtime mismatch** — [useInternalLinks.ts#L182](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/src/experiments/internal-links/hooks/useInternalLinks.ts#L182).
Every read of the global goes through `(window as any).aiInternalLinksData` instead of the typed `window.aiInternalLinksData` declared in `index.tsx`, and `maxSuggestions` — typed `number` there — is run through `parseInt(...)`, implying the localized value is actually a string at runtime. The casts mean TS can't catch this mismatch at any of the 4 call sites.

**10. Full post content re-serialized/re-counted on every keystroke** — [useInternalLinks.ts#L186](https://github.com/Infinite-Null/ai/blob/5b4ed5db81991995241718b9ab5fa1e0a978abf2/src/experiments/internal-links/hooks/useInternalLinks.ts#L186).
`useInternalLinks()` subscribes unconditionally to `editor.getEditedPostContent()` (re-serializes the whole block tree on every content change) from an always-mounted sidebar panel, and recomputes the content-length check on every render without memoizing on `content`. On a long post, every keystroke does a full serialize + character-count scan just to toggle a disabled state.

---
*Reviewed via 8 independent finder passes (correctness, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions) followed by direct source verification of each candidate — using Claude Code.*

PR #891, opened 2026-07-23
Feat: Add semantic search experiment to AI plugin

### LINKED ISSUE #844: New Experiment: Semantic search in wp admin

The built-in admin search matches exact words, so if you remember writing "that post about pricing changes" but the post was titled "Updates to our plans" then searching for "pricing" in the posts list finds nothing.  On sites with hundreds of posts, users can end up scrolling through pages of results or guessing at old titles which as a small friction that can compound daily for anyone actively managing content.

The in-progress native vector search experiment in #683 solves this for site visitors on the frontend search page, but the people managing the site face the same problem in wp-admin and aren't covered by it.

**_NOTE:_** This is blocked until the underpinning embeddings work is available, specifically the PHP AI Client's embedding generation API (https://github.com/WordPress/php-ai-client/issues/244) and the native vector search experiment that builds the content index on top of it (#683).  This issue is intended as a follow-on to that foundation, not a parallel implementation.

Extend semantic search to the admin side, reusing the same content index that #683 builds:

* Posts list search: searching from the post/page list tables returns results based on what content means, not just exact word matches, falling back gracefully to default search when no AI provider is configured (matching our existing no-provider degradation pattern)
* Command palette: the editor's command palette (Cmd/Ctrl+K) gains a "search content semantically" path, so authors can jump to the post they're thinking of without leaving the editor

Ideally the retrieval piece is exposed as an ability so other features (and potentially third-party plugins following our lead) can reuse it.

Editors, admins, and multi-author teams hit this constantly and it's arguably a more frequent touchpoint than frontend search, since content managers search their own back catalog far more often than any single visitor does.  It also extends the value of the vector-search investment to a second audience at relatively low incremental cost, since indexing, storage, and embedding generation are already handled by #683.

- [x] Yes

priyanshuhaldar007:
## What?
Closes https://github.com/WordPress/ai/issues/844

Adds a new **Semantic Search** experiment to the AI plugin, extending semantic/vector search from the frontend into wp-admin. This covers two surfaces:

- **Posts list search** — a "Search semantically" option on `edit.php` list tables that re-ranks results by meaning instead of exact keyword matching, with graceful fallback to default WordPress search when no provider is configured or nothing is indexed yet.
- **Command palette (Cmd/Ctrl+K)** — the block editor's command palette gains live semantic search results as the author types, letting them jump directly to a post they're thinking of.

Also adds a **Tools → Semantic Search Index** admin page for configuring an embedding provider, testing connectivity(for Google Gemini), and batch-indexing existing content.

## Why?

The built-in admin search only matches exact words. If an editor remembers "that post about pricing changes" but the post is titled "Updates to our plans," searching "pricing" in the posts list returns nothing. This friction compounds daily for anyone actively managing a large site, and content managers hit it more often than any single frontend visitor hits search — the same problem #683 already solves for frontend search but not for wp-admin.

This PR is intended as a **follow-on to #683**, not a parallel implementation. It reuses the same conceptual approach (embed content, compare via cosine similarity) but is currently backed by a temporary, self-contained implementation since the two foundational pieces this issue is blocked on are not yet merged:

- `php-ai-client#244` (embedding generation API) — still open, changes requested
- `ai#683` (native vector index) — still in draft

Rather than wait, this PR ships a fully working version now, with the two upstream-dependent pieces isolated behind clear `TODO` markers so migration is a small, contained change once each lands.

## How?

Implemented as a new `Semantic_Search` experiment following the existing Feature/Experiment pattern used by `Comment_Moderation`, `Summarisation`, and `Abilities_Explorer`.

**New classes** (`includes/Experiments/Semantic_Search/`):
- `Semantic_Search.php` — experiment entry point: metadata, settings registration, hook wiring
- `Embedding_Api.php` — HTTP wrapper supporting OpenAI, Jina AI, Ollama, and Google Gemini embedding endpoints (temporary stand-in until `php-ai-client#244` lands)
- `Embedding_Store.php` — reads/writes embeddings as post meta (`_wpai_semantic_search_embedding`) (temporary stand-in until `ai#683`'s native VECTOR column lands)
- `Indexer.php` — orchestrates the API → store pipeline, batched at 5 posts per request to stay within PHP's execution timeout
- `Vector_Search.php` — computes cosine similarity in PHP against all indexed posts; filters by a per-provider score threshold (model-dependent, user-configurable — higher-dimensional models produce lower absolute similarity scores for equivalent semantic pairs, so a single hardcoded threshold is wrong across providers)
- `List_Table_Integration.php` — hooks `pre_get_posts` on `edit.php`; re-ranks results via `post__in` + `orderby: post__in` when semantic search is available and returns results, otherwise leaves the query untouched
- `REST_Controller.php` — four routes under `/ai/v1/semantic-search/...` for search, indexing, index status, and connection testing
- `Index_Page.php` — admin UI at Tools → Semantic Search Index

**New assets:**
- `build-scripts/experiments/semantic-search.js` — registers a Gutenberg command loader via `registerCommandLoader`, debounced on the palette's search string

**Modified:**
- `Experiments.php` — registers `Semantic_Search::class` in `EXPERIMENT_CLASSES`

**Design decisions worth flagging for review:**
- All registered settings fields include `show_in_rest: true` so the React settings page at Settings → AI can read/write them via the core `/wp/v2/settings` route (matches the `Comment_Moderation` pattern).
- The command palette embeds the search term directly into each command's `name` field, since cmdk (the library backing Gutenberg's palette) filters items on `name + label`, and whether `keywords` is forwarded to `Command.Item` varies by Gutenberg version. This is a defensive, version-agnostic fix; `keywords` is also set as a secondary safeguard.
- Command `icon` is explicitly `null` rather than a string — Gutenberg expects `null` or a React element; a plain string silently fails to render with no visible error.
- No `echo`/output is used inside REST callbacks or any code path they invoke, since any stray output corrupts the JSON body returned to `apiFetch`.
- Both classes with upstream dependencies (`Embedding_Api.php`, `Embedding_Store.php` + `Vector_Search.php`) are marked with inline `TODO` comments describing the exact swap to `php-ai-client#244`'s embedding API and `ai#683`'s native VECTOR INDEX query, respectively. No other class in this PR should need to change across either migration.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude (Anthropic)
Model(s): Claude Sonnet 5
Used for: Architecture planning, initial implementation, score threshold calibration across embedding providers, and documentation. All code was reviewed, tested against a live WordPress install, and adjusted by me based on real test results before inclusion in this PR.

## Testing Instructions

1. Activate the AI plugin and enable the **Semantic Search** experiment under Settings → AI.
2. Configure an embedding provider (OpenAI, Jina AI, Ollama, or Google Gemini) and, if required, an API key.
3. Go to **Tools → Semantic Search Index**, click **Test Connection** to confirm the provider responds with a model name and dimension count.
4. Click **Index All Posts** and confirm the progress counter reaches "Done" with no errors.
5. **Posts list test:** go to Posts, check "Search semantically," and search for a concept that's present in a post's content but not its title/keywords (e.g. search "pricing" against a post titled "Updates to our plans" that discusses subscription costs). Confirm the post is returned.
6. **Command palette test:** open any post in the block editor, press Cmd/Ctrl+K, and type a concept-based query. Confirm semantically related posts appear as clickable results and navigate correctly on click.
7. **Fallback test:** clear the API key in Settings and repeat steps 5–6. Confirm both surfaces fall back gracefully to default WordPress search behaviour with no errors or blank screens.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
|  Screenshot 2026-07-23 at 9 54 45 PM |  Screenshot 2026-07-23 at 9 54 52 PM |

## Changelog Entry

> Added - Semantic search experiment for wp-admin: posts list search and command palette results ranked by content meaning rather than exact keyword match, with configurable embedding provider support (OpenAI, Jina AI, Ollama, Google Gemini) and a Tools admin page for indexing.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-891-4c2d6071673a5ad411ec38cb9164973131b93a52.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter (changes requested):
So we don't want to have to maintain provider integrations within this plugin. Those integrations should be done upstream in their individual provider plugins if at all possible.

Also note we have https://github.com/WordPress/ai/pull/892 open now which brings over the embedding work from the PHP AI Client so we can start taking advantage of that prior to that being released into WordPress 7.2 (and allows us to support older versions of WordPress even after that point). So ideally this PR is refactored to remove any provider integration and any embedding generation should filter through the PHP AI Client changes in that PR

jeffpaul:
@priyanshuhaldar007 please see the new tracking issue for embedding integration in the AI plugin (#962) and whether you might be interested in coordinating with @artpi and @ColinM-sys who've also worked on #683 and #943 respectively on the smaller, more reviewable PRs on the path toward implementing embedding support... thanks!

PR #897, opened 2026-07-24
Prototype : Implement slug generation feature with UI and tests

### LINKED ISSUE #876: New Experiment: Suggest permalink slugs

Well-crafted permalink slugs carry real SEO value (concise, keyword-relevant URLs), but authors often leave the auto-generated slug as-is or hand-edit it without much thought.  This is a good candidate for AI assistance, similar in spirit to this plugin's existing Title Generation and Excerpt Generation experiments with their same "one-click, review, apply" pattern, but applied to the permalink slug instead.

Add a new experiment that suggests a slug for the post permalink based on the post's title and/or content, following the same generate >> review >> select pattern as Title Generation.

Still open, and worth settling with the community before/while scoping this: where the trigger and suggestion(s) surface.  A few options, not mutually exclusive:

1. Pre-publish flow: slug suggestion appears as a step in the pre-publish panel, alongside other pre-publish checks.
2. Post editor sidebar, alongside the slug/permalink editing UI: a "Generate/Regenerate" affordance next to the existing slug field, matching how Title Generation surfaces above the title field.
3. Both: sidebar trigger for on-demand use while drafting, pre-publish surfacing as a nudge for anyone who hasn't touched the slug yet.

### Scope for v1

* Ability to generate a slug suggestion (or a short list of options) from post title/content.
* Review-and-select only; nothing overwrites the existing slug without the user choosing it, consistent with how every other experiment in this plugin requires explicit user action.
* Filterable: number of suggestions returned, making use of the existing minimum content length requirement before suggestions are offered (similar to the existing content-length gate on Summarization).

### Open questions

* Pre-publish panel, sidebar-next-to-slug, or both?  See above, this affects where the ability gets wired into the editor UI.  _[I think starting with just the sidebar-next-to-slug would suffice]_
* Should slug suggestions consider existing permalinks on the site to avoid near-duplicate slugs, or is that out of scope for v1?
* Does this warrant a WP-CLI command for bulk slug review/regeneration on existing content, similar to the alt-text bulk command?

_No response_

- [x] Yes

milindmore22:
## What?

Closes #876 

This PR introduces the **Slug Generation** experimental feature (`ai/slug-generation` ability & experiment). It provides AI-assisted permalink suggestions directly inside the WordPress Block Editor based on post title, content, and additional context.

Both entry points (sidebar popover + pre-publish panel) are currently supported in prototype.

> Should slug suggestions consider existing permalinks on the site to avoid near-duplicate slugs, or is that out of scope for v1?

WordPress core natively enforces slug uniqueness on post save, not sure if we should cover this in Slug generation

> Does this warrant a WP-CLI command for bulk slug review/regeneration on existing content, similar to the alt-text bulk command?

Not covered in prototype, but we can cover if required.

## Why?

Creating concise, keyword-rich, and SEO-friendly permalink slugs often requires extra effort from content creators. This feature streamlines the publishing workflow by automatically suggesting URL-safe slugs within the permalink inspector popover and the pre-publish panel.

## How?

- **Abilities API Integration**:
  - Implemented `WordPress\AI\Abilities\Slug_Generation\Slug_Generation` extending `Abstract_Ability`.
  - Added system instructions (`system-instruction.php`) enforcing URL-safe, lowercase, hyphenated slug generation.
  - Sanitizes and validates outputs using `sanitize_title()`.

- **Experiment Framework**:
  - Added `WordPress\AI\Experiments\Slug_Generation\Slug_Generation` under `Experiment_Category::EDITOR`.
  - Localized configuration settings (minimum content length requirement, number of suggestions).

- **Block Editor UI Components**:
  - **Permalink Inspector Button**: Injects a "Generate Slug" button into the Block Editor permalink popover (`.editor-post-url`).
  - **Interactive Modal (`SlugGenerationModal`)**: Renders a modal interface allowing users to preview, manually edit, select, or regenerate slug options.
  - **Pre-Publish Panel (`SlugPrePublishPanel`)**: Adds a `PluginPrePublishPanel` sidebar card to review or change the post slug prior to publishing.

- **Automated Tests**:
  - Added PHP integration test coverage (`Slug_GenerationTest.php`) for ability execution, argument validation, and permissions.
  - Updated E2E helper routines (`helpers.ts`) for reliable REST API save synchronization during bulk toggle tests.

### Use of AI Tools
AI assistance: Yes  
Tool(s): Antigravity AI  
Model(s): Gemini 3.6 Flash  
Used for: Initial implementation of ability and experiment classes, React UI components, unit/integration tests, and code documentation. Final changes reviewed and verified by human developer.

## Testing Instructions

1. Ensure global AI features and an AI Connector with text generation capabilities are active.
2. Navigate to **Settings > AI** and enable the **Slug Generation** experiment under *Editor Experiments*.
3. Create a new post in the Block Editor and add a post title and content (at least 250 characters).
4. Click on the Permalink / URL field in the document sidebar to open the permalink popover.
5. Click **Generate Slug**. Verify the modal pops up with 3 generated slug suggestions.
6. Select a suggested slug (or edit it in the input field) and click **Insert**. Verify the post permalink updates.
7. Click **Publish** to open the pre-publish panel. Verify the *Suggested Slugs* pre-publish panel appears and allows applying suggestions.

## Screenshots or screencast

https://github.com/user-attachments/assets/06fff2c6-7090-4d8b-84d4-d59fef558581

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-897-0bf98238bbd234cb91a54233c69772446953db71.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Abilities/Slug_Generation/Slug_Generation.php
```diff
@@ -0,0 +1,299 @@
...
+class Slug_Generation extends Abstract_Ability {
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since x.x.x
+	 */
+	protected function guideline_categories(): array {
+		return array( 'site', 'copy' );
+	}
```
dkotter: Open to thoughts here but not sure if generating a slug benefits from these guidelines. May be fine to just not pass these in

includes/Abilities/Slug_Generation/Slug_Generation.php
```diff
@@ -0,0 +1,299 @@
...
+			$prompt_input .= "<content>{$post_content}</content>";
+		}
+		if ( ! empty( $context ) ) {
+			if ( is_array( $context ) ) {
+				$context = implode( "\n", $context );
+			}
+			$prompt_input .= "\n\n<additional-context>{$context}</additional-context>";
+		}
+
+		$number_of_suggestions = (int) apply_filters( 'wpai_slug_generation_number_of_suggestions', (int) $args['number_of_suggestions'] );
```
dkotter: We should add a filter docblock here

includes/Abilities/Slug_Generation/Slug_Generation.php
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+		if ( empty( $result ) ) {
+			return new WP_Error(
+				'no_results',
+				esc_html__( 'No slug suggestion was generated.', 'ai' )
+			);
+		}
+
+		// Parse the output lines into clean, sanitized WordPress slugs.
```
dkotter: Any reason to approach it this way instead of using structured outputs? Typically that will work better than just hoping the LLM returns things in the structure we want and avoids us having to do this type of parsing

includes/Abilities/Slug_Generation/system-instruction.php
```diff
@@ -0,0 +1,29 @@
...
+
+The slug suggestions should follow these requirements:
+- Be concise (typically 2 to 5 words) and optimized for SEO.
+- Focus on key concepts and relevant keywords.
+- Use only lowercase letters, numbers, and hyphens.
+- Do not include any file extensions (e.g., .html, .php).
+- Output exactly %d suggestions, one per line.
+- Do not include any markdown, bullets, numbering, or formatting.
+- Output only the raw slug text. Respond directly without preamble. Do not wrap the output in quotes. Do not add closing remarks or follow-up questions.',
+	$wpai_slug_num
```
dkotter: Would we want a line in here asking it to match the language of the title given? We do this for other abilities to ensure the content returned matches the language of the content passed in

includes/Experiments/Slug_Generation/Slug_Generation.php
```diff
@@ -0,0 +1,116 @@
...
+		// Ensure the post type supports titles and isn't an attachment screen.
+		if (
+			! $screen ||
+			! post_type_supports( $screen->post_type, 'title' ) ||
+			in_array( $screen->post_type, array( 'attachment' ), true )
+		) {
+			return;
+		}
+
+		// Enqueue frontend scripts, styles, and pass localized configuration settings to window.
```
dkotter: This claims "frontend" but this only loads in the admin

includes/Experiments/Experiments.php
```diff
@@ -43,6 +43,7 @@ final class Experiments {
 		\WordPress\AI\Experiments\Meta_Description\Meta_Description::class,
 		\WordPress\AI\Experiments\Title_Generation\Title_Generation::class,
 		\WordPress\AI\Experiments\Type_Ahead\Type_Ahead::class,
+		\WordPress\AI\Experiments\Slug_Generation\Slug_Generation::class,
```
dkotter: We've been trying to organize these alphabetically so they display that way in the settings page

tests/e2e/specs/admin/settings.spec.js
dkotter, line 1: Looks like we're just adjusting existing E2E tests in this PR. We should also ensure we have new E2E tests to cover all this new functionality

src/experiments/slug-generation/index.tsx
```diff
@@ -0,0 +1,296 @@
...
+				isAttached = true;
+				return;
+			}
+
+			// Create wrapper container for the Generate button
+			container = document.createElement( 'div' );
+			container.className = 'ai-slug-generation-container';
+			container.style.marginTop = '8px';
+			container.style.display = 'flex';
+			container.style.justifyContent = 'flex-end';
```
dkotter: Do we need these inline styles? Looks like they're part of the CSS file we load

dkotter: And personal opinion here but right now all of our buttons are either full width or left aligned. This button is not full width and right aligned. I'd suggest taking one of those two approaches just to match the UI we use elsewhere

src/experiments/slug-generation/components/SlugGenerationButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+	const tooShortLabel = sprintf(
+		/* translators: %d: minimum number of characters required. */
+		__(
+			'Slug suggestions will be available when the post content has at least %d characters.',
+			'ai'
+		),
+		minContentLength
+	);
+
+	const buttonLabel = __( 'Generate Slug', 'ai' );
```
dkotter: To match how all of our other buttons work, I think this should be `Generate Slug` if the post doesn't have a slug. If it does, it should be `Regenerate Slug`

src/experiments/slug-generation/components/SlugGenerationButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+		// Dispatch the trigger event to open the modal and start generation
+		window.dispatchEvent(
+			new CustomEvent( 'ai-trigger-slug-generation', {
+				detail: { postId, title, content },
+			} )
+		);
+
+		// Close the slug popover immediately by simulating a click on its close/toggle button
+		const closeButton = document.querySelector(
+			'.editor-post-url button[aria-label="Close"], .editor-post-url button[aria-label="Close popover"], .editor-post-url button[aria-label="Close dialogue"]'
```
dkotter: I don't think we can rely on these labels here. For instance, what if the site is translated, I'm assuming the button labels will be translated as well. We should find a better approach to finding the close button

includes/Abilities/Slug_Generation/Slug_Generation.php
dkotter: Also, we should probably have some validation here after the filter runs. What if someone uses the filter and doesn't return a number, or returns something like 1000 or 0? Ideally we set a minimum and maximum (1 to 10?) and ensure this is always an integer between those two

dkotter: If we do the above, we can also update the schema with minimum and maximum values

includes/Experiments/Slug_Generation/Slug_Generation.php
```diff
@@ -0,0 +1,116 @@
...
+		Asset_Loader::enqueue_script( 'slug_generation', 'experiments/slug-generation', array( 'include_core_abilities' => true ) );
+		Asset_Loader::enqueue_style( 'slug_generation', 'experiments/slug-generation' );
+		Asset_Loader::localize_script(
+			'slug_generation',
+			'SlugGenerationData',
+			array(
+				'enabled'             => $this->is_enabled(),
+				'minContentLength'    => get_min_content_length( 'slug-generation', 250 ),
+				'numberOfSuggestions' => (int) apply_filters( 'wpai_slug_generation_number_of_suggestions', 3 ),
```
dkotter: I don't like that we use the same filter here and in the ability itself. What this means is if a 3rd party directly calls the ability and passes in `number_of_suggestions` but another plugin has a filter on this hook, that filter will override. I'd suggest we leave the filter here but remove it from the ability call, which allows sites to set their default but allows anyone directly using the ability to pass in the value they want without having that modified by a filter

includes/Abilities/Slug_Generation/Slug_Generation.php
```diff
@@ -0,0 +1,299 @@
...
+			$prompt_input .= "<title>{$post_title}</title>\n\n";
+		}
+		if ( ! empty( $post_content ) ) {
+			$prompt_input .= "<content>{$post_content}</content>";
+		}
+		if ( ! empty( $context ) ) {
+			if ( is_array( $context ) ) {
+				$context = implode( "\n", $context );
+			}
+			$prompt_input .= "\n\n<additional-context>{$context}</additional-context>";
```
dkotter, line 161: So our context here loses it's keys, so no way to know what each piece of data actually represents. Ideally we build a `Key: value` string

includes/Abilities/Slug_Generation/Slug_Generation.php
```diff
@@ -0,0 +1,299 @@
...
+		// Parse the output lines into clean, sanitized WordPress slugs.
+		$lines = explode( "\n", $result );
+		$slugs = array();
+		foreach ( $lines as $line ) {
+			$line = trim( $line, " \t\n\r\0\x0B\"'" );
+			if ( empty( $line ) ) {
+				continue;
+			}
+
+			$slugs[] = sanitize_title( $line );
```
dkotter: I believe `sanitize_title` allows underscores though I don't think we want those in slugs. Should verify if that's the case and if so, use a different sanitizer here

src/experiments/slug-generation/components/SlugGenerationModal.tsx
```diff
@@ -0,0 +1,172 @@
...
+				) : (
+					<Flex
+						direction="column"
+						gap="2"
+						align="stretch"
+						role="group"
+						aria-labelledby={ labelId }
+					>
+						{ suggestions.map( ( slug, index ) => (
+							<Button
```
dkotter: Should these be buttons? Seems like `RadioControl` or `ToggleGroupControl` will be better for accessibility

src/experiments/slug-generation/components/SlugGenerationModal.tsx
```diff
@@ -0,0 +1,172 @@
...
+								{ slug }
+							</Button>
+						) ) }
+					</Flex>
+				) }
+			</div>
+
+			<TextControl
+				label={ __( 'Selected slug', 'ai' ) }
+				value={ selectedSlug }
```
dkotter, line 100: There's a bug here on regeneration. On initial generation, seems this input is set to the first slug suggestion in the modal. If I regenerate in the modal, no slug is selected and this input shows the previous first slug

src/experiments/slug-generation/components/SlugGenerationButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+import type { SlugGenerationData } from '../types';
+
+const MINIMUM_CONTENT_COUNT_DEFAULT = 250;
+const NUMBER_OF_SUGGESTIONS_DEFAULT = 3;
+
+/**
+ * Helper to fetch localized settings passed from PHP to the global window object.
+ */
+const getSettings = (): SlugGenerationData => {
+	const settings = ( window as any ).aiSlugGenerationData ?? {};
```
dkotter: We've been trying to move away from ever use `as any`. I think we can move this to the types file to avoid doing this and update this across all three files

src/experiments/slug-generation/index.tsx
```diff
@@ -0,0 +1,296 @@
...
+			root.render( <SlugGenerationButton /> );
+
+			isAttached = true;
+		};
+
+		// Run initial check
+		findAndAttach();
+
+		// Create observer to listen for sidebar renders/toggles and popover display.
+		observer = new MutationObserver( () => {
```
dkotter: So I understand why we're doing this but I think this could cause some performance issues, as this is constantly running. Could we maybe add debouncing or narrowing the observed subtree to just the sidebar or gating on a store selector for the URL popover being open? If there's a way to narrow the scope here so this isn't constantly firing for any and all changes, that would be great

tests/Integration/Includes/Experiments/Slug_Generation/Slug_GenerationTest.php
```diff
@@ -0,0 +1,181 @@
...
+	 * Test that enqueue_assets() returns early for post types that do not support title.
+	 */
+	public function test_enqueue_assets_returns_early_for_post_type_without_title_support() {
+		register_post_type(
+			'no_title_cpt',
+			array(
+				'supports' => array( 'editor' ),
+			)
+		);
+		set_current_screen( 'no_title_cpt' );
```
dkotter, line 160: We need to ensure we clean up the screen in `tearDown` to avoid this leaking into other tests

dkotter (changes requested):
Overall this looks good, tested well. I've left a handful of comments but let me know if there's questions on any of those. In addition, we'll want to ensure we add proper documentation that matches all of our other experiment docs.

includes/Abilities/Slug_Generation/Slug_Generation.php
milindmore22: Agreed! Structured outputs are definitely better. I originally set it up this way because my local testing model didn't support JSON schemas. I'll update the code and push a new commit.

jeffpaul:
@milindmore22 are you able to make the updates from Darin's review above?

milindmore22:
@jeffpaul Yes, I’m addressing the PR reviews. The ones with 👍🏼 have been addressed, and I’ll work on the rest of them today.

jeffpaul (changes requested):
In the Slug suggestions modal, when the Regenerate button is loading let's ensure the text there matches other loading states (e.g. "Generating...").

dkotter (approved):

PR #898, opened 2026-07-26
Refactor experiments to leverage ai service layer

theaminulai:
## What?
Closes #233

Refactors the Excerpt Generation, Title Generation, Content Summarization, Alt Text Generation, and Image Generation (Generate_Image + Generate_Image_Prompt) abilities to route their AI calls through the `AI_Service` layer introduced in #101, instead of calling `wp_ai_client_prompt()` directly.

## Why?
With #101 merged and part of the 0.2.1 release, the various experiments/abilities still constructed prompt builders by calling the global `wp_ai_client_prompt()` function directly and duplicated model-preference logic themselves. Centralizing that through `AI_Service` gives us one place to manage default model configuration and keeps each ability's prompt-builder setup consistent.

## How?
- Added `AI_Service::create_imagegen_prompt()`, mirroring the existing `create_textgen_prompt()`, but seeded with `get_preferred_image_models()` instead of the text-generation model list, since image generation needs a different default model preference.
- Updated the following abilities to call `AI_Service::get_instance()->create_textgen_prompt( $prompt )` instead of `wp_ai_client_prompt( $prompt )` as the first call in their private `get_prompt_builder()` method:
  - `Excerpt_Generation`
  - `Title_Generation`
  - `Summarization`
  - `Alt_Text_Generation`
  - `Generate_Image_Prompt`
- Updated `Generate_Image` to call the new `AI_Service::get_instance()->create_imagegen_prompt( $prompt )`.
- Left the rest of each ability's builder chain (`using_system_instruction()`, `using_temperature()`, `with_file()`, `filter_prompt_builder()`, etc.) untouched, since `Abstract_Ability::set_provider_model_preference()` already handles developer provider/model overrides and reapplies the correct fallback model list (text vs. vision vs. image) after `AI_Service` seeds the default — so behavior is unchanged, only the entry point moved to the service layer.
- `Abilities Explorer` required no change: `Ability_Handler` only invokes arbitrary registered abilities generically via `wp_get_ability()->execute()` and never calls the AI client directly, so there was nothing to migrate there.
- Other experiments not listed in #233 (Editorial Updates, Editorial Notes, Content Classification, Type Ahead, Content Resizing, Meta Description, Suggest Reply, Comment Moderation) still call `wp_ai_client_prompt()` directly and are intentionally out of scope for this PR.

## Testing Instructions
1. Confirm no regressions in behavior — model preference, developer overrides, and prompt filters are unchanged:
   - `composer test` / run the PHPUnit integration suite, in particular:
[...]

PR #905, opened 2026-07-29
Deprecate the AI_Service layer

### LINKED ISSUE #233: Refactor experiments to leverage AI_Service layer

With #101 merged and part of the [0.2.1 release](https://github.com/WordPress/ai/releases/tag/0.2.1), we'll want to refactor the various experiments to leverage that layer.

- [ ] refactor Abilities Explorer to use AI_Service layer
- [ ] refactor Alt Text Generation to use AI_Service layer
- [ ] refactor Content Summarization to use AI_Service layer
- [ ] refactor Excerpt Generation to use AI_Service layer
- [ ] refactor Image Generation to use AI_Service layer
- [ ] refactor Title Generation to use AI_Service layer

theaminulai:
## What?
See #233

Deprecates the `AI_Service` class (`includes/Services/AI_Service.php`) and the `get_ai_service()` helper, rather than removing them outright.

## Why?
This started as an outright removal, since `AI_Service` (introduced in #101 / 0.2.1) isn't used anywhere in the plugin and a follow-up attempt to adopt it (#898) was closed as not wanted. @dkotter pointed out that both the class and the helper are public — third-party code could be relying on them — so removing them immediately would cause fatal errors for anyone using them. Deprecating first and removing in the next major release is the safer path, consistent with how this plugin already deprecates hooks in `includes/Deprecated.php`.

## How?
- `AI_Service::__construct()` now calls `_deprecated_class( self::class, 'x.x.x', 'wp_ai_client_prompt()' )`, matching how WordPress core signals class deprecation. The notice fires once per request via the singleton constructor rather than on every `get_instance()` call.
- `get_ai_service()` now calls `_deprecated_function( __FUNCTION__, 'x.x.x', 'wp_ai_client_prompt()' )` before returning the instance.
- Both carry `@deprecated x.x.x` docblock tags and updated descriptions pointing callers at `wp_ai_client_prompt()` directly, with a short snippet showing the one bit of behavior the helper added (applying `get_preferred_models_for_text_generation()` by default).
- `docs/ARCHITECTURE_OVERVIEW.md` now notes `AI_Service` as deprecated in the `Services/` folder description.
- `tests/Integration/Includes/Services/AI_ServiceTest.php` uses `setExpectedDeprecated()` around every call into the deprecated surface, so the suite fails if the deprecation notice goes missing *or* if an unexpected notice fires. The singleton is reset between tests via reflection so the notice (which the constructor only emits once per process) stays deterministic.
- Added a changelog entry under Unreleased/Deprecated; left the historical 0.2.1 "Added" entry untouched.

## Testing Instructions
1. Run the PHPUnit integration suite; `tests/Integration/Includes/Services/AI_ServiceTest.php` should pass and confirm deprecation notices fire correctly:
   - `composer test`
2. With `WP_DEBUG` enabled, manually call `WordPress\AI\get_ai_service()` or `AI_Service::get_instance()` from a test snippet and confirm a deprecation notice is logged, while the returned instance still works (`create_textgen_prompt()` still returns a usable prompt builder).
3. Confirm no other code in the plugin still calls `AI_Service` or `get_ai_service()` outside the deprecated files themselves.

## Changelog Entry

> Deprecated - The `AI_Service` class and the `get_ai_service()` helper introduced in 0.2.1 will be removed in the next major release. Neither is used anywhere in the plugin; experiments and abilities call `wp_ai_client_prompt()` directly. 

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-905-b05347cd9edb301ed6cf0c726749fd302d56fe12.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter (commented):
@theaminulai Thanks for the PR! In thinking about it a bit more, since the class is public and the helper function is also public, better to deprecate those and remove later than just remove entirely. Otherwise anyone that is relying on those will hit fatal errors.

I've pushed changes to handle that deprecation process but let me know if all of that looks good to you. In addition, would be great to get the PR title and description updated since we're not removing anything in this release, just deprecating

jeffpaul:
@theaminulai checking in to see if you're able to make those changes that Darin referenced above?

theaminulai:
Thanks for the nudge, @jeffpaul — and thanks for pushing those changes, @dkotter. Deprecating rather than removing outright makes sense given both are public.

I've reviewed the deprecation approach and it looks good to me — `_deprecated_class()` / `_deprecated_function()` calls, `@deprecated x.x.x` tags, and the rewritten test suite with `setExpectedDeprecated()` all match the conventions already used elsewhere in the plugin (`includes/Deprecated.php`, etc.).

One thing I noticed was missing: no changelog entry for the deprecation. Just pushed one under `Unreleased`/`Deprecated`. I've also updated the PR title and description to reflect that this deprecates `AI_Service` and `get_ai_service()` rather than removing them, since we're not shipping a removal in this release.

Ready for another look whenever you get a chance.

dkotter:
> One thing I noticed was missing: no changelog entry for the deprecation

We don't add changelog entries to that file on an individual PR basis; those should be added to the PR description and we pull those out when generating a release. I've updated the PR description with that changelog and pushed a commit to remove that from the file

dkotter (approved):

PR #914, opened 2026-08-01
Logging: add a public API for recording MCP tool and ability requests

### LINKED ISSUE #906: Request Logging: no public API to record the reserved `mcp_tool` and `ability` log types

The AI Request Logging experiment's read contract already declares three log types, but only one of them can ever exist.

- `AI_Request_Log_Controller::get_collection_params()` exposes `type` with `'enum' => array( '', 'ai_client', 'mcp_tool', 'ability' )` (`includes/Logging/REST/AI_Request_Log_Controller.php`).
- The client mirrors that contract — `LogEntry.type` is typed `'ai_client' | 'mcp_tool' | 'ability'` (`src/admin/ai-request-logs/types.ts`).
- But the only writer is `Log_Data_Extractor::extract_request_data()`, which hardcodes `'type' => 'ai_client'`. Nothing in the plugin ever produces an `mcp_tool` or `ability` row.

There is also no supported way for code outside the experiment to write one. The manager is private at every level:

- `AI_Request_Logging::$manager` and `::get_manager()` are `private`
- `Logging_Integration::$log_manager` is `private static` with no accessor

So a consumer's only option is to construct its own `new AI_Request_Log_Manager()`. That technically works, but it is the wrong shape: it bypasses the experiment-enabled check entirely, and calling `init()` on it re-runs `maybe_upgrade_table()` and re-negotiates the daily `wp_schedule_event()` cleanup cron. Skipping `init()` instead risks writing into a table that hasn't been upgraded. Either way the caller ends up coupled to internals that carry no backward-compatibility promise.

Two additive changes:

1. **A public entry point for recording a request**, so consumers don't construct their own manager — for example a `wpai_log_ai_request( array $data ): string|false` function, or a static accessor on `Logging_Integration`. It would no-op when the experiment is disabled, and otherwise delegate to the already-public `AI_Request_Log_Manager::log()`.
2. **Validate/normalize `type` against the same set the REST controller advertises**, so the write side and the read side can't drift apart.

Optionally, an action fired at write time would let consumers observe rows without polling the REST endpoint.

I'm happy to open a PR for this if the direction looks right.

The MCP Adapter is the obvious first consumer — an MCP tool call is exactly the kind of AI-initiated request a site owner wants in the log, and `mcp_tool` looks reserved with that in mind. The same applies to direct ability invocations, which today are invisible unless they happen to go out through the AI Client's HTTP transporter.

I maintain a plugin that exposes abilities over MCP, and I'd like its tool calls to appear in the site's existing AI Request Log rather than in a parallel log of my own. Currently that means reaching into private internals. If this isn't solved centrally, each MCP or ability surface will end up shipping its own logging table — which seems to be the outcome the reserved type values were meant to prevent.

Related but distinct: #732 covers requests that bypass the SDK HTTP transporter (a *provider transport* gap, with #757 in flight). This issue is about *log types* the schema already anticipates but nothing can produce. The two would compose well rather than overlap.

- [x] Yes

azizulhasan:
## What?
Closes #906

Adds a public API so consumers that surface abilities themselves — an MCP server, or code invoking an ability directly — can record requests in the AI Request Logging experiment's log, instead of reaching into the experiment's internals or shipping a parallel log of their own.

- Adds `WordPress\AI\log_ai_request()`.
- Adds `AI_Request_Log_Manager::get_types()` as the single source of truth for the supported log types, and derives the REST `type` enum from it.
- `AI_Request_Log_Manager::log()` now rejects an unsupported `type`.
- Fires `wpai_ai_request_logged` after a successful write.

## Why?

The log's read contract already declares three types, but only one of them could ever exist:

- `AI_Request_Log_Controller::get_collection_params()` exposes `type` with `enum => array( '', 'ai_client', 'mcp_tool', 'ability' )`.
- The client mirrors it — `LogEntry.type` is typed `'ai_client' | 'mcp_tool' | 'ability'` in `src/admin/ai-request-logs/types.ts`.
- But the only writer is `Log_Data_Extractor::extract_request_data()`, which hardcodes `'type' => 'ai_client'`. Nothing ever produced an `mcp_tool` or `ability` row.

There was also no supported way for anything outside the experiment to write one: `AI_Request_Logging::$manager` and `::get_manager()` are private, and `Logging_Integration::$log_manager` is `private static` with no accessor. A consumer's only option was to construct its own `AI_Request_Log_Manager`, which bypasses the experiment-enabled check and re-runs `maybe_upgrade_table()` plus the cleanup cron negotiation on `init()`.

The practical consequence: an MCP tool call that publishes a post is exactly the kind of AI-initiated request a site owner expects to find in **Tools → AI Request Log**, and it was invisible. Left unsolved centrally, every MCP or ability surface ends up shipping its own logging table.

## How?

**`WordPress\AI\log_ai_request()`** in `includes/helpers.php` returns the log ID on success, or `false` when the experiment is disabled, so callers can invoke it unconditionally. It follows the namespaced convention used throughout that file (`has_ai_credentials()`, `get_post_context()`, …) rather than the `wpai_`-prefixed name sketched in the issue.

**`Logging_Integration::get_log_manager()`** exposes the shared manager, returning `null` when nothing has initialised it.

**`AI_Request_Log_Manager::get_types()`** returns the supported types, and `AI_Request_Log_Controller::get_collection_params()` now builds its enum from it, so the write and read sides cannot drift apart.

**`AI_Request_Log_Manager::log()`** validates `type` and refuses anything outside that set with `_doing_it_wrong()`. This is a behaviour change on a public method and worth calling out: a row carrying a type the REST API cannot filter is unreachable through the UI, so it is refused at the point of writing rather than stored where nothing can retrieve it. Any existing caller passing a custom type would begin losing rows.

**One existing test changed.** `AI_Request_Log_ManagerTest::test_log_persists_entry()` passed `'type' => 'ui'`, which was never one of the advertised values — a row typed `ui` can never be filtered through the REST API. It is updated to `ai_client`. Happy to revisit if `ui` was intentional.

**Not included:** validation is limited to `type`. Normalising other fields felt like scope creep for this issue.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Investigating the gap, drafting the implementation and the integration tests, and running the verification described below. I reviewed every line, decided the API shape and naming, and tested the change end-to-end on a local WordPress 7.1 site against a real MCP server before opening this PR.

## Testing Instructions

**Setup:** enable the **AI Request Logging** experiment (the API intentionally no-ops while it is off).

**1. A row can be recorded, and the reserved type is reachable**

Drop this in an mu-plugin and load any admin page:

```php
add_action( 'admin_init', function () {
	\WordPress\AI\log_ai_request( array(
		'type'      => 'mcp_tool',
		'operation' => 'example/tool-call',
		'status'    => 'success',
	) );
} );
```

Go to **Tools → AI Request Log** — the entry appears. Confirm the type by requesting `/wp-json/ai/v1/logs?type=mcp_tool`; before this change that filter could never match a row.

**2. An unsupported type is refused**

Change `'type'` to `'not-a-real-type'`. With `WP_DEBUG` on you get a `_doing_it_wrong()` notice, the call returns `false`, and no row is written.

**3. It no-ops when the experiment is disabled**

Turn the experiment off and reload. The call returns `false` and nothing is written — no fatal, no notice.

**4. The write action fires**

```php
add_action( 'wpai_ai_request_logged', function ( $log_id, $data ) {
	error_log( "logged {$log_id} as {$data['type']}" );
}, 10, 2 );
```

**Integration tests:** `tests/Integration/Includes/Logging/AI_Request_Log_ManagerTest.php` and `Log_Ai_RequestTest.php` cover every supported type, rejection of unsupported and missing types, the action firing (and not firing when rejected), the inactive-experiment path, and that `get_types()` matches the REST enum.

**Real-world check:** I verified this against an MCP server exposing abilities on WordPress 7.1. On unpatched `develop` the MCP tool call ran and left no trace; with this branch the same call recorded an `mcp_tool` entry with a real duration, and a failing call recorded `status: error` carrying the actual `WP_Error` message.

> **Note:** I could not run PHPUnit locally (no Docker on this machine), so the integration tests here have their first execution in CI. `vendor/bin/phpcs` and `vendor/bin/phpstan analyse` (level 8) both pass clean.

## Screenshots or screencast

No UI changes. Entries render through the existing AI Request Log screen.

## Changelog Entry

> Added - Public `WordPress\AI\log_ai_request()` API so MCP servers and ability consumers can record requests in the AI Request Log.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-914-772f1ebbec58996a2d62befa635f12b7ca3eb6b7.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/experiments/ai-request-logging.md
+++ b/docs/experiments/ai-request-logging.md
@@ -25,6 +25,41 @@ The logging system uses the decorator pattern to wrap the SDK's HTTP transporter
 
 This approach uses the SDK's public API rather than reflection or internal hacks, making it resilient to SDK updates.
 
+## Public API
+Consumers that surface abilities themselves — an MCP server, or code invoking an ability directly — can record their own requests instead of shipping a parallel log.
+
+### `WordPress\AI\log_ai_request()`
+Writes an entry to the same log the experiment populates. Returns the log identifier on success, or `false` when the experiment is disabled or the write failed, so it is safe to call unconditionally.
+
+```php
+\WordPress\AI\log_ai_request( array(
+    'type'        => 'mcp_tool',       // Required. One of AI_Request_Log_Manager::get_types().
+    'operation'   => 'example/tool',   // Required. Non-empty string.
+    'status'      => 'success',        // Required. Non-empty string.
+    'provider'    => 'openai',
+    'model'       => 'gpt-5-nano',
+    'duration_ms' => 120,
+    'context'     => array( 'input_preview' => '…' ),
+) );
+```
+
+`type`, `operation`, and `status` are validated: anything else triggers `_doing_it_wrong()` and no row is written. `type` must be one of the values `AI_Request_Log_Manager::get_types()` returns — `ai_client` (generations made through the AI Client SDK), `mcp_tool`, or `ability`. A row carrying an unlisted type could never be filtered through the REST API, and `operation` and `status` are stored in columns that cannot be null.
+
+The REST `type` filter enum is derived from `get_types()`, so the read and write sides cannot drift apart.
+
+## Action Hooks
+
+### `wpai_request_logged`
+Fires after an entry is written, with the log identifier and the stored row. Lets consumers react to logged requests without polling the REST endpoint.
+
+```php
+add_action( 'wpai_request_logged', function( $log_id, $data ) {
+    if ( 'error' === $data['status'] ) {
+        // Alert on failed AI requests.
+    }
+}, 10, 2 );
+```
+
 ## Filter Hooks
 The logging system exposes several filter hooks for extensibility:
 
--- a/includes/Logging/AI_Request_Log_Manager.php
+++ b/includes/Logging/AI_Request_Log_Manager.php
@@ -180,23 +180,6 @@ class AI_Request_Log_Manager {
 
 		$log_id = $this->repository->insert( $data );
 
-		if ( false === $log_id ) {
-			return false;
-		}
-
-		/**
-		 * Fires after an AI request has been written to the log.
-		 *
-		 * Lets consumers react to logged requests without polling the REST
-		 * endpoint.
-		 *
-		 * @since x.x.x
-		 *
-		 * @param string              $log_id The identifier of the log entry.
-		 * @param array<string,mixed> $data   The logged data.
-		 */
-		do_action( 'wpai_ai_request_logged', $log_id, $data );
-
 		return $log_id;
 	}
 
--- a/includes/Logging/AI_Request_Log_Manager.php
+++ b/includes/Logging/AI_Request_Log_Manager.php
@@ -160,7 +160,59 @@ class AI_Request_Log_Manager {
 	 * @return string|false The log ID on success, false on failure.
 	 */
 	public function log( array $data ) {
-		return $this->repository->insert( $data );
+		$type  = $data['type'] ?? '';
+		$types = self::get_types();
+
+		if ( ! in_array( $type, $types, true ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				sprintf(
+					/* translators: 1: the supplied log type, 2: comma-separated list of supported log types. */
+					esc_html__( 'Unsupported log type "%1$s". Supported types are: %2$s.', 'ai' ),
+					esc_html( is_scalar( $type ) ? (string) $type : gettype( $type ) ),
+					esc_html( implode( ', ', $types ) )
+				),
+				'x.x.x'
+			);
+
+			return false;
+		}
+
+		$log_id = $this->repository->insert( $data );
+
+		if ( false === $log_id ) {
+			return false;
+		}
+
+		/**
+		 * Fires after an AI request has been written to the log.
+		 *
+		 * Lets consumers react to logged requests without polling the REST
+		 * endpoint.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param string              $log_id The identifier of the log entry.
+		 * @param array<string,mixed> $data   The logged data.
+		 */
+		do_action( 'wpai_ai_request_logged', $log_id, $data );
+
+		return $log_id;
+	}
+
+	/**
+	 * Gets the log types that may be recorded.
+	 *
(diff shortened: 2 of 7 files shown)


azizulhasan:
CI is green except for one E2E test — `settings.spec.js:264` › "Can turn on all experiments in a group" (188/189 passed). I think it's unrelated to this PR:

- The call log contradicts itself: the locator resolved 14 times to `<input checked type="checkbox" id="inspector-toggle-control-7" ...>` while the assertion reported `unexpected value "unchecked"`.
- It failed on a different toggle each attempt — `nth(1)`, then `.first()`, then `nth(1)` — rather than the same one consistently.
- This PR is PHP-only under `includes/Logging/` and `includes/helpers.php`; it touches no settings code, experiment registration, or JavaScript.

PHPCS, PHPStan and all 12 PHPUnit jobs (PHP 7.4–8.4, WP latest and trunk) pass.

I can't re-run the job from a fork — could someone with write access kick it off? Happy to dig further if it fails again in the same place.

dkotter:
> I can't re-run the job from a fork — could someone with write access kick it off? Happy to dig further if it fails again in the same place.

I've re-triggered that workflow so hopefully it passes. It is a flaky test, some fixes were added in https://github.com/WordPress/ai/pull/897 that we should maybe look to extract to it's own PR to get that merged in quicker

dkotter (approved):

PR #916, opened 2026-08-05
Content Classification: Dynamicize taxonomy error messages

Infinite-Null:
## What?
This PR updates the error and notice messages in the `Content_Classification` ability to dynamically display the specific taxonomy label (e.g., "category" or "tag") rather than generic hardcoded "taxonomy" terminology.

## Why?
When users interacted with specific taxonomy panels (such as "Suggest Categories" under the Categories sidebar panel), error notices displayed generic phrasing like *"No taxonomy suggestions were generated."* or *"You do not have permission to generate taxonomy suggestions."* This caused confusion as users were explicitly generating suggestions for categories.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 4.6

## Testing Instructions
1. Open a Post.
2. Open the **Categories** accordion panel in the right sidebar.
3. Trigger Category suggestion generation with empty content or under conditions with no results/no permission.
4. Verify the notice banner displays `"Content is required to generate category suggestions."` or `"No category suggestions were generated."` (instead of generic `"No taxonomy suggestions were generated."`).

#### Before

https://github.com/user-attachments/assets/e857b9bc-c954-4283-b920-17e5fd0868e1

#### After

https://github.com/user-attachments/assets/4eaadfcc-af24-4fe7-9736-58d02b397242

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-916-0549debefa8a3625cefdb7dfbd6b60ec21be74a5.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (dismissed):
Tests well on Playground, will await @dkotter's review before merging

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -744,4 +752,34 @@ private function get_top_terms( string $taxonomy, int $limit = 100 ): array {
...
+	 * @since x.x.x
+	 *
+	 * @param string $taxonomy The taxonomy slug.
+	 * @return string Lowercase taxonomy label (e.g., 'category', 'tag', 'taxonomy').
+	 */
+	private function get_taxonomy_label( string $taxonomy ): string {
+		$taxonomy = sanitize_key( $taxonomy );
+
+		if ( '' === $taxonomy || ! taxonomy_exists( $taxonomy ) ) {
+			return 'taxonomy';
```
dkotter: Not sure if this fallback is ever hit but if it is, this string isn't passed through translation now

includes/Abilities/Content_Classification/Content_Classification.php
```diff
@@ -744,4 +752,34 @@ private function get_top_terms( string $taxonomy, int $limit = 100 ): array {
...
+
+		if ( ! $tax_object instanceof WP_Taxonomy ) {
+			return 'taxonomy';
+		}
+
+		$label = ! empty( $tax_object->labels->singular_name )
+			? $tax_object->labels->singular_name
+			: ( ! empty( $tax_object->labels->name ) ? $tax_object->labels->name : $taxonomy );
+
+		return strtolower( trim( wp_strip_all_tags( (string) $label ) ) );
```
dkotter: Do we need `strtolower` here? I think in some languages the taxonomy name is likely better left capitalized so I'd remove that here to avoid that

tests/Integration/Includes/Abilities/Content_ClassificationTest.php
```diff
@@ -1446,4 +1446,24 @@ static function ( $prompt ) use ( &$captured_prompt ) {
 
 		$this->assertStringNotContainsString( '<available-terms>', $captured_prompt, 'Prompt should not contain available terms for allow_new strategy' );
 	}
+
+	/**
+	 * Test get_taxonomy_label returns expected singular labels.
+	 *
+	 * @since x.x.x
+	 */
+	public function test_get_taxonomy_label_returns_correct_labels(): void {
```
dkotter, line 1455: This test is fine but I think we can add some better/additional tests. For instance, we can add a test that actually invokes the callbacks via reflection to ensure the messages are accurate, like:

```php
$result = $method->invoke( $this->ability, array( 'taxonomy' => 'category' ) );
$this->assertStringContainsString( 'category suggestions', $result->get_error_message() );
```

We also don't have tests for the `no_results` error code with a non-default taxonomy, a situation where a missing or non-string taxonomy is passed to `permission_callback` (which is new behavior in this PR that defaults to `post_tag`), or a custom taxonomy that has a multi-word label

dkotter (changes requested):

Infinite-Null:
Thanks for the feedback! I’ve reworked the changes. Could you please review the updated changes when you get a chance. Thank you.

cc: @jeffpaul @dkotter

dkotter (approved):

PR #917, opened 2026-08-06
Content Resizing: Focus Accept button when suggested content generation completes

Infinite-Null:
## What?
Ensures focus moves directly to the **Accept** button in the Content Resizing modal once suggested content generation finishes, rather than defaulting focus to any focusable elements (such as links) rendered in the "Original" content section.

## Why?
If the original block text contains focusable HTML elements (e.g., hyperlinks), modal updates cause focus to land on those elements in the "Original" section.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 4.6

## Testing Instructions
Below is the video demonstrating the testing instructions.

https://github.com/user-attachments/assets/7805e550-66a4-433b-b4df-b867567c6698

### Before:

https://github.com/user-attachments/assets/e1010794-2e31-4046-abef-133f214e31fa

### After

https://github.com/user-attachments/assets/3c1f1e09-ddfb-439d-916d-1eb320132d9f

## Changelog Entry

> Fixed - Ensure focus moves to the Accept button when suggested content generation completes in the Content Resizing modal.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-917-074f1dd73952140a284ca0fc00f616e6955fcf30.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (dismissed):
Tests well via Playground, will await @dkotter's review before merging

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
```diff
@@ -83,7 +83,11 @@ export default function ContentResizingToolbar( {
 			block: 'start',
 			behavior: 'auto',
 		} );
-	}, [ isModalOpen, suggestedContent ] );
+
+		if ( ! isLoading && suggestedContent !== null ) {
+			acceptButtonRef.current?.focus();
```
dkotter: Correct me if I'm wrong but I believe doing `focus()` will automatically scroll an element into view, which we already do the line above: `acceptButtonRef.current?.scrollIntoView`. Likely not a problem but should maybe do:

```ts
acceptButtonRef.current?.focus( { preventScroll: true } );
```

to prevent conflicts between the two?

src/experiments/content-resizing/components/ContentResizingToolbar.tsx
```diff
@@ -83,7 +83,11 @@ export default function ContentResizingToolbar( {
...
+		}
+	}, [ isModalOpen, isLoading, suggestedContent ] );
```
dkotter, line 94: By adding `isLoading` here, we re-trigger the scroll on every loading transition, which is new behavior. Might be better to introduce a new effect for the focus state to avoid this

dkotter (changes requested):
Works fine but left a few minor comments.

Also would be nice to update our existing E2E test to ensure the right button is focused after generation, to ensure this issue doesn't come back (and likely something we've missed in our other E2E tests and focus fixes)

Infinite-Null:
Thanks for the feedback! I’ve reworked the changes. Could you please review the updated changes when you get a chance. Thank you.

cc: @jeffpaul @dkotter

dkotter (approved):

PR #930, opened 2026-08-12
Editorial Updates: Fix missed updates for Pullquote and other value-based blocks

yogeshbhutkar:
## What, Why, and How?

The `Pullquote` block can receive notes through the `Generate Editorial Notes` flow, but clicking `Apply Editorial Updates` generates a successful response without updating the block.

This happens because `Pullquote` stores its primary text in the `value` attribute ([Ref](https://github.com/WordPress/gutenberg/blob/cc8e2c958b252da7eec9ef8c91c3e982cff060a4/packages/block-library/src/pullquote/block.json#L10)). Editorial Updates previously handled Image blocks through `alt` and defaulted all other supported blocks to `content`, causing updates for Pullquote blocks to be written to the wrong attribute.

This PR detects whether a block uses `value` and writes the generated update to that attribute when present. This fixes Pullquote updates and supports other value-based blocks in the future.

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.6
Used for: Code reviews and drafting the e2e test.

## Testing Instructions

1. Enable the Editorial Notes and Editorial Updates experiments.
2. Create or edit a post.
3. Add a Pullquote block containing some text.
4. Generate an editorial note for the Pullquote.
5. Apply Editorial Updates.
6. Confirm that the Pullquote text is updated.
7. Save and reload the post.
8. Confirm that the updated Pullquote content persists.

### Before

https://github.com/user-attachments/assets/9e4e0770-8033-424d-9652-3d3917a045cb

### After

https://github.com/user-attachments/assets/f810bb53-b9c9-4342-be1a-b78749f19692

## Changelog Entry

> Fixed - Apply editorial updates to blocks that store editable text in the value attribute.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-930-b62b2bba05b50c5ba3e1672d35180f2ec6c25db9.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
+++ b/src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
@@ -53,6 +53,26 @@ interface Block {
 }
 
 /**
+ * Returns the attribute Editorial Updates should use for a block's primary
+ * editable text. Most text blocks use `content`, while blocks such as
+ * Pullquote use `value` and Image uses `alt`.
+ *
+ * @param block The block to inspect.
+ * @return The editable text attribute.
+ */
+function getEditableTextAttribute( block: Block ) {
+	if ( block.name === 'core/image' ) {
+		return 'alt';
+	}
+
+	if ( Object.hasOwn( block.attributes, 'value' ) ) {
+		return 'value';
+	}
+
+	return 'content';
+}
+
+/**
  * Updates a Note status to approved.
  *
  * @param noteId Note ID.
@@ -407,11 +427,8 @@ export function useEditorialUpdates(): {
 								refinedContent &&
 								refinedContent !== blockText
 							) {
-								// For heading and paragraph it's content, image is alt
 								const attributeToUpdate =
-									block.name === 'core/image'
-										? 'alt'
-										: 'content';
+									getEditableTextAttribute( block );
 
 								dispatch(
 									blockEditorStore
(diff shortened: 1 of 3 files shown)


src/experiments/editorial-updates/hooks/useEditorialUpdates.ts
```diff
@@ -52,6 +52,26 @@ interface Block {
...
 
+/**
+ * Returns the attribute Editorial Updates should use for a block's primary
+ * editable text. Most text blocks use `content`, while blocks such as
+ * Pullquote use `value` and Image uses `alt`.
+ *
+ * @param block The block to inspect.
+ * @return The editable text attribute.
+ */
+function getEditableTextAttribute( block: Block ) {
```
dkotter: Couple suggestions:

1. The `getBlockText` function we use elsewhere defaults to `content` and then `value`. Here we default to `value` then `content`. I think ideal would be to standardize the two so there's never a mismatch
2. I'd recommend moving this function into the `src/utils/blocks.ts` file alongside the other block functions (like `getBlockText`)

dkotter (changes requested):

dkotter (approved):

PR #932, opened 2026-08-12
Slug Generation: fix silent no-op and stale applied state for edited slugs

ekamran:
## Description

The Slug Generation experiment lets users hand-edit a suggested slug before applying it, in both the permalink modal and the pre-publish panel. Edits are normalized with `cleanForSlug()` on apply, and two problems came out of that:

1. Input that normalizes to an empty slug (for example `!!!`) was silently dropped. The modal still closed as if the slug had been applied, with no feedback anywhere, and the panel button did nothing on click.
2. The pre-publish panel compared the raw input against the stored slug. After applying an edited value like `My Cool Slug` the post correctly received `my-cool-slug`, but the button stayed on an enabled Apply forever and the Applied state was unreachable.

### Before

Modal, input that cannot become a slug, Apply closes the modal with nothing applied and no feedback:

before-modal-invalid

Panel, edited slug is applied but the button never reflects it:

before-panel-custom

### After

Modal stays open with the Apply button disabled and an explanation:

after-modal-invalid

Panel previews the normalized form and flips to a disabled Applied:

after-panel-custom

### How it is fixed

Everything rides on existing core pieces. `applySlug()` now returns the slug it applied, or an empty string when nothing was applied. A small shared helper describes how the current input will be applied, and both surfaces use it the same way:

- A help text under the field (the `help` prop of `TextControl`) previews the normalized form when it differs from the input, for example `Will be applied as "my-cool-slug".`
- When the input cannot become a slug, the help explains why and Apply is disabled.
- The Applied state in the panel is derived from the normalized form, so it is reachable for edited input, and the modal only closes when something was actually applied.

Percent-encoded slugs (how non-Latin suggestions are stored after `sanitize_title()`) are excluded from the preview, since `cleanForSlug()` does not account for octets and the preview would show a garbled form for a suggestion the user never edited. The underlying apply behaviour for those slugs is pre-existing and is now noted in the doc's Limitations section.

## Testing

- Three new e2e tests: blocked input keeps the modal open and the slug untouched even on a forced click, an edited slug shows the preview and applies the normalized form, and the panel blocks unusable input and flips to a disabled Applied. Full slug spec passes (8 tests).
- Verified by mutation: reverting either fix makes the matching test fail.
- Manually verified both surfaces in wp-env, including edge cases: empty input, spaces only, emoji only, script tags, percent-encoded input, literal percent text, Hindi text, re-editing after apply, selecting a radio after a custom edit, and regenerating with custom text typed.
- `npm run lint:js` and the TypeScript check pass. No PHP changes.

The docs page is updated to match, including the Manual Testing steps and a Limitations entry.

## Use of AI Tools

This PR was developed with the assistance of AI tools for local testing, code review and drafting, with all changes reviewed and verified by me.

## Changelog Entry

> Fixed - Slug Generation: applying a hand-edited slug now previews the normalized form, blocks input that cannot become a slug, and correctly shows the Applied state.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-932-4bbdf7079fbbc695d9306059ef658968e1fe3f2a.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/experiments/slug-generation.md
+++ b/docs/experiments/slug-generation.md
@@ -19,7 +19,7 @@ When enabled, the experiment adds two entry points for permalink suggestions:
 - Multiple suggestions per request (3 by default, configurable from 1 to 10)
 - Every suggestion is sanitized with `sanitize_title()` so it is always a valid WordPress slug
 - Suggestions are checked against existing content, so a suggestion that is already taken comes back with a numeric suffix
-- Hand-editable before insertion; edits are normalized with `cleanForSlug()`
+- Hand-editable before insertion; edits are normalized with `cleanForSlug()`, a help text below the field previews the normalized form, and input that normalizes to an empty slug cannot be applied
 - Gated behind a minimum content length (250 characters by default) so the model has something to work with
 - Language-aware — slugs use the language of the supplied title/content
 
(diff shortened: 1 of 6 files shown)


dkotter (approved):

dkotter:
Note that E2E tests aren't running right know due to WordPress 7.0.4 not being properly tagged on GitHub:

PR #935, opened 2026-08-13
Fix: make connector validity capability-aware instead of text-only

### LINKED ISSUE #933: Connector validity check assumes text generation, so speech-only and image-only connectors report as invalid
### Description

`has_valid_ai_credentials()` decides whether *any* configured connector is valid by asking a single question: can it generate text? A site whose only connector is a speech or image provider is therefore told its connectors may be invalid, when they are correctly configured and working.

The notice shown is:

> The AI plugin requires a valid AI Connector to function properly. Please review the AI Connectors you have configured to ensure they are valid.

That reads as "your connector is broken." The connector is fine; it simply does not do text.

**Expected:** the plugin explains that its features need a text-generation connector, and that the configured connector does not provide one.

**Actual:** a red error notice asking the user to review their connectors "to ensure they are valid."

Note the UI already distinguishes two cases — "no connectors configured" versus "configured but not valid" — and a non-text connector lands in the second. So the connector is detected correctly; it is the verdict and the wording that mismatch.

1. Install WordPress 7.0.4 and the AI plugin 1.2.0, with no AI connector configured.
2. Activate a connector plugin that registers a non-text capability only, and give it a working API key. I used an ElevenLabs provider registering `TEXT_TO_SPEECH_CONVERSION` and `SPEECH_GENERATION`.
3. Confirm in Settings → Connectors that the connector shows as Connected.
4. Visit the AI plugin settings screen.

The red "ensure they are valid" notice appears.

Run against the environment above, with the connector configured and authenticated:

```php
$p = wp_ai_client_prompt( 'Test' );
var_dump( $p->is_supported_for_text_generation() );            // false
var_dump( $p->is_supported_for_text_to_speech_conversion() );  // true
var_dump( WordPress\AI\has_valid_ai_credentials() );           // false
```

The connector is registered, authenticated, and usable for the capability it advertises. Only the text-generation question fails.

The verdict comes from `includes/helpers.php`, in `has_valid_ai_credentials()` (line 617 on `develop` at the time of writing):

```php
// See if we have credentials that give us access to generate text.
try {
    return wp_ai_client_prompt( 'Test' )->is_supported_for_text_generation();
} catch ( Throwable $t ) {
    return false;
}
```

That result reaches the settings screen as `PAGE_DATA.hasValidCredentials`.

### Why this may be worth separating

Most current features — alt text, summarisation, classification, comment moderation — genuinely do need text generation, so gating them is right. The issue is that a capability-specific requirement is reported as a global validity judgement.

This will affect more sites over time. The AI Client SDK defines capabilities for text-to-speech, speech generation, image generation, video generation, and embeddings, and connectors implementing the non-text ones are starting to appear. Any such connector, alone on a site, produces this notice.

There is precedent in this repository for a finer-grained approach: PR #679 ("Gate image generation UI on provider support detection") gates the image generation UI on whether a provider actually supports it, and PR #748 refined that detection. The global credentials check has not yet picked up the same capability awareness.

Possible directions, not prescriptive:

1. Keep `has_valid_ai_credentials()` as the "can this plugin do its text-based work" gate, but reword the notice to name the missing capability, e.g. "These features need a connector that generates text. Your configured connectors provide speech only."
2. Split the check: a connector is *valid* if it authenticates and registers any capability; *sufficient for a feature* is then decided per feature, the way image generation already is.
3. Report per-capability status on the settings screen, so a site can see that speech is available even while text is not.

Happy to work on a PR if there is a preferred direction.

### Environment info

- WordPress 7.0.4
- AI plugin 1.2.0. The same code is present on `develop`, so this is not specific to the released version.
- PHP AI Client 1.3.1 (bundled in core)
- Connector under test: a third-party ElevenLabs provider, v0.3.0
- Environment: `wp-env`, block theme (Twenty Twenty-Five), Chrome on macOS

### Confirmations

- **Searched existing issues:** yes. I looked for `valid connector`, `has_valid_ai_credentials`, and capability/connector wording. The closest existing work is about whether credentials are *present* (#197, #534, #731, #799) rather than about a valid connector of a non-text modality, and PRs #603, #679, #748 on capability detection. I did not find an existing report of this behaviour, but I may have missed one.
- **Tested with all plugins deactivated except the AI plugin:** not applicable, and I want to be straightforward about why rather than tick the box. This behaviour requires a second plugin — a connector — to be active. With every other plugin deactivated there is no connector at all, which produces the *other* branch of this notice ("Verify you have one or more AI Connectors configured"). The minimal reproduction is the AI plugin plus exactly one non-text connector.
- **Theme type:** Block.

---

*AI assistance: Yes. Tool(s): Claude Code (Claude Opus 5). Used for: reproducing the behaviour in a local wp-env environment, tracing it to `has_valid_ai_credentials()`, searching for duplicates, and drafting this report. The reproduction commands above were executed and their output is quoted verbatim.*

whyisjake:
## What?
Closes #933

Makes the connector validity check capability-aware, so a site whose only connector provides speech or image generation is no longer told its connectors may be invalid.

## Why?

`has_valid_ai_credentials()` decided whether *any* configured connector was valid by asking a single question — can it generate text:

```php
return wp_ai_client_prompt( 'Test' )->is_supported_for_text_generation();
```

That verdict reached the settings screen as `PAGE_DATA.hasValidCredentials` and rendered a red error: *"Please review the AI Connectors you have configured to ensure they are valid."*

A connector registering only `TEXT_TO_SPEECH_CONVERSION` / `SPEECH_GENERATION` is authenticated, registered, and working. It simply does not do text. The connector was detected correctly — the verdict and the wording were what mismatched.

This repository had already solved the narrower version of the problem: #679 gated the image generation UI on real provider support, and #748 refined that into `has_image_generation_support()`, a request-free scan over connector model metadata. The global credentials check never picked up the same capability awareness.

## How?

Direction 1 of the three the issue proposed — generalize the check and reword the notice — chosen because it is the smallest correct fix and mirrors the existing #679/#748 precedent. `has_valid_ai_credentials()` keeps its name, signature, return type, and semantics as the "can this plugin do its text-based work" gate.

- **`has_capability_support( string $capability, bool $reset_cache = false )`** generalizes the scan out of `has_image_generation_support()`, memoized per capability, with a new `wpai_has_capability_support` filter. `has_image_generation_support()` becomes a delegate and applies `wpai_has_image_generation_support` *after* the general filter, so existing third-party integrations stay authoritative for image generation.
- **`has_valid_ai_credentials()`** is re-backed with that scan for `TEXT_GENERATION`. The `has_ai_credentials()` early return and the `wpai_pre_has_valid_credentials_check` short-circuit keep their exact positions. Its docblock now states that it is a capability-specific gate, not a judgement about connector validity.
- **`get_supported_capabilities()`** reports what the connectors *can* do, and is exposed to the settings page as `capabilities` alongside `hasCredentials` and `hasValidCredentials`. Together the three signals distinguish "no connectors" from "connectors that don't generate text".
- **The settings notice** names the requirement and what the connectors provide, at `warning` rather than `error` — nothing is broken. The "no connectors configured" branch is unchanged and stays an `error`. Capability values map to human-readable labels in the UI; unrecognized values are dropped so a raw identifier can never reach user-visible copy, and an empty or absent list falls back to a capability-free sentence.
- **`wp ai alt-text generate`** splits its message the same way.

`CapabilityEnum` is a constants class extending `AbstractEnum`, not a native PHP enum, so the helper takes the constant's `string` value rather than an enum instance.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: tracing the behaviour to `has_valid_ai_credentials()`, drafting the implementation plan, implementing the change, and writing the tests.

## Testing Instructions

Reproducing the original report needs the AI plugin plus exactly one non-text connector. The e2e suite simulates that without shipping a speech provider, via a capability override endpoint added to the `e2e-testing` support plugin.

Automated:

```sh
npm run wp-env:test start
npm run test:php
npm run test:e2e -- tests/e2e/specs/admin/connector-capabilities.spec.js
```

Manual:

1. Configure any connector with a working API key.
2. Force it to advertise a non-text capability, e.g.
   `add_filter( 'wpai_has_capability_support', fn( $s, $c ) => 'speech_generation' === $c, 10, 2 );`
3. Visit Settings → AI. The notice should name text generation as the requirement and speech generation as what is available, and should not say the connectors may be invalid.
4. Remove all connector credentials and reload — the original "Verify you have one or more AI Connectors configured" error should be unchanged.

Tests executed for this PR:

- Full PHPUnit suite — 1395 tests, 0 failures, 39 skipped
- `HelpersTest` — new `has_capability_support`, `get_supported_capabilities`, and `has_valid_ai_credentials` coverage; the 7 pre-existing `has_image_generation_support` tests pass with unmodified bodies
- `Settings_PageTest::test_script_module_data_includes_capability_status`
- `Alt_Text_CommandTest` — 28 tests
- Playwright `connector-capabilities.spec.js` (3) and `settings.spec.js` (17), 20/20 on a clean environment
- `phpcs`, `phpstan`, `eslint`, `tsc` — all clean

## Screenshots or screencast

Both captured on a speech-only connector.

| Before | After |
| ------ | ----- |
|notice-before        |notice-after       |

## Changelog Entry

> Fixed - Connectors that provide only non-text capabilities, such as speech or image generation, are no longer reported as possibly invalid. The settings screen and the `wp ai alt-text` command now explain that a text-generating connector is required and name what the configured connectors do provide.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-935-9b1d7285c58bf4378b3846d8f256578a665cc3e4.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter:
We were already considering just removing the `has_valid_ai_credentials` function as we really only use that right now to show an admin notice (as the attached Issue mentions) and that doesn't really seem to be that helpful.

So in looking at this PR, I guess where I'm struggling is if it's worth 1000+ lines of code to make that error message more useful or if the right answer is to just get rid of that entirely. If we were actually gating functionality based on that check we'd obviously want that to be more robust but when it's just used for an admin notice, trying to decide if this is worth keeping or not. Any thoughts?

whyisjake:
Fair question. I went and checked what actually calls it, and the picture is different from what I'd assumed too.

`has_valid_ai_credentials()` has two non-notice callers on develop:

- `includes/CLI/Alt_Text_Command.php:94` hits `WP_CLI::error()` when it returns false, which exits. So a site whose only connector does speech or images gets `wp ai alt-text generate` aborted with "No valid AI credentials found."
- `includes/Settings/Settings_Page.php:84` passes it to AI Home, where `routes/ai-home/stage.tsx:1159` gates rendering on `! PAGE_DATA.hasValidCredentials`.

I think the "only used for an admin notice" description fits `has_ai_credentials()`, the one without "valid". That backs the dashboard widget and Site Health, both display surfaces. The two names are one word apart and I mix them up constantly.

On size: 1,083 of the 1,298 added lines are tests. Production is 215, across `helpers.php` (116), `stage.tsx` (83), `Settings_Page.php` (36) and `Alt_Text_Command.php` (7).

So the trade is 215 lines of production code against a CLI command that exits on a working connector. Your words were that if we were "actually gating functionality based on that check" you'd want it more robust. The CLI path is that case.

One more input: #1004 wires `Prompt_Model_Client::supports_text_generation()` to this function, so the AI Workspace uses it to decide whether a turn can run at all. That PR went ready for review today.

Removing the function is still a reasonable call and I'll take it if you'd rather. It isn't free, though. You'd need to decide what `wp ai alt-text` does when nothing text-capable is configured, and what AI Home renders in that state.

If you want this smaller instead, the capability-aware helper without the settings-copy rework gets most of the value in a much shorter diff. Which way do you want to go?

dkotter:
> `includes/Settings/Settings_Page.php:84` passes it to AI Home, where `routes/ai-home/stage.tsx:1159` gates rendering on `! PAGE_DATA.hasValidCredentials`.

So this isn't fully accurate. Yes we pass `hasValidCredentials` here but it's only used to render the notice, it does not gate the rendering of the settings page.

It is used in the Alt Text WP-CLI command but that should have never been used there, as you've discovered that only checks for text generation and when generating alt text, we care about vision capabilities. So that check was never useful there in the first place (so I'd recommend removing it).

I'd personally still lean to just removing this check (probably removing the function as well or at least deprecating it). But we are considering some onboarding changes (see https://github.com/WordPress/ai/issues/994) and depending on what direction we go there, we may want some version of this (for example, to provide specific instructions depending on if you have a valid connector in place vs not having one in place).

So I think for now, I'd leave this as-is and we can re-evaluate once we have some sort of onboarding in place

PR #941, opened 2026-08-14
Content Translation: Add user-triggered retry

yogeshbhutkar:
## What, Why, and How?

This PR adds a “Retry” action to the warning notice shown when a title or block translation fails.

For example, if 2 out of 10 blocks fail, the user can retry only those failed blocks instead of translating the entire post again. The PR tracks the client IDs of failed blocks and uses them to retry only the affected translations.

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.6 Sol
Used for: Code reviews and implementation suggestions.

## Testing Instructions

1. Create a post with a title and several paragraphs, then generate a translation.
2. Simulate one or more block translation failures and verify a "Retry" action appears.
3. Click "Retry" and confirm only the failed blocks are translated.
4. Simulate a title translation failure and confirm Retry title translation retries only the title.

## Screencast

https://github.com/user-attachments/assets/e4968fef-a60d-4345-93f5-b0a311f6b4da

## Changelog Entry

> Added - User-triggered retry for Content Translation.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-941-ff9b0dd3c7dba0195bafc56df939b96e5b22783b.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/experiments/content-translation/hooks/useContentTranslation.ts
+++ b/src/experiments/content-translation/hooks/useContentTranslation.ts
@@ -5,7 +5,7 @@ import { useState } from '@wordpress/element';
 import { store as blockEditorStore } from '@wordpress/block-editor';
 import { store as editorStore } from '@wordpress/editor';
 import { store as noticesStore } from '@wordpress/notices';
-import { useDispatch, useSelect } from '@wordpress/data';
+import { select, useDispatch, useSelect } from '@wordpress/data';
 import { __, _n, sprintf } from '@wordpress/i18n';
 
 /**
@@ -37,6 +37,7 @@ type UseContentTranslationReturn = {
 
 type TranslateOptions = {
 	translateTitle?: boolean;
+	targetClientIds?: readonly string[] | undefined;
 };
 
 // Notice IDs for the content translation process.
@@ -59,13 +60,9 @@ export function useContentTranslation(): UseContentTranslationReturn {
 
 	const { minContentLength } = getSettings();
 
-	const { postId, allBlocks, title, content } = useSelect( ( sel ) => {
+	const { postId, content } = useSelect( ( sel ) => {
 		return {
 			postId: sel( editorStore ).getCurrentPostId() as number,
-			allBlocks: sel( blockEditorStore ).getBlocks(),
-			title: sel( editorStore ).getEditedPostAttribute(
-				'title'
-			) as string,
 			content: sel( editorStore ).getEditedPostContent(),
 		};
 	}, [] );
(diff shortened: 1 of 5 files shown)


src/experiments/content-translation/hooks/useContentTranslation.ts
```diff
@@ -78,20 +75,24 @@ export function useContentTranslation(): UseContentTranslationReturn {
...
 		languageCode: string,
 		options?: TranslateOptions
 	) => {
-		const { translateTitle = false } = options || {};
+		const { translateTitle = false, targetClientIds = undefined } =
+			options || {};
 
 		// Remove any existing error notices.
 		noticeDispatch.removeNotice( TRANSLATION_NOTICE_ID );
-		noticeDispatch.removeNotice( TRANSLATION_NOTICE_ID_TITLE );
```
dkotter, line 93: So this is likely an edge case but since we no longer clear all the notices here, there's a potential bug here:

1. Trigger translation, including the title
2. If this fails, an error notice is shown to retry title translation
3. If I don't click the retry button but instead click the sidebar translation button again and choose a different language and choose to not translate the title this time, the existing failure notice sticks around. If I then click the retry button, it now translates into the first language and not the second language

Again, this is likely very edge case so may not be worth fixing

src/experiments/content-translation/hooks/useContentTranslation.ts
```diff
@@ -177,6 +187,18 @@ export function useContentTranslation(): UseContentTranslationReturn {
...
 				id: TRANSLATION_NOTICE_ID_TITLE,
+				actions: [
+					{
+						label: __( 'Retry title translation', 'ai' ),
+						onClick: () => {
+							if ( ! ensureProvider( TRANSLATION_NOTICE_ID ) ) {
+								return;
+							}
+
+							translatePostTitle( languageCode );
```
dkotter: The retry action here triggers `translatePostTitle` directly and thus we never run `setIsTranslating( true )`. This means the busy state is never triggered and someone could trigger translation again while the first is still loading

dkotter (changes requested):

src/experiments/content-translation/hooks/useContentTranslation.ts
yogeshbhutkar: Thanks for the review and apologies for missing this. IMO, although it’s an edge case, it could trigger an unnecessary API call and consume tokens (considering the generation might not be relevant anymore), so I'm inclined to fix this. Please let me know if you think we should handle this differently.

I’ve updated the implementation to use a single retry action for both title and block translation failures. Separate retry buttons felt confusing because translation begins as a single user action.

Also, notices are now standardized. Configuration, input, or validation problems produce an error notice (or if the data prevents translation from starting), so the user can resolve the issue and initiate translation again. Warnings report issues encountered during processing, such as skipped blocks in a batch due to size limitations or failed translation requests.

dkotter (dismissed):

dkotter:
The E2E tests keep timing out on this PR. I bumped that timeout up a bit and can continue to bump that up if needed but seems like something else must be going wrong here. I ran these tests locally and they all pass fine so some additional investigation is needed

yogeshbhutkar:
I've tried both the previous 30-minute timeout and the new 35-minute timeout, and the tests seem to pass in both cases. I've kept the timeout at 35 minutes since we've come quite close to hitting the 30-minute mark in Actions on a few other PRs.

Regarding the previous 35-minute timeout failure here, the checkout step itself took [6m 14s](https://github.com/WordPress/ai/actions/runs/32404306714/job/96539636834), whereas it normally completes in under 10 seconds. The run before that also seemed unusually slow overall, so this may have been an issue with the runner.

The newly added test, in particular, passes consistently when run locally.

dkotter (approved):

PR #945, opened 2026-08-16
Add core/edit-content ability for targeted post content edits

seedprod:
## What?
See #40.

Adds a `core/edit-content` Ability — the write twin of `core/read-content` (#739). It applies a server-side, exact-match text replacement to a post's title, excerpt, or content, instead of requiring clients to round-trip the whole field.

## Why?
Whole-field writes force an AI client to fetch an entire content blob, mutate it in context, and send it all back — expensive, and every untouched byte the model re-emits is a byte it can corrupt. A server-side anchored patch means only the changed snippet ever transits the client. This is the editing contract AI coding tools have broadly converged on (exact string, unique match, fail loudly), applied to WordPress content with capability checks and revisions.

## How?
- `old_content` is matched byte-for-byte against the stored raw value; the match count must equal `expected_matches` (default 1) or the edit is refused with the actual count. No regex, no fuzzy matching — a stale or wrong snippet fails loudly rather than editing the wrong place.
- Every refusal (unknown/unexposed post, unsupported field, serialized value, no match, count mismatch) leaves the post unchanged.
- Persists via `wp_update_post()`, so revisions and normal save filters apply. The response reports `replaced`, `status`, `modified_gmt`, and an `exact_persistence` flag (false when save filters such as KSES altered the saved value) — deliberately not the full field value.
- Permission gate: authenticated + post type exposed via `show_in_abilities` + `current_user_can( 'edit_post' )`, resolved before execution; all denials are uniform.
- Registers through its own gated class (`Gated\Edit_Content`) rather than inside `Read_Content`, so write access is an individually removable unit via the `wpai_gated_abilities` filter. Maintainers may prefer a separate experiment for write abilities — open to that.

Known limitations (documented in code): the match check and save are not atomic (patch-on-snapshot, not compare-and-swap); `is_serialized()` guards whole-value serialization only; the input schema's post type enum snapshots exposure at registration (same as read). Natural follow-up: a bounded raw search/snippet ability so clients can locate exact anchors without reading whole fields.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code (Fable); Codex CLI (independent design review)
Model(s): Claude Fable 5, GPT-5.6-sol
Used for: Implementation and tests, built to my specification and design decisions; the design and final patch were adversarially reviewed with independent AI reviewers before submission. I take responsibility for this contribution.

## Testing Instructions
2. `npm run test:php -- --filter EditContentTest` (18 tests: happy paths for all three fields with revision assertions, every refusal code, capability and exposure denials, serialized refusal, literal `$`/`\` handling, fail-closed ID parsing, expected-match semantics)
3. Full suite: `npm run test:php` — 1391 tests passing locally; `composer lint` and `composer phpstan` clean.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-945-6c5b185fe75cdb78f050dadcd37c50f94c75200a.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Content/Content.php
+++ b/includes/Abilities/Content/Content.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * The `core/read-content` WordPress Ability.
+ * The `core/read-content` and `core/edit-content` WordPress Abilities.
  *
  * @package WordPress\AI
  *
@@ -26,6 +26,13 @@ defined( 'ABSPATH' ) || exit;
  * readable post by ID or by post type and slug, or querying multiple readable posts filtered
  * by post type, status, author, parent, or included IDs. Raw fields are only returned for
  * posts the current user can edit.
+ *
+ * Also provides the `core/edit-content` ability, which applies a server-side text
+ * replacement to the title, excerpt, or content of a single post the current user can
+ * edit, saving through the standard update flow so a revision is created when the post
+ * type supports revisions. The edit ability registers separately through its own gated
+ * class (see {@see self::init_edit()}), keeping write access an individually gateable
+ * unit.
  *
  * This class is kept almost identical to the WordPress core class `WP_Content_Abilities`
  * so the two implementations stay in sync. Differences from the core class are marked with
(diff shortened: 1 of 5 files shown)


jeffpaul:
@jorgefilipecosta I'll leave this one for you to review/approve/close based on the updated abilities architectural approach you're putting in place

PR #947, opened 2026-08-18
Fix: Increase the default http timeout for `wp/v2/settings` for `put` and `post` request

### LINKED ISSUE #940: Introduce new custom endpoint for settings save and by pass core /wp/v2/settings endpoint for saving AI options
### Description

Extracting the issue from here - https://github.com/WordPress/ai/pull/881#issuecomment-5269459468

Upstream Core Trac - https://core.trac.wordpress.org/ticket/65867
Upstream PR - https://github.com/WordPress/wordpress-develop/pull/13031

Needs discussion on wether to change the settings endpoint and create new endpoint based in AI plugin for settings.

1. On a site with Connectors and an AI provider plugin active, configure and save a valid AI provider API key.
2. Make any POST/PUT to /wp/v2/settings that does not include that key — e.g. change the Site Title, or save an unrelated plugin's settings through the same endpoint.
3. Observe that Core revalidates the stored AI key (a provider network call fires).
4. If that validation returns anything other than true (provider down, rate limited, transient error), the stored key is overwritten with an empty string, even though it was never submitted.

- None

### Environment info

- WordPress latest.
- AI plugin latest

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/940

## Why?

- PR resolves the issue where settings save is removing the `openAI` key while checking the authentication on each save request, which sometimes throws the timeout error.

## How?

- PR adds the support of increased http timeout on filter which only added when the route called is `wp/v2/settings`.

### Use of AI Tools

- Yes, Claude Code, Opus 5.
- Used for timeout implemenation and for unit tests to add all scenarios.

## Testing Instructions

1. Add the openAI key in connectors screen.
2. Go to `Settings --> AI`.
3. Quickly update some settings multiple times.
4. Check for response, it should not remove the openAI key.

## Changelog Entry

> Fixed - Prevent stored AI provider API keys from being cleared during unrelated `/wp/v2/settings` saves by extending the HTTP timeout while core revalidates them.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-947-58839c5bc295d271ddc7e28c5d59f292674930cf.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Settings/Settings_Registration.php
```diff
@@ -75,6 +75,46 @@ public function init(): void {
...
+	/**
+	 * Returns a longer HTTP timeout, never lowering an existing one.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param mixed $timeout The current timeout in seconds.
+	 * @return int The timeout to use, in seconds.
+	 */
+	public function extend_revalidation_timeout( $timeout ): int {
+		return max( (int) $timeout, 30 );
```
dkotter: Looking at the Core `http_request_timeout` filter, it claims `$timeout_value` is a `float`, but we cast to `int` here. Should we change that to cast to `float`?

includes/Settings/Settings_Registration.php
```diff
@@ -75,6 +75,46 @@ public function init(): void {
...
+	 * @param mixed $server   The REST server instance.
+	 * @param mixed $request  The REST request.
+	 * @return mixed The unchanged response.
+	 */
+	public function maybe_extend_revalidation_timeout( $response, $server, $request ) {
+		if ( $request instanceof \WP_REST_Request
+			&& '/wp/v2/settings' === $request->get_route()
+			&& in_array( $request->get_method(), array( 'POST', 'PUT' ), true )
+		) {
+			add_filter( 'http_request_timeout', array( $this, 'extend_revalidation_timeout' ) ); // phpcs:ignore WordPressVIPMinimum.Hooks.RestrictedHooks.http_request_timeout
```
dkotter, line 104: We add this filter here but never remove it, so if there happens to be other requests that fire in the same session, they'll all inherit this timeout. We should likely try and remove this filter after this particular request runs, if possible, something like:

```php
add_filter( 'rest_post_dispatch', array( $this, 'restore_default_timeout' ), 11, 3 );

public function restore_default_timeout( $response, $server, $request ) {
      remove_filter( 'http_request_timeout', array( $this, 'extend_revalidation_timeout' ) );
      return $response;
}
```

dkotter (changes requested):

includes/Settings/Settings_Registration.php
hbhalodia: Sure, thanks. We can cast it to float type.

includes/Settings/Settings_Registration.php
hbhalodia: Resolved in - https://github.com/WordPress/ai/pull/947/changes/58839c5bc295d271ddc7e28c5d59f292674930cf

includes/Settings/Settings_Registration.php
hbhalodia: Resolved in - https://github.com/WordPress/ai/pull/947/changes/58839c5bc295d271ddc7e28c5d59f292674930cf

dkotter (approved):

PR #948, opened 2026-08-18
Order the core/read-users collection by display name

gziolo:
## What?

Split out of #931 so it can be reviewed on its own.

Two small changes to `core/read-users` and one test change for `core/read-content`:

1. `core/read-users` collections are now ordered by display name, ascending, and the output schema says so.
2. The comments on `test_include_order_does_not_control_the_result_order` explain the new order.
3. `test_query_lean_projection_does_not_prime_the_post_meta_cache` becomes `test_query_lean_projection_does_not_request_post_meta_priming` and checks what the ability asks for.

## Why?

**Collection order.** `core/read-users` returned users in whatever order `WP_User_Query` picked by default, which is by user login. The output schema documented no order at all, so a caller had no way to know what to expect, and the order could change under them. Ordering by display name matches `/wp/v2/users`, and it is the name a person actually sees in the response. Setting it here and writing it into the schema turns an accident into a promise.

**The lean projection test.** The old test counted the post meta queries that ran after the request and asserted zero. That measures the wrong layer. Whether `update_post_meta_cache` is honored belongs to whoever runs the query, and that layer is free to prime more for its own reasons. What this ability owns is the request it builds, so that is what the test now checks.

## How?

`core/read-users` sets `orderby => display_name` and `order => ASC` on the collection query, and the `users` property description in the output schema gains "ordered by name, A to Z".

The include list keeps working the way it did. It selects which users come back, it does not order them, which is also what the REST users controller does unless a caller asks for the include order.

`test_query_lean_projection_does_not_request_post_meta_priming` hooks `pre_get_posts`, picks out the query the ability built (the one with a non-empty `post__in`), and asserts `update_post_meta_cache` is off on it.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Extracting these changes from #931, and drafting this description. Verified locally against the test suite. I reviewed and edited the result, and I take responsibility for it.

## Testing Instructions

2. `npm run test:php` — the full suite passes (1373 tests, 3914 assertions, 44 skipped).
3. `npm run lint:php` and `npm run lint:php:stan` — both clean.

To confirm the rewritten test still guards the behavior, make `Content::should_prime_post_caches()` return `true` and run it again. It fails.

## Changelog Entry

> Changed - The `core/read-users` ability now returns collections ordered by display name, A to Z, and the output schema documents that order.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-948-e6b325750c7fb5d489e6d04d3bd16f13eb70e6f7.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Users/Users.php
+++ b/includes/Abilities/Users/Users.php
@@ -204,16 +204,22 @@ final class Users {
 		$per_page = $this->normalize_per_page( $input );
 		$page     = isset( $input['page'] ) ? max( 1, $this->input_int( $input['page'] ) ) : 1;
 
+		// Collections are ordered by display name, ascending, matching the REST users
+		// controller. The output schema documents that order, so it is set here rather than
+		// left to a query default.
 		$query_args = array(
 			'number'      => $per_page,
+			'orderby'     => 'display_name',
+			'order'       => 'ASC',
 			'offset'      => ( $page - 1 ) * $per_page,
 			'count_total' => true,
 		);
 
 		$include = $this->normalize_include( $input );
 		if ( array() !== $include ) {
-			// The include order is not applied as `orderby`. Keeping the default
-			// ordering lets WP_User_Query share cached results with other queries.
+			// The include list selects which users are returned; it does not order them.
+			// This mirrors the REST users controller, which orders by the include list only
+			// when a caller asks for it.
 			$query_args['include'] = $include;
 		}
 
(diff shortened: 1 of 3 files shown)


jorgefilipecosta (approved):
LGTM 👍

dkotter (approved):

PR #211, opened 2026-02-04
Add Service Account experiment

Jameswlepage:
## What?

Exploring the concept of a true "service account"; something a real user can create as a way for a service, like Claude Code, an automation tool, or an API can use to interact with WP in a safe, auditable manner. It's an actual new user type, and I've hacked a bit to get some of the UI "core-like". If it moves closer to a core feature, we'll need to rethink but for now, I'm interested in thinking around ... "do we need this"? 

- Adds the Service Account experiment (role, admin UI, REST endpoints).
- Registers the experiment in the loader.
- Adds integration coverage and documentation.

## Why?
Service accounts enable automation and non-interactive access while keeping user activity traceable in logs and revisions. This is core-aligned functionality that needs dedicated UX and API support.

## How?
- Introduce the Service Account experiment classes and manager.
- Add admin UI customizations for service account creation/editing.
- Add REST endpoints for service account CRUD and app-password regeneration.
- Add docs and integration tests.

## Testing Instructions
1. Enable the Service Accounts experiment.
2. Go to Users > Add New, select the Service role, enter a service account name.
3. Confirm username/email auto-populate; create the user.
4. In Users list, verify the service account shows and the Service Accounts view has an accurate count.
5. Edit the service account and create an Application Password in the section below.
6. REST: `POST /wp/v2/service-accounts`, `GET /wp/v2/service-accounts`, `POST /wp/v2/service-accounts/{id}/app-password`.

### Testing Instructions for Keyboard
1. Use `Tab` to reach the Role dropdown on Users > Add New.
2. Use arrow keys to select Service role, `Tab` to the Service account name field.
3. Type a name, `Tab` through username/email (read-only) and password.
4. `Tab` to submit and press `Enter`.
5. On the edit screen, `Tab` to the Application Passwords section and generate a password.

## Screenshots or screencast
CleanShot 2026-02-04 at 14 06 53@2x

CleanShot 2026-02-04 at 14 07 25@2x

[...]

PR #961, opened 2026-08-20
Feat[Experiment]: Add Agent Users, dedicated accounts for AI agents

### LINKED ISSUE #923: Introduce agent users to give AI agents an auditable identity

Today, AI agents interact with WordPress exclusively by borrowing a human user's identity. Whether through the MCP Adapter, the Abilities API REST endpoints, or WP-CLI (`--user`), an agent authenticates with a user's credentials, typically an application password, and every ability's `permission_callback` runs against that user. This works, but it has a structural gap:

* Agents always run on behalf of users. There is no other option, even for external or autonomous agents where no human is present at execution time.
* Access to abilities is controlled through permission checks (`current_user_can()` via `permission_callback`). That is the right enforcement layer, but it can only answer what a given user may do, never what a given agent may do.
* There is no way to distinguish users from agents. Not in `post_author`, not in activity logs, not in the new `wp_ability_invoked` hook ([Abilities API improvements in 7.1](https://make.wordpress.org/core/2026/07/31/abilities-api-improvements-in-wordpress-7-1/)). A site owner cannot answer what the agent changed yesterday, or revoke an agent's access without touching a human account.

This has come up repeatedly in recent discussions. The [August 5 core AI chat](https://make.wordpress.org/ai/2026/08/06/ai-contributor-weekly-summary-5-august-2026/) leaned toward representing agents as a type of user, partly because that is where regulation around AI disclosure is heading. The trust ladder framing @JasonTheAdams presented there, with gradual and task-specific autonomy backed by observability and recoverability, requires knowing who the agent is before any of its levels can be enforced. Slack discussions in #core-ai about permission sets for agents point the same way, both as an enterprise requirement and as a trust-building measure for smaller sites. The currently documented workaround of [manually creating a dedicated limited user for MCP access](https://developer.wordpress.org/news/2026/02/from-abilities-to-ai-agents-introducing-the-wordpress-mcp-adapter/) demonstrates both the demand and the gap. Those are ordinary users, with a login, full role privileges, and no marker that they are automated.

Extend the existing user concept rather than inventing a parallel identity system:

* Way to designate a user account as an agent, provisioned deliberately, marked as such, and visible as an agent everywhere attribution matters.
* No interactive login (application passwords or a similar mechanism instead), excluded by default from user listings and author dropdowns, plus other constraints appropriate for automated actors.
* Existing roles adapted where needed. Agents inherit the role system as their capability ceiling, so an "editor agent" is immediately understandable, with adjustments where human role defaults are not appropriate for automated actors.
* Dual usage – the current model stays fully valid, and agents acting interactively on behalf of a logged-in user keep working exactly as today. External and autonomous agents, such as remote MCP clients and scheduled jobs, get explicit agent accounts that can be granted, audited, and revoked independently of any human account.

This builds directly on what has already shipped. The `permission_callback` remains the enforcement boundary, [`wp_get_abilities()` filtering](https://make.wordpress.org/core/2026/08/05/filtering-registered-abilities-with-wp_get_abilities-in-wordpress-7-1/) provides the substrate for exposing only the abilities an agent's role permits, and `wp_ability_invoked` provides the audit hook that an agent-aware actor model would make meaningful.

### Related

* #354 covers unified abilities exposure controls, where agent users add the missing dimension of who is consuming.
* #736 covers role and user access controls per feature.
* WordPress/mcp-adapter#176 and WordPress/mcp-adapter#228 cover approval and audit workflows, both of which need actor attribution.
* The [Application Passwords integration guide](https://make.wordpress.org/core/2020/11/05/application-passwords-integration-guide/) has listed credential scoping as future work since the 5.6 merge.

Intentionally out of scope for this issue: credential improvements such as expiry and OAuth, trust levels and autonomy tiers, and approval workflows. Each deserves its own discussion once the identity question is settled.

_No response_

- [x] Yes

gziolo:
## What?

See #923.

This PR adds the experimental **Agent Users** feature. It gives external software—AI agents, MCP clients, scheduled jobs, and similar tools—dedicated WordPress identities instead of requiring them to borrow a human account.

This PR covers the identity model. Richer audit trails and provenance can build on that identity separately.

## Why?

An external agent currently has two poor identity choices: reuse a human credential, which mixes attribution and revocation, or run without a user, which loses authorship and produces surprising capability behavior.

An agent account reuses WordPress's existing user primitive. Roles, content ownership, revisions, comments, deletion with reassignment, and user-based logs therefore continue to work. The agent's access can be changed or revoked without modifying a human account.

## How?

An agent is a regular WordPress user marked by `wpai_agent` user meta. The agent acts as its own principal: the creator is recorded as provenance in `wpai_agent_created_by`, not as the person the agent acts for. The marker changes the account's identity and authentication contract while preserving core user behavior:

- **Core user management surfaces.** Users → Add Agent reuses the Add User form. The profile remains the canonical place to change the role and manage credentials. The Users list labels roles such as `Editor (agent)`, provides account-type filtering, and replaces password reset with an Application Passwords shortcut.
- **Every agent username ends with `_agent`.** Provisioning appends the suffix when it is missing, including for programmatic callers. That makes agents easy to spot wherever only the login is shown, such as WP-CLI output, author names, and logs.
- **Non-interactive authentication.** Password login and password resets are blocked. Credentials are issued and revoked through core's REST-backed Application Password flows under their normal permission checks, preserving the one-time plaintext reveal. Other authentication mechanisms can participate when they resolve the request to the agent user.
- **The role defines authority.** There are no agent-specific roles or hardcoded blocked capabilities. An agent receives the same capabilities WordPress grants a human with the assigned role. An Administrator agent is therefore fully trusted and carries the same operational risk as any other Administrator account; lower roles remain limited by WordPress's normal capability mapping.
- **Non-administrator agents cannot write unfiltered HTML.** This is the one deliberate exception to role parity. Some roles below Administrator carry `unfiltered_html`, most notably Editor on single site, and model output stored with it becomes stored XSS (scripts saved into content). The capability is stripped from every agent without `manage_options`, so their content passes core's normal KSES filtering. The check matches the resolved primitive, which also covers meta capabilities such as `edit_css`.
- **Provisioning cannot widen the creator's access.** Creating an agent requires `create_users`, `promote_users`, and the primitive `edit_users` capability needed to manage the resulting account. The selected role must be within the provisioner's effective capabilities. That comparison runs first while listing roles and again against the real marked agent after creation, so user-specific capability filters cannot give the agent a role capability denied to its provisioner. A failed final check rolls the account back before core's `edit_user_created_user` compatibility and notification action runs. Capabilities that are inert for the agent, such as the stripped `unfiltered_html`, do not count against the provisioner.
- **Agents stay visible.** Normal user queries continue to return agents because they own content and participate in capability checks; hiding them is a display concern, never a query concern. REST user responses expose a read-only `wpai_is_agent` field, and the global `wpai_is_agent_user()` helper gives plugins one documented way to detect agent accounts.
- **Identity safeguards outlive the experiment UI.** Provisioning and presentation follow the WordPress AI support check, the global AI features setting, and the experiment setting. Existing agents remain unable to log in interactively or reset passwords whenever the plugin is active, and the unfiltered HTML restriction stays in force even if those settings are later disabled.

### Multisite model

WordPress stores user identity and Application Passwords across the network, while site memberships and roles are site-specific. Agent accounts follow that core model instead of adding a parallel boundary:

- One agent can be a member of several sites, and its role on each site defines what it can do there.
- The same credential identifies the network user everywhere. It authenticates the agent as a logged-in user on every site, like any WordPress credential, but a site the agent does not belong to only sees an authenticated user without capabilities there.
- Agents are provisioned from a site so the initial role has site context. Adding an agent to another site, removing it, changing its role, and deciding who may manage it all use core's normal multisite permission and invitation flows.
- Core lets only accounts with the network-level `manage_network_users` capability edit other users on multisite. The same rule decides who can provision agents and manage their credentials.
- The only agent-specific multisite restriction is that agents cannot become super admins, because that status bypasses the role system.

The plugin must be network-activated when this experiment is used on multisite. The agent marker is network-wide, so every site has to enforce the login and password-reset safeguards.

Core hardcodes several user-screen labels and action positions. Small inline scripts progressively enhance those labels for agents; without JavaScript, the underlying core forms continue to work.

Out of scope: assistants running inside a logged-in human session, credential protocols such as OAuth and expiry, per-run audit correlation, trust tiers, and approval workflows.

### Use of AI Tools

- AI assistance: Yes
- Tool(s): Claude Code, OpenAI Codex
- Model(s): Claude Fable 5, Claude Opus 5, GPT-5 (Codex)
- Used for: drafting and iterating on the implementation, tests, documentation, core-screen integration, security review, multisite behavior, capability safeguards, and validation in wp-env. I directed the scope and design decisions and reviewed the results.

## Testing Instructions

Application Passwords require HTTPS or `WP_ENVIRONMENT_TYPE` set to `local`. With wp-env, add a git-ignored `.wp-env.override.json` containing `{"config": {"WP_ENVIRONMENT_TYPE": "local"}}` and restart.

### Single site

1. Enable AI features and the Agent Users experiment in Settings → AI.
2. Go to Users → Add Agent, enter a username and unique email, select a role, and create the account. Confirm the stored username ends with `_agent`.
3. On the agent profile, create and copy an Application Password.
4. Authenticate through REST: `curl -u 'USERNAME:PASSWORD' 'https://your-site/wp-json/wp/v2/users/me'`.
5. Confirm permissions match the selected role. For example, an Editor remains unable to administer the site, while an Administrator agent has normal Administrator authority and risk.
6. Create content through REST and confirm the agent is the attributed author.
7. As an Editor agent, save content containing a `<script>` tag and confirm KSES strips it. A human Editor keeps core's single-site behavior.
8. Confirm interactive login and password reset are rejected.
9. Confirm the Users list marker, account-type filter, and Application Passwords row action.

### Multisite

1. Network-activate the plugin, then enable AI features and the experiment on the site under test.
2. As a network administrator, create an agent from that site's Users → Add Agent screen.
3. Confirm its Application Password authenticates on that site through REST.
4. Add the agent to a second site through core's Add Existing User flow and give it a role there. Confirm the same credential now works on both sites, each within that site's role.
5. Remove the agent from one site and confirm it loses its authority there while keeping its role on the other site.
6. Confirm a site administrator without network permissions cannot edit the agent, matching core's rule for human accounts.
7. Confirm the agent cannot be granted super admin.

Run the focused suites:

```sh
npm run test:php -- --filter Agent_UsersTest
npx wp-env --config=.wp-env.test.json run cli --env-cwd=wp-content/plugins/ai bash -c "WP_MULTISITE=1 vendor/bin/phpunit -c phpunit.xml.dist --filter Agent_UsersTest"
```

## Screenshots or screencast

Users screen with the Add Agent action, the `(agent)` marker, and the account-type filter:

Users screen with the Add Agent action, agents marked in the Role column, and the agents filter open

Users → Add Agent, using core's Add User form in agent mode:

Add Agent screen with username, email, names, website, and role fields

An agent profile with the account note and human-only sections hidden:

Edit Agent screen with the account note and the name, contact, and about sections

## Changelog Entry

> Added - New Experiment: Agent Users; provision dedicated accounts for external agents so their work is attributable and independently revocable.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-961-dcae15b63c3062f2fce01299882fc4baef2e6589.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Experiments/Agent_Users/Admin_Page.php
```diff
@@ -0,0 +1,417 @@
...
+	 */
+	private function render_agents_table(): void {
+		$agents = get_users(
+			array(
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key -- Bounded admin-screen query; agents are a small set.
+				'meta_key'     => Agent_Account::META_KEY,
+				'meta_compare' => 'EXISTS',
+				'orderby'      => 'registered',
+				'order'        => 'DESC',
+				'number'       => 100,
```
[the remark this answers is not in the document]
gziolo: Good catch, thanks. The silent cut-off was a real gap.

I kept this table as a short list and did not add pagination here. The paginated, searchable list already exists: the "AI Agents" view on the Users screen. Building a second one would duplicate the core list table.

What changed in ffeada82:
- The heading now shows the total, for example "Recent Agents (137)".
- The table shows the 20 most recent agents. When more exist, a line says how many are hidden and links to the AI Agents view.
- The description under the table always links to that view, so the relationship between the two screens is easy to find.
- The query orders by registration date and then ID, so the list is stable.

includes/Experiments/Agent_Users/Agent_Account.php
```diff
@@ -0,0 +1,475 @@
...
+	 * escalating privileges through an existing one, for example by changing
+	 * an administrator's email and password.
+	 *
+	 * @since x.x.x
+	 *
+	 * @var array<int, string>
+	 */
+	private const BLOCKED_CAPABILITIES = array( // phpcs:ignore SlevomatCodingStandard.Classes.DisallowMultiConstantDefinition -- This is used as an array const.
+		'unfiltered_html',
+		'create_users',
```
jorgefilipecosta, line 77: Not having the ability to create and edit users maybe be too restrictive. I imagine an admin using an agent to bulk add subscribers or authors, and other scenarios where user management is useful. User management abilities were some of the prioritized abilities not allowing agents users to use them may be too limiting. But I can understand being restrictive at the start as long as the list is filterable.

includes/Experiments/Agent_Users/Agent_Account.php
```diff
@@ -0,0 +1,475 @@
...
+		if ( is_wp_error( $login ) ) {
+			return $login;
+		}
+
+		$email = $this->validate_email( $email );
+		if ( is_wp_error( $email ) ) {
+			return $email;
+		}
+
+		$user_id = wp_insert_user(
```
jorgefilipecosta, line 172: Should we apply existing core filters like edit_user_created_user?

includes/Experiments/Agent_Users/REST_Field.php
```diff
@@ -0,0 +1,67 @@
...
+			'user',
+			self::FIELD_NAME,
+			array(
+				'get_callback' => static function ( array $user_data ): bool {
+					return Agent_Account::is_agent( (int) ( $user_data['id'] ?? 0 ) );
+				},
+				'schema'       => array(
+					'description' => __( 'Whether this is an agent account, used by software rather than a person.', 'ai' ),
+					'type'        => 'boolean',
+					'context'     => array( 'embed', 'view', 'edit' ),
```
jorgefilipecosta, line 61: We expose on view and embed contexts without permission checks I think even an anonymous visitor of the site can use REST API to check if an user is an agent, in certain cases for labelling of ai etc it may be needed. But I'm not sure if we should publicly expose on the endpoint which users are agents.

includes/Experiments/Agent_Users/Profile_Screen.php
```diff
@@ -0,0 +1,266 @@
...
+	 * Core has no hook between the title and the form, so the note is printed
+	 * with the notices, hidden, and moved under the title by the footer
+	 * script. It is plain text, not a notice: it describes the account, it
+	 * does not ask for attention.
+	 *
+	 * @since x.x.x
+	 */
+	public function render_account_type(): void {
+		$screen = get_current_screen();
+		if ( ! $screen || 'user-edit' !== $screen->id || null === self::profile_agent() ) {
```
jorgefilipecosta, line 165: Should we also include user-edit-network here for requests on multi site to /wp-admin/network/user-edit.php?user_id=<agent>?

includes/Experiments/Agent_Users/Agent_Account.php
```diff
@@ -0,0 +1,475 @@
...
+	 * an administrator's email and password.
+	 *
+	 * @since x.x.x
+	 *
+	 * @var array<int, string>
+	 */
+	private const BLOCKED_CAPABILITIES = array( // phpcs:ignore SlevomatCodingStandard.Classes.DisallowMultiConstantDefinition -- This is used as an array const.
+		'unfiltered_html',
+		'create_users',
+		'edit_users',
```
jorgefilipecosta, line 78: We don't include edit_user so an agent can add or change its own application passwords. Is that something we should block? I guess it is fine but wanted to confirm.

includes/Experiments/Agent_Users/Users_Screen.php
```diff
@@ -0,0 +1,214 @@
...
+	/**
+	 * Registers the Users screen hooks.
+	 *
+	 * @since x.x.x
+	 */
+	public function register(): void {
+		add_filter( 'get_role_list', array( $this, 'mark_agent_roles' ), 10, 2 );
+		add_action( 'manage_users_extra_tablenav', array( $this, 'render_filter' ) );
+		add_filter( 'users_list_table_query_args', array( $this, 'filter_list_table' ) );
+		add_filter( 'user_row_actions', array( $this, 'filter_row_actions' ), 10, 2 );
```
jorgefilipecosta, line 50: Multisite src/wp-admin/includes/class-wp-ms-users-list-table.php uses ms_user_row_actions, it also does not relies calls manage_users_extra_tablenav. So I think on multi sites agent users will look like normal users.

jorgefilipecosta (commented):
From the code/technical point of view this seems to be on a very good direction. I think multisite presents a whole set of challenges and we are not handling things to deal with multisite yet, It is probably better to disable this experiment on multisite and then have a follow up PR focused on multisite.

In terms of UX, we are forcing agents to have an email, I think email could be optional defaulting to the agent creator email, or even have the email field prefielled with the original author email. It may be tricky if emails are unique per user if that's the case I think it should be ok for agent users to not have an email.

gziolo:
Thank you for the review, @jorgefilipecosta!

Agreed on multisite. It brings its own set of questions, like the network Users screen, the network profile screen, and the fact that user meta is shared across the whole network. I will keep this PR focused on the general mechanism and handle multisite in a dedicated follow-up PR. To avoid a half-working state, this PR will disable the experiment when multisite is detected. The safeguards for existing agent accounts stay active either way.

On the email, it has to stay unique. WordPress enforces that in `wp_insert_user()` itself, so defaulting to the creator's email would fail on the second agent. We could auto-generate unique placeholder addresses, but I would rather not. The email gives us a communication layer for observing the agent: it is where notifications about its activity and account changes arrive. A placeholder address would silently drop all of that.

One idea for a follow-up: offer a button to fill or simply prefill the field with a plus-addressed suggestion based on the creator's email, for example `you+agent-name@your-domain`. It is unique, passes validation, and still delivers to the creator's inbox. For now, I would keep the field required and unchanged.

includes/Experiments/Agent_Users/REST_Field.php
gziolo: Publicly labeling AI actors is precisely the transparency aspect of the experiment. Agents will be clearly marked as such when listed as authors in posts. My assumption is that the flag provides limited information that an attacker could exploit.

includes/Experiments/Agent_Users/Profile_Screen.php
gziolo: Handled in a5216813. The profile integration now recognizes `user-edit-network`, so agent accounts are identified correctly on the network profile screen as well.

includes/Experiments/Agent_Users/Users_Screen.php
gziolo: Handled in a5216813. The network Users list now uses `ms_user_row_actions` and links agent management back to the assigned site.

gziolo:
I changed direction from disabling the experiment on multisite to supporting site-bound agents in this PR. Each agent records the site where it was created and can authenticate only while it remains a member of that site. Roles on other sites do not expand access. Agents cannot be added elsewhere or become super admins, and site administrators can manage only agents assigned to their site. This is implemented and covered by multisite tests in a5216813.

includes/Experiments/Agent_Users/Agent_Account.php
gziolo: After reflecting on this a bit, I removed the agent-specific capability blacklist in b74dca18. Agent authority now follows the assigned role: an Administrator agent can manage users, while lower roles remain limited by core’s normal capability mapping. The account creator accepts the risk associated with the selected role.

includes/Experiments/Agent_Users/Agent_Account.php
gziolo: Added in b74dca18. Successful provisioning now fires core’s `edit_user_created_user` action with admin-only notification, after final role authorization. This is covered by an integration test.

includes/Experiments/Agent_Users/Agent_Account.php
gziolo: This is now redesigned with the check for blocked capabilities removed. Agents follow core’s normal capability mapping, so they may manage their own Application Passwords when core permits it. The role-parity behavior is covered by an integration test.

gziolo:
Summary of the latest changes after review:

- Removed the agent-specific blocked-capability list. The assigned WordPress role now defines the agent's authority: an Administrator agent has normal Administrator access and risk, while lower roles remain limited by core's capability mapping.
- Strengthened provisioning so it requires `create_users`, `promote_users`, and the primitive `edit_users` capability. The selected role cannot exceed the provisioner's effective access.
- Added a final comparison against the real marked agent after creation. This catches user-specific capability filters that could otherwise grant the agent a role capability denied to its provisioner; failed authorization rolls the account back before `edit_user_created_user` runs.
- Preserved core behavior by firing `edit_user_created_user` for successful creation and returning an actual HTTP 403 for unauthorized submissions.
- Confirmed that agents may manage their own Application Passwords when core permits it.
- Kept the multisite identity boundary: agents are bound to one site, require network activation for provisioning, cannot become super admins, and cannot authenticate or gain membership outside their assigned site.
- Updated the UI copy, experiment documentation, tests, and PR description to reflect this model.

artpi:
All this sounds fantastic!

> Membership alone would therefore allow the same agent credential to follow the user to every site where it receives a role. (....) Each multisite agent records one assigned site in wpai_agent_site_id. 

I am not sure this is necessary and breaks some of the WordPress logic.
Roles are site-specific so that provides a boundary: you would have to explicitly add an agent to the site. this would follow established WP checks

One thing that I like in telegram is that when you are creating a telegram bo, the login has to end with `_bot`.
I think this would allow to quickly distinguish these accounts when browsing them in wpcli or somewhere.
KMy login:
arturpiszek
My. bot login
arturpiszek_mcp_agent


The other thing is that Linear makes it super explicit when something was done by the agent **and on behalf of whom** - I think this would be neat.
If we choose this route, we have to decide:
- Is the agent working permanently on my "behalf"
- Is the agent usrable by other people and acting "independently" and I am only its creator?

gziolo:
Thanks @artpi, all good points!

You are right about multisite. I picked the strict site binding because I did not know this area well enough. After checking core again, site roles already give the boundary we need. Only super admins can create an active user without email confirmation, and adding an existing user to another site sends an invitation that the agent's email owner has to confirm. So I will remove the site binding and treat agents as regular network users. Roles decide where an agent can act, and core decides who can manage the account. The only extra rule that stays is that an agent cannot become a super admin.

I like the `_bot` idea. Every agent login will end with `_agent`. It gets added on creation, so it also applies when agents are created from code. It should make agents easy to spot in WP-CLI, the post author name, and logs.

On "acting on behalf": as I noted earlier, I would keep two cases separate for now. Agent accounts always act on their own behalf, with their own role and attribution. That is what this PR covers. The person who creates the agent is only its creator, not the one the agent acts for. The other case is a human account used by an external agent. There the work belongs to the person, and we would mark it as agent assisted. That needs a different attribution model and fits better with the audit trail work.

webmyc:
## Compatibility note: agent users and code that assumes a principal is human

Following up on my #961 test with the compatibility note I offered. These are the ecosystem-side seams I have hit or reproduced where "an agent is a normal, queryable WordPress user" meets code that quietly assumed a principal was a person. None are defects in the experiment; they are the adjustments the surrounding code needs once agent users exist. I found all of these against real sites, most of them the hard way before agent users existed.

### 1. Resolving a principal by role can now resolve to an agent

Any code that answers "who should I act as" by querying for a role will, on some sites, get an agent. The canonical shape:

```php
get_users( array( 'role' => 'administrator', 'number' => 1, 'orderby' => 'ID', 'order' => 'ASC' ) );
```

Reproduced against #961: with an Administrator agent whose ID is lower than any human administrator (original admin deleted or renamed, or the site provisioned agent-first via WP-CLI or a host), this returns the agent, and the caller adopts it as the content-owning principal. `post_author` then points at an agent, and any later "is this author human" check loops.

This is correct behaviour from the experiment (agents must stay queryable so ownership and capability code keeps working). The fix belongs in the resolver: exclude agents when picking a principal to act as.

```php
// Before
get_users( array( 'role' => 'administrator', 'number' => 1, 'orderby' => 'ID', 'order' => 'ASC' ) );

// After: a human principal, not another agent
get_users( array(
    'role'       => 'administrator',
    'number'     => 1,
    'orderby'    => 'ID',
    'order'      => 'ASC',
    'meta_query' => array( array( 'key' => 'wpai_agent', 'compare' => 'NOT EXISTS' ) ),
) );
```

A stable `WordPress\AI\...\Agent_Account::is_agent()` (which #961 provides) is the thing every such resolver should consult. A raw meta check invites drift; a documented function does not.

### 2. Role-based user queries already miss admins; agents widen the blast radius

Two failures I shipped before agent users existed, both caused by a principal the query could not see. They matter here because agent users are one more principal these queries mishandle, and any resolver touched by note 1 above should account for both.

- `get_users( array( 'role' => 'administrator' ) )` only enumerates a site's own member list. A network super admin who runs a subsite without being explicitly added to it does not appear, so the query concludes there is no administrator.
- On installs where the administrator comes from a custom role or a runtime `user_has_cap` / `map_meta_cap` filter rather than the stored role, the same query returns empty while the site plainly has an admin.

The guidance worth stating out loud for anyone touching user queries: hiding an agent is a display concern (author dropdowns, participant lists), never a query concern. Normal queries must keep returning agents; only the UI should omit them.

### 3. A single-site Editor agent carries `unfiltered_html`

On a single-site install, both Editor and Administrator hold `unfiltered_html` by default. Reproduced against #961: an Editor agent's REST write stored a `<script>` tag, a `<style>` block, and a third-party custom element verbatim, with no kses filtering. The README already names this ("model output combined with it means stored XSS"); the note is only that the current experiment does not strip the capability yet, and that the role default carries it, so closing it needs an adjustment tied to the agent flag rather than the choice of role.

The corollary matters for anyone whose agents write structured builder markup: stripping `unfiltered_html` from an agent is the right default, and it reintroduces the kses wall a minimal human hits, where allowed elements are filtered and `<style>` contents can be corrupted because kses reads CSS selectors as markup. An allowlist (`wp_kses_allowed_html`) says which tags are safe here, but it cannot be authored by the party doing the writing, because the safe set is whatever plugins that specific site has installed. Whatever ships, agents that produce site-specific markup will need a way for the site, not the agent, to declare the content shapes its agents may write, held separately from what the agent may do.

### Verified positive controls (no action needed)

For completeness, against #961 these behaved exactly as designed: interactive login blocked, password reset blocked, Application Passwords issued and authenticated, `wpai_agent_created_by` recorded, `is_agent()` and the `wpai_is_agent` REST field correct, and agents correctly returned by ordinary user queries.

Happy to keep running patches against the fleet as this evolves and report back.

gziolo:
One finding while adding the `_agent` suffix. Agent provisioning uses `wp_insert_user()`, which allows `_` in usernames, so the suffix works in this PR. But multisite validates signups with [`wpmu_validate_user_signup()`](https://developer.wordpress.org/reference/functions/wpmu_validate_user_signup/), which only allows lowercase letters and digits (`/[^a-z0-9]/`). Any flow that goes through it, like the REST users endpoint on multisite, would reject `_agent` (and `-agent` too). That rule is very strict and may cause some friction for agents on multisite. I will keep it in mind for the multisite rework.

gziolo:
Thanks @webmyc for testing and for the detailed notes!

On `unfiltered_html`, this is a matter of decision. The reasoning you referred to is from an earlier revision. I removed the blocked capability list on purpose, so an agent now gets exactly what its role has. If you give an agent the Editor or Administrator role, `unfiltered_html` comes with it. We could remove this one capability for agents. In some cases, that is useful, because it stops scripts and styles in content. But it also blocks legitimate cases where the agent adds a safe script or style. And it would not remove the bigger risks, because an Administrator agent can still install a plugin or change the theme. I am open to either option, so more opinions are welcome.

The points about role queries need further investigation. Some of the reported cases look like expected WordPress behavior. An agent is a regular user with a role, so a query for that role will return it, and a super admin who is not a member of a site does not show up in that site's user list.

webmyc:
@gziolo on `unfiltered_html`, since you asked for opinions: strip it from agents by default, and here is the reasoning from someone whose agents write scripts and styles for a living.

the Administrator-agent case is not the interesting one. an admin agent is fully trusted, can install plugins, and reasoning about one capability there is deck-chair arrangement, agreed. the interesting tier is below admin, because the whole promise of this proposal, in your words on #923, is a principal smaller than an administrator that can still own content. Editor is the natural role for that principal, and on single-site Editor ships `unfiltered_html`, so the smaller principal arrives with stored-XSS-grade caps by default. a site owner who provisions an "editor agent" is not expecting that.

the "blocks legitimate cases" concern is real, my fleet is the evidence: agents legitimately write builder markup, embeds, styles all day. but i have run both sides of this tradeoff in production, and the answer that worked is not a binary cap. it is: safe default, plus a site-owned escape hatch that is narrower than `unfiltered_html`. concretely, strip the cap on the agent flag, and give the experiment one filter through which the site (or its plugins) can extend the kses allowlist for agent writes, the way `wp_kses_allowed_html` works but scoped to agent principals. the set of safe tags is a property of what the site has installed, so the site is the only party that can author it. the agent never gets to grant itself scripts; the site can, deliberately, per element. that covers my builder-markup case without reopening the XSS default, and an admin who truly wants an unfiltered agent can still grant the cap explicitly and own that decision.

on the role queries, no disagreement: both behaviors are core semantics working as designed, which is what the note says too. i included them as integrator context, not as bugs, because the resolver-adoption case is code that exists in the wild today (mine included) and will misfire the day agent users ship. if it is useful, i am happy to reshape that section into a short "guidance for integrators" block for the experiment README: resolvers that pick a principal to act as should consult `is_agent()`, hiding is a UI concern never a query concern. that puts the guidance where plugin authors will actually find it.

the `_agent` login suffix helps here too: it gives resolvers a second, dependency-free way to skip agents even when the AI plugin is not active to provide `is_agent()`. worth documenting as part of the convention.

gziolo:
@artpi @webmyc thank you, both rounds of feedback are now in the branch:

- Multisite follows core (80ea8632). The site binding is gone. Site membership and roles are the boundary, and core decides who can manage agents through `manage_network_users`. The only extra rule left is that agents cannot become super admins. Network activation is still required, because the agent marker is shared across the network, so every site must block interactive login for the same accounts.
- Usernames end with `_agent` (7d2210c1). The suffix is added on creation, also when agents are created from code. The README documents it as a convention.
- Agents below administrator lose `unfiltered_html` (2772edfb, cf5555f5). The capability is removed from every agent without `manage_options`, so their content goes through the normal KSES filtering. The check looks at the resolved primitive capability, so `edit_css` is covered as well. Administrator agents keep core behavior.
- New `wpai_is_agent_user()` helper (e61e49bf). It is an early entry point, the name may still change before the experiment settles. Code that picks a user to act as can call one documented function, instead of a raw meta query.

Two things I decided to postpone:

- The site-owned KSES allowlist filter. The experiment adds no custom hooks until the identity model is confirmed, because a public filter is hard to remove later. It is the first hook I would consider after that.
- Integrator guidance in the README. @webmyc, your offer to write that section still stands, and I would be happy to take it as a follow-up.

webmyc:
> * Integrator guidance in the README. @webmyc, your offer to write that section still stands, and I would be happy to take it as a follow-up.

it's here https://github.com/WordPress/ai/pull/1010

nerrad:
I like the direction. But have you considered introducing the concept of parent/child user accounts (introducing DB column `parent`)? Agent users could then always be attached to a parent user account (and be only child users). 

By default the agent user would then:

- inherit the capabilities of the parent user (but keep certain global restrictions in the case of the parent being an administrator/super-admin).
- could open up the ability for the parent user limiting what their agent can do _within the constraints of the capabilities the parent user has_
- Site owner could restrict which users can add agent (sub) users.
- Allows for application keys, additional auth etc to still be attached to individual agents.
- Automatic cleanup when parent users caps/role is changed or if that user is removed, all agent users are removed as well.
- Allows for things like linking agent activity to actual users ("{Agent A} on behalf of {Parent User}")
- ...etc

jeffpaul:
@gziolo note that I just closed #211 in favor of your work here, let's ensure we credit @Jameswlepage on the resulting PR merge here for his earlier work

PR #972, opened 2026-08-25
Limit comment moderation bulk analysis queue size

ekamran:
## Description

The plugin bounds AI bulk actions with `get_bulk_action_max_items()`, a filterable, feature-keyed limit that defaults to 100. Alt text generation and summarization both use it. Comment moderation registers a bulk action but does not, so it marks every selected comment as pending regardless of how many were selected.

Each pending comment becomes a model call once it is analyzed, which is the same reason the other two bulk actions are bounded.

## Changes

- Comment moderation now applies `get_bulk_action_max_items()` to its bulk action.
- Selected comments are validated before the limit is applied, so IDs for comments that no longer exist cannot consume slots and push out valid ones.
- Comment IDs are deduplicated, so the same comment selected twice counts once.
- Comments left out are reported in the admin notice, which switches to a warning when anything was left out.
- `wpai_analysis_truncated` is registered as a removable query arg and scrubbed from the request URI, matching the existing notice args.

## Testing

- `npm run test:php -- --filter Comment_ModerationTest` passes with 31 tests and 130 assertions. Three new tests cover the limit, deduplication, and invalid IDs not consuming slots, plus one covering the notice when comments were left out but none were queued.
- Full PHP test suite passes.
- PHPCS passes on both changed files.
- PHPStan passes on both changed files.

Manual wp-admin verification was performed with a temporary local filter that set the Comment Moderation bulk limit to 2 and satisfied the credential gate. Selecting 5 comments and running the real bulk action queued 2 comments, left 3 untouched, showed a warning notice for the truncated count, and scrubbed the notice query args after load.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-972-ef49eea1f51834af823564aa9a2b4689c59df308.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/experiments/alt-text-generation.md
+++ b/docs/experiments/alt-text-generation.md
@@ -266,7 +266,7 @@ add_filter( 'wpai_bulk_action_max_items', function ( int $max_items, string $fea
 }, 10, 2 );
 ```
 
-Values below 1 are clamped to 1. The same filter governs the bulk summarization cap, so check `$feature_id` when you only mean to change one.
+Values below 1 are clamped to 1. The same filter governs the bulk summarization and comment moderation caps, so check `$feature_id` when you only mean to change one.
 
 ### Adding Custom UI Elements
 
--- a/docs/experiments/comment-moderation.md
+++ b/docs/experiments/comment-moderation.md
@@ -130,6 +130,18 @@ add_filter( 'wpai_comment_moderation_should_moderate', function( $should_moderat
 }, 10, 3 );
 ```
 
+### Adjusting the Bulk Batch Cap
+
+Each comment queued by the **Analyze Sentiment and Toxicity** bulk action becomes one billed model call once it is analyzed, so `handle_bulk_action()` caps how many comments a single request will queue. The default is 100; comments beyond the cap are not queued, and the surplus is reported to the user in an admin notice.
+
+```php
+add_filter( 'wpai_bulk_action_max_items', function ( int $max_items, string $feature_id ): int {
+    return 'comment-moderation' === $feature_id ? 25 : $max_items;
+}, 10, 2 );
+```
+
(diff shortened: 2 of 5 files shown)


dkotter (changes requested):
Thanks for the PR! Overall this looks good and works as expected just have two things that I think we should add here:

1. We strip query args on the client-side (see `clearQueuedAnalysisQueryArg`) and we should likely extend that to remove the new `wpai_analysis_truncated` query arg. That query arg could end up being left in the URL otherwise
2. We should look to update our docs around this experiment to mention this new limit and how to change the limit

ekamran:
Thanks for the review. I pushed a follow-up commit addressing both points.

`clearQueuedAnalysisQueryArg()` now removes `wpai_analysis_truncated` alongside `wpai_analysis_queued`. I also updated the guard so the URL is cleaned if either notice arg is present.

I added an "Adjusting the Bulk Batch Cap" section to the Comment Moderation docs, including the default limit, the `wpai_bulk_action_max_items` filter, and the `comment-moderation` feature id. I also updated the existing Alt Text docs note so the shared filter wording stays accurate now that it also applies to Comment Moderation.

dkotter (approved):

PR #973, opened 2026-08-25
Fix: Meta Description: SEO plugin detection cache never expires and skips the "no plugin" case

### LINKED ISSUE #970: Meta Description: SEO plugin detection cache never expires and skips the "no plugin" case
### Description

`SEO_Integration::detect_active_plugin()` caches the active SEO plugin (to pick the meta-description meta key) in the `wpai_active_seo_plugin` transient,

1. The positive result is stored with **no expiry** and is only cleared on `deactivated_plugin` - a hook that's registered **only while the Meta Description experiment is enabled**. So the cache can go stale and never self-heal.
2. The **"no SEO plugin active" result is never cached**, so a full plugin scan runs on every call, including the front-end `output_meta_description()` path.

## Code
`includes/Abilities/Meta_Description/SEO_Integration.php:78-96`
```php
$active_plugin = get_transient( 'wpai_active_seo_plugin' );
if ( ! empty( $active_plugin ) ) {
	return $active_plugin;
}
// ...
foreach ( self::get_supported_plugins() as $slug => $info ) {
	if ( is_plugin_active( $info['file'] ) ) {
		set_transient( 'wpai_active_seo_plugin', $slug ); // no expiry
		return $slug;
	}
}
return null; // never cached
```
Invalidation (only when the experiment is on): `includes/Experiments/Meta_Description/Meta_Description.php:60` (`deactivated_plugin`) and `:165` (`delete_transient`). No `activated_plugin` handling.

1. Activate a supported SEO plugin (e.g. Yoast) + enable the Meta Description experiment; trigger meta handling once to cache the slug.
2. Disable the Meta Description experiment.
3. Deactivate the SEO plugin.
4. Re-enable the experiment → detection still returns the deactivated plugin's slug, and never corrects itself (no expiry).

## Impact
- **Stale cache:** if the SEO plugin changes while the experiment is disabled (so the clear hook isn't registered), the cached slug persists forever, and meta descriptions are read/written under the wrong meta key.
- **Performance:** sites with no supported SEO plugin re-scan all plugins (`is_plugin_active()`) on every meta-description call.

## Affected files
- `includes/Abilities/Meta_Description/SEO_Integration.php` (78-96)
- `includes/Experiments/Meta_Description/Meta_Description.php` (60, 165)

## Suggested fix
- Give the positive transient a TTL (e.g. `DAY_IN_SECONDS`).
- Cache the "none detected" result too (sentinel value + TTL).
- Also invalidate on `activated_plugin`, or make invalidation independent of the experiment toggle.

https://github.com/user-attachments/assets/f34fbb35-6de4-4991-9cf1-ceb88bc1c2bd

### Environment info

- WordPress Latest.
- AI plugin develop branch

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

hbhalodia:
## What?

Closes https://github.com/WordPress/ai/issues/970

## Why?

1. The positive result is stored with no expiry and is only cleared on deactivated_plugin - a hook that's registered only while the Meta Description experiment is enabled. So the cache can go stale and never self-heal.
2. The "no SEO plugin active" result is never cached, so a full plugin scan runs on every call, including the front-end output_meta_description() path.

## How?

- Updates the transient to cache it for a day.
- Update the activated, deactivated hook outside of experiment, so a cache would invalidate regardless of experiment is enabled or not.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 5
Used for: Code, Unit Test under the direction provided by me. Implementation reviwed by me.

## Testing Instructions

1. Activate a supported SEO plugin (e.g. Yoast) + enable the Meta Description experiment; trigger meta handling once to cache the slug.
2. Disable the Meta Description experiment.
3. Deactivate the SEO plugin.
4. Re-enable the experiment, This should use the plugin's meta key, i.e wpai_meta_description.

## Changelog Entry

> Fixed - Cached the active SEO plugin detection - with a TTL and immediate invalidation on any plugin activation/deactivation - so meta description meta-key lookups no longer re-scan active plugins on every request.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-973-b482d0bd1f34675dc2c1b0ba67420399795c3c10.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Abilities/Meta_Description/SEO_Integration.php
+++ b/includes/Abilities/Meta_Description/SEO_Integration.php
@@ -29,6 +29,22 @@ class SEO_Integration {
 	public const FALLBACK_META_KEY = 'wpai_meta_description';
 
 	/**
+	 * Transient key caching the detected SEO plugin.
+	 *
+	 * @since x.x.x
+	 * @var string
+	 */
+	public const CACHE_KEY = 'wpai_active_seo_plugin';
+
+	/**
+	 * Cached value meaning "no supported SEO plugin is active".
+	 *
+	 * @since x.x.x
+	 * @var string
+	 */
+	private const CACHE_NONE = 'none';
+
+	/**
 	 * Returns the list of supported SEO plugins and their meta keys.
 	 *
 	 * @since 0.7.0
@@ -69,16 +85,22 @@ class SEO_Integration {
 
 	/**
 	 * Detects the currently active SEO plugin.
-	 * Cache is flushed when a plugin is deactivated.
+	 *
+	 * Result is cached with a TTL and flushed on plugin activation/deactivation.
 	 *
 	 * @since 0.7.0
 	 *
 	 * @return string|null The slug of the active SEO plugin, or null if none detected.
 	 */
 	public static function detect_active_plugin(): ?string {
-		$active_plugin = get_transient( 'wpai_active_seo_plugin' );
-		if ( ! empty( $active_plugin ) ) {
-			return $active_plugin;
+		$cached = get_transient( self::CACHE_KEY );
+
+		if ( self::CACHE_NONE === $cached ) {
+			return null;
+		}
+
+		if ( ! empty( $cached ) ) {
+			return $cached;
 		}
 
 		if ( ! function_exists( 'is_plugin_active' ) ) {
@@ -87,12 +109,36 @@ class SEO_Integration {
 
 		foreach ( self::get_supported_plugins() as $slug => $info ) {
 			if ( is_plugin_active( $info['file'] ) ) {
-				set_transient( 'wpai_active_seo_plugin', $slug );
+				set_transient( self::CACHE_KEY, $slug, DAY_IN_SECONDS );
 				return $slug;
 			}
 		}
 
+		// Cache the "none active" result so the scan is not repeated every call.
+		set_transient( self::CACHE_KEY, self::CACHE_NONE, DAY_IN_SECONDS );
 		return null;
+	}
+
+	/**
+	 * Registers cache invalidation on plugin activation and deactivation.
+	 *
+	 * Runs regardless of the Meta Description experiment state so the cache is
+	 * never left stale when a plugin changes while the experiment is disabled.
+	 *
+	 * @since x.x.x
+	 */
+	public static function register_cache_invalidation(): void {
+		add_action( 'activated_plugin', array( self::class, 'clear_cache' ) );
+		add_action( 'deactivated_plugin', array( self::class, 'clear_cache' ) );
+	}
+
+	/**
+	 * Clears the detected SEO plugin cache.
+	 *
+	 * @since x.x.x
+	 */
+	public static function clear_cache(): void {
+		delete_transient( self::CACHE_KEY );
 	}
 
 	/**
--- a/includes/Abilities/Meta_Description/SEO_Integration.php
+++ b/includes/Abilities/Meta_Description/SEO_Integration.php
@@ -128,17 +128,55 @@ class SEO_Integration {
 	 * @since x.x.x
 	 */
 	public static function register_cache_invalidation(): void {
-		add_action( 'activated_plugin', array( self::class, 'clear_cache' ) );
-		add_action( 'deactivated_plugin', array( self::class, 'clear_cache' ) );
+		add_action( 'activated_plugin', array( self::class, 'clear_cache_on_plugin_change' ), 10, 2 );
+		add_action( 'deactivated_plugin', array( self::class, 'clear_cache_on_plugin_change' ), 10, 2 );
 	}
 
 	/**
-	 * Clears the detected SEO plugin cache.
+	 * Clears the cache when a plugin is activated or deactivated. Also
+	 * supports multisite and network activated/deactivated plugins.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $plugin       Path to the plugin file, relative to the plugins directory.
+	 * @param bool   $network_wide Whether the change applied network-wide.
+	 */
+	public static function clear_cache_on_plugin_change( string $plugin, bool $network_wide = false ): void {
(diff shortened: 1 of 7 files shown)


includes/Abilities/Meta_Description/SEO_Integration.php
```diff
@@ -87,14 +109,38 @@ public static function detect_active_plugin(): ?string {
 
 		foreach ( self::get_supported_plugins() as $slug => $info ) {
 			if ( is_plugin_active( $info['file'] ) ) {
-				set_transient( 'wpai_active_seo_plugin', $slug );
+				set_transient( self::CACHE_KEY, $slug, DAY_IN_SECONDS );
```
dkotter, line 112: We may want to add an upgrade routine (see `Upgrades.php` for how we've done this) to automatically clear this transient out on existing sites. Otherwise this could stay cached "forever" on those sites (though likely will naturally clear over time)

includes/Abilities/Meta_Description/SEO_Integration.php
```diff
@@ -87,14 +109,38 @@ public static function detect_active_plugin(): ?string {
...
 				return $slug;
 			}
 		}
 
+		// Cache the "none active" result so the scan is not repeated every call.
+		set_transient( self::CACHE_KEY, self::CACHE_NONE, DAY_IN_SECONDS );
```
dkotter, line 118: I think there's likely a multisite bug here, as this transient will be set at the site-level but some sites may have an SEO plugin network-activated. So if a site on a network ends up caching `none`, that value potentially won't be cleared until the cache expires (24 hours). Likely need to ensure we clear this cache on network plugin events

dkotter (changes requested):

includes/Abilities/Meta_Description/SEO_Integration.php
hbhalodia: Thanks, will fix and remove the transient on upgrade.

includes/Abilities/Meta_Description/SEO_Integration.php
hbhalodia: Thanks, checking for this and will update.

includes/Abilities/Meta_Description/SEO_Integration.php
hbhalodia: Fixed in commit - https://github.com/WordPress/ai/pull/973/changes/6add0845630f62658c18bbaf8f1437e65116f1c4

includes/Abilities/Meta_Description/SEO_Integration.php
hbhalodia: Fixed in commit - https://github.com/WordPress/ai/pull/973/changes/b482d0bd1f34675dc2c1b0ba67420399795c3c10

dkotter (approved):
Things look good to me, thanks! Holding on merging this for now though as we add a new upgrade routine here that has the file name hardcoded to v1.4.0. If we end up doing a minor release (v1.3.1 as an example) prior to releasing 1.4.0, this code will be inaccurate. Will merge this once we know for sure the version we'll be releasing this as

dugyen (commented):
Reviewed this in detail (diff, tests, and interaction with the surrounding code). This looks like a solid, correct fix — didn't find a blocking issue.

**What it does right, matching both parts of #970:**
- Adds a `DAY_IN_SECONDS` TTL to the positive cache in `SEO_Integration.php`, and now also caches a `'none'` sentinel for "no SEO plugin active" so the front-end `output_meta_description()` path stops re-scanning every plugin on every request.
- Moves cache invalidation off `Meta_Description::register()` and onto an unconditional `SEO_Integration::register_cache_invalidation()` called from `Main::load()` (hooked to `plugins_loaded`), so the cache now self-heals via `activated_plugin`/`deactivated_plugin` regardless of whether the experiment is enabled — exactly the "invalidation independent of experiment toggle" fix the issue asked for.
- Goes a bit further than requested: the new `V1_4_0` upgrade routine deletes the old non-expiring transient once on update, so sites with an already-stale infinite cache get a clean slate rather than waiting on the new logic. Also handles multisite network-wide activation by clearing the cache on every site in the network.
- Test coverage is thorough: positive/negative caching, TTL-independent invalidation on both hooks, the upgrade routine's version-gating (skips on fresh installs and already-upgraded sites), and multisite network-wide vs. single-site invalidation scope.
- I checked for stale references to the removed `clear_active_plugin_cache()` method/hook elsewhere in the codebase, including the pre-existing `Meta_DescriptionTest.php` — none found, so no regression there.

**Two minor, non-blocking observations:**
1. The multisite network-wide path (`clear_cache_for_network()`) fetches every site (`get_sites(['number' => 0])`) and does a `switch_to_blog()` round-trip per site on every plugin activate/deactivate, not just SEO-relevant ones. Fine for typical networks; could be slow on a network with many thousands of sites.
2. The `phpcs:ignore` comment on `switch_to_blog()` names a specific VIPMinimum sniff — worth confirming CI is happy with that exact sniff name for this repo's ruleset, but harmless either way.

Nice work, @hbhalodia.

---
AI assistance: Yes. Tool(s): Claude Code (Sonnet 5). Used for: reading the diff and tracing it against the rest of the codebase to write this review; I read and take responsibility for the assessment above.

PR #832, opened 2026-07-05
chore(deps): drop @wordpress dependencies to wp-7.0 versions

justlevine:
## What?

- Followup to https://github.com/WordPress/ai/pull/831

This PR drops the `@wordpress` production `dependencies` to the versions that shipped in WordPress 7.0

> [!IMPORTANT]
> This PR is stacked on https://github.com/WordPress/ai/pull/831 which should be merged first.
> Relevant diff: https://github.com/WordPress/ai/pull/832/commits/23fd12f7802741102d6eff10b808715498fc3a4f

## Why?

`@wordpress/dependency-extract-wordpress-plugin` strips our production dependencies from the build, so to confirm we're not consuming incompatible functionality we want to lock them whenever possible.

## How?

Removed the `devDependencies` section to prevent conflicts, and then ran `wp-scripts packages-update --dist-tag=wp-7.0`

### Use of AI Tools

none

## Testing Instructions

1. `npm ci` - confirm you can still install
2. `npm run build` - confirm building still works
3. `npm run lint:js && npm run typecheck` - confirm linting and typechecking still passes.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
|        |       |

## Changelog Entry

> Added - New feature.
> Changed - Existing functionality.
> Deprecated - Soon-to-be removed feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.
> Developer - Development related updates.
> Developer - Drop @wordpress dependencies to versions included in WordPress Core 7.0.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-832-23fd12f7802741102d6eff10b808715498fc3a4f.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #984, opened 2026-08-28
Fix: Update DataViews to fix settings page breakage in Gutenberg trunk

aduth:
## What?

Updates `@wordpress/dataviews` and `@wordpress/ui` to their latest versions.

## Why?

The AI settings page is currently broken on Gutenberg `trunk`, and will be broken in the next release of Gutenberg unless fixed.

The issue stems back to https://github.com/WordPress/gutenberg/pull/81478 removing `@wordpress/dataviews` as an allowlisted package in `@wordpress/private-apis`, as the result of the work done in https://github.com/WordPress/gutenberg/issues/81230 to remove private APIs usage in DataViews. 

By removing DataViews from the allowlist, any consumer which bundles an older version of `@wordpress/dataviews` that still relies on those private APIs will fail. From the consumer point of view, the mitigation is to upgrade to the latest version.

## How?

Upgrades DataViews to the latest version.

The `package-lock.json` diff is much larger than I'd anticipate or hope, but I'm not sure it's avoidable. This is what results from `npm install` after bumping the versions in `package.json`. 

### Use of AI Tools

AI assistance: Yes
Tool(s): Cursor IDE
Model(s): Auto (likely Composer)
Used for: Research of impact based on related pull request and equivalent fixes in other projects like https://github.com/Automattic/jetpack/pull/51701, as well as actual bumping of versions. Changes were reviewed by myself.

## Testing Instructions

Test with changes alongside Gutenberg either `trunk` or [23.9.0 RC1 prerelease](https://github.com/WordPress/gutenberg/releases/tag/v23.9.0-rc.1).

To test locally, I created a `.wp-env.override.json` with the following contents to load from my sibling `gutenberg` code directory:

```json
{
	"$schema": "https://schemas.wp.org/trunk/wp-env.json",
	"plugins": [
		".",
		"../gutenberg",
		"https://downloads.wordpress.org/plugin/ai-provider-for-google.zip",
		"https://downloads.wordpress.org/plugin/ai-provider-for-openai.zip"
	]
}
```

Then run `npm run build && npm run wp-env start`.

Go to http://localhost:8888/wp-admin/options-general.php?page=ai-wp-admin

Confirm page loads as expected.

Before|After
---|---
image|image

## Changelog Entry

See include CHANGELOG.md edit.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-984-1acd99bb0daa42fcf968e3a7476d013e8826c9c8.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file, per [the Ke
 
 ## [Unreleased] - TBD
 
+### Fixed
+
+- Fixed AI settings page failing to render on Gutenberg 23.9+ after `@wordpress/dataviews` was removed from the private-apis allowlist.
+
 ## [1.3.0] - 2026-08-18
 ### Added
 - New Experiment: Content Translation; translates Paragraph and Heading blocks—and optionally the post title—into a selected language directly from the post editor ([#747](https://github.com/WordPress/ai/pull/747)).
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
 				"@wordpress/core-abilities": "0.10.0",
 				"@wordpress/core-data": "7.44.0",
 				"@wordpress/data": "10.44.0",
-				"@wordpress/dataviews": "14.1.0",
+				"@wordpress/dataviews": "18.1.0",
 				"@wordpress/date": "5.44.0",
 				"@wordpress/dom": "4.47.0",
 				"@wordpress/dom-ready": "4.44.0",
@@ -32,7 +32,7 @@
 				"@wordpress/notices": "5.44.0",
 				"@wordpress/plugins": "7.44.0",
 				"@wordpress/primitives": "4.45.0",
-				"@wordpress/ui": "0.20.0",
+				"@wordpress/ui": "0.21.0",
 				"@wordpress/url": "4.44.0",
 				"@wordpress/wordcount": "4.44.0",
 				"react": "18.3.1"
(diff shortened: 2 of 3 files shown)


youknowriad:
Why are the imports here not using the "/wp" bundle. Without this, there are other potential issues as WP gets updated.

aduth:
> Why are the imports here not using the "/wp" bundle. Without this, there are other potential issues as WP gets updated.

It probably should, per [the documented note](https://github.com/WordPress/gutenberg/tree/trunk/packages/dataviews#:~:text=If%20you%27re%20trying%20to%20use%20the%20DataViews%20component%20in%20a%20WordPress%20plugin%20or%20theme%20and%20you%20are%20building%20your%20scripts%20using%20the%20%60%40wordpress/scripts%60%20package%2C%20you%20need%20to%20import%20the%20components%20from%20%60%40wordpress/dataviews/wp%60%20instead%20of%20%60%40wordpress/dataviews%60.). My initial instinct is that might be a bigger lift, but given the lockfile turmoil I'm facing here and potential risk of updates from a newer version of DataViews & UI packages, that might be a good alternative. Assuming it's mostly a matter of updating the imports, I'll push an alternative pull request for that.

aduth:
Aside: The `/wp` note being buried in the DataView's documentation might contribute to it being missed. Since it mentions `@wordpress/scripts`, it's also not clear if it applies to projects using `@wordpress/build`.

aduth:
Alternative updating to use `/wp` entrypoint: #989

dkotter:
@aduth With #989 now merged, can this one be closed?

aduth:
@dkotter I'll go ahead and close, though I imagine "keeping dependencies up to date" is still valid independent of #989. But since #989 addresses the immediate concern of crashing, that can happen on a more relaxed timeline.

dkotter:
> @dkotter I'll go ahead and close, though I imagine "keeping dependencies up to date" is still valid independent of #989. But since #989 addresses the immediate concern of crashing, that can happen on a more relaxed timeline.

Worth noting we are purposely not updating most of these dependencies as we want to ensure we're using the same versions that are in our lowest supported WordPress version (7.0). See https://github.com/WordPress/ai/pull/831 and https://github.com/WordPress/ai/pull/832 for more on that

aduth:
> Worth noting we are purposely not updating most of these dependencies as we want to ensure we're using the same versions that are in our lowest supported WordPress version (7.0). See #831 and #832 for more on that

Interesting, and good to know! I don't think it's a great situation if plugins are intentionally not using the latest versions of packages, and one of the intended benefits of these ["bundled" packages](https://github.com/WordPress/gutenberg/blob/trunk/packages/README.md#when-to-omit-or-set-wpscript-to-false) like `@wordpress/dataviews`, `@wordpress/ui`, and `@wordpress/admin-ui` is that they're more self-contained. In fact, that's what the changes in #989 are all about, since it flips DataViews to bundle everything it relies on into the AI plugin rather than relying on `window.wp` globals which may or may not be present depending on the version of WordPress being run (to your point of a 7.0 floor). But this isn't universal, and a package like `@wordpress/ui` is "bundled" but doesn't have a `/wp` entrypoint, and so some of [its internal dependencies](https://github.com/WordPress/gutenberg/blob/4624b867b2c92b2c60d255c0c1dfceea5ea21f40/packages/ui/package.json#L48-L58) do end up relying on `window.wp` globals if it's built using `@wordpress/scripts` or `@wordpress/build`. This is probably something to be solved / standardized in Gutenberg so that consumer projects like this can rely on using the latest versions.

dkotter:
Probably worth noting that `@wordpress/ui` and `@wordpress/admin-ui` are the two packages right now that we don't follow this approach with and instead keep those more or less up-to-date (via Dependabot bumps). So maybe `@wordpress/dataviews` should also follow that approach? Probably worth a discussion on #832 for those types of things though

simison:
Since DataViews gets bundled, doing updates now lets you absorb breaking changes incrementally, in smaller batches, rather than facing them all at once during a larger future version jump. That should end up being less (or at least easier) work in total.

PR #985, opened 2026-08-30
Remove Enable AI header toggle

### LINKED ISSUE #600: `Enable AI` header toggle doesn't reflect aggregate state of sub-features
### Description

 
The `Enable AI` toggle in the page header shows a binary on/off state that doesn't accurately represent the underlying configuration. When only some sub-features are enabled, the toggle still renders as fully ON, with no visual distinction from the "all enabled" state.
 
## Actual behavior
 
The header toggle renders as fully ON (solid blue), identical to the state where every sub-feature is enabled across every section.
 
## Expected behavior
 
The header toggle should accurately represent the state of the features it governs. Two options worth considering:
 
### Option A — tri-state header toggle
 
The header toggle reflects aggregate state across all sub-features:
 
- **Off** — nothing enabled anywhere.
- **Partial / indeterminate** — at least one but not all sub-features enabled. Distinct visual treatment (partial fill, dashed track, or indeterminate indicator).
- **On** — everything enabled.
### Option B — move the toggle to each section (preferred)
 
Drop the global header toggle and give each section (Content Experiments, Admin Experiments, future sections) its own master toggle with the same tri-state behavior scoped to that section. This avoids a single control trying to summarize an ever-growing surface, keeps the aggregate state legible at the section level, and pairs naturally with the existing `Enable all` / `Disable all` buttons. As more experiment groups are added, the model scales without the header toggle becoming increasingly meaningless.

1. Go to the AI settings screen.
2. Under Content Experiments, enable only a subset of features (e.g. Alt Text Generation and Meta Description Generation, leave the rest off).
3. Observe the `Enable AI` toggle in the page header.

Image

### Environment info

The latest: WP (7.1), AI plugin (1.0), Safari.

- [x] Yes

- [x] Yes

- [x] Block
- [ ] Classic
- [ ] Not sure

wprashed:
## What?

Closes #600

Removes the `Enable AI` header toggle from the AI settings page (`routes/ai-home/stage.tsx`), cleans up remaining `globalEnabled` UI disabling logic, and allows individual feature and experiment group toggles to directly control plugin functionality.

## Why?

The `Enable AI` header toggle displayed a binary ON/OFF state that did not accurately reflect individual feature configurations. Per maintainer consensus in #600, this PR removes the global header toggle entirely, allowing individual feature toggles and category group toggles (or deactivating the plugin / using core filters) to manage functionality.

## How?

- Removed the `Enable AI` header `<ToggleControl>` and `InfoTip` from `routes/ai-home/stage.tsx`.
- Removed unused `globalToggleDescription`, `GLOBAL_FIELD_ID`, `noop`, and `DisabledToggle` component from `stage.tsx`.
- Removed `globalEnabled` prop and conditions from `SectionActions` and `VisualCardToggle` in `stage.tsx`, ensuring individual toggles and bulk action buttons remain interactive.
- Removed `.ai-showcase-card--disabled` CSS rule from `routes/ai-home/style.scss`.
- Updated `is_globally_enabled()` in `includes/Abstracts/Abstract_Feature.php` to default to `true` when option is not explicitly set.
- Updated Playwright E2E helpers (`tests/e2e/utils/helpers.ts`) in `disableExperiments` and `enableExperiments` to wait for section action buttons to render before counting, correctly capture settings save responses, and toggle showcase cards.
- Updated `tests/e2e/specs/admin/settings.spec.js` to assert section button states and removed the obsolete test for disabled global bulk actions.

### Use of AI Tools

AI assistance: Yes
Tool(s): Antigravity
Model(s): Gemini 2.5 Pro
Used for: Diagnosing E2E test failures, updating settings UI components and E2E test assertions, and structuring the PR according to repository guidelines.

## Testing Instructions

1. Go to the AI settings page (`options-general.php?page=ai-wp-admin`).
2. Confirm the header renders cleanly with Docs, Contribute, and options links, without the `Enable AI` header toggle.
3. Confirm section group toggles ("Enable all" / "Disable all") and individual feature toggles are enabled and continue to work as expected.
4. Verify that running `npm run typecheck`, `npm run lint:js`, and `npm run build` pass without errors.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| Header includes "Enable AI" toggle | Header has no "Enable AI" toggle; features are managed via section buttons and toggles |

## Changelog Entry

> Removed - The `Enable AI` header toggle from the AI settings page.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-985-e3cf1aeb40340afdd78698ddf199c5800f39f1cd.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter (changes requested):
E2E tests are failing here.

In addition, please ensure the PR description contains everything from our PR template, as already mentioned [here](https://github.com/WordPress/ai/pull/987#pullrequestreview-5068498885)

includes/Abstracts/Abstract_Feature.php
```diff
@@ -169,7 +169,7 @@ public function get_category(): string {
 	 * {@inheritDoc}
 	 */
 	final public function is_globally_enabled(): bool {
-		return (bool) get_option( Settings_Registration::GLOBAL_OPTION, false );
+		return (bool) get_option( Settings_Registration::GLOBAL_OPTION, true );
```
dkotter, line 172: So I've not tested things yet so I may be wrong but if someone currently has things set to be globally disabled, it seems like after this update there will be no way for them to actually enable things, since we're removing the UI.

So for new installs and existing installs that were already enabled, those should be fine. But existing installs that purposely disabled things, I think they'll be permanently stuck in that disabled state.

We use the result of this method in the `is_enabled` method and so that method will always return false in that scenario. And with no UI to enable things, the `Settings_Registration::GLOBAL_OPTION` value will never be updated.

Honestly not sure on the best approach here. We could change this method to always return true. We could remove this method all together. We could have an upgrade routine to set `Settings_Registration::GLOBAL_OPTION` to true. And probably other approaches as well I'm not thinking of right now

routes/ai-home/stage.tsx
```diff
@@ -1019,19 +938,6 @@ function AISettingsPage() {
 					) }
 					actions={
 						<>
-							<Stack align="center" gap="xs">
-								<ToggleControl
-									label={ __( 'Enable AI', 'ai' ) }
```
dkotter, line 1024: So we remove the display of the setting here but we don't actually remove the registration of the setting in `includes/Settings/Settings_Registration.php`. Should we?

tests/e2e/utils/helpers.ts
```diff
@@ -163,47 +163,58 @@ export const clearConnector = async (
...
-	if ( await globalToggle.isChecked() ) {
-		return;
-	}
-	await globalToggle.check();
-	await expect(
-		page.locator( '.components-snackbar__content', {
-			hasText: 'AI enabled.',
-		} )
-	).toBeVisible();
+export const enableExperiments = async ( admin?: Admin, page?: Page ) => {
```
dkotter, line 215: Any reason to keep this? Since this does nothing now, thinking it would be better to just get rid of it

dkotter (changes requested):

jeffpaul:
@wprashed this ideally gets into the 1.4.0 release, noting that are you able to make updates here from latest code review?

PR #989, opened 2026-08-31
Update DataViews usage to use `/wp` entrypoint

aduth:
## What?

Updates imports of `@wordpress/dataviews` to use its `/wp` entrypoint.

Alternative to #984

## Why?

Fixes broken AI settings page when running Gutenberg 23.9.0 or newer.

This is the documented recommendation for using DataViews in plugins built using `@wordpress/scripts`:

> If you're trying to use the DataViews component in a WordPress plugin or theme and you are building your scripts using the `@wordpress/scripts` package, you need to import the components from `@wordpress/dataviews/wp` instead of `@wordpress/dataviews`.

Source: https://github.com/WordPress/gutenberg/tree/trunk/packages/dataviews

The reason this is recommended is also the reason I'm proposing this update now: Since not using the `/wp` entrypoint causes DataViews to use shared WordPress scripts (instead of bundling those dependencies), DataViews' internal references to APIs could become unreliable if it depends on newer or removed APIs. See also #984

### Use of AI Tools

See AI disclosure in #984, though in this pull request I did not use any AI for updating the imports.

## Testing Instructions

Repeat Testing Instructions from #984

## Screenshots or screencast

Before|After
---|---
image|image

## Changelog Entry

> Fixed - Settings page failing to render on Gutenberg 23.9+ after `@wordpress/dataviews` was removed from the private-apis allowlist.

cc @youknowriad @ntsekouras @oandregal

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-989-890b912d816d5f83cf61f97b865313df75b18143.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -97,12 +97,6 @@ export default [
 				'error',
 				{
 					selector:
-						'ImportDeclaration[source.value=/^@wordpress\\u002F.+\\u002F/]',
-					message:
-						'Path access on WordPress dependencies is not allowed.',
-				},
-				{
-					selector:
 						'JSXAttribute[name.name="id"][value.type="Literal"]',
 					message:
 						'Do not use string literals for IDs; use withInstanceId instead.',
--- a/routes/ai-home/components/DeveloperSettings.tsx
+++ b/routes/ai-home/components/DeveloperSettings.tsx
@@ -2,8 +2,8 @@
  * WordPress dependencies
  */
 import { Button, Spinner } from '@wordpress/components';
-import { DataForm } from '@wordpress/dataviews';
-import type { Field, Form } from '@wordpress/dataviews';
+import { DataForm } from '@wordpress/dataviews/wp';
+import type { Field, Form } from '@wordpress/dataviews/wp';
 import {
 	useCallback,
 	useEffect,
--- a/routes/ai-home/components/FeatureToggle.tsx
+++ b/routes/ai-home/components/FeatureToggle.tsx
@@ -4,7 +4,7 @@
 import apiFetch from '@wordpress/api-fetch';
 import { Spinner, ToggleControl } from '@wordpress/components';
 import { useDispatch } from '@wordpress/data';
-import type { DataFormControlProps } from '@wordpress/dataviews';
+import type { DataFormControlProps } from '@wordpress/dataviews/wp';
 import { useCallback, useEffect, useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { store as noticesStore } from '@wordpress/notices';
--- a/routes/ai-home/stage.tsx
+++ b/routes/ai-home/stage.tsx
@@ -21,7 +21,11 @@ import {
 } from '@wordpress/components';
 import { store as coreStore } from '@wordpress/core-data';
 import { useDispatch, useRegistry, useSelect } from '@wordpress/data';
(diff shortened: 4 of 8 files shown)


youknowriad (dismissed):
Approving but a check from folks working on this project more actively might be good.

youknowriad (dismissed):

aduth:
There's a remaining lint error on "Path access on WordPress dependencies is not allowed".

https://github.com/WordPress/ai/blob/0833eec2091dff12e86a34b0ba7435ce171be143/eslint.config.mjs#L98-L103

This specific restricted syntax may have been inherited originally from Gutenberg and was meant to prevent access to specific files in packages. These days, Gutenberg packages use [subpath exports](https://nodejs.org/api/packages.html#subpath-exports), which strictly prevents (at a Node.js/bundler level) any access to paths not listed in the exports. For this reason, the equivalent rule was removed in Gutenberg in https://github.com/WordPress/gutenberg/pull/72978 and I'd recommend doing the same in this project.

dkotter:
> This specific restricted syntax may have been inherited originally from Gutenberg and was meant to prevent access to specific files in packages. These days, Gutenberg packages use [subpath exports](https://nodejs.org/api/packages.html#subpath-exports), which strictly prevents (at a Node.js/bundler level) any access to paths not listed in the exports. For this reason, the equivalent rule was removed in Gutenberg in [WordPress/gutenberg#72978](https://github.com/WordPress/gutenberg/pull/72978) and I'd recommend doing the same in this project.

Looks like this was originally brought in via #172 and reading through that, seems those rules were pulled from other WordPress repos (I image Gutenberg was definitely an inspiration there). If that rule is no longer valid/needed, I'm fine with us removing that here

aduth:
Thanks for taking a look @dkotter ! I removed the restricted syntax pattern in https://github.com/WordPress/ai/pull/989/commits/890b912d816d5f83cf61f97b865313df75b18143 . For what it's worth, the import in `tests/e2e/playwright.config.ts` feels like it was a true positive of what the pattern was trying to prevent 😅 Though in the case of the scripts package, (1) I believe this was the intended usage pattern (e.g. there's no documentation for Playwright config, but see similar ["Extending the webpack config"](https://github.com/WordPress/gutenberg/tree/trunk/packages/scripts#extending-the-webpack-config)), and (2) the `@wordpress/scripts` package doesn't define `exports`. I guess I spoke too confidently in my previous comment when I said that "Gutenberg packages use subpath exports" since there are about a half-dozen or so that don't define exports, though those are likely intentional where subpath references like the one that's used in this project are intended.†

† Although my personal opinion is that we should still define those `exports` even if they're pointing to the exact paths that people are already expecting to be using.

dkotter (approved):

Issue #354, opened 2026-03-30
Unifiied Abilities exposure controls


artpi:
After implementing a few consumers of abilities, it seems that we need a unified effort to control **what abilities are available where**.

In the simplest implementation:

- Plugin can declare a "Surface" for abilities (`wp_abilities_register_consumer` or something like this)
- User can choose which abilities are available where

## Why?

Abilities will come from plugins and we have no control over their quality and behavior - some of them will be splendid, some of them - not so much. 
We need a way for the user to say "i want this, this and this abilities to available in Claude via MCP, but i want these avialable in my AI agent available on this site"

**If we don't do this centrally, I predict every plugin with a "surface" shipping its own version and we'll have a mess** 

## Rough idea

- You may use abilities as curently  - then you have access to entire catalog
- You also may register a "surface" for your use case for the abilities
- When you do, a UI automatically appears in WP-Admin so you can decide which abilities are available on that surface
- You can query abilities API via surface 
- Every MCP server automatically becomes a surface

It is kind of related to https://github.com/WordPress/ai/issues/348 

CC @gziolo  @jorgefilipecosta

gziolo:
Related: I opened [Abilities API: Add filtering support to `wp_get_abilities()`](https://core.trac.wordpress.org/ticket/64990) ticket to add filtering support to `wp_get_abilities()`. The [Observed need](https://core.trac.wordpress.org/ticket/64990#Observedneed) section documents exactly the fragmentation you're describing, presenting the MCP adapter, WooCommerce, and WebMCP adapter, all of which have built their own ad-hoc filtering because the API doesn't offer it natively.

The "surface" concept here maps well to the filtering + hooks pipeline proposed in that ticket. A registered surface could translate into `$args` (category, namespace, meta) plus ecosystem-scoped hooks that enforce per-surface visibility. That way `wp_get_abilities()` becomes the single query point regardless of consumer, and surface-level controls layer on top via filters rather than each plugin reimplementing the selection logic.

gziolo:
A first building block for this shipped in WordPress 7.1: `wp_get_abilities()` now supports unified filtering — declarative args (`category`, `namespace`, nested `meta` with strict comparison), per-call `item_include_callback` / `result_callback`, and global `wp_get_abilities_item_include` / `wp_get_abilities_result` filters, all running in a single registry pass. The REST collection endpoint exposes the same filters as query params, with boolean coercion for the built-in `readonly` / `destructive` / `idempotent` annotations.

[...]

PR #992, opened 2026-09-02
Add MCP Access experiment: adapter auto-install and ability exposure controls

andreilupu:
## What?

See WordPress/mcp-adapter#178 (AI-plugin integration discussed in the comments); related to #354.

Adds an **MCP Access** experiment that wires a site up to the [MCP Adapter](https://github.com/WordPress/mcp-adapter) companion plugin and gives site owners a place to control which abilities are exposed to AI agents over the Model Context Protocol:

- Enabling the experiment **installs and activates the MCP Adapter plugin from WordPress.org** if it is not already active. The experiment description states this explicitly.
- A **Tools → MCP Access** screen lists registered abilities with per-ability expose/hide checkboxes and a reset-to-default affordance, shows the MCP server endpoint URL, and surfaces the adapter's install state.

## Why?

Discussed in WordPress/mcp-adapter#178: the adapter stays a standalone, headless plugin (reusable by other consumers, no admin UI of its own — see WordPress/mcp-adapter#184), while the AI plugin provides the user-facing control layer. Abilities come from arbitrary plugins; site owners need a clear place to decide what is reachable by AI agents (the same need behind #354).

## How?

**Experiment** (`includes/Experiments/MCP_Adapter/`): standard `Abstract_Feature` experiment (`mcp-adapter`, admin category, no AI-capability requirement).

**Auto-install** (`Plugin_Installer`): runs on `admin_init` while the experiment is enabled; installs via core's `plugins_api()` + `Plugin_Upgrader` (silent skin) and activates. Guards: skips when already active; requires `install_plugins`/`activate_plugins` (which `DISALLOW_FILE_MODS` strips via `map_meta_cap`); a failed attempt sets a 1-hour transient lock so admin pages never stall on repeated downloads; a `wpai_pre_mcp_adapter_autoinstall` filter short-circuits the attempt for tests and externally managed hosts. The attempt runs on the first admin load after enabling because the experiments framework does not run disabled experiments' code in the request that enables them. While the plugin is missing/inactive, the screen shows the state, the last auto-attempt error if any, and a manual install/activate button (driving core's `wp/v2/plugins`) as the immediate retry path. The slug is filterable (`wpai_mcp_adapter_plugin_slug`) so the flow can be exercised before the adapter is published on WordPress.org.

**Exposure overrides** (`Exposure_Overrides`): stored in the `wpai_mcp_exposed_abilities` option as an ability-name → bool map; applied by injecting `meta.mcp.public` through the `wp_register_ability_args` filter (priority 100). The adapter's default server resolves exposure from exactly that key (`McpAbilityExposure`), so overrides take effect with no hard dependency between the plugins. The registration-time default is stashed before injection so the screen can always report and restore it. Abilities without an override keep their developer-declared visibility.

**Settings REST** (`Settings_Controller`, `ai/v1/mcp/settings`, `manage_options`): GET returns adapter/plugin state, endpoint URL (read from the adapter's registered servers, not hardcoded), and per-ability effective + default exposure — delegating to `McpAbilityExposure::is_meta_public()` when the adapter is active so the screen always agrees with the server. POST persists overrides (`null` removes one); values are validated with `rest_is_boolean` and coerced via `sanitize_callback`, so form-encoded clients work.

**UI** (`src/experiments/mcp-adapter/`): React screen following the existing experiment app patterns; recoverable inline error notices, `getErrorMessage()` for REST errors.

### Open questions for reviewers

1. **Enforcement layer for site-owner overrides.** Registration-time injection has the limitations the adapter's own `McpAbilityExposure` docblock warns about: consumers that trigger ability registration before `init:15` see unfiltered defaults for that request, later `wp_register_ability_args` callbacks can overwrite the injected meta, and the mutation is visible to every meta consumer. I checked whether the WP 7.1 `wp_get_abilities()` filtering ([#64990](https://core.trac.wordpress.org/ticket/64990), `wp_get_abilities_item_include`) could carry the overrides instead: it covers **discovery** (the default server lists via `wp_get_abilities()`), but not **enforcement** — `execute-ability` and `get-ability-info` fetch by name via `wp_get_ability()` and gate on `McpAbilityExposure::is_public()`, which reads meta unfiltered. Would a resolution-time filter in the adapter — e.g. `apply_filters( 'mcp_ability_is_public', $resolved, $ability )` inside `is_public()` — be acceptable? It would give site-owner controls (and eventually the "surfaces" idea from #354) a reliable enforcement point at the adapter layer, matching what 7.1 did for queries at the core layer. I'm happy to PR it to mcp-adapter; this PR's option format and screen would carry over unchanged.
2. **Relationship to #354.** This screen is deliberately scoped to MCP exposure. If the unified "surfaces" model lands, the stored option can migrate into it — flagging so we align early rather than fork the UX.
3. **Disable semantics.** Disabling the experiment stops enforcing overrides while the adapter keeps serving its defaults (the framework does not run disabled experiments' code). The description and screen state this explicitly. If reviewers prefer harder guarantees, that also lands on the resolution-time filter from question 1.
4. **Install timing.** `admin_init` (first admin load after enabling) vs. hooking the settings-save moment — happy to change if there's a preference.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Fable 5
Used for: Implementation, integration tests, and an adversarial self-review pass (which produced the hardening in the second commit). All code was reviewed, exercised locally, and is submitted under my responsibility.

## Testing Instructions

1. `npm install && composer install && npm run wp-env start`, `npm run build`.
2. Since `mcp-adapter` is not on WordPress.org yet, point the installer at a stand-in plugin — e.g. as an mu-plugin:
   ```php
   add_filter( 'wpai_mcp_adapter_plugin_slug', static fn (): string => 'hello-dolly' );
   ```
3. Enable **Settings → AI → Experiments → MCP Access**. Note the description mentions the WordPress.org install.
4. Load any wp-admin page: the stand-in plugin is installed from WordPress.org and activated automatically. (Deactivate + delete it and revisit to watch it again; as a non-admin or with `DISALLOW_FILE_MODS`, no install is attempted.)
5. Visit **Tools → MCP Access**: toggle an ability's exposure, save, confirm the "Overridden / Reset to default" state; reset and confirm the developer default returns.
6. With the real adapter active (e.g. mounted via `.wp-env.override.json`), confirm the endpoint URL row and that `WP\MCP\Abilities\McpAbilityExposure::is_public()` agrees with the screen for an overridden ability.
7. `npm run test:php -- --filter "MCP_Adapter|Settings_Controller|Plugin_Installer"` — 24 tests.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| _(n/a — new screen)_ | _TODO: add screenshot of Tools → MCP Access (missing-plugin state + ability table)_ |

## Changelog Entry

> Added - MCP Access experiment: installs and activates the MCP Adapter plugin from WordPress.org when enabled, and adds a Tools → MCP Access screen to control which abilities are exposed to AI agents over MCP.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-992-b9a2fea14439d8351c4b41b0ff9a327c95b28bff.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/Experiments.php
+++ b/includes/Experiments/Experiments.php
@@ -42,6 +42,7 @@ final class Experiments {
 		\WordPress\AI\Experiments\Editorial_Notes\Editorial_Notes::class,
 		\WordPress\AI\Experiments\Editorial_Updates\Editorial_Updates::class,
 		\WordPress\AI\Experiments\Excerpt_Generation\Excerpt_Generation::class,
+		\WordPress\AI\Experiments\MCP_Adapter\MCP_Adapter::class,
 		\WordPress\AI\Experiments\Meta_Description\Meta_Description::class,
 		\WordPress\AI\Experiments\Slug_Generation\Slug_Generation::class,
 		\WordPress\AI\Experiments\Title_Generation\Title_Generation::class,
--- a/includes/Experiments/MCP_Adapter/Admin_Page.php
+++ b/includes/Experiments/MCP_Adapter/Admin_Page.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * Admin page for the MCP Access screen.
+ *
+ * @package WordPress\AI\Experiments\MCP_Adapter
+ */
+
+declare( strict_types=1 );
+
+namespace WordPress\AI\Experiments\MCP_Adapter;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Registers and renders the MCP Access admin page.
+ *
+ * @since 0.9.0
+ */
+class Admin_Page {
+	/**
+	 * The admin page slug.
+	 *
+	 * @since 0.9.0
+	 * @var string
+	 */
+	public const PAGE_SLUG = 'ai-mcp-access';
+
+	/**
+	 * Initializes the admin page hooks.
+	 *
+	 * @since 0.9.0
+	 */
+	public function init(): void {
+		add_action( 'admin_menu', array( $this, 'add_admin_menu' ) );
+	}
+
+	/**
+	 * Adds the Tools submenu page.
+	 *
+	 * @since 0.9.0
+	 */
+	public function add_admin_menu(): void {
+		add_submenu_page(
+			'tools.php',
+			__( 'MCP Access', 'ai' ),
+			__( 'MCP Access', 'ai' ),
+			'manage_options',
+			self::PAGE_SLUG,
+			array( $this, 'render_page' )
+		);
+	}
+
+	/**
+	 * Renders the page mount point for the React app.
+	 *
+	 * @since 0.9.0
+	 */
+	public function render_page(): void {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			wp_die( esc_html__( 'You do not have sufficient permissions to access this page.', 'ai' ) );
+		}
+
+		echo '<div class="wrap">';
+		echo '<h1>' . esc_html__( 'MCP Access', 'ai' ) . '</h1>';
+		echo '<div id="ai-mcp-access-root"></div>';
+		echo '</div>';
+	}
+}
--- a/includes/Experiments/MCP_Adapter/Exposure_Overrides.php
+++ b/includes/Experiments/MCP_Adapter/Exposure_Overrides.php
@@ -1,0 +1,118 @@
+<?php
+/**
+ * Site-owner overrides for MCP ability exposure.
+ *
+ * @package WordPress\AI\Experiments\MCP_Adapter
+ */
+
+declare( strict_types=1 );
+
+namespace WordPress\AI\Experiments\MCP_Adapter;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Applies stored per-ability exposure overrides to ability registration.
+ *
+ * The MCP Adapter plugin resolves exposure from `meta.mcp.public` (falling
+ * back to `meta.public`). Overrides saved from the MCP Access screen are
+ * injected into that meta key at registration time, so the adapter's default
+ * server picks them up without the two plugins depending on each other.
+ *
+ * @since 0.9.0
+ */
+final class Exposure_Overrides {
+	/**
+	 * Option storing the per-ability exposure overrides.
+	 *
+	 * Map of ability name => bool (true: exposed, false: hidden). Abilities
+	 * not present in the map keep their registration-time default.
+	 *
+	 * @since 0.9.0
+	 * @var string
+	 */
+	public const OPTION_NAME = 'wpai_mcp_exposed_abilities';
+
+	/**
+	 * Filters ability registration args to apply a stored exposure override.
+	 *
+	 * @since 0.9.0
+	 *
+	 * @param array<string, mixed> $args Ability registration args.
+	 * @param string               $name Ability name.
+	 *
+	 * @return array<string, mixed> Filtered args.
+	 */
+	public static function filter_ability_args( array $args, string $name ): array {
+		$overrides = self::get_overrides();
+
+		if ( ! array_key_exists( $name, $overrides ) ) {
+			return $args;
+		}
+
+		if ( ! isset( $args['meta'] ) || ! is_array( $args['meta'] ) ) {
+			$args['meta'] = array();
+		}
+
+		if ( ! isset( $args['meta']['mcp'] ) || ! is_array( $args['meta']['mcp'] ) ) {
+			$args['meta']['mcp'] = array();
+		}
+
+		$args['meta']['mcp']['public'] = $overrides[ $name ];
+
+		return $args;
+	}
+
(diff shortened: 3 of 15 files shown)


dugyen:
Tested this in WordPress Playground against the CI build artifact (`pr-992-b9a2fea1...zip`). Since `mcp-adapter` isn't on WordPress.org yet, I used the `wpai_mcp_adapter_plugin_slug` filter pointed at `hello-dolly` as the stand-in, per the testing instructions, and drove the flow via a custom blueprint (`setSiteOptions` for `wpai_features_enabled` / `wpai_feature_mcp-adapter_enabled`, plus a `writeFile` mu-plugin for the slug filter) to land directly on **Tools → MCP Access**.

**Verified working:**
- Enabling the experiment auto-installs and activates the stand-in plugin (`hello-dolly`) with no manual step — confirmed via the plugin footer text and the `plugin` object in `GET ai/v1/mcp/settings` (`status: "active"`, `can_install`/`can_activate: true`).
- **Tools → MCP Access** renders the ability table correctly, populated with the three core abilities (`core/get-site-info`, `core/get-user-info`, `core/get-environment-info`), all exposed by default.
- Toggling an ability's exposure off + **Save changes** persists correctly: `POST ai/v1/mcp/settings` succeeds, the row flips to "Overridden" with a "Reset to default" link, and the override survives a full page reload.
- **Reset to default** correctly reverts the row to "Default" only after Save is clicked (client-only state until then) — `overrides` in the REST response goes back to `{}` and the ability reverts to its developer-declared default.
- Graceful degradation when the real adapter class isn't present: `adapter_active: false` and `endpoint: null` in the settings response, rather than a fatal error — matches the "no hard dependency between the plugins" design described in the PR.
- No PHP fatals or console errors surfaced during any of the above; the only console errors were expected CORS noise from Playground's own artifact-download retries.

Didn't test against the real `mcp-adapter` plugin itself (not published yet) so I can't confirm the `McpAbilityExposure::is_public()` agreement mentioned in step 6 of the testing instructions, or the actual MCP server endpoint URL rendering — that needs the real plugin mounted via `.wp-env.override.json` as noted.

Nice, clean implementation — the fallback behavior when the adapter isn't active is a good touch.

dugyen:
Reviewed the diff end to end (PHP + REST controller + React screen + tests). Didn't find a confirmed bug — the second/third/fifth commits already cover an adversarial self-review pass, and it shows. A few things I specifically checked and found solid:

- `Exposure_Overrides::filter_ability_args` correctly stashes the pre-override registration default before injecting `meta.mcp.public`, and only touches abilities that actually have a stored override.
- `Plugin_Installer` capability gating (`install_plugins`/`activate_plugins`, stripped by `DISALLOW_FILE_MODS` via `map_meta_cap`), the 1-hour failure lock, and the `wpai_pre_mcp_adapter_autoinstall` test short-circuit all behave as documented.
- `Settings_Controller`'s validate/sanitize split correctly handles form-encoded boolean-like strings, the ability-name regex, and null-removal semantics — and POST responses are built from the fresh overrides map rather than stale registered-ability meta.
- The React state machine for pending/override/reset (`McpAccessApp.tsx`) and the install-vs-activate branching against `/wp/v2/plugins` line up with the PHP-side `status`/`can_install`/`can_activate` semantics.

The remaining open items are the ones already called out in the PR description itself (registration-time injection not being the final enforcement point, `admin_init`-triggered auto-install running on every admin page load, disable semantics) — I don't have anything to add beyond what's already flagged there.

---

jeffpaul:
Dropping this into the 1.5.0 milestone assuming the MCP Adapter plugin will be available on WPORG by then (as a prerequisite for this PR).

galatanovidiu:
A few things I noticed:

- The adapter can't stay deactivated while the experiment is on. `Plugin_Installer::maybe_install_and_activate()` runs on every `admin_init`, so after deactivating MCP Adapter on the Plugins screen, the redirect to `plugins.php?deactivate=true` activates it again. I think deactivating the adapter should keep it off, so no MCP endpoint is exposed.
- The three default adapter abilities (`mcp-adapter/discover-abilities`, `mcp-adapter/get-ability-info`, `mcp-adapter/execute-ability`) show up on the MCP Access screen with a checkbox. The user should not be able to enable or disable them. They are on whenever the adapter's default server is enabled.

I haven't gone through all the details of this PR. I will get back to this as soon as I have more time to test.

PR #975, opened 2026-08-25
Bring over latest embeddings code from the PHP AI Client

dkotter:
## What?

Part of #962

Brings over the latest embeddings code from the PHP AI Client and adds a temporary WP-CLI command to test that.

## Why?

In #892 we brought over the embeddings code from the PHP AI Client into a custom `Vendor` directory, allowing us to start building on top of that prior to that being in WordPress. It was then decided that some changes were needed on that code (currently happening in https://github.com/WordPress/php-ai-client/pull/274) and so we disabled the loading of that code prior to the last release of this plugin (see #946).

While the upstream PR hasn't been merged yet, I think it's close so this PR gets us started on bringing over those latest changes and turning support back on.

In addition, a temporary WP-CLI command was brought back to easily test the changes here. Same as before, plan is to remove that prior to merge.

## How?

- Pull the changes from https://github.com/WordPress/php-ai-client/pull/274 into our custom `Vendor` directory
- Update our `generate_embeddings` helper function to support the new approach (a model is required)
- Load the `SDK_Overlay` class again, which loads in our vendored PHP AI Client
- Add a temporary WP-CLI command to test embeddings

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 5 xHigh effort
Used for: Analyzing the changes made upstream and pulling those in as well as bringing back the removed WP-CLI command and adjusting everything accordingly to account for the new changes. Final review and testing by me

## Testing Instructions

Easiest way to test is via the custom WP-CLI command this PR has. Note you'll also need to install an AI Provider plugin that supports embeddings.

1. Pull this PR down
2. Ensure you have an AI Provider plugin installed, activated and configured that supports embeddings: OpenAI and Ollama released support in the last version; [Google PR](https://github.com/WordPress/ai-provider-for-google/pull/30)
3. Open up a terminal window
4. Run `wp ai embeddings generate 'This is some text'`
5. You should get a message around missing a model and provider
6. Run `wp ai embeddings generate 'This is some text' --provider=openai --model=text-embedding-3-small`
7. You should get a message about being in dry run mode
8. Run `wp ai embeddings generate 'This is some text' --provider=openai --model=text-embedding-3-small --dry-run=false`
9. You should see a success message and information about the provider and model used as well as the embedding value (trimmed)
10. Try other versions of the command and ensure they work as expected:
- `wp ai embeddings generate --post-id=42 --provider=openai --model=text-embedding-3-small --dry-run=false`
[...]

PR #993, opened 2026-09-02
Add embedding similarity calculations

### LINKED ISSUE #962: Tracking: Implement Embedding Support
## Overview

This issue tracks the implementation of embedding support in the WordPress AI plugin.  Existing embedding work (#683, #891, #943) should be divided into smaller, independently reviewable PRs rather than combining foundational architecture and user-facing features into large feature PRs.

## Proposed Work

* [x] Add embedding support from the PHP AI Client #975 (https://github.com/WordPress/php-ai-client/pull/274)
* [x] Implement the storage and CRUD layer for embeddings #976 
* [ ] Add background synchronization and large-scale processing
* [ ] Add low-level similarity and calculation utilities #993
* [ ] Build individual features, such as semantic search (#844) or frontend chat (#142), on top of the shared foundation
* [ ] Document the architecture and developer-facing APIs
* [ ] Complete embedding model declarations or support in connector plugins where it remains missing (https://github.com/WordPress/ai-provider-for-google/pull/30)

Each PR should generally:

* Branch independently from the `develop` branch
* Be reviewable and mergeable on its own
* Avoid including user-facing features before the required foundation is available
* Include focused tests and documentation

Related PRs may be stacked when a direct implementation dependency makes that more practical.

## Testing

Local models, including Ollama, should be supported as a practical way to test embedding workflows without ongoing provider API costs.

Embeddings generally only need to be regenerated when their source content changes.  Similarity calculations and other operations using stored embeddings should not require additional provider requests.

Links to the relevant issues and PRs should be added to the checklist above as the work is defined.

dkotter:
## What?

Partially closes #962

Adds a `Vector_Math` class with various public methods that can be used to run similarity calculations against two vector embeddings. Also adds a `Vector_Ranker` class to score vectors against one another.

## Why?

#892 and #975 brought in embedding code from the PHP AI Client. #976 brought in the data layer to store embeddings. This PR is one of the next steps in the overall embedding goal, allowing us to run comparisons from one vector to another.

## How?

- Adds a `Vector_Math` class with various public methods that can be used to run similarity calculations against two vector embeddings
- Adds a `Vector_Ranker` class to score vectors against one another
- Adds another command to our temporary embeddings WP-CLI command to test things

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Fable 5.1, Opus 5
Used for: Planning approach and executing on that. Review of approach and final review and testing by me

## Testing Instructions

Easiest way to test is the custom WP-CLI command added as part of this PR:

1. Open a terminal window
2. Ensure you have embeddings generated for at least two posts
```bash
wp ai embeddings generate --post-id=1 --dry-run=false --chunk --provider=ollama --model=nomic-embed-text:latest
wp ai embeddings generate --post-id=2 --dry-run=false --chunk --provider=ollama --model=nomic-embed-text:latest
```
3. Compare the same post against itself:
```bash
wp ai embeddings compare 1 1
```
4. You should see a cosine similarity of 1 and euclidean distance of 0
5. Compare the posts against each other:
```bash
wp ai embeddings compare 1 2
```
6. You should see similarity numbers for each chunk as well as a single number for the overall comparison

## Changelog Entry

> Added - New `Vector_Math` and `Vector_Ranker` classes, allowing users to run similarity and ranking queries against vectors.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-993-863fe51628f33282471df15563d1b1379ec99f0c.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


andreilupu:
Great work on this 👏  To me, the math, classes, and tests are really solid. One thing I hit while reading through the `compare` command:

**When two posts share more than one provider/model pair, there's currently no way to compare them.**

In `resolve_stored_model()` ([includes/CLI/Embeddings_Command.php#L708-L714](https://github.com/WordPress/ai/blob/d9990d5d103b18f60522e5fa202f6c0cc607dc75/includes/CLI/Embeddings_Command.php#L708-L714)), the multiple-shared-models branch prints the shared pairs formatted as flags:

```
Posts 42 and 99 share more than one model:
  --provider=openai --model=text-embedding-3-small
  --provider=ollama --model=nomic-embed-text:latest
Error: Posts have more than one model in common. Embed both with the same model, then compare.
```

Two issues with this:

1. The `--provider=... --model=...` lines read as flags the user should pass to `compare`, but `compare` doesn't accept those flags — its synopsis only has `--metric` and `--pairs`.
2. The suggested fix ("Embed both with the same model, then compare") doesn't resolve the situation: the posts already share a model — two of them — and since `save()` upserts on `(object_type, object_id, provider, model, chunk_index)`, re-embedding with either model won't remove the other shared pair. The only way out today is deleting rows by hand (SQL or `Embedding_Repository::delete_for_object()` via `wp eval`), since no delete command is exposed.

This is easy to hit in practice: embed a couple of posts with Ollama while testing locally, later re-embed them with OpenAI, and `compare` becomes unusable for those posts.

**Suggestion:** add optional `--provider=<provider>` and `--model=<model>` flags to `compare` (mirroring `generate`) and use them to disambiguate in `resolve_stored_model()`. To be clear, this is still a same-model comparison on both sides; the flags only select *which* shared model's vectors to use, not a way to compare across models. Auto-detection can stay as the default for the common single-shared-model case, and the current error output already prints exactly the flag values the user would need to pass, so the UX would compose nicely.

dkotter:
> When two posts share more than one provider/model pair, there's currently no way to compare them

Thanks for the review. I did initially have model and provider as arguments you could pass in (thus the discrepancy in that error message) but removed that to keep things as simple as possible.

This WP-CLI command is only there for test purposes and the plan is to remove all that code prior to the next release. So I'm happy to make changes there as needed but not overly concerned with making it perfect as it's just a test tool.

Curious if you're seeing actual issues in the embedding code or if this is just an issue with that WP-CLI command? Also worth noting the work is not finished here (as you mention, there isn't a real way to delete embeddings yet). All of the actual sync work will be coming in a new PR.

andreilupu:
Oh, just the WP-CLI command, the embedding code itself looks solid to me. `Vector_Math` and `Vector_Ranker` both check out (I specifically verified the cosine clamp, zero-vector handling, sort directions, and key preservation in `rank()`).

That explains the error message, now it makes sense, and given the command is throwaway, I wouldn't polish it further.

I'll admit I was hoping these CLI commands would actually ship, though. As a plugin developer, I'd probably reach for `wp ai embeddings compare` more often than the classes themselves. It's the fastest way to check "did my content get embedded, do these two posts actually look related to the model" without writing SQL against a table of binary blobs. 

That said, I get that freezing a debug tool into a public contract before the sync layer exists would be premature. Totally fine waiting for a proper `wp ai embeddings` suite later.

dkotter:
> I'll admit I was hoping these CLI commands would actually ship, though. As a plugin developer, I'd probably reach for `wp ai embeddings compare` more often than the classes themselves. It's the fastest way to check "did my content get embedded, do these two posts actually look related to the model" without writing SQL against a table of binary blobs.
> 
> That said, I get that freezing a debug tool into a public contract before the sync layer exists would be premature. Totally fine waiting for a proper `wp ai embeddings` suite later.

So I guess no reason we couldn't leave these types of commands in, just wasn't sure if they'd provide enough value to justify having in place (and having to maintain going forward).

I do think once we have all the pieces in place here, most of the functionality that's in the command will be moved out into various other embedding classes (and the command can then be updated to use as much of that as possible). So your note here around the same post being embedded twice with different models is something we'll want to account for in the larger scope here.

jeffpaul (approved):

PR #1002, opened 2026-09-04
Rename the content and users abilities and remove ai/get-post-details

gziolo:
## What?

Renames two abilities so the names describe what they do, and removes one ability that duplicated the new content query:

- `core/read-content` is now `core/content-query`
- `core/read-users` is now `core/users-query`
- `ai/get-post-details` is removed. Use the single-post mode of `core/content-query` instead.

## Why?

Both renamed abilities query a set of items with filters and pagination, or fetch a single item. "Query" describes this better than "read". It also lines up with the query-style naming used elsewhere in WordPress.

`ai/get-post-details` fetched one post by ID and returned title, slug, content, excerpt, author, and type. The single-post mode of `core/content-query` returns the same data plus more, with read-level access rules instead of requiring edit access. Keeping both meant two ways to do the same thing.

## How?

Rename:

- The ability registrations, unregister guards, labels, and docblocks use the new names.
- The gated wrapper classes are renamed to `Content_Query` and `Users_Query`.
- The Content ability methods that embedded the old name now use `content_query` (for example `execute_content_query`).
- Integration tests, the e2e helper plugin, and the e2e specs follow the new names. The two spec files are renamed too.

Removal:

- The `ai/get-post-details` registration is gone. The gated `Post_Utilities` wrapper now only registers `ai/get-post-terms`.
- The static `Posts::get_post_details()` helper and its `wpai_get_post_details` filter stay, because `get_post_context()` uses them for the generation abilities. The helper tests call the static method directly now.
- The client-side `getContext()` helper in the image generation feature was not used anywhere, so it is deleted. `generateImage()` already reads the title and type from the editor store.
- The Custom Abilities e2e spec now checks the gate through `core/content-query`.
- Past entries in `CHANGELOG.md` and `readme.txt` keep the old names because they describe released versions.

### Deprecated aliases

The old names keep working for a few releases, so nobody has to deal with a second breaking change right away:

- `core/read-content` and `core/read-users` are registered as deprecated aliases. Each alias copies the label, description, category, schemas, and meta of its replacement, and forwards execution and permission checks to it. The description starts with a "Deprecated" note, and the meta carries a `deprecated` entry with the version and the replacement name.
- `ai/get-post-details` stays registered with the same deprecation note in its description and meta.
- Executing any of the three triggers a deprecation notice through `_deprecated_function()`, so developers see it with `WP_DEBUG` on.
- A new `register_deprecated_ability_alias()` helper in `includes/helpers.php` builds the aliases, so future renames can reuse it.
- The aliases are marked with `@todo` comments for removal after a few releases.

## Testing Instructions

1. Enable the Custom Abilities experiment on the AI settings screen.
2. Visit `/wp-json/wp-abilities/v1/abilities` and check that `core/content-query` and `core/users-query` are listed. The old names `core/read-content`, `core/read-users`, and `ai/get-post-details` are still listed, with "(deprecated)" in the label and a `deprecated` entry in the meta.
3. Run `core/read-content` with an `id` and confirm it returns the same result as `core/content-query`. With `WP_DEBUG` on, a deprecation notice is logged.
4. Run the PHP tests:
   ```
   npm run test:php -- --filter 'ContentTest|UsersTest|Gated_AbilitiesTest|Custom_AbilitiesTest|Show_In_AbilitiesTest|HelpersTest|PostsTest'
   ```
5. Run the e2e specs:
   ```
   npm run test:e2e -- tests/e2e/specs/abilities tests/e2e/specs/experiments/custom-abilities.spec.js
   ```

## Changelog Entry

> Changed - The `core/read-content` and `core/read-users` abilities are renamed to `core/content-query` and `core/users-query`. The old names keep working as deprecated aliases for now.
> Deprecated - The `ai/get-post-details` ability. Use the single-post mode of `core/content-query` instead.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1002-3fef3b20b145d3740863e2809422b4e01f6c7e97.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter:
> We could also register the old names as deprecated aliases for one or two releases. What do you think? Is a clean rename fine here, or should we add aliases?

Personally I'd suggest going the alias route here as we already shipped a breaking change with these abilities in the last release so anyone using these would have to deal with breaking changes two releases in a row.

That said, these are now behind an experiment toggle and we don't claim to provide backwards compat for experiments so could be justified in not doing that

gziolo:
Thanks for the input, that is a fair point about two breaking changes in a row. I went with the alias route.

- `core/read-content` and `core/read-users` are now deprecated aliases. They copy the replacement ability and forward execution and permission checks to it.
- `ai/get-post-details` stays registered with a deprecation note.
- All three trigger a deprecation notice through `_deprecated_function()` when executed, and their meta carries a `deprecated` entry with the version and the replacement name.
- A small `register_deprecated_ability_alias()` helper builds the aliases, so we can reuse it for future renames.

I marked the aliases with `@todo` comments so we can remove them after a few releases. Let me know if you would rather see a fixed target version.

dkotter (approved):

justlevine:
(I know this already merged, but as there's no matching GH issue for this that I see, so leaving the feedback here) 

- I fail to see how `*-query` is a better pattern than `get-*` which is what core currently uses, even if it's better than `read-*` (_Both_ still fall short when compared to slug fragments 🤷)
- Regarding breaking changes, I'd also recommend using `ai/*` as the namespace instead of `core/`, both because 
   - it's _in_ the AI plugin, 
   - if these ever do make it to WordPress `core` with this naming structure intact (full disclosure I'll be fighting hard against), that would allow us to prevent _another_ breaking change and give us a better way to handle the migration path since we'll have the `ai/*` alias.
   
Recommending instead:
- `ai/get-user-data` (until an eventual merge as `core/user/get` )
- `ai/get-post-object-data` (until an eventual merge `core/post-object/get`. Separately, I would _love_ if we could redo the last decade of WP and the "Everything is a Post", but until we can think of a holistic reason as to why a Category description, Product Brand image, or any other taxonomic meta data isn't also "content", IMO there's no responsible way to use that word in reference to `WP_Post` and not `WP_Post->post_content`)

Issue #1003, opened 2026-09-04
Let the AI Workspace's tool surface come from the Abilities API


whyisjake:
# Let the AI Workspace's tool surface come from the Abilities API

## Problem

The AI Workspace (#282) ships with a hand-maintained allowlist of two abilities: `ai/search-content` and `ai/propose-drafts`. Everything else registered on the site — including every ability a third party registers — is invisible to the assistant.

That makes the workspace a chat feature. The plugin's actual thesis is that the Abilities API is the extension point, and the workspace is its first real conversational consumer. A site that registers a WooCommerce ability, a custom CPT ability, or an internal tool gets no benefit from it today.

The gap shows up in #282's own examples. "Help me configure my site navigation" and "build a custom 404 page that finds the three most likely posts" are not reachable with a search tool and a draft proposer. The ticket asks for outcomes that only a broad tool surface delivers, and then the implementation narrows the surface to two.

The narrowing was deliberate and correct for a first cut — every ability the model can call is also reachable by an instruction embedded in content someone else wrote. This issue is about replacing that hand-curation with a policy.

## Proposed behaviour

Admit abilities by what they declare, not by a list someone maintains.

`WP_Ability` already carries semantic annotations — `readonly`, `destructive`, `idempotent` — and this plugin already sets them on its own abilities. A policy built on them:

- **`readonly: true` and not `destructive`** — admitted automatically. The assistant can call it; results are wrapped as untrusted, provenance-tagged data as they are today.
- **Writes** — admitted only behind the propose-then-confirm path the workspace already implements for drafts, where a person approves the resolved values before anything is written.
- **`destructive: true`** — never admitted. This matches the workspace's existing product decision that the assistant lists things for a person to act on rather than acting itself.
- **Unannotated** — *not* admitted by default. Annotations default to `null`, not `false`, so "not declared readonly" must mean "unknown", never "safe".

That last point matters more than it looks. The annotations are documented as "hints for tooling and documentation": they are the ability author's claim about their own code, not a property WordPress enforces. A policy can use them to decide a **default**, but it cannot treat them as a security boundary by itself. Two things follow:

1. The site owner needs to see the resulting list and be able to change it — which is also the most literal reading of "surface the Abilities API" for a non-developer.
2. An exposure axis is worth considering, so registering an ability does not silently publish it to every consumer at once. Today `show_in_rest` and `meta.mcp.public` control REST and MCP exposure; a workspace surface is a third audience with a different risk profile.

## Scope boundaries

[...]

PR #255, opened 2026-02-24
Remove duplicate error display in generate Alt text

juanfra:
## What?

Removes the error message from the controls area when AI summary generation fails.

## Why?

Errors were being displayed twice. Once in the top notice bar and again right above the button. This redundancy was confusing and you had to dismiss the two independently.

## How?

Removed the error display from the controls area

## Testing Instructions

1. Be sure that you don't have any AI provider with text generation capabilities configured (maybe just try without any configured)
2. Go to a page or post.
3. Add an image.
4. In the image controls, head over "Settings" and click "Generate Alt text"
5. Confirm the error notice is showing only once on the top bar.

## Screenshots or screencast 

**Before**

https://github.com/user-attachments/assets/46be73f5-0899-4523-9fb9-532c18cffe76

**After**

Screenshot 2026-02-24 at 15-57-48 Edit Page “WordPress in the AI era ” ‹ Stories® — WordPress

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginZipFile%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-255-6dbea05475f2dfcadd91db1c22ca9be3ad2cdc50.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #1004, opened 2026-09-04
Add the Global AI Workspace

### LINKED ISSUE #282: Chat experiment: Integration outside the editor and outside single-task AI use

whyisjake:
Implements #282: a full-screen conversational admin screen where a site owner talks to an AI that can read their content — under their own capabilities, never above them.

## What this adds

A new `ai-workspace` experiment (Tools → AI Workspace, `manage_options`), built on the Abilities API and core's ability-backed tool loop:

- **A turn endpoint and tool loop.** Each function call runs through the resolver's single-call `execute_ability()` rather than the batch path, so every invocation — including denials — produces exactly one log row and a provenance envelope.
- **Two read abilities.** `ai/search-content` returns titles and excerpts only; `ai/read-content-bodies` returns full bodies for at most five posts named by ID. Both filter row by row at execute time against the requesting user's own capabilities.
- **Propose-then-confirm writes.** Draft creation is proposed with resolved values a person approves before anything is written. No write ability is registered.
- **Streaming**, forward-ported through the existing SDK overlay pattern with an Anthropic SSE mapper, gated so a site that cannot stream falls back cleanly and says so.
- **A block editor handoff**, opening the workspace seeded with the current post's identity — never its body, which would be a second, unenforced way in.
- **A retrieval trace**, reporting per invocation what was searched, what came back, and what permission withheld.

## Notes for review

**Permission filtering is execute-time, not declaration-time.** The coarse capability on a tool decides whether it is *declared*; which rows come back is decided per row, per user, on every call. Both matter, and the tests mutate each independently to prove neither is inert.

**Withheld counts are measured, not derived — a deliberate divergence from the plan.** The plan's U13 prescribed the opposite, specifying the withheld count as "the difference between the ability's total and its returned rows". Implementing it revealed that verification to be wrong, so it was not followed. `total - count(results)` conflates pagination with permission filtering: page one of fifty matches would announce forty posts hidden by the person's role on an ordinary search. `ai/search-content` counts withholding as the permission walk drops rows from the page it builds. `ai/read-content-bodies` deliberately reports no count at all — it answers for IDs the caller named, so counting unknown and unreadable apart would say whether those exact posts exist.

**Tool surface is a hand-maintained allowlist of three abilities**, which is deliberate for a first cut and is the subject of follow-up #1003 — every ability the model can call is also reachable by an instruction embedded in content someone else wrote.

## Vendored SDK code (please review deliberately)

This PR adds **11 files / +2,119 lines under `includes/Vendor/AiClient/`**. Reviewers should weigh this separately from the rest of the diff, because it is third-party code rather than plugin code.

**What it is.** `WordPress\AiClient` is the PHP AI Client, bundled *inside WordPress core* at `wp-includes/php-ai-client/` — it is not a Composer dependency of this plugin, which has no runtime dependencies at all. `includes/Vendor/AiClient/` is not a copy of that library: the files declare the SDK's own `WordPress\AiClient\…` namespace and are loaded by `SDK_Overlay`'s **prepended** autoloader, so the plugin can answer a class request before core's autoloader and supply classes the bundled release does not have yet. Each is gated by a sentinel so the overlay stands down once core catches up.

**This is an existing pattern, not a new one.** `develop` already vendors 10 files this way for the `embeddings` feature. Streaming is the second consumer and follows the same shape. Nothing in core or in `vendor/` is modified.

**The part that genuinely differs from `embeddings`, and the reason to look closely:** the `embeddings` overlay front-ports classes from upstream `trunk` — code that exists and will land. The `streaming` overlay is **pinned to the head of an unmerged PR (php-ai-client#255) on a fork**, not to upstream trunk. If that PR changes shape or never merges, this branch carries 2,119 lines of a future version that never arrives. That is a defensible trade for a feature-flagged experiment, but it should be an explicit decision rather than something discovered inside a large diff.

**Measured against this project's own expectations for an Experiment.** `docs/FEATURE_EXPERIMENT_LIFECYCLE.md` lists, under *"Expectations for Experiments at this stage"*, that "Code is intentionally scoped and easy to remove or revise"; and under *"When reviewing an Experiment, maintainers typically consider"*, "Whether the Experiment is sufficiently self-contained." Neither is a hard gate — the same document says "Early implementations do not need to be production-ready, but they should be reviewable, testable, and understandable by others" — but the streaming overlay sits further from both than `embeddings` does, and that is a maintainer's call to make knowingly rather than to discover.

**Two reversal costs, and they are not the same.** Turning streaming *off* is a one-line filter: the seam already exists at `includes/Experiments/AI_Workspace/REST/Turn_Controller.php:290` (`wpai_workspace_stream_emitter`), it is already used to suppress streaming in `tests/e2e-testing/e2e-testing.php:22`, and the buffered fallback in `Prompt_Model_Client::generate()` is exercised on every non-streaming host. *Removing* the 2,119 vendored lines is a separate and larger act that no filter performs — until php-ai-client#255 merges, the plugin carries a dependency on an unmerged upstream fork whether streaming is enabled or not. The cheap answer addresses behaviour; the lifecycle expectation is about code. Both are stated here so the first is not mistaken for the second.

**The re-vendor trigger.** When php-ai-client#255 merges, re-vendor the streaming subset from upstream `trunk` and re-check the guard methods — if a later release adds `Response::getStream()` or `RequestOptions::isStream()` for an unrelated reason, the guards stop discriminating and must be re-picked.

Two consequences are recorded in `includes/Vendor/AiClient/README.md`:

- **A re-vendor obligation** once #255 merges.
- **A guard fragility.** The overlay discriminates on the presence of `Response::getStream()` and `RequestOptions::isStream()`. If a later SDK release adds either method for an unrelated reason, the guards silently stop discriminating and must be re-picked.

Every symbol the trunk-era vendored classes reference was checked against the bundled 0.3.1 signatures, and the modified `Response` is a strict superset: the constructor's `?string $body` widens to `string|StreamInterface|null`, and `getBody()`/`getData()`/`toArray()` behave identically for a string body.

**On the Codecov report:** those 2,119 vendored lines are ~21% of this patch and are its worst-covered files (`RequestOptions` 19%, `Response` 40%, `ChunkAccumulator` 69%), because they are upstream code carrying paths this plugin never calls. Coverage currently measures all of `includes/` with no exclusions, so third-party code is scored as if it were ours. Project coverage still rises with this PR (75.37% → 75.99%).

## Testing

- `npm run test:php` — **1717 tests, 5076 assertions, 39 skipped, 0 failures.**
- `npm run test:e2e` — **32 AI Workspace specs passing**, against a mocked Anthropic provider with sequenced tool-calling turns. No network access.
- `composer run-script lint` (PHPCS, WordPress-VIP-Go + Slevomat) — clean.
- `vendor/bin/phpstan analyse --memory-limit=3G` (level 8) — no errors.
- `npm run typecheck`, `npm run lint:js` — clean.

**Test suites executed by name**, as the [WordPress AI Guidelines](https://make.wordpress.org/ai/handbook/ai-guidelines/) require:

PHPUnit (`tests/Integration/`) — `AI_WorkspaceTest`, `Turn_ControllerTest`, `Tool_SelectorTest`, `Proposal_ControllerTest`, `Propose_DraftsTest`, `Prompt_Model_ClientTest`, `Search_ContentTest`, `Read_Content_BodiesTest`, `SDK_OverlayTest`, `Streaming_Turn_DriverTest`, `Streaming_Http_TransporterTest`, `Anthropic_Stream_MapperTest`, `Anthropic_Streaming_Text_Generation_ModelTest`, `Fopen_Stream_OpenerTest`, `UninstallTest`.

Playwright (`tests/e2e/specs/experiments/`) — `ai-workspace.spec.js`, `ai-workspace-tools.spec.ts`, `ai-workspace-post-results.spec.ts`, `ai-workspace-markdown.spec.ts`, `ai-workspace-handoff.spec.ts`.

Individual tests named in the mutation evidence below: `test_execute_callback_clamps_to_five_without_schema_validation`, `test_private_body_of_another_author_is_withheld_from_author`, `test_draft_body_of_another_author_is_withheld_from_lower_roles`, `test_unexposed_post_types_are_never_read`, `test_mixed_request_returns_only_readable_posts`, `test_retrieval_summary_never_counts_paginated_rows_as_withheld`, `test_pagination_alone_reports_nothing_withheld`, `test_withheld_counts_the_rows_the_permission_walk_dropped`, `test_vendored_files_use_the_prefixed_psr_dependencies`.

Security-relevant tests were verified by mutation rather than by passing alone. Forcing `Read_Content_Bodies::check_read_permission()` open and removing its clamp fails five tests, including one asserting an unexposed post type's body sentinel appears nowhere in the encoded result. Replacing the withheld count with subtraction fails `test_retrieval_summary_never_counts_paginated_rows_as_withheld` and `test_pagination_alone_reports_nothing_withheld`.

## Still open

Opening as a draft — the presentation layer is being worked separately, and three decisions are unresolved: the confirmation modal's default selection state, the full-screen gap (`is-fullscreen-mode` is applied but `@wordpress/interface` CSS is never enqueued), and the editor handoff's placement in the Options menu.

---

AI assistance: Yes. Tool(s): Claude Code (Opus 5). Used for: implementation of all units in the linked plan, test authoring including the mutation testing described above, and this description.

> **@whyisjake — this line needs your sign-off and I have deliberately not written it for you.** The [WordPress AI Guidelines](https://make.wordpress.org/ai/handbook/ai-guidelines/) require the contributor to understand every line submitted and be able to explain it under review. That attestation is yours to make or reword, and it is not something an agent can satisfy on your behalf. Replace this block before marking the PR ready.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1004-b22cafb966bf3b0be7ca9668837157591dc1ca10.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
@whyisjake we'll likely want the streaming PR in the PHP AI client into a tagged release there as a precursor to this PR (and perhaps something in the Requests library as well?), so if you could help on those upstream effort(s) and then come back to this PR that'll help keep things unblocked.

whyisjake:
Thanks @jeffpaul, agreed on the sequencing.

WordPress/php-ai-client#255 is rebased and mergeable and is waiting on @JasonTheAdams's review. The streaming subset vendored here is byte-identical to the current head of that PR, apart from the core-prefixed PSR imports, and the AI Workspace test suites run it under core's prefixed environment. I'll post that as a consumer review there and weigh in on the two open design questions (the opt-in streaming interface, and aborting a stream mid-way).

On Requests: WP's HTTP API buffers the whole body, so this branch opens the stream with `fopen()` and re-applies `wp_http_validate_url`, connector approval, and logging itself. Requests already fires `request.progress` per chunk, exposed in WP as the `requests-request.progress` action, but that is push-based. The `StreamedGenerativeAiResult` in WordPress/php-ai-client#255 pulls from a PSR-7 stream. Bridging the two needs Requests or WP_Http to return a lazy body stream. If that's the upstream change you had in mind, I'll open an issue proposing it before doing anything more here.

This stays a draft until WordPress/php-ai-client#255 is in a tagged release, then I'll re-vendor from trunk.

thelovekesh:
> Bridging the two needs Requests or WP_Http to return a lazy body stream. If that's the upstream change you had in mind, I'll open an issue proposing it before doing anything more here.

Hey @whyisjake, I’ve proposed adding streaming support to the Requests library in https://github.com/WordPress/Requests/issues/1056, and there’s also an open PR for it.

PR #1005, opened 2026-09-04
Import core's prefixed PSR EventDispatcher in the vendored EmbeddingBuilder

whyisjake:
The vendored `EmbeddingBuilder` imports an unprefixed PSR interface that does not exist under a WordPress bootstrap.

## The bug

WordPress core scopes the PHP AI Client's PSR dependencies under `WordPress\AiClientDependencies\`. Core ships the interface at `wp-includes/php-ai-client/third-party/Psr/EventDispatcher/EventDispatcherInterface.php`, declaring `namespace WordPress\AiClientDependencies\Psr\EventDispatcher`. The vendored file imports the bare name instead:

```php
use Psr\EventDispatcher\EventDispatcherInterface;
```

**Why it has not blown up yet.** The symbol appears only as a nullable typed property and constructor parameter:

```php
private ?EventDispatcherInterface $eventDispatcher;
public function __construct( …, ?EventDispatcherInterface $eventDispatcher = null )
```

PHP does not resolve a nullable type while the value is null, and nothing currently passes a dispatcher. The first caller to pass a real one gets a `TypeError` — core's prefixed interface does not satisfy the unprefixed name. The failure lands when the feature is first *used*, not when it was introduced, which is the worst time to find it.

## Not an upstream change

This is a vendoring adaptation, not a defect in [php-ai-client](https://github.com/WordPress/php-ai-client). Upstream installs its PSR packages through Composer, where the bare name is correct — prefixing there would break their builds to fix a problem only WordPress has. There is nothing to send upstream.

## The missing guard

Nothing asserted this property, so a future re-vendor could reintroduce it silently. This adds `SDK_OverlayTest::test_vendored_files_use_the_prefixed_psr_dependencies()`, which scans the vendored tree and fails on any unprefixed `Nyholm\` or `Psr\` import.

It deliberately matches **every** `Psr\` namespace rather than `Psr\Http\` alone: core prefixes `Psr\EventDispatcher\` and `Psr\SimpleCache\` the same way, and a PSR-7-only pattern is narrower than the hazard it exists to prevent — which is precisely how this import went unnoticed.

## Testing

- `npm run test:php` — **1571 tests, 4487 assertions, 39 skipped, 0 failures.**
- `composer run-script lint` — clean.
- `vendor/bin/phpstan analyse --memory-limit=3G` (level 8) — no errors.

The guard was verified load-bearing rather than assumed: reverting the import to the bare name fails the new test, naming the offending file. Restored, it passes across all vendored files.

## Relationship to #1004

These same changes are also present in #1004 (AI Workspace), where the guard is needed for the streaming files that PR vendors. They are byte-identical, so **whichever merges first makes the other a no-op.** This PR exists so the fix is not blocked behind that much larger, still-draft branch.

---

AI assistance: Yes. Tool(s): Claude Code (Opus 5). Used for: diagnosis, the fix, the guard test, and this description.

> **@whyisjake — this attestation is deliberately left for you.** The [WordPress AI Guidelines](https://make.wordpress.org/ai/handbook/ai-guidelines/) require the contributor to understand every line submitted and explain it under review; that is not something an agent can satisfy on your behalf. Replace this block before merging.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1005-eca91ea150cb61096427244359c3512ec0b95eb9.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Vendor/AiClient/README.md
+++ b/includes/Vendor/AiClient/README.md
@@ -55,7 +55,25 @@ upstream whenever a later PR touches the embedding execution path.
 
 ## Modifications applied
 
-**None.** Unlike the `Secrets` vendor directory, these files keep their real
+**One prefixed import.** WordPress core scopes the PHP AI Client's PSR dependencies under
+`WordPress\AiClientDependencies\`, so the unprefixed names do not exist under a WordPress
+bootstrap. `src/Builders/EmbeddingBuilder.php` imports
+`Psr\EventDispatcher\EventDispatcherInterface`, and the vendored copy uses core's prefixed name.
+Only the `use` line differs from upstream; the class body is byte-identical.
+
+| File | Upstream import | Vendored import |
+| --- | --- | --- |
+| `src/Builders/EmbeddingBuilder.php` | `Psr\EventDispatcher\EventDispatcherInterface` | `WordPress\AiClientDependencies\Psr\EventDispatcher\EventDispatcherInterface` |
+
+This is a vendoring adaptation, not an upstream bug: upstream installs its PSR packages through
+Composer, where the bare name is correct. It is not a change to send upstream.
+
+`SDK_OverlayTest::test_vendored_files_use_the_prefixed_psr_dependencies()` asserts that no vendored
+file imports an unprefixed `Nyholm\…` or `Psr\…` symbol, so a future re-vendor cannot silently
+reintroduce one. It matches every `Psr\` namespace rather than `Psr\Http\` alone, because core
+prefixes `Psr\EventDispatcher\` and `Psr\SimpleCache\` the same way.
+
+Otherwise unchanged. Unlike the `Secrets` vendor directory, these files keep their real
 `WordPress\AiClient\…` namespace unchanged. That is required: the updated OpenAI/Google/Ollama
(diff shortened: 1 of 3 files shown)


dkotter (approved):
Changes look correct and test well.

For future reference, we do have a [specific PR template](https://github.com/WordPress/ai/blob/develop/.github/PULL_REQUEST_TEMPLATE.md) we want people to use and it doesn't appear this one uses that (seems to just be AI generated copy). That template is helpful to follow as it includes things like the changelog entry, actual test steps (not just automated tests) and AI usage.

PR #1007, opened 2026-09-05
Stop measuring vendored SDK code as this plugin's coverage

whyisjake:
Coverage measures `includes/Vendor/`, which is third-party SDK code copied verbatim from upstream [php-ai-client](https://github.com/WordPress/php-ai-client). Those files carry paths this plugin never calls, so the score reflects upstream's test suite rather than ours.

`phpcs.xml.dist` and `phpstan.neon.dist` already exclude that directory. This adds the same exclusion for coverage: an `<exclude>` in `phpunit.xml.dist` so the files never enter the clover report, and a new `codecov.yml` so they are absent from Codecov's view of it.

The percentage can move in either direction, since this removes covered lines as well as uncovered ones.

## Testing

`npm run test:php` passes at 1571 tests, which confirms PHPUnit still parses the config; a malformed `<coverage>` block fails the run outright. Codecov's validator accepts `codecov.yml` and compiles the matcher to `(?s:includes/Vendor/.*)\Z`.

Not verified locally: that the clover report itself omits the files. Neither wp-env container has Xdebug or PCOV, so no coverage run is possible here. CI will show it on this PR.

---

AI assistance: Yes. Tool(s): Claude Code (Opus 5). Used for: identifying the gap during a code review of #1004, both edits, and this description.

> **@whyisjake — this attestation is deliberately left for you.** The [WordPress AI Guidelines](https://make.wordpress.org/ai/handbook/ai-guidelines/) require the contributor to understand every line submitted and explain it under review; that is not something an agent can satisfy on your behalf. Replace this block before merging.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1007-7e28ba32ac9063173a65707fe3455b662f8bbcff.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


codecov.yml
```diff
@@ -0,0 +1,13 @@
...
+#
+# `includes/Vendor/` holds third-party SDK code copied verbatim from upstream
+# (see includes/Vendor/AiClient/README.md). It is not written or maintained
+# here, and it carries code paths this plugin never calls, so scoring it
+# measures upstream's test suite rather than this one. The repository already
+# treats that directory as not-our-code: phpcs.xml.dist and phpstan.neon.dist
+# both exclude it. This does the same for coverage.
+#
+# phpunit.xml.dist carries the matching exclusion so the underlying report
+# never contains those files in the first place.
```
justlevine: This reads like doc-slop. If you think `ignore: includes/Vendor` isn't self documenting, then a single-line "Locally-vendored dependencies" should be enough to communicate why this is being ignored.
```suggestion

```

phpunit.xml.dist
```diff
@@ -22,6 +22,16 @@
...
 			<directory suffix=".php">includes</directory>
 		</include>
+		<!--
+			Vendored third-party SDK code, forward-ported verbatim from upstream
+			php-ai-client. It carries code paths this plugin never calls, so
+			measuring it reports how complete upstream's own test suite would need
+			to be rather than how well this plugin is tested. Excluded here for the
+			same reason phpcs.xml.dist and phpstan.neon.dist already exclude it.
+		-->
+		<exclude>
```
justlevine: More doc-slop. If a comment is actually needed (IMO it isn't), then a simple one-line `<!-- Vendored 3rd-party code -->` should more than suffice.

```suggestion
		<exclude>
```

.github/.codecov.yml
justlevine, line 1: Can we relocate this file to `.github/.codecov.yml`?

- `.github/*` keeps it from cluttering up the repo repo and helps with self-documentation: this file is only used by our GitHub CI
- `.codecov.yml` dotfiles makes it visually explicit that this is a config file. 

(supported per https://docs.codecov.com/docs/codecov-yaml#can-i-name-the-file-codecovyml )

justlevine (changes requested):
Thanks so much for this PR @whyisjake 🙇

Please be mindful of doc-slop, both in in the PR description (that project maintainers need to read and evaluate) and in the diffs themselves (where it also becomes infectious to future AI-led contributions).

PS: Also big thanks for adding the `AI Disclosure` to the bottom of the PR description, it was immensely helpful to allow me not to worry about human "intent" and made it quick and easy to share feedback 🙌

whyisjake:
Fair hit, thank you. The description ran 588 words for a 23-line change, and the diff carried eleven lines of comment over two lines of config.

Both are trimmed. The config comments now point at `includes/Vendor/AiClient/README.md` rather than restating the reasoning, which also removes a third copy of it that was free to drift from the other two.

The infectious point is the one that landed for me. A verbose comment in a config file is what the next contributor copies.

Kept the AI disclosure, since it was useful.

justlevine (approved):
🙌
(Proven working by the codecov report )

dkotter (approved):

PR #747, opened 2026-06-19
Feat[Experiment]: Add AI-Powered Content Translation

yogeshbhutkar:
## What, Why, and How?

Closes https://github.com/WordPress/ai/issues/187

This PR adds a new Content Translation experiment that lets users translate post content directly from the editor. It introduces a backend ability, `ai/content-translation`, to achieve this.

The feature is needed so editors can generate translations without leaving the post editing flow. It supports common target languages, gates access through the experiment settings, checks provider support for text generation, validates the selected language, and requires the user to have edit permissions for the post being translated.

Future explorations:

- Add a dedicated action for translating the post title independently.
- Add per-block translation controls through the Content Resize toolbar.
- Extend the support to multiple content blocks.

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-5.5
Used for: Test suggestions and frequent AI-assisted code reviews.

## Testing Instructions

1. Enable the Content Translation experiment in `Settings > AI`.
2. Confirm that an AI Connector with text generation support is configured.
3. Create or open a post that contains translatable paragraph or heading content.
4. In the post sidebar, click Generate Translation, choose a target language, and start the translation.
5. Verify that the translated content is applied to the expected blocks.

## Screencast

https://github.com/user-attachments/assets/cf3b4d06-4440-48b3-a782-9b8c0315da3e

## Changelog Entry

> Added - New Experiment: Content Translation; translates Paragraph and Heading blocks—and optionally the post title—into a selected language directly from the post editor

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-747-7a65b29753590670c2bd87db25ac1025ce820ee9.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #1009, opened 2026-09-07
Content Translation: support list item, verse, preformatted, and pullquote blocks

### LINKED ISSUE #979: Increase the types of blocks Content Translation supports

Currently the Content Translation experiment (introduced in #747) will only translate heading and paragraph blocks (as well as optionally the title). While this is good enough to test the feature out, ideally we support as many blocks as possible.

Increase block type support from only headings and paragraphs to as many Core blocks (that contain text) as we can. Worth noting some of our other experiments do already process additional blocks (see Editorial Notes as an example). So I think there's already functionality built-in that we can look to take advantage of here.

_No response_

- [x] Yes

dugyen:
## What?
Closes #979

Expands the Content Translation experiment's editor UI to support translating `core/list-item`, `core/verse`, `core/preformatted`, and `core/pullquote` blocks, in addition to the existing `core/paragraph` and `core/heading`.

## Why?
The Content Translation experiment (#747) only translates paragraph and heading blocks, even though the plugin already ships shared infrastructure for reading and writing a block's primary text attribute regardless of block type — the Editorial Updates experiment already uses it for a wider set of blocks (`getEditableTextAttribute()` in `src/utils/blocks.ts`). This PR takes advantage of that existing infrastructure for Content Translation, as suggested in #979.

## How?
- `src/experiments/content-translation/constants.ts`: added `core/list-item`, `core/verse`, `core/preformatted`, and `core/pullquote` to `TRANSLATION_SUPPORTED_BLOCK_TYPES`.
- `src/experiments/content-translation/utils/index.ts`: `getTranslatableBlock()` now also resolves and returns the block's editable attribute via the shared `getEditableTextAttribute()` helper.
- `src/experiments/content-translation/hooks/useContentTranslation.ts`: the write-back path previously hard-coded the `content` attribute when applying a translation. That's a latent bug for any block whose text lives elsewhere — `core/pullquote` uses `value` — so adding it to the supported list without this fix would have translated the block but silently failed to apply the result. The write-back now uses the attribute resolved above.
- Updated `docs/experiments/content-translation.md` to reflect the new supported blocks and explain the RichText-attribute constraint.
- Added an e2e test (`tests/e2e/specs/experiments/content-translation.spec.js`) covering all four new block types, including that the pullquote's `value` attribute (not `content`) gets updated.

**Deliberately out of scope**, with reasoning left as a code comment in `constants.ts`:
- `core/table` and `core/image`: their text isn't a single RichText attribute, so they need dedicated read/write handling this experiment doesn't have.
- `core/quote`'s and `core/pullquote`'s `citation` field, and `core/details`'s `summary`: secondary text the current one-attribute-per-block write-back path doesn't target.
- `core/code`: source code, not prose to translate.
- Container blocks (`core/quote`, `core/list`) need no direct entry: their own text lives in child blocks (`core/paragraph`, `core/list-item`), which are already reached because the block tree is flattened before filtering.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Surveying open issues to find an unclaimed, well-scoped one; tracing the existing block-support and write-back code across the Content Translation and Editorial Updates experiments to find the shared `getEditableTextAttribute()` infrastructure; identifying the value-attribute write-back bug; implementing the fix, docs, and e2e test. I reviewed the diff, ran `npm run typecheck` and `wp-scripts lint-js` locally (both clean), and take responsibility for the change.

## Testing Instructions
1. Enable Experiments globally, then enable the Content Translation experiment.
2. Create a post with a `core/list-item` (inside a `core/list`), a `core/verse`, a `core/preformatted`, and a `core/pullquote` block, each with enough content to pass the minimum length.
3. Open the post sidebar, click **Generate Translation**, pick a target language, and translate.
4. Verify all four blocks are translated in place — in particular, verify the pullquote's quote text updates (not just paragraph/heading).
5. `npm run typecheck` and `npx wp-scripts lint-js` on the changed files.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1009-a1293b093f4146ce2f1221c708307bca552fb04d.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dugyen:
## Test report

Tested commit `47979acc0882738d3cf238dd56c3c8a5edc4154b` against `develop`.

**Verdict: looks correct, recommend approve.** Static checks are clean and the code change matches the stated fix. Live, end-to-end translation in a running editor could not be exercised in this environment (details below).

### Verified
- ✅ Plugin builds and installs cleanly — installed the PR's CI-artifact zip into a fresh WordPress Playground instance; Plugins page shows `AI 1.3.0` active, no activation errors.
- ✅ Content Translation experiment surfaces and toggles on under Settings → AI → Editor Experiments.
- ✅ `npm run typecheck` — 0 errors.
- ✅ `npx wp-scripts lint-js` on the four changed files — 0 errors, 0 warnings.
- ✅ Write-back logic reviewed against the claimed fix: `getEditableTextAttribute()` resolves `content` for list item/verse/preformatted and `value` for pullquote; `useContentTranslation.ts` now writes to that resolved key instead of a hard-coded `content`. Matches the PR description.
- ✅ New e2e spec (`tests/e2e/specs/experiments/content-translation.spec.js`) reviewed — inserts all four new block types, translates via a mocked response, and asserts each block's own attribute updates, including that pullquote checks `value` not `content`. Well-targeted.

### Not run
- The included Playwright e2e suite (needs `wp-env`/Docker — unavailable in this environment).
- Live "Generate Translation" click-through against a real AI connector — no API key was entered (out of scope for this session), and separately, the Playground editor's iframe stopped accepting synthetic input partway through testing (reproduced across two fresh Playground sessions; unrelated to this PR's code — ordinary wp-admin settings pages responded fine).

### Minor finding (non-blocking)
`includes/Experiments/Content_Translation/Content_Translation.php:47` — the experiment's admin-facing description (Settings → AI → Editor Experiments card) still reads *"Translate paragraph and heading blocks into a different language."* `docs/experiments/content-translation.md` was updated to list all six supported blocks in this PR, but this PHP string wasn't. Cosmetic, but now inconsistent with the docs this same PR updated.

---
🤖 Tested and reported by [Claude Code](https://claude.com/claude-code)

dugyen:
Fixed the stale description string flagged above in eac36cc — the Settings → AI card now lists all six supported block types.

dugyen:
Ran the new e2e coverage (`tests/e2e/specs/experiments/content-translation.spec.js`) for real against a live WordPress site (bypassing the Docker/wp-env requirement noted in my earlier comment, by pointing Playwright at an already-running instance with the `ai-provider-for-openai` connector + the `tests/e2e-testing` mock plugin active, matching the project's own `.wp-env.json` test setup).

**Found one real bug**, isolated to the test itself (not the PR's translation code): the *"Translates list item, verse, preformatted, and pullquote blocks"* test asserted on `getByRole('document', { name: 'Block: Verse' })`, but current Gutenberg has renamed `core/verse`'s display title to **"Poetry"** (`wp.blocks.getBlockType('core/verse').title === 'Poetry'`) while keeping the block name `core/verse` for backward compatibility. So the block's accessible name is actually `"Block: Poetry"`, and the locator never matched — even though the translation itself was applied correctly (verified manually via the block editor's data store). Since `.wp-env.json` tracks WP "latest stable," this would very likely have failed the same way in CI.

Fixed in eac36cc6..8b9d22a8 by updating the locator to `Block: Poetry`.

**Result: 10/10 e2e tests passing** on the current branch tip (`8b9d22a8`), including full coverage of all four new block types (list item, verse, preformatted, pullquote) and the retry-on-failure flow.

Verdict: looks correct, recommend approve.

dkotter:
Thanks for the PR @dugyen. Increasing the blocks we support is great but wondering if there's a better way to ensure we support all Core blocks, rather than only supporting a subset? I understand not all blocks have text to translate so we may need to maintain some sort of list but it would be ideal if any text within any Core block can be translated.

I just fear we'll have to continue to open new PRs to increase our block support here and would be ideal if we can just support everything (more or less). I know our Editorial Notes feature does already support most blocks (including tables) so that's an area I'd suggest looking at for ideas.

PR #1010, opened 2026-09-08
Docs: add integrator guidance to the Agent Users README

webmyc:
Follow-up to https://github.com/WordPress/ai/pull/961#issuecomment-5518462186, where you said you would take the integrator section as a follow-up. Based on `add/agent-users-experiment` rather than `develop`, because the README does not exist there yet.

Scope is exactly the section that was agreed: guidance for third-party code, no hooks, no changes to the identity model. It points at `wpai_is_agent_user()` and the documented `wpai_agent` key and stops there. I have deliberately not touched the site-owned KSES allowlist, since you postponed that until the identity contract settles and i think that is the right order.

### Why these four parts

I shipped the same bug this documents, in three places, so the section is written from the repair rather than from theory.

Respira resolves a WordPress user for any request that arrives without one, so writes carry an author. Three resolvers picked the lowest-numbered administrator, which on a site provisioned agent-first by a host or WP-CLI is an agent: the run-as for a whole MCP session, the duplicator adopting an admin identity so `wp_insert_post()`'s internal `unfiltered_html` check passes, and the account an API key binds to.

The branch that most needed the fix is the one i would not have predicted. Our resolver falls back to a capability scan, and that fallback exists because the role query came back empty on two real customer sites: a multisite subsite whose administrator was a network super admin never added to the site, and a single-site install where admin rights came from a custom role plugin through `user_has_cap`. Both looped OAuth clients forever on a 401. The repair in both cases was to resolve by capability instead of by role.

Which is the standard advice, and which is exactly the branch an Administrator agent satisfies. Fixing only the role query would have moved the bug one branch down. That is why the section spends a paragraph on it rather than a line.

For scale: that code runs on 1,335 WordPress sites in the last 90 days, 1,001 of which took at least one successful write from an assistant, 187,050 successful writes since 10 June. Fixed on our side in 8.8.26, including the part where the security scanner deliberately keeps listing agent administrators, because an agent with admin rights that nobody remembers provisioning is the finding rather than noise.

### Two things i left out on purpose

No recommended default for attribution. Whether an agent's display name should reach a visitor is a site's editorial decision, and a plugin guessing at it is how you get surprising bylines.

No hook, per the above.

One observation worth having in your pocket: the multisite rule that agents cannot be super admins gives integrators a fallback path that cannot land on an agent. It is a nice second-order property of a restriction you added for a different reason, and the section says so.

Happy to cut this down, split it, or move it under `Developer reference` instead of after it. Voice is aimed at matching the rest of the file rather than my own, so say the word if it reads off.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1010-c828d7da55547c5bbc973b101fdf9fd4fddc0463.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Experiments/Agent_Users/README.md
+++ b/includes/Experiments/Agent_Users/README.md
@@ -71,3 +71,76 @@ Disable the experiment in code:
 ```php
 add_filter( 'wpai_feature_agent-users_enabled', '__return_false' );
 ```
+
+## Integrator guidance
+
+Agent users change the answer to a question a lot of plugin code asks without noticing: given this site, which user should my code act as? Until now every account that could be resolved that way belonged to a person. Some of them no longer do, and the code that resolves one is usually old, several layers down, and written by someone who is no longer looking at it.
+
+The rule that keeps the rest of this short: an agent is hidden from presentation, never from queries. Ownership, capability, revision, and comment code must keep seeing agents or it will draw the wrong conclusions about content they own. Author dropdowns and participant lists are where they should be filtered out.
+
+### Choosing a user to act as
+
+The two common shapes both resolve by authority, and authority is what an agent has:
+
+```php
+// By role.
+get_users( array( 'role' => 'administrator', 'number' => 1, 'orderby' => 'ID', 'order' => 'ASC' ) );
+
+// By capability: the usual repair when the role query proves unreliable.
+foreach ( get_users( array( 'number' => 50, 'orderby' => 'ID' ) ) as $user ) {
+	if ( user_can( $user, 'manage_options' ) ) {
+		return $user->ID;
+	}
+}
+```
+
+Either can return an Administrator agent. The second one matters more, because moving a resolver from roles to capabilities is the standard advice for making it reliable, and it does not help here. An Administrator agent holds `manage_options` exactly as a human Administrator does, which is the identity model working as designed rather than a gap in it.
+
+Both shapes also tend to be reached for by ID order, and ID order is not a proxy for "the site owner". An agent provisioned before the humans, on a site set up agent-first by a host or by WP-CLI, sorts first. So does an account created by an attacker who backdated it, which is a pattern security scanners already look for; agent users add a legitimate account with the same sorting behaviour.
+
+Where the resolved user will own content or stand in for a person, exclude agents:
+
+```php
+get_users( array(
+	'role'       => 'administrator',
+	'number'     => 1,
+	'orderby'    => 'ID',
+	'order'      => 'ASC',
+	'meta_query' => array(
+		array(
+			'key'     => 'wpai_agent',
+			'compare' => 'NOT EXISTS',
+		),
+	),
+) );
+```
+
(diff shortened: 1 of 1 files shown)


webmyc:
@gziolo this is the README section from https://github.com/WordPress/ai/pull/961#issuecomment-5518462186, as a follow-up against the experiment branch. Scoped to what was agreed: integrator guidance only, no hooks, nothing that touches the identity model.

Since i last commented i went and fixed the same bug in my own plugin, so the section is written from the repair. The part i would not have guessed is that the capability-based fallback is the one that most needs the agent check. Our resolver has that fallback because the role query came back empty on two real customer installs, and resolving by capability is the standard repair for it. An Administrator agent satisfies it exactly. Fixing only the role query moves the bug one branch down, so the section spends a paragraph there rather than a line.

Also worth noting: agents not being able to become super admins means a resolver that falls back to `get_super_admins()` cannot land on one. That is a useful second-order property of a restriction you added for a different reason, and the section points it out.

Nothing here needs a decision from you on the KSES allowlist. That still reads to me like the right thing to hold until the identity contract settles.

webmyc:
The six failing `Test PHP` jobs here are not from this PR, which changes one markdown file and no code.

They fail the same way on the base branch's own head (`dcae15b`), and repo-wide: of the last 30 `Test` runs there is one success on 5 September and nothing green since.

The cause is the wp-env image build, not the test suite:

```
27 |     RUN apt-get clean
28 | >>> RUN apt-get -qy update
target wordpress: failed to solve: process "/bin/sh -c apt-get -qy update"
  did not complete successfully: exit code: 100
```

Three retries, same result each time. apt exit 100 at image build is a package-source problem rather than anything in the branch, so it will most likely need either a base-image bump or a rebuild once the mirror settles.

Flagging it because a docs PR sitting under six red checks looks like it broke something, and it did not.

gziolo:
Thank you for opening the PR. I will have a closer look in the next few days. I experienced similar issues with CI jobs on another PR.

dkotter:
> I experienced similar issues with CI jobs on another PR.

Yeah, there's a known issue with `wp-env` where the PHP 7.4 and 8.0 images are no longer working (see https://github.com/WordPress/ai/pull/1011#issuecomment-5590833519).

Edit: I believe this has been fixed upstream now in https://github.com/WordPress/gutenberg/pull/82478 and should be in the next release of `wp-env`

PR #1011, opened 2026-09-08
Remove unneeded URL validation

dkotter:
## What?

Remove some URL validation code that is no longer needed as we can rely on changes introduced in WP 7.0.3

## Why?

We recently updated the validation we run on URLs when someone passes an image URL into our Alt Text Generation Ability. Some of the validation introduced there is no longer needed as WordPress added the same changes in v7.0.3 so this PR simplifies things a bit to rely on Core as much as possible.

## How?

- Rely more on `wp_http_validate_url` and remove code that we no longer need
- Bump our minimum WordPress version to 7.0.3 which is where those additional fixes were added
- Remove some IPv6 checks that were dead code

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 5
Used for: Evaluating our current code against WordPress Core to determine what could be cleaned up. All changes tested and reviewed by me

## Testing Instructions

Ensure the Alt Text Generation feature still works as expected.

If desired, directly invoke the Alt Text Generation Ability with a custom `image_url` passed in to ensure that works

## Changelog Entry

> Changed - Bump our minimum WordPress version to 7.0.3.
> Removed - Unnecessary URL validation that is now handled by WordPress core.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1011-286dd118d35738acfcd02e96540e2f30a5f71ffe.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dkotter:
Note there's currently an issue in `wp-env` where the PHP package it loads for PHP 7.4 and 8.0 has been shutdown and so any environment that uses those versions won't start up. This causes 6 failing workflows here (and on all other PRs). Ideally this is fixed upstream in Gutenberg but in the meantime, we could drop PHP 7.4 and 8.0 from our test matrix if we want green tests across the board.

justlevine:
> Note there's currently an issue in `wp-env` where the PHP package it loads for PHP 7.4 and 8.0 has been shutdown and so any environment that uses those versions won't start up. This causes 6 failing workflows here (and on all other PRs). Ideally this is fixed upstream in Gutenberg but in the meantime, we could drop PHP 7.4 and 8.0 from our test matrix if we want green tests across the board.

PS: @dkotter this how @galatanovidiu is handling it in `WordPress/mcp-adapter` 
- https://github.com/WordPress/mcp-adapter/pull/322

jeffpaul (approved):

PR #1027, opened 2026-09-14
[WIP 🚧] Content Translation: Enable per-block translation

yogeshbhutkar:
## What, Why, and How?

Content Translation currently only supports translating an entire post, with or without the title. Editors cannot translate an individual content block, or the title on its own.

This limitation shows up in a few practical cases. A post may need different blocks in different languages. The title may need to be translated independently of the body (probably in a different language). Failed blocks are also hard to recover later: after a refresh, the retry notice would be gone, so there is no way to re-run translation for just those blocks.

This PR adds per-block translation. It extends the existing `content resizing` toolbar so that, with the experiment enabled, a compatible block can be translated without running the full-post flow. Title-only translation is not included in this PR.

Next steps:

1. Add independent title translation
2. Decouple the content resizing toolbar into a shared AI toolbar, which different experiments/third-party consumers can use via slot-fills
3. Surface the AI toolbar on multi-block selection

## Screencast

https://github.com/user-attachments/assets/309f699d-ff52-4c86-977f-076fd48b437e

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-6 Astra
Used for: Implementation suggestions and code reviews.

## Changelog Entry

> Added - Per-block content translation.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1027-823b195c765e2cef137b1c1c305935aeb98a69a7.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


yogeshbhutkar:
In an attempt to polish the content translation experiment further, this PR now includes per-block translation as an initial POC. I've outlined in the PR description why I think there’s a need for per-block translation, along with a few possible next steps. I’d appreciate some thoughts on the approach and whether we’re interested in moving in this direction.

I’d also like to better understand the direction we want to take with this feature and the broader toolbar implementation, as well as the overall interest in pursuing it. A shared toolbar (more on it in the PR description) could give us a consistent place to expose multiple per-block AI features in the future, while also leaving room for other consumers to extend it based on their needs. Based on the feedback, I can refactor this PR accordingly and open a separate issue to formalize and track the toolbar work.

cc: @dkotter @jeffpaul

dkotter:
@yogeshbhutkar This all sounds good to me and feels like the right approach.

> Decouple the content resizing toolbar into a shared AI toolbar, which different experiments/third-party consumers can use via slot-fills

This would be great to solve as I do think there are other things that could be added to this toolbar.

> Add independent title translation

For Title Generation, we do render a block toolbar when focus is on the title input. Likely something we can copy from there or even something we can extract and make more generic to work for this and potentially other things in the future.

> Surface the AI toolbar on multi-block selection

This would be great to figure out as it would also be nice for Content Resizing

PR #1029, opened 2026-09-14
Add loading indicator for type-ahead suggestions

### LINKED ISSUE #791: Add loading animation/custom cursor for Type Ahead

As discussed starting in https://github.com/WordPress/ai/pull/151#issuecomment-4247280108 let's see what community feedback is when using the Type Ahead experiment and try some explorations via PRs on adding in some sort of loading animation or custom cursor when the experience is enabled.  So instead of the native blinking cursor perhaps something that let's you know Type Ahead is waiting or working.

No proposed solution as of yet, merely to see how usage of Type Ahead goes and whether we should look into something to better denote in the UI that Type Ahead is waiting/working.

prasadkarmalkar:
## What?

Closes #791

## Why?

Adds a "breathing dots" loading indicator, next to the text caret, that shows while a type-ahead suggestion request is in progress in the Type Ahead feature.

## How?

- Created a new component `TypeAheadLoadingIndicator` which renders the three dots breathing indicator in editor iframe using `createPortal` 
- It uses the same caret `DOMRect` that the ghost-text overlay already uses, following the existing `useCaretData`/`ownerDocument` pattern in this codebase.
- Added a `isLoading` indicator in `useTypeAheadSuggestion` hook, Which is set `true` for the duration of a request and cleared on success, error, cancellation, and on the existing request timeout.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Creating a POC, comments and reviewing the code. All the code changes are self-reviewed as well.

## Testing Instructions

1. Enable the Type Ahead experiment and open a `core/paragraph` block.
2. Type text that triggers a suggestion (auto debounce, or `Ctrl`/`Cmd`+Space for manual trigger) and confirm the three dots appear near the caret while the request is pending, then disappear once the ghost text/suggestion renders.
3. Trigger a request and quickly continue typing/move the caret away. Confirm the indicator disappears and doesn't get stuck.

## Screenshots or screencast

https://github.com/user-attachments/assets/dd15484c-02e7-4b46-97f8-4b336d4086fd

## Changelog Entry

> Added - Loading indicator for type-ahead suggestion requests.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1029-fe829e302dac283f8d6003635266927a3ffe7c7e.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
```diff
@@ -0,0 +1,80 @@
...
+	ownerDocument: Document | null;
+	rect: DOMRect | null;
+	visible: boolean;
+};
+
+const LOADING_INDICATOR_OFFSET_TOP = -3;
+const LOADING_INDICATOR_OFFSET_LEFT = 4;
+
+/**
+ * Portal-rendered breathing-dots marker anchored to the caret while a
```
dkotter, line 27: So I think this surfaces an existing bug we should look to fix. `useCaretData` recreates on every scroll/resize/ResizeObserver tick which means if I scroll while loading is in progress, this new loading indicator disappears and it seems the request itself is stopped. I don't think that is how it should work, scrolling doesn't change where my cursor is at

src/experiments/type-ahead/components/TypeAheadBlock.tsx
```diff
@@ -324,6 +326,11 @@ const TypeAheadBlock = ( {
 				container={ editable ?? null }
 				text={ caretAtEnd ? suggestion?.text ?? null : null }
 			/>
+			<TypeAheadLoadingIndicator
+				ownerDocument={ caret?.ownerDocument ?? document }
+				rect={ caret?.rect ?? null }
+				visible={ isLoading && ! suggestion?.text }
```
dkotter, line 333: If a request finishes but I don't like the suggestion and I manually trigger it again (control + space), the loading indicator doesn't show again because of this conditional

dkotter, line 333: If I manually trigger this in the middle of a paragraph, the loading indicator shows over existing text. Maybe this is fine but wondering if there's a better way to achieve this

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+	if ( ! body || ! style ) {
+		return null;
+	}
+
+	return createPortal(
+		<span
+			className="ai-type-ahead-loading-indicator"
+			style={ style }
+			aria-hidden="true"
```
dkotter: For accessibility, might be nice if this announces to screen readers that it's loading

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		const scrollX = win?.scrollX ?? win?.pageXOffset ?? 0;
+		const scrollY = win?.scrollY ?? win?.pageYOffset ?? 0;
+
+		return {
+			position: 'absolute',
+			zIndex: 1,
+			top: rect.bottom + LOADING_INDICATOR_OFFSET_TOP + scrollY,
+			left: rect.left + LOADING_INDICATOR_OFFSET_LEFT + scrollX,
```
dkotter: I don't think this works on RTL languages

dkotter (changes requested):

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
prasadkarmalkar: For this issue, can we follow the approach below?

- As `useCaretData` exposes `caret` which has `rect` inside it, and `rect` changes on every scroll/resize
- Can we separate it from it, so instead of exporting a single `caret` object containing `rect` as well, we expose two values, `caret` and `rect` separately?
- This way we can prevent the loading request cancellation from scrolling/resizing

src/experiments/type-ahead/components/TypeAheadBlock.tsx
prasadkarmalkar: I have updated it to set the suggestions to `null` whenever the user makes the manual trigger, as we will need to see whether the suggestion text is present or not, or else it will collide with the indicator

prasadkarmalkar: The loading indicator is overflowing and not an inline element in the content now. Maybe we can hide it when it's in the middle of a paragraph? Or should we show it inline in the content? This might require more refactoring of the current implementation, as we have for suggestions when it's in the middle of a paragraph.
Ref - https://github.com/WordPress/ai/blob/a254b06e53f441151db156e3aae05247c09e3880/src/experiments/type-ahead/components/TypeAheadBlock.tsx#L145

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
prasadkarmalkar: Updated it to have `aria-label={ __( 'Loading suggestions', 'ai' ) }` or should we announce it via [speak](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-a11y/#speak) function from `@wordpress/a11y` ?

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
prasadkarmalkar: Updated the position based on the direction of the `editable` element by passing it to the `TypeAheadLoadingIndicator`

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
dkotter: That sounds reasonable to me

src/experiments/type-ahead/components/TypeAheadBlock.tsx
dkotter: I think likely fine to leave it for now (have it overlap existing text). If we get complaints we can consider other options

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
dkotter: I think this is likely good enough

src/experiments/type-ahead/components/TypeAheadLoadingIndicator.tsx
prasadkarmalkar: Thanks @dkotter , I have made the changes as per above approach.

dkotter (approved):

PR #1031, opened 2026-09-15
Alt Text Generation: Integrate with Decorative Image settings

### LINKED ISSUE #1030: Integrate Alt Text Generation with WP 7.1 decorative image support

WordPress 7.1 adds a "Mark as decorative" option to the Image block.  When enabled, WordPress clears the alt text, renders `role="none"`, and disables captions and links for that image.  The Alt Text Generation experiment should account for this new core behavior so that we do not generate unnecessary alt text for images the user has already identified as decorative.  The `ai/alt-text-generation` ability can also already identify images as decorative via its `is_decorative` response, so we should consider connecting that result to the new core setting.

Related: #44 and #557.

Update Alt Text Generation in the Image block so that:
* If an image is already marked decorative, we skip alt text generation and do not offer the Generate Alt Text action
* If AI determines that an image is decorative, suggest that the user mark the image as decorative rather than generating alt text
* Ideally, provide a simple action alongside that suggestion to enable the core "Mark as decorative" setting

I would lean toward suggesting the change rather than automatically enabling it.  Whether an image is decorative can depend on its context, and enabling the setting also affects captions and links, so the user should retain control over that decision.

_No response_

- [x] Yes

yogeshbhutkar:
## What, Why, and How?

Closes https://github.com/WordPress/ai/issues/1030

This PR integrates Alt Text Generation with the Image block’s decorative setting. It disables generation for images marked as decorative and prevents applying an existing suggestion after the image is marked decorative. When AI identifies an image as decorative, it presents a suggestion with a “Mark as decorative” action, keeping the decision with the user.

The action mirrors core by clearing alt text and setting caption and link attributes to `undefined`. If supported, it also sets `isDecorative`, leaving core to render `role="none"` on the front-end. On editors without support, it clears the same attributes but omits the flag (`isDecorative`).

### Use of AI Tools

AI assistance: Yes
Tool(s): Codex
Model(s): GPT-6 Astra
Used for: Multiple rounds of code-reviews to stress test the changes, and to generate the PR description.

## Testing Instructions

1. Enable Alt Text Generation and configure an AI provider.
2. Add an Image block with alt text, a caption, and a link.
3. Mark it decorative and verify generation is disabled.
4. Unmark it and verify generation is enabled.
5. Trigger a decorative AI response and verify the suggestion appears.
6. Accept the suggestion and verify alt text, caption, and link are cleared.
7. Save and reload. Verify the decorative setting persists and core renders `role="none"`.
8. On an unsupported editor, verify the same fields clear without setting `isDecorative`.

## Screencast

https://github.com/user-attachments/assets/fc5c04d6-115e-4c09-aa82-2aee31028321

## Changelog Entry

> Changed - Integrates Alt Text Generation with the Image block’s decorative setting.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1031-b69779d1ac909f97c196a2af24daa9fbf46d08bf.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


src/experiments/alt-text-generation/components/AltTextControls.tsx
```diff
@@ -122,13 +107,13 @@ export function AltTextControls( {
 	 * Handles the generate button click.
 	 */
 	const handleGenerate = async () => {
-		if ( ! ensureProvider( NOTICE_ID ) ) {
+		if ( ! ensureProvider( NOTICE_ID ) || !! attributes?.isDecorative ) {
```
dkotter: Couple things to potentially consider:

1. Cores `Mark as decorative` option exists during generation. So if I click `Generate Alt Text` and while that is processing I click `Mark as decorative`, the request still runs and I see results but I can't apply those results. Thinking maybe we disable the `Mark as decorative` option when generation is in progress?
2. If I generate alt text and then after the results load I then click `Mark as decorative`, the results still show. Should likely hide those

src/experiments/alt-text-generation/components/AltTextControls.tsx
```diff
@@ -236,23 +250,25 @@ export function AltTextControls( {
...
-					<div style={ { marginBottom: '12px' } }>
-						<DecorativeNotice />
-						<div
-							style={ {
-								display: 'flex',
-								gap: '8px',
-								marginTop: '8px',
-							} }
-						>
+				{ isFoundDecorative && (
```
dkotter, line 238: There's a combination here we support that Core does not. If an image is set to open in a lightbox (in the block toolbar, click the link icon and then click `Enlarge on click`) Core automatically removes the `Mark as decorative` option. Basically any image that opens in a lightbox can't be decorative. But our code doesn't consider that so if you generate alt text and it comes back as decorative, we still prompt the user to mark it as decorative. Not a big deal but likely something we want to align with Core on

dkotter (changes requested):
Left a few comments to discuss.

May be nice also to add E2E tests specific to these decorative paths to ensure they work. And documentation could also be updated to mention this.

src/experiments/alt-text-generation/components/AltTextControls.tsx
yogeshbhutkar: > **_Cores Mark as decorative option exists during generation. So if I click Generate Alt Text and while that is processing I click Mark as decorative, the request still runs and I see results but I can't apply those results. Thinking maybe we disable the Mark as decorative option when generation is in progress?_**

Core does not expose a stable API to disable the Image block’s `Mark as decorative` checkbox like this, and there is no stable selector for that control. Any attempt to hide or disable it from this plugin would depend on inspector DOM that can change, thereby making this approach a bit fragile.

> **_If I generate alt text and then after the results load I then click Mark as decorative, the results still show. Should likely hide those_**

I believe with the current approach, in this scenario, even if the results are shown, the "Apply" button will be disabled. However, I do acknowledge the issue of displaying an unexpected entity without clarifying the rationale behind disabling its primary action button.

How about we simply hide generate, the suggestion, and the action buttons whenever `isDecorative` is on, and only show the notice? Uncheck it, and those controls come back. This should help fix both issues. Same idea as in the screencast.


https://github.com/user-attachments/assets/845a9c88-9951-4c3e-9e18-3bd8f7897f3e

src/experiments/alt-text-generation/components/AltTextControls.tsx
yogeshbhutkar: In this scenario, if we choose to respect the AI-generated output, we could prompt the user to leave the alt text empty. Alternatively, we could prioritize the lightbox behavior and force-generate alt text by augmenting `image_meta`.

From an accessibility standpoint, I’m hesitant to suggest keeping the lightbox image decorative with an empty alt, since assistive technologies would ignore it entirely even though it triggers a prominent lightbox interaction that effectively displays the image full-screen.

I also understand the rationale behind Core’s decision that a lightbox image cannot be decorative, since `role="none"` doesn’t really align with such a prominent interaction.

Please let me know which approach we should prefer.

src/experiments/alt-text-generation/components/AltTextControls.tsx
dkotter: > How about we simply hide generate, the suggestion, and the action buttons whenever isDecorative is on, and only show the notice? Uncheck it, and those controls come back

Yeah, I think that's a good solution here

src/experiments/alt-text-generation/components/AltTextControls.tsx
dkotter: So I don't know all the thought/discussion that went into the current Core approach but I'm assuming it's something they thought through and so seemed like a good idea to match them. That said, this is also likely an edge case so maybe we just ignore it for now?

dkotter (approved):

PR #1028, opened 2026-09-14
Feat/suggest reply comment edit page

Infinite-Null:
## What?
Closes #1024

Adds "Suggest Reply" support to the WordPress Edit Comment admin screen.

## Why?
Previously, the Suggest Reply experiment was only available on the Comments list table and the Dashboard Activity widget. When site editors opened the dedicated Edit Comment screen, the Suggest Reply action was missing, preventing them from generating AI reply suggestions directly in the full comment editor.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 4.6
Used for: Validating bug, suggesting a fix.

## Testing Instructions
1. Navigate to **AI > Experiments** and ensure the **Suggest Reply** experiment is enabled.
2. Go to **Comments** and click **Edit** on an existing comment.
3. Verify that the **Suggest Reply** button is displayed beneath the comment editor textarea.
4. Click the dropdown chevron button to verify that tone options (**Friendly**, **Professional**, **Casual**) are available and selectable (with **Friendly** selected by default).
5. Click **Suggest Reply**.
6. Verify the button transitions to a loading state and, once completed, the generated reply populates the comment textarea.

## Screenshots or screencast

https://github.com/user-attachments/assets/5085bf7f-407f-48a1-9c6b-a9bfd66c64ba

https://github.com/user-attachments/assets/d45e2f20-58c3-4f43-a0e7-926b824abad9

## Changelog Entry

> Added - Suggest Reply support to the Edit Comment admin screen.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1028-793565203453bfc3b8715149a277b19731b5534a.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #1049, opened 2026-09-19
Fix: Hide Suggest Reply controls in Quick Edit mode

Infinite-Null:
## What?
Related: #1028 

Hides the "Suggest Reply" split button when the inline comment form is opened in **Quick Edit** mode.

## Why?

The "Suggest Reply" button was incorrectly appearing in Quick Edit mode, where it makes no sense, since the user is editing the comment itself, not composing a reply to it.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 4.6

## Testing Instructions

1. Ensure the **Suggest Reply** experiment is enabled under **AI → Experiments**.
2. Navigate to `/wp-admin/edit-comments.php`.
3. Hover over any comment row to reveal its row actions.

**Quick Edit (should be hidden):**

4. Click **Quick Edit**.
5. Confirm the "Suggest Reply" split button is **not visible** in the button area.
6. Click **Cancel**.

**Reply (should be visible):**

7. Hover over the same comment row and click **Reply**.
8. Confirm the "Suggest Reply" split button **is visible** in the button area.
9. Verify the tone dropdown and generation flow still work as before.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| image | Screenshot 2026-09-19 at 8 50 58 PM |

## Changelog Entry
Fixed - Hide the "Suggest Reply" button when the inline comment form is in Quick Edit mode.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1049-ca6187e9c540f98418be4a0847f382c887d8bcd8.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


src/experiments/suggest-reply/components/SuggestReply.ts
```diff
@@ -296,6 +296,29 @@ function createSplitButtonControls(): HTMLElement {
 	return container;
 }
 
+/**
+ * Returns true when the WP inline form is currently open in Quick Edit mode
+ * (i.e. editing an existing comment). WordPress signals this by adding the
+ * `inline-edit-comment` class to #replyrow.
```
dkotter: This comment seems inaccurate. It mentions reading `inline-edit-comment` class but actually reads `edithead` ID

src/experiments/suggest-reply/components/SuggestReply.ts
```diff
@@ -468,6 +494,23 @@ export function init(): void {
 
 	commentList.setAttribute( INIT_FLAG_ATTR, 'true' );
 
+	/**
+	 * Watch for class changes on #replyrow so we can hide the Suggest Reply
+	 * controls when WordPress switches the inline form into Quick Edit mode
+	 * (class `inline-edit-comment`) and show them again in Reply mode
+	 * (class `inline-reply-comment`).
```
dkotter: Same thing here, this comment doesn't seem to be accurate

tests/e2e/specs/experiments/suggest-reply.spec.js
```diff
@@ -112,6 +112,44 @@ test.describe( 'Suggest Reply Experiment', () => {
 		).toHaveAttribute( 'aria-selected', 'true' );
 	} );
 
+	test( 'Suggest Reply controls are hidden in Quick Edit mode and visible in Reply mode', async ( {
```
dkotter: Minor but this test claims to test both Quick Edit and Reply mode but seems it only tests Quick Edit. Either update this test description or add the Reply mode test

Infinite-Null:
Thanks for the review, @dkotter! I've addressed all points. Happy to hear if anything else needs adjusting. Thanks again!

dkotter (approved):

PR #1053, opened 2026-09-21
Editorial Notes: Fix block selection when Show Template is active

### LINKED ISSUE #1052: Editorial Notes: Only reviews template blocks instead of post content when "Show template" is enabled
### Description

The Editorial Notes experiment is not working properly when "Show template" mode is enabled in the block editor. 

It is expected by Editorial Notes to analyze the actual blocks belonging to the post/page being edited. Instead, it only analyzes the surrounding template blocks, completely skipping the post content itself. 

For example, on a post containing 13 content blocks:
- When "Show template" is disabled, it counts and reviews all 13 blocks: `Reviewing blocks… (0 of 14)`.
- When "Show template" is enabled, it only counts the 4 template wrapper blocks: `Reviewing blocks… (0 of 4)`.

1. Enable the "Editorial Notes" experiment under AI > Experiments.
2. Create or edit a post with several blocks.
3. In the editor header / options, enable "Show template".
4. Open the Post settings sidebar on the right.
5. Under the Post Status panel, look at the "Generate Editorial Notes" button and click it.
6. Observe that the progress indicator only counts the template blocks (e.g. 0 of 4) rather than the actual post blocks (e.g. 0 of 14).

- Without "Show template": Reviews all 13 post blocks (`0 of 14`).

Image

- With "Show template": Only reviews the 4 template blocks (`0 of 4`).

Image

### Environment info

_No response_

- [x] Yes

- [x] Yes

- [ ] Block
- [ ] Classic
- [ ] Not sure

Infinite-Null:
## What?
Closes: #1052

Ensures Editorial Notes analyzes the actual post and page content blocks instead of the surrounding template blocks when the "Show template" option is enabled in the WordPress block editor.

## Why?
When "Show template" is turned on in WordPress, the editor wraps the post inside template elements (such as the Header, Post Title, Featured Image, and Footer). Editorial Notes was fetching blocks from the root of the template rather than the post content itself. As a result, it only reviewed the template blocks (e.g., 4 blocks) instead of the actual post blocks (e.g., 14 blocks).

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 4.6

## Testing Instructions
1. Enable the **Editorial Notes** experiment under **AI > Experiments**.
2. Open or create a post with several paragraphs, headings, and lists.
3. With "Show template" turned off, click **Generate Editorial Notes** in the Post sidebar. Confirm that it reviews all blocks in the post.
4. In the editor header/options, enable **Show template**.
5. Click **Generate Editorial Notes** again.
6. Verify that it still reviews the full count of post blocks, rather than just the template blocks.

## Screenshots or screencast

| Before | After |
| ------ | ----- |
| Screenshot 2026-09-21 at 6 10 48 PM | Screenshot 2026-09-21 at 6 16 37 PM |

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fadmin.php%3Fpage%3Dai-wp-admin%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1053-b49742580a5b317fffcb354b98c12208dfa45bea.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


src/experiments/editorial-notes/hooks/useEditorialNotes.ts
```diff
@@ -72,6 +72,75 @@ interface NoteRecord {
...
 }
 
+/**
+ * Returns the client ID of the `core/post-content` block if present in the editor canvas.
+ * When "Show template" is enabled, post blocks reside inside this block.
+ *
+ * @param selectFn The select function (defaults to @wordpress/data select).
+ * @return The post content block client ID, or null if not found.
+ */
+export function getPostContentClientId(
```
dkotter, line 82: Seems like a lot of the functions introduced here could be useful in other contexts (see https://github.com/WordPress/ai/pull/1055) so wondering if we pull them out of this and into a more general utility file?

src/experiments/editorial-notes/hooks/useEditorialNotes.ts
```diff
@@ -72,6 +72,75 @@ interface NoteRecord {
...
+	selectFn: ( store: any ) => any = select
+): string | null {
+	const blockEditor = selectFn( blockEditorStore );
+	const postContentClientIds: string[] | undefined =
+		blockEditor.getBlocksByName?.( 'core/post-content' );
```
dkotter, line 87: See #1055 but I think we need to ensure we only find the main `core/post-content` block here and not one that may be nested inside a Query Loop

dkotter (changes requested):

src/experiments/editorial-notes/hooks/useEditorialNotes.ts
```diff
@@ -72,6 +72,75 @@ interface NoteRecord {
...
 
+/**
+ * Returns the client ID of the `core/post-content` block if present in the editor canvas.
+ * When "Show template" is enabled, post blocks reside inside this block.
+ *
+ * @param selectFn The select function (defaults to @wordpress/data select).
+ * @return The post content block client ID, or null if not found.
+ */
+export function getPostContentClientId(
+	selectFn: ( store: any ) => any = select
```
yogeshbhutkar, line 83: We can use the `SelectFunction` type exported from `@wordpress/data` here and at other similar places.

src/experiments/editorial-notes/hooks/useEditorialNotes.ts
yogeshbhutkar, line 87: Once we type the store, we can remove optional chaining and use type inference for `postContentClientIds` and other similar places.

src/experiments/editorial-notes/hooks/useEditorialNotes.ts
```diff
@@ -72,6 +72,75 @@ interface NoteRecord {
...
+): T[] {
+	const blockEditor = selectFn( blockEditorStore );
+	const postContentClientId = getPostContentClientId( selectFn );
+
+	if ( postContentClientId ) {
+		return ( blockEditor.getBlocks( postContentClientId ) ?? [] ) as T[];
+	}
+
+	return ( blockEditor.getBlocks() ?? [] ) as T[];
```
yogeshbhutkar, line 116: The `?? []` might not execute as `getBlocks` should return a stable empty array regardless (per its return type). Also, the first param of `getBlocks` (`rootClientId`), is typed `string | undefined`, so we can collapse both into a single call and if `postContentClientId` becomes `undefined`, it should return blocks from the root canvas. This will become apparent once the store is typed. However, please check this logic at your end.

tests/e2e/utils/helpers.ts
```diff
@@ -590,3 +590,48 @@ export const disableAdvancedSettings = async ( page: Page ) => {
...
 	await page.keyboard.press( 'Escape' );
 };
+
+/**
+ * Helper to enable or disable "Show template" mode.
+ *
+ * @param page   Playwright page.
+ * @param enable Whether to enable or disable template mode.
+ */
+export const setShowTemplate = async ( page: Page, enable: boolean ) => {
```
yogeshbhutkar, line 600: I believe we can simplify this to the following (in which case this helper becomes unnecessary):

```ts
// Enable the template mode.
await page.getByRole( 'button', { name: 'View', exact: true } ).click();
await page
	.getByRole( 'menuitemcheckbox', { name: 'Show template' } )
	.click();
```

The issue you're running into probably would be that the `Show template` value is being stored in user preferences and might already be enabled. To get around it, you can use `await requestUtils.resetPreferences();` before each test.

You can also refer to the GB codebase for similar implementation here:
https://github.com/WordPress/gutenberg/blob/f1de34dc43f4f53f5e71dbd668109cacfc384c77/test/e2e/specs/editor/various/post-content-focus-mode.spec.js#L15-L19

yogeshbhutkar (changes requested):
Adding a few minor nits:

Infinite-Null:
Hi @dkotter and @yogeshbhutkar, I've addressed all the feedback. Can you please have a look at your convenience. Thank you!

src/experiments/editorial-notes/components/EditorialNotesPlugin.tsx
yogeshbhutkar: If we're bailing out early on `! clientId` above, then this `if` becomes redundant.

src/utils/blocks.ts
```diff
@@ -238,3 +239,86 @@ export function getEditableTextAttribute(
 
 	return undefined;
 }
+
+/**
+ * Context returned for the post content blocks.
+ */
+export interface PostContentBlockContext< T = BlockWithContent > {
```
yogeshbhutkar: I'd prefer declaring interface/types at the start of the file along with other interface/types until we decide to extract them in a separate file.

src/utils/blocks.ts
```diff
@@ -238,3 +239,86 @@ export function getEditableTextAttribute(
...
+ *
+ * If in template-locked mode but no valid `core/post-content` block is found, `isMissingPostContent`
+ * is set to true and `allBlocks` is empty to prevent operating on template wrapper blocks.
+ *
+ * @template T Block type extending BlockWithContent.
+ * @param {Function} [selectFn] The WordPress data select function (defaults to @wordpress/data select).
+ * @return {PostContentBlockContext<T>} The post content block context.
+ */
+export function getPostContentBlockContext< T = BlockWithContent >(
+	selectFn: ( store: any ) => any = select
```
yogeshbhutkar:
```suggestion
	import type { SelectFunction } from '@wordpress/data';
	selectFn: SelectFunction = select
```

src/utils/blocks.ts
```diff
@@ -238,3 +239,86 @@ export function getEditableTextAttribute(
...
+): PostContentBlockContext< T > {
+	const editor = selectFn( editorStore );
+	const isShowingTemplate =
+		editor?.getRenderingMode?.() === 'template-locked';
```
yogeshbhutkar: Is there a reason to use optional chaining? I believe all optional chainings here are unnecessary.

src/utils/blocks.ts
```diff
@@ -238,3 +239,86 @@ export function getEditableTextAttribute(
...
+	if ( ! isShowingTemplate ) {
+		return {
+			rootClientId: null,
+			allBlocks: ( blockEditor?.getBlocks?.() ?? [] ) as T[],
+			isMissingPostContent: false,
+		};
+	}
+
+	const postContentClientIds: string[] | undefined =
+		blockEditor?.getBlocksByName?.( 'core/post-content' );
```
yogeshbhutkar:
```suggestion
	const postContentClientIds = blockEditor.getBlocksByName( 'core/post-content' );
```

The return type of `getBlocksByName` is `string[]` so the `undefined` in explicit types seems off. Regardless, let's prefer implicit types returned by the selectors everywhere.

src/utils/blocks.ts
```diff
@@ -238,3 +239,86 @@ export function getEditableTextAttribute(
...
+
+	const rootClientId =
+		postContentClientIds?.find( ( clientId: string ) => {
+			if (
+				typeof blockEditor?.getBlockParentsByBlockName === 'function'
```
yogeshbhutkar: This check seems unnecessary. This will become apparent once we use proper types for `selectFn`.

src/utils/blocks.ts
```diff
@@ -238,3 +239,86 @@ export function getEditableTextAttribute(
...
+			) {
+				const queryParents = blockEditor.getBlockParentsByBlockName(
+					clientId,
+					[ 'core/query', 'core/post-template' ]
```
yogeshbhutkar: Is there a reason to include `core/post-template` here? From the `block.json` ([Ref.](https://github.com/WordPress/gutenberg/blob/d1969ff400278263deed21bf2f206fd4b0b42a3e/packages/block-library/src/post-template/block.json#L7)), it appears that this already has `core/query` as the only ancestor, so this path should already be covered via `core/query`?

src/experiments/editorial-notes/hooks/useEditorialNotes.ts
```diff
@@ -222,8 +223,15 @@ export function useEditorialNotes(): {
...
 
-			// Get all blocks and flatten the tree.
-			const allBlocks = select( blockEditorStore ).getBlocks();
+			// Get all blocks belonging to the post and flatten the tree.
+			const { allBlocks, isMissingPostContent } =
+				getPostContentBlockContext< Block >();
+
+			if ( isMissingPostContent || allBlocks.length === 0 ) {
+				setLastRunCount( 0 );
+				return;
```
yogeshbhutkar: I believe this displays the message `No new suggestions found`, but this seems incorrect considering we didn't try to find the suggestions and bailed out early due to unexpected data. In this case, we should ideally clarify this to the user with an appropriate message.

tests/e2e/specs/experiments/editorial-notes.spec.js
```diff
@@ -397,4 +397,83 @@ test.describe( 'AI Editorial Notes Experiment', () => {
...
 		resolveRequest();
 	} );
+
+	test.describe( 'Show template mode', () => {
+		test.beforeAll( async ( { requestUtils } ) => {
+			await requestUtils.activateTheme( 'twentytwentyfour' );
+		} );
+
+		test.afterAll( async ( { requestUtils } ) => {
+			await requestUtils.activateTheme( 'twentytwentyone' );
```
yogeshbhutkar: I believe we should also do `await requestUtils.resetPreferences();` at `afterAll` to prevent the template state from leaking into other tests. Also, let's follow natural order, i.e. `beforeAll`, `beforeEach`, and then `afterAll` so it's easier to read.

Ref: 
https://github.com/WordPress/gutenberg/blob/f1de34dc43f4f53f5e71dbd668109cacfc384c77/test/e2e/specs/editor/various/post-content-focus-mode.spec.js#L22-L24

src/utils/blocks.ts
```diff
@@ -238,3 +239,86 @@ export function getEditableTextAttribute(
 
 	return undefined;
 }
+
+/**
+ * Context returned for the post content blocks.
+ */
+export interface PostContentBlockContext< T = BlockWithContent > {
+	rootClientId: string | null;
+	allBlocks: T[];
```
yogeshbhutkar: IMO, this generic might be tech debt. Since this is a shared helper, `allBlocks` should reflect what the selector returns.

yogeshbhutkar (changes requested):

